PR #2, opened 2017-08-18
Add basic docs structure to agent

ruflin:
This is the basic doc structure needed for the elastic docs build. For convenience I added a script that can be executed with `npm run docs` to build the docs locally to see if the doc build passes and how the docs will look like.

As soon as we have Jenkins added as CI we should also add a build step for `npm run docs` to make sure we do not break the doc build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 node_modules
+build
+
+.DS_Store
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,4 @@
+build
+docs
+script
 test
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -1,0 +1,6 @@
+= APM Node.js Agent Reference
+
+
+== APM Node.js Agent
+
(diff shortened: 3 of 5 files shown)


watson:
@ruflin thanks for taking the time to do this 😄 

Regarding the `Makefile`: Yes, in Node.js land we normally use npm to run scripts, e.g. `npm run docs`. You can read more about that here: https://docs.npmjs.com/cli/run-script

The scripts in the `test` folder is purely for testing - that's why they are there. So I think it's fine to either create a `script` folder like you did, or maybe just put the script in the `docs` folder.

One last thing: When publishing to npm, the npm CLI will not package up files and folders excluded by `.gitignore`. But if you have things that you want in git, but don't want in the package, you can add it to `.npmignore` which works just like `.gitignore` but just for npm. I suggest doing that with both `docs` and `script`.

ruflin:
@watson 
* Command is now `npm run docs`
* Directories added to `.npmignore`

PR #5, opened 2017-08-24
Add docs build to travis

ruflin:


.travis.yml
```diff
@@ -161,6 +161,17 @@ jobs:
       env: TAV=knex,ws,graphql,express-graphql
       script: tav --quiet
 
+    # Node.js 0.10
+    - stage: 'test'
```
watson: Have this been placed last on purpose? Since the `test` stage always will come before any other stage, it would make it easier to read if this was moved up above the `dependencies` stage

.travis.yml
```diff
@@ -161,6 +161,17 @@ jobs:
       env: TAV=knex,ws,graphql,express-graphql
       script: tav --quiet
 
+    # Node.js 0.10
```
watson: I'd probably change this to `# Docs` or something similar 😄

.travis.yml
```diff
@@ -161,6 +161,17 @@ jobs:
...
+    - stage: 'test'
+      script: make docs
```
watson: This should be `npm run docs` instead (interesting that `make docs` exists with exit code zero when there's no Makefile!)

.travis.yml
ruflin: No specific reason, will move it up.

.travis.yml
ruflin: lol, copy paste fail :-( really interesting that it passes.

.travis.yml
ruflin: Fixed ;-)

ruflin:
@watson could you have a look again?

PR #7, opened 2017-08-24
[Docs] Add banner for preliminary docs

dedemorton:
Adding this file to the docs directory will add a banner at the top of each page. I forgot that we only get the banner by default on master if we also have a released version.

image

I'll also create banners in the other repos.

### DIFF
--- a/docs/page_header.html
+++ b/docs/page_header.html
@@ -1,0 +1,1 @@
+You are looking at preliminary documentation for a future release.


watson:
@dedemorton Thanks for the help with setting this up 😃 The text looks fine to me.

To whoever merges this (might be my self): Please change the commit message to `docs(*): add banner for preliminary docs`. I haven't had time to document this rule anywhere, but we're currently trying out the [Angular styled commit messages](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit) in this repo which makes the CHANGELOG easier to maintain.

PR #9, opened 2017-08-28
Don't rely on native database client binaries

watson:
By not expecting the `psql` and `mysql` binaries to be present on the host system, we make it easier for people to run the test suite.


watson:
@ruflin PR updated: Fixing a raise condition in the new database reset logic.

PR #10, opened 2017-08-31
Add simple Elasticsearch instrumentation

watson:
This adds traces for all Elasticsearch queries. The query is logged as well, but for now the index is not logged as that isn't easily available from the Elasticsearch driver.

Example of new Elasticsearch trace:

```js
{
  name: 'Elasticsearch: POST /_search',
  type: 'db.elasticsearch.request',
  start: 2.241679,
  duration: 8.883929,
  stacktrace: [...],
  context: {
    db: { statement: '{"q":"pants"}', type: 'elasticsearch' }
  }
}
```

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -65,3 +65,6 @@ koa-router:
   peerDependencies: koa@2
   versions: '>=5.2.0 <8'
   commands: node test/instrumentation/modules/koa-router/index.js
+elasticsearch:
+  versions: '>=8.0.0'
+  commands: node test/instrumentation/modules/elasticsearch.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ services:
 - mongodb
 - redis-server
 - postgresql
+- elasticsearch
 
 # There's a bug npm@5 which ships with Node 8 that means we get a currupted
 # install when using the test-all-versions command. For details see:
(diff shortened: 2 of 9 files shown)


roncohen:
Cool @watson. I'm wondering if we can think of a better `name` for the traces. Currently, it would be:

     Elasticsearch: POST /_search

We don't usually prefix the name with the name of the service (we don't do: `postgres: SELECT FROM table`), but i can see why you did it here - to distinguish it from regular outgoing http requests. We'll use different colors for external HTTP requests thought.

We might be able to name the trace as follows:

    `query.term`
    `query.match`

by looking at the payload. But I couldn't really come up with a good way to represent indexing a new document without using the REST nomenclature. Maybe the right approach is a mix:

    POST /_search - query.term

watson:
@roncohen I'm not exactly sure what you mean by `query.term`  and `query.match`. Could you come with an example?

roncohen:
@watson For example, here is a match query: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-match-query.html

This could result in the trace being called `query.match`, for example.

watson:
@roncohen I agree that the `Elasticsearch: ` prefix isn't ideal and doesn't follow our normal naming conventions. But I fear that trying to analyse the query to improve the trace name can quickly become complicated as I imagine there's a lot of edge cases we then need to take care of?

Also we have to remember that this trace will always be followed by the normal HTTP trace which will be named `POST /_search` in this case. That's why I preferred some kind of prefix so it would be clear when skimming over the traces, which was which (a post-fix is easier to miss).

I'd obviously like to improve this naming over time and for instance also using the index name as part of the name.

Do you have any suggestions for how we could prefix the trace instead or don't you share my concern about the similarity with normal HTTP traces?

watson:
Merging after 👍 from @roncohen 

Conclusion is that we can always improve the trace name later

PR #13, opened 2017-09-04
Fix setting of custom flushInterval

watson:
Previously the queue was initialized prior to the agent configuration being processed. This meant that agent._flushInterval wasn't available. This commit moves the initialization of the queue to a later stage where the agent configuration have already been processed.


lib/instrumentation/index.js
```diff
@@ -15,18 +15,23 @@ module.exports = Instrumentation
 function Instrumentation (agent) {
   var self = this
```
ruflin: seems like travis is not happy as this variable is never used?

watson: Fixed

PR #16, opened 2017-09-04
Don't start agent if appName is invalid

watson:


lib/agent.js
```diff
@@ -99,6 +99,10 @@ Agent.prototype.start = function (opts) {
     logger.error('Elastic APM isn\'t correctly configured: Missing appName')
     this.active = false
     return this
+  } else if (!/^[a-zA-Z0-9 _-]+$/.test(this.appName)) {
+    logger.error('Elastic APM isn\'t correctly configured: appName contained invalid characters! (allowed: a-z, A-Z, 0-9, _, -, <space>)')
```
ruflin: Could you also log the appName itself here? Something like `appName "in_valid" contains invalid characters.` This could make it easier for the users to see what he actually has set or in case the logs are read by a third paryt, see what it is set to.

I would use `contains` instead of `contained` as at the time of logging, it is still the case.

watson: Fixed 👍

watson:
> As far as I understand, the app is still started, but the agent isn't?

Yes

> Would it be an option to also not start the app in this case? Or is this too intrusive?

The way to not start the app, would be to simply throw this error instead of logging it. That will bring down the app. But most node hosting environments will see this and restart the app, in which case it will crash again and so forth.

My thinking however was that this is too intrusive, but I'm happy to hear counter arguments.

> Are there any options that a user could validate his config before starting?

Not at the moment no. But since the config options most likely change between dev, staging, production etc, validating them locally in your dev env doesn't mean that they are valid in production etc. So I don't really know how valuable such a feature will be?

PR #17, opened 2017-09-04
Only truncate config options once

watson:
Previously these options were truncated for every request to the intake API. Now they are only truncated once the agent is started.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -5,6 +5,8 @@ var fs = require('fs')
 var path = require('path')
 var objectAssign = require('object-assign')
 var bool = require('normalize-bool')
+var trunc = require('unicode-byte-truncate')
+var request = require('./request')
 
 module.exports = config
 
@@ -76,6 +78,7 @@ function config (opts) {
 
   normalizeIgnoreOptions(opts)
   normalizeBools(opts)
+  truncate(opts)
 
   return opts
 }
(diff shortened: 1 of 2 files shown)


lib/config.js
```diff
@@ -5,6 +5,9 @@ var fs = require('fs')
 var path = require('path')
 var objectAssign = require('object-assign')
 var bool = require('normalize-bool')
+var trunc = require('unicode-byte-truncate')
+
+var ES_KEYWORD_MAX_SIZE = 1024
```
ruflin: I would not prefix this with `ES_` as from the agent perspective, it will be the apm-server that will complain and not elasticsearch as it's in the json schema. So also if someone uses Kafka as output on the apm-server side, he still gets an error.

What about just `KEYWORD_MAX_LENGTH`?

watson: I know it's defined in the schema, but I just thought that the concept of a keyword and the limit of 1024 chars was specific to Elasticsearch. So it was more of a "note" to the reader of why the limit existed. But if we want to remove any reference to Elasticsearch, then we shouldn't call it anything with keyword either. I just can't come up with any good name that doesn't mention any of those name.

ruflin: `STRING_MAX_LENGTH` ?

watson: Cool, let's go with `INTAKE_STRING_MAX_LENGTH` to underline that it's related to the intake API.

watson:
@ruflin I ended up not updating the PR with the `INTAKE_STRING_MAX_LENGTH` constant as it's better to unify all locations where we refer to `ES_KEYWORD_MAX_SIZE` to instead use `INTAKE_STRING_MAX_LENGTH`. That doesn't fit into this PR, and the only way is to merge this and then make another one afterwards.

PR #18, opened 2017-09-05
Unify intake API field max length

watson:
~~Depends on #17~~


watson:
@ruflin If we were to have a single location for truncating the payload, it would be inside the `lib/request.js` file (where most of it is today). The reason we have the logic scattered throughout the code, is to minimise the number of strings we truncate more than once. In the future I might even optimise this even more than it is now, so all strings inside `app` and `system` are also only truncated once.

PR #19, opened 2017-09-06
Dockerize tests and add Jenkins support

simitt:
First attempt for dockerizing tests and bringing them to Jenkins.

This currently only runs the `docs` test and the tests with different `node` versions.

To Follow: 
* Running the tests with the `TAV` settings.
* Making use of `npm cache`.
* Dockerizing ES 

See the outcome at [Jenkins Node Agent](https://apm-ci.elastic.co/job/apm-node-agent-test/)


watson:
@simitt Thanks a lot for all the effort and time you have put into getting us set up on Jenkins. I can't wait to see how fast we can get our tests to run now!

I'll go over the PR in detail ASAP, but one thing just caught my eye:

Instead of using the environment variable `TEST_ENV=ci`, could you possibly use `CI=true`? That will work with Travis out of the box (and many other CI services). Then we can change the check for `TRAVIS` to simply a check for `CI`. That way we only have to check one environment variable.

simitt:
Thanks for pointing that out, I will change that.

test/instrumentation/modules/pg/_utils.js
```diff
@@ -1,12 +1,13 @@
...
 
+var host = process.env.POSTGRES_HOST || 'localhost'
 var Client = require('pg').Client
 
 exports.reset = reset
 exports.loadData = loadData
 
 function reset (cb) {
-  var client = new Client({database: 'postgres'})
+  var client = new Client({host: host, database: 'postgres', user: 'postgres'})
```
simitt: @watson I changed the user to `postgres` - which user is usually used when you run the tests locally?

watson: I use the default user which is either the value of the environment variable `PGUSER`, or if that isn't set, then the value of the environment variable `USER`. So in my case it's the username of the user I'm logged in as on my laptop.

watson:
@simitt All files inside the test folder is already ignored by npm when publishing a new version of the Node module. But the new files you've added in the root of the project are not. To fix that, just add them to `.npmignore`. That way we don't increase the size of the `elastic-apm` module unnecessarily 😃

test/instrumentation/modules/ioredis.js
```diff
@@ -12,7 +13,7 @@ var Redis = require('ioredis')
 test('not nested', function (t) {
   resetAgent(done(t))
 
-  var redis = new Redis()
+  var redis = new Redis('6379', host)
```
watson: You don't need to specify the default port. If the host is the first argument it will work fine as well 😃

If you simply inline the environment variable, then it will default to `localhost` if it's not set. I.e:

```js
new Redis(process.env.REDIS_HOST)
```

test/instrumentation/modules/ioredis.js
```diff
@@ -55,7 +56,7 @@ test('not nested', function (t) {
 test('nested', function (t) {
   resetAgent(done(t))
 
-  var redis = new Redis()
+  var redis = new Redis('6379', host)
```
watson: You don't need to specify the default port. If the host is the first argument it will work fine as well 😃

If you simply inline the environment variable, then it will default to `localhost` if it's not set. I.e:

```js
new Redis(process.env.REDIS_HOST)
```

test/instrumentation/modules/redis.js
```diff
@@ -39,7 +41,7 @@ test(function (t) {
     t.end()
   })
 
-  var client = redis.createClient()
+  var client = redis.createClient('6379', host)
```
watson: If you change the argument to an object, you can get away with only supplying the hostname and if you don't need to provide a fallback to `localhost` either (since that's the default), so you can simplify it like so:

```js
redis.createClient({host: process.env.REDIS_HOST})
```

test/instrumentation/modules/mysql/pool-release-1.js
```diff
@@ -32,6 +32,7 @@ test('release connection prior to transaction', function (t) {
 function createPool (cb) {
   setup(function () {
     var pool = mysql.createPool({
+      host: process.env.MYSQL_HOST || 'localhost',
```
watson: No need to specify `localhost` as the fallback as it will happen automatically if `MYSQL_HOST` isn't set

test/instrumentation/modules/mysql/mysql.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+var host = process.env.MYSQL_HOST || 'localhost'
```
watson: No need to specify `localhost` as the fallback as it will happen automatically if `MYSQL_HOST` isn't set

test/instrumentation/modules/mysql/_utils.js
```diff
@@ -1,11 +1,12 @@
 'use strict'
 
+var host = process.env.MYSQL_HOST || 'localhost'
 var mysql = require('mysql')
 
 exports.reset = reset
 
 function reset (cb) {
-  var client = mysql.createConnection({user: 'root', database: 'mysql'})
+  var client = mysql.createConnection({user: 'root', database: 'mysql', host: host})
```
watson: No need to specify `localhost` as the fallback as it will happen automatically if `MYSQL_HOST` isn't set, so you can just inline the environment variable to keep things together as `host` isn't used anywhere else. Potentially split the object into one line per property if you feel the line is getting too long.

docker-compose.yml
```diff
@@ -0,0 +1,64 @@
...
+      - "9200:9200"
+
+  node_tests:
+    build: 
+      context: .
+      dockerfile: Dockerfile
+    environment: 
+      MONGODB_HOST: 'mongodb'
+      REDIS_HOST: 'redis'
+      ELASTICSEARCH_HOST: 'elasticsearch'
```
watson: In Elasticseach there seem to be a de facto standard of calling this environment variable `ES_HOST`. It doesn't really matter for us right now, but in the future it might help us if we sticked with the "standard".

Jenkinsfile
```diff
@@ -0,0 +1,62 @@
+#!groovy
+def nodejs_versions = ['8','7.9.0','6', '5', '4', '0.10', '0.12']
+def test_jobs = [:]
+def env_nodejs_versions = ['8','7','6', '5', '4', '0.10', '0.12']
+def environments = ['generic-pool,mysql,redis,koa-router','ioredis,pg','bluebird','knex,ws,graphql,express-graphql']
```
watson: Would it make more sense to call this `tav_jobs` or something like that?

watson:
@simitt sorry for the mountain of review comments 😅

watson:
I'm not entirely sure how the npm cache works. We have to make sure it doesn't go stale and we then up testing old dependencies. More research is needed 😄

Jenkinsfile
```diff
@@ -0,0 +1,62 @@
...
+                        }finally{ 
+                            sh("./test/script/docker/run_tests.sh ${nodejs_version}")
+                        }
+                    }
+                }
+            }
+            parallel test_jobs
+        }
+
+        //stage("Environments Tests"){
```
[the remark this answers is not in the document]
ruflin: @watson Are the dependencies tests something that should be run for each commit / PR or could it also be run after merging or even daily base? Meaning in general this breaks and if it breaks it's probably less related to the PR but more something that changed in these dependencies?

test/posttest.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-if [ "$TRAVIS" != "true" ]; then
+if [[ "$TRAVIS" != "true" && "$CI" != "true" ]]; then
```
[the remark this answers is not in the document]
ruflin: I'm kind of hoping that travis runs exactly the same in the future as the Jenkins env, means environment is also set up by containers. Not for this PR but thought it's worth noting here.

simitt:
I did not read up too much about npm caching, but I did look up if it expires or not. According to [npm cache docs](https://docs.npmjs.com/cli/cache#details) it should not be necessary to expire something manually. 
But yes we should have an eye on it for the start.

simitt:
@watson thank you for the review, I adapted the usage of `ENVs` in the tests according to your suggestions, and fixed some other issues.

simitt:
As you can see jenkins built got green. I suggest we merge this, then I can also trigger the infra team to add the automatic build.
In the next step I will try to get the `Environment` tests running.

test/instrumentation/modules/redis.js
```diff
@@ -39,7 +39,7 @@ test(function (t) {
     t.end()
   })
 
-  var client = redis.createClient()
+  var client = redis.createClient('6379', process.env.REDIS_HOST)
```
simitt: @watson fyi the suggested handling `var client = redis.createClient({host: process.env.REDIS_HOST})` failed for redis drivers <= 2.0.1.

simitt:
[jenkins build successful](https://apm-ci.elastic.co/job/apm-node-agent-test/43/)

watson:
@simitt this broke the test-cli because the folder `test/scripts` was renamed to `test/script` as part of this PR. What was the reason for this change?

simitt:
I'm sorry this broke it, I wasn't. aware of that since all the tests passed. 
The reason for the change was to be aligned with the general structure. There is a `script` folder and other folders also seemed to be singular rather than plural. Thus it just felt right to have a `test/script` rather than a `test/scripts` and it looked like a minor change to me.

PR #21, opened 2017-09-06
Don't kill all databases after testing

watson:
Be smarter about stopping database after local testing.  

For MongoDB we can use a pid file and for PostgreSQL we instead use the `pg_ctl` command to control the database as a daemon.

### DIFF
--- a/test/posttest.sh
+++ b/test/posttest.sh
@@ -1,8 +1,8 @@
 #!/usr/bin/env bash
 
 if [[ "$CI" != "true" ]]; then
-  killall postgres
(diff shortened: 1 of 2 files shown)


watson:
Just pushed a few changes and updated the description accordingly

watson:
I personally don't run Docker locally for performance reasons. But if you do, this script should be able to detect it and abort. So this is only for people who run the databases "manually"

ruflin:
@simitt Requesting your review on this as you best know if this could have an affect on Jenkins CI or not.

PR #25, opened 2017-09-13
Test that request URL is parsed correctly for hapi

watson:


test/instrumentation/modules/hapi.js
```diff
@@ -15,6 +16,28 @@ var test = require('tape')
...
+    t.equal(request.url.pathname, '/captureError')
+    t.equal(request.url.search, '?foo=bar')
+    t.equal(request.url.raw, '/captureError?foo=bar')
+    t.equal(request.url.hostname, 'localhost:' + server.info.port)
+    t.equal(request.socket.encrypted, false)
+    server.stop()
+    t.end()
+  })
+
+  agent.captureError = originalCaptureError
```
simitt: Should this set the `captureError` to the original value _after_ running the specified assertions in `resetAgent`?

watson: No, `resetAgent` will add a mock `captureError` function, which we need in all the other tests. But in this particular test we actually want the real thing. So after running `resetAgent`, but before running the assertions, we just restore the `captureError` function to its original value.

simitt: ah got it, thx for the explanation.

PR #32, opened 2017-09-18
Add validateServerCert config option

watson:
By default the agent will validate the SSL/TLS certificates used by the APM Server (if running on HTTPS). This validation can now be turned off by setting the `validateServerCert` option to `false`. This is particularly useful if the APM Server is using self-signed certificates.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -90,6 +90,17 @@ The secret token optionally expected by the APM Server.
 
 The URL to where the APM Server is deployed.
 
+[[validate-server-cert]]
+===== `validateServerCert`
+
+* *Type:* Boolean
+* *Default:* `true`
+* *Env:* `ELASTIC_APM_VALIDATE_SERVER_CERT`
+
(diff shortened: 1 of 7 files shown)


watson:
Is `validateServerTLS` a good name?

watson:
@simitt I have tests in the dependency: https://github.com/elastic/apm-nodejs-http-client/commit/7353fbdc9701ebfcbaae5248db55bf12df907533 - but you're right that it might make sense to have tests here as well in case I don't use the dependency correctly or a new version of the dependency is released which contains a bug. I'll look into it

simitt:
For me `validateServerTLS` sounds fine, but if you want to make it even more clear you could extend it to `validateServerTLSCertificate`.

watson:
@simitt or maybe just `validateServerCert`?

watson:
@simitt added tests

watson:
@simitt updated

watson:
I think it's good after the last commit. Just WFG now

watson:
Just force pushed again. But only changes to the documentation

PR #33, opened 2017-09-18
Clean up package.json dependency sorting

watson:


watson:
The Travis build was holding up other important builds, so I cancelled it as this doesn't have any influence on any tests

PR #38, opened 2017-09-19
Don't run dependency tests on greenkeeper branches

watson:
The greenkeeper.io service will create branches starting with `greenkeeper/` when ever a new dependency is released on npm. In this case we only need to test aganst the latest version of the dependency. So there's no reason to run the full dependency build stage in that case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,127 +48,127 @@ jobs:
     - stage: dependencies
       node_js: '8'
       env: TAV=generic-pool,mysql,redis,koa-router
-      script: tav --quiet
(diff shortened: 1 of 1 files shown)


watson:
We currently don't use Greenkeeper on this project, but do on the opbeat/opbeat-node repo. The plan is to move Greenkeeper over here if this trick works. Before merging this I'll be running some tests on the Opbeat repo as well.

watson:
Just tested this on a one off branch called `greenkeeper/testing` and it worked flawlessly. So I'd say we can merge this into master now

PR #40, opened 2017-09-19
Don't sample transactions

### LINKED ISSUE #30: Only one transaction is sent.
I have created a very basic nodejs test app. It is basically just
```
var apm = require('elastic-apm').start({
  // Set required app name (allowed characters: a-z, A-Z, 0-9, -, _, and space)
  appName: 'testNode App',

  // Use if APM Server requires a token
  // secretToken: '',

  // Set custom APM Server URL (default: http://localhost:8200)
  // serverUrl: '',
})


var http = require('http');

var server = http.createServer(function(req, res) {
  res.writeHead(200);
  res.end('Test APM');
});
server.listen(8080);
```

When calling `curl localhost:8080` a few times and waiting until the requests get pushed to the server I only have one document in the database. It looks like it always only sends 1 transaction to the server (at least for the same route).

watson:
In the old Opbeat agent we would sample transactions. This is no longer done and this commit removes the old sampling logic.

Fixes #30

### DIFF
--- a/lib/instrumentation/queue.js
+++ b/lib/instrumentation/queue.js
@@ -12,8 +12,7 @@ function Queue (opts, onFlush) {
   if (typeof opts === 'function') return new Queue(null, opts)
   if (!opts) opts = {}
   this._onFlush = onFlush
-  this._samples = []
-  this._sampled = {}
+  this._transactions = []
   this._timeout = null
   this._flushInterval = (opts.flushInterval || 60) * 1000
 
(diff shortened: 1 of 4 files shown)


watson:
Ignore the Jenkins CI failures for now

simitt:
In the tests you changed the name from `samples` to `transactions` but did not actually test that it can contain more than 1 sample. Thus, you still could use the sampling method and the tests would pass. 
I'd appreciate one test actually testing for non-sampling, since this is a key feature.

watson:
I didn't have a test previously testing that I sampled either. Just changed the property name because I felt it was better 😃 The actual sampling code is removed it `queue.js`. But I agree that we can improve the test suite by adding more tests. Sampling can be hard to test for though as it's easy to get false positives as it all depends on the sampling algorithm used. But I'll add a simple test that hopefully will catch most cases 😅

watson:
@simitt added test

PR #49, opened 2017-10-02
Ensure order of properties in context.request.url

watson:
~~Depends on #48~~


watson:
I had a hard time coming up with one that made just a little sense. What I meant was that this change would mean that `raw`, `protocol`, `hostname`, `port`, `pathname`, `search` came in that specific order in the generated JSON instead of `pathname` coming before `raw`. This in it self isn't important at all, but it just made the code easier to read - so maybe that is what I should have said 😅

PR #51, opened 2017-10-02
Always log DEBUG_PAYLOAD info to STDOUT

watson:
Before you'd need to run with `DEBUG=*` as well to get any output. Since this is only used during development it's ok to log directly to STDOUT.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -189,8 +189,8 @@ function truncEach (arr, len) {
 function capturePayload (endpoint, payload) {
   var dumpfile = path.join(os.tmpdir(), 'elastic-apm-' + endpoint + '-' + Date.now() + '.json')
   fs.writeFile(dumpfile, JSON.stringify(payload), function (err) {
-    if (err) debug('could not capture intake payload: %s', err.message)
-    else debug('intake payload captured: %s', dumpfile)
+    if (err) console.log('could not capture intake payload: %s', err.message)
+    else console.log('intake payload captured: %s', dumpfile)
   })
 }
 


lib/request.js
```diff
@@ -189,8 +189,8 @@ function truncEach (arr, len) {
 function capturePayload (endpoint, payload) {
   var dumpfile = path.join(os.tmpdir(), 'elastic-apm-' + endpoint + '-' + Date.now() + '.json')
   fs.writeFile(dumpfile, JSON.stringify(payload), function (err) {
-    if (err) debug('could not capture intake payload: %s', err.message)
-    else debug('intake payload captured: %s', dumpfile)
+    if (err) console.log('could not capture intake payload: %s', err.message)
+    else console.log('intake payload captured: %s', dumpfile)
```
simitt: I am not sure if this is a good change. Even when only used in development env, you basically break with the logger usage you use everywhere else. Could you use `logger.debug` and set the logger to `console`?

watson: Good question. The main reason why I made this change was to make it simpler/easier to use the payload debug feature.

Before this change, it was a bit annoying that setting `DEBUG_PAYLOAD=1` wouldn't actually show you where it wrote the payload files. It basically meant that `DEBUG_PAYLOAD` by it self was completely useless (on macOS for instance, you cannot predict where the files will be written as random folder names are used for each run, e.g. `/var/folders/8y/k4pwpl7515v1n1tdynxl1xr80000gn/T/`).

So to have any use of this feature you'd have to set both `DEBUG` and `DEBUG_PAYLOAD` - which btw would produce a lot more output than just the intended payload debug output. 

I want a way to enable the feature without having to also enable/set another thing.

While I could change `console.log` to `logger.debug` (or one of the other levels), the downside of that would be that it potentially would require the user to change the agent configuration if that particular log level was suppressed. That would put the user in the exact same position as the one I'm trying to avoid with this PR.

That being said. Using the logger would allow the user to redirect the output, which is nice if used in production. But I'm not sure this should ever be used in production.

So though I agree in theory with your suggested change, I'm not sure it's that user friendly in practice. Thoughts?

simitt: I have not actually looked into what output is logged here to the console, but together with #52 where you make it clear that it only writes the info where to find the log file to the console, this makes sense and sounds reasonable.

PR #53, opened 2017-10-02
test(jenkins): switch from Jenkinsfile to JJB.

simitt:
Add .jenkins files for different Jenkins setup.

fyi @elasticdog


watson:
@simitt should I be concerned about that Jenkins is failing on this PR?

watson:
@simitt Where are these new yaml files used?

simitt:
@watson as discussed last week the Jenkins setup is going to change from using Pipelines (basically the Jenkinsfile) to using the infra standard. The currently failing Jenkins does not need to concern you, since the new handling isn't deployed yet.
The new files will still allow to have the definition of dependencies and nodejs versions in the agent repo. Jenkins is basically fetching those files and using them as parameters for building the jobs.

test/.jenkins_tav_envs.yml
```diff
@@ -0,0 +1,5 @@
+ENVS:
```
watson: This is probably just my lacking knowledge of Jenkins shining though, but I'm not sure if `ENVS` means something in Jenkins-terms? If not, I think this would be better named `TAV` as that is the name of [the associated environment variable](https://github.com/watson/test-all-versions#whitelist-tests-with-environment-variables) these values will be used for.

If on the other hand `ENVS` this makes perfect sense for a person more familiar with Jenkins, then just ignore this comment 😄

simitt: I've renamed it to `TAV`. 
The naming was not Jenkins specific but aligned with how it was called in the `.travis.yml`: `env: TAV=xx`.

watson:
@simitt Merged 😃 Once again, thank you for taking the lead on the Jenkins setup 🙏

PR #54, opened 2017-10-03
Simplify queue API

watson:
The queue shouldn't have to know about transactions and what to do with them when it's flushed.

This commit makes the queue generic so it can hold any data. This makes it easier to test and is good practice in general (separation of concerns).

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -5,6 +5,7 @@ var path = require('path')
 var hook = require('require-in-the-middle')
 var Transaction = require('./transaction')
 var Queue = require('./queue')
+var protocol = require('./protocol')
 var debug = require('debug')('elastic-apm')
 var request = require('../request')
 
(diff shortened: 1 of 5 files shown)


lib/instrumentation/index.js
```diff
@@ -25,12 +26,14 @@ Instrumentation.prototype.start = function () {
   var self = this
   this._started = true
 
-  this._queue = new Queue({flushInterval: this._agent._flushInterval}, function (err, transactions) {
-    if (err) {
-      debug('queue returned error: %s', err.message)
-    } else if (self._agent.active && transactions) {
-      request.transactions(self._agent, transactions)
-    }
+  this._queue = new Queue({flushInterval: this._agent._flushInterval}, function onFlush (transactions) {
```
ruflin: Not part of this PR, but I'm hoping in the future we also get `flush_size` means when the size is reached, it flushes.

watson: It's already done, but I need this PR to be merged first before I can commit it to not get merge conflicts 😄

watson:
@simitt yeah I was actually thinking about this the other day and came to the conclusion that it should hold `_items` - just never got around to update the PR. I'll do that now

watson:
@ruflin there's no plans to use this queue for errors. At the moment it's only for transactions.

watson:
@simitt updated - do you like this better?

lib/instrumentation/index.js
watson: FYI: #56

PR #56, opened 2017-10-07
Add maxQueueSize config option

watson:
Use the maxQueueSize option to force a flush of the queue when it reaches a certain size.

### DIFF
--- a/lib/instrumentation/queue.js
+++ b/lib/instrumentation/queue.js
@@ -14,6 +14,9 @@ function Queue (opts, onFlush) {
   this._items = []
   this._timeout = null
   this._flushInterval = (opts.flushInterval || 60) * 1000
+  this._maxQueueSize = Number.isSafeInteger(opts.maxQueueSize) && opts.maxQueueSize >= -1
+    ? opts.maxQueueSize
+    : -1
 
   // The purpose of the boot flush time is to be lower than the normal flush
   // time in order to get a result quickly when the app first boots. But if a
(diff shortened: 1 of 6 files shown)


lib/instrumentation/queue.js
```diff
@@ -25,7 +26,8 @@ function Queue (opts, onFlush) {
 
 Queue.prototype.add = function (obj) {
   this._items.push(obj)
-  if (!this._timeout) this._queueFlush()
+  if (this._maxQueueSize !== -1 && this._items.length >= this._maxQueueSize) this._flush()
```
jahtalab: if `this._maxQueueSize === 0` also disables this option but user might expect otherwise (since documentation says -1 disables it). In any case it's very minor, just making sure it's intended.

watson: If `maxQueueSize` is set to `0` then both expressions should be `true` - as `this._items.length` will be `1` - in which case it will flush the queue.

jahtalab:
LGTM 👍

lib/instrumentation/queue.js
jahtalab: Should have made the comment on line 17, Sorry about the confusion: 
`this._maxQueueSize = opts.maxQueueSize || -1` if `opts.maxQueueSize===0` then `this._maxQueueSize` would be -1 which disables this option or am I missing something?

watson: Hmm good point. On that note, having a `maxQueueSize` of `0` and `1` have essentially the same effect. I wonder if that's ok, or we should change `0` to mean disabled instead of `-1` 🤔

watson: Fixed

PR #63, opened 2017-10-25
Add fuzziness to flushInterval

watson:
Randomize flush time to avoid servers started at the same time to all connect to the APM server simultaneously.

### Is this even a problem today?

If the queue flush timer was started as soon as the server booted, this PR would solve the problem where multiple servers started at the same time and hence also flushed their queue at the same time.

But since our queue flush timers actually doesn't start until the first transaction have been ended, we might not even have a problem? So this PR might be trying to solve a problem that doesn't exist. Thoughts?

### Better algorithm?

As noted by the TODO in the code, there might be an existing algorithm we can leverage instead of inventing our own

### DIFF
--- a/lib/instrumentation/queue.js
+++ b/lib/instrumentation/queue.js
@@ -31,6 +31,11 @@ Queue.prototype.add = function (obj) {
 Queue.prototype._queueFlush = function () {
   var self = this
   var ms = boot ? MAX_FLUSH_DELAY_ON_BOOT : this._flushInterval
+
+  // Randomize flush time to avoid servers started at the same time to
+  // all connect to the APM server simultaneously
+  ms = fuzzy(ms, 0.05) // +/- 5%
+
   debug('setting timer to flush queue: %dms', ms)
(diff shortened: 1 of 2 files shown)


lib/instrumentation/queue.js
```diff
@@ -50,3 +55,9 @@ Queue.prototype._clear = function () {
   this._items = []
   this._timeout = null
 }
+
+// TODO: Check if there's an existing algorithm for this we can use instead
+function fuzzy (n, pct) {
+  var variance = n * pct * 2
```
roncohen: Untested, but this seems simpler:
`return (Math.random() * 2 - 1) * pct * n + n`

roncohen:
Tiny comment otherwise LGTM 👍

watson:
@roncohen thanks 😃 Do you have any thoughts on the "Is this even a problem today?" part of the PR description?

roncohen:
We haven't had this exact problem but Cloud asked us to do something like this previously.

PR #65, opened 2017-10-25
Drop support for Node.js <4

watson:


sqren:
Is this gonna be a major version bump? #semverGate

watson:
No, it's a minor bump. As long as the version is `0.x`, a minor bump is also considered breaking. It's only when you reach `1.0.0` that only major bumps are breaking (which is why it's normally a good idea to always start at version `1.0.0` for new modules as it gives more flexibility).

lib/instrumentation/modules/https.js
```diff
@@ -9,17 +9,14 @@ module.exports = function (https, agent, version) {
   debug('shimming https.Server.prototype.emit function')
   shimmer.wrap(https && https.Server && https.Server.prototype, 'emit', shared.instrumentRequest(agent, 'https'))
 
-  // From Node.js v0.11.12 until v9.0.0, https requests just uses the
-  // http.request function. So to avoid creating a trace twice for the same
-  // request, we'll only instrument the https.request function if the Node
-  // version is less than 0.11.12 or >=9.0.0
+  // From Node.js v9.0.0 and onwards, https requests no longer just calls the
```
roncohen: "call" should be singular here.

watson: Fixed :)

PR #68, opened 2017-10-26
Log HTTP results in groups of 100

watson:
When recording an HTTP transaction we set its result to the HTTP status code. With this change we now only record the major group, e.g. `2xx` instead of `204` and `5xx` instead of `500` etc.

**Question:** This will obviously only work for HTTP status codes that are above 99 and less than 1000. Can we think of any case where people use an HTTP status code not in this range?

### DIFF
--- a/lib/instrumentation/modules/http.js
+++ b/lib/instrumentation/modules/http.js
@@ -27,7 +27,9 @@ module.exports = function (http, agent) {
       var trans = agent._instrumentation.currentTransaction
 
       if (trans) {
-        trans.result = this.statusCode
+        // It shouldn't be possible for the statusCode to be falsy, but just in
+        // case we're in a bad state we should avoid throwing
+        trans.result = 'HTTP ' + (this.statusCode || '').toString()[0] + 'xx'
 
         // End transacton early in case of SSE
         if (headers && typeof headers === 'object' && !Array.isArray(headers)) {
--- a/test/instrumentation/modules/hapi.js
+++ b/test/instrumentation/modules/hapi.js
@@ -276,7 +276,7 @@ test('request error logging with Error', function (t) {
   var customError = new Error('custom error')
 
   resetAgent(function (endpoint, headers, data, cb) {
-    assert(t, data, { status: '200', name: 'GET /error' })
+    assert(t, data, { status: 'HTTP 2xx', name: 'GET /error' })
 
     server.stop()
   })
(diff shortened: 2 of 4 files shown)


AdriVanHoudt:
Why only record the major group?

watson:
@AdriVanHoudt We still record the actual status code under `context.response.status_code`. The transaction result is used more for grouping the transactions into logical groups. In the past we would make a range query on the server to group all 200->299 together etc. But it's both expensive to do at query-time and it makes assumptions about a transaction being related to an HTTP request which is not always the case.

AdriVanHoudt:
ah makes sense, if the actual status code is still available then 👍

watson:
I just updated this to prefix HTTP transactions with `HTTP `, e.g. `HTTP 2xx`

lib/instrumentation/modules/http.js
```diff
@@ -27,7 +27,7 @@ module.exports = function (http, agent) {
       var trans = agent._instrumentation.currentTransaction
 
       if (trans) {
-        trans.result = this.statusCode
+        trans.result = 'HTTP ' + this.statusCode.toString()[0] + 'xx'
```
sqren: Can statusCode be `undefined` and therefore fail?

watson: It shouldn't be possible to get to that state. Node.js would normally throw if you set an invalid status code, e.g:

```
_http_server.js:188
    throw new RangeError(`Invalid status code: ${originalStatusCode}`);
    ^

RangeError: Invalid status code: undefined
    at ServerResponse.writeHead [as writeHeader] (_http_server.js:188:11)
    at ServerResponse.wrappedWriteHead [as writeHead] (/Users/watson/code/node_modules/elastic-apm/lib/instrumentation/modules/http.js:25:29)
    at ServerResponse._implicitHeader (_http_server.js:179:8)
    at ServerResponse.end (_http_outgoing.js:752:10)
    at Server.onRequest (/Users/watson/code/node_modules/elastic-apm/test/instrumentation/modules/http/basic.js:110:7)
    at emitTwo (events.js:125:13)
    at Server.emit (events.js:213:7)
    at Server.emit (/Users/watson/code/node_modules/elastic-apm/lib/instrumentation/http-shared.js:54:19)
    at parserOnIncoming (_http_server.js:602:12)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:116:23)
```

But I guess it doesn't hurt to guard against it in case the behavior of Node.js changes in the future. I'll update the PR.

watson: Updated... in case `this.statusCode` is `undefined`, it will produce the result `HTTP undefinedxx`. Not pretty, but we're talking about a state that's technically not possible to end up with in Node.js today. So I think it's fair to not add too much logic to handle this - as long as we're not throwing.

PR #74, opened 2017-11-17
Fix npm module name

antonmry:
Doc and examples are pointing to https://www.npmjs.com/package/elastic-apm-node (Not found) instead of https://www.npmjs.com/package/elastic-apm . Feel free to close the PR if it's not the case.

Probably also the file for configuration should be changed from:
`ELASTIC_APM_CONFIG_FILE=/path/to/elastic-apm-node.js`
to 
`ELASTIC_APM_CONFIG_FILE=/path/to/elastic-apm.js`

I can create an issue/PR for that if it makes sense.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -6,7 +6,7 @@ For getting started,
 we recommend that you take a look at our framework specific documentation for either <<express,Express>>, <<hapi,hapi>>, <<koa,Koa>>, or <<custom-stack,custom frameworks>>.
 
 The Elastic APM agent for Node.js is a singleton.
-You get the agent instance by either requiring `elastic-apm-node` or `elastic-apm-node/start`.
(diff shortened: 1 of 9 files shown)


watson:
@antonmry Thanks for noticing and making the PR. You caught us in the middle of a rename of the module (we're renaming it from "elastic-apm" to "elastic-apm-node").

It wasn't our intention that the new documentation should be live without the renaming having taken effect - which unfortunately happened. I'll make sure to fast track this so that the documentation matches reality.

watson:
As of version 0.5.0, the module have now been published under its new name. If you try to install any of the old versions, you should see this deprecation warning:

```
npm WARN deprecated elastic-apm@0.4.0: elastic-apm have been renamed to elastic-apm-node, please install that module instead
```

antonmry:
@watson I was guessing something like that but I've preferred to open the issue just in case. Thanks for the update.

PR #77, opened 2017-11-24
Use async_hooks for Node.js 8.2.0 and above

watson:
Todo:

- [x] Investigate if it's worth it filtering out `TIMERWRAP` - See https://github.com/elastic/apm-agent-nodejs/pull/77#discussion_r152994525
- [x] Investigate if we should add a hook for `promiseResolve` as well
- [x] ~~Benchmark the difference in overhead between this solution and the monkey-patching solution in use pre-v8.2.0~~

### DIFF
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -9,7 +9,7 @@ module.exports = function (ins) {
   Object.defineProperty(ins, 'currentTransaction', {
     get: function () {
       const asyncId = asyncHooks.executionAsyncId()
-      return transactions.get(asyncId)
+      return transactions.has(asyncId) ? transactions.get(asyncId) : null
     },
     set: function (trans) {
       const asyncId = asyncHooks.executionAsyncId()
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -1,0 +1,37 @@
+'use strict'
+
+const asyncHooks = require('async_hooks')
+
+module.exports = function (ins) {
+  const asyncHook = asyncHooks.createHook({init, before, after, destroy})
+  const initState = new Map()
+  const beforeState = new Map()
+
+  asyncHook.enable()
+
+  function init (asyncId, type, triggerAsyncId, resource) {
+    // We don't care about the TIMERWRAP, as it will only init once for each
+    // timer that shares the timeout value. Instead we rely on the Timeout
+    // type, which will init for each scheduled timer.
+    if (type === 'TIMERWRAP') return
+
+    initState.set(asyncId, ins.currentTransaction)
+  }
+
+  function before (asyncId) {
+    if (!initState.has(asyncId)) return // in case type === TIMERWRAP
+    beforeState.set(asyncId, ins.currentTransaction)
+    ins.currentTransaction = initState.get(asyncId)
+  }
+
+  function after (asyncId) {
+    if (!initState.has(asyncId)) return // in case type === TIMERWRAP
+    ins.currentTransaction = beforeState.get(asyncId)
+  }
+
+  function destroy (asyncId) {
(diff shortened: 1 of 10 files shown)


lib/instrumentation/async-hooks.js
```diff
@@ -1,641 +1,37 @@
...
-              throw result.errorVal
-            } else {
-              return result.returnVal
-            }
-          })
-        }
-      }
-    }
+  function before (asyncId) {
+    if (!initState.has(asyncId)) return // in case type === TIMERWRAP
```
watson: We need to consider if the benefit of filtering out `TIMERWRAP` types outweigh the overhead of these `has()` calls

watson:
Concern: We'd probably lose context if using AsyncHooks if a call goes through a native module that isn't using n-api. There's work going on to get AsyncHooks to work with Nan in the Node.js Diagnostic WG, but it's far from ready.

watson:
@Qard I just took a look at [`WeakMap`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) and as far as I can see we can't use it as the weak part is the key and not the value, and so the key needs to be an object (`asyncId` is a number):

> The `WeakMap` object is a collection of key/value pairs in which the keys are weakly referenced.  The keys must be objects and the values can be arbitrary values.

Qard:
Ah, right...maybe we can just do some testing around the `clear*` functions to make sure the references are getting cleared from the maps properly?

watson:
> Concern: We'd probably lose context if using AsyncHooks if a call goes through a native module that isn't using n-api. There's work going on to get AsyncHooks to work with Nan in the Node.js Diagnostic WG, but it's far from ready.

On second thought, this really doesn't matter as it's the same issue with our monkey patched version of the code.

lib/instrumentation/async-hooks.js
```diff
@@ -0,0 +1,35 @@
...
+    // We don't care about the TIMERWRAP, as it will only init once for each
+    // timer that shares the timeout value. Instead we rely on the Timeout
+    // type, which will init for each scheduled timer.
+    if (type === 'TIMERWRAP') return
+
+    transactions.set(asyncId, ins.currentTransaction)
+  }
+
+  function destroy (asyncId) {
+    if (!transactions.has(asyncId)) return // in case type === TIMERWRAP
```
watson: I imagine leaving this up to the `Map.prototype.delete` function is just as efficient

lib/instrumentation/async-hooks.js
```diff
@@ -3,9 +3,19 @@
...
-  const asyncHook = asyncHooks.createHook({init, before, after, destroy})
-  const initState = new Map()
-  const beforeState = new Map()
+  const asyncHook = asyncHooks.createHook({init, destroy})
+  const transactions = new Map()
+
+  Object.defineProperty(ins, 'currentTransaction', {
+    get: function () {
+      const asyncId = asyncHooks.executionAsyncId()
+      return transactions.get(asyncId)
```
Qard: Not sure if it's still the case, but previously `map.get(...)` could deopt if the item did not exist. It might be a good idea to put a `if (!transactions.has(asyncId)) return` before the get.

lib/instrumentation/async-hooks.js
Qard: Same as `get`, this map deopt if it tries to delete an item that doesn't exist.

Qard:
Looks like `async_hooks.executionAsyncId()` was actually not added until 8.2.

watson:
@Qard Ah good catch regarding v8.2. I'll update the version requirement 😃

watson:
Here's how the new Compatibility top section looks:

image

And here's how the new section in the Agent API docs looks:

image

watson:
@Qard I think this is ready now. If you're ok with it I'll merge it into master and release a new version to npm

PR #82, opened 2017-11-28
Lint commit messages

watson:
Make sure that new commit messages follow our commit message conventions.

For now we'll use the config-conventional rule-set which is adapted from the Angular commit message conventions.

For an overview of the rules see:
https://github.com/marionebl/commitlint/tree/master/@commitlint/config-conventional

## To do

- [x] Make it run on Travis CI ~~(currently the PR is trying out a few different approaches. Before merging we should only have one)~~
- [ ] Make it run on Jenkins


watson:
Actually I'm not really sure why the Jenkins tests are failing right now. It shouldn't be running the commit linter at all on Jenkins at the moment, but it does. So maybe I don't have the `JENKINS` environment variable available?

package.json
```diff
@@ -123,6 +127,11 @@
       "/test/sourcemaps/fixtures/lib"
     ]
   },
+  "commitlint": {
+    "extends": [
+      "@commitlint/config-conventional"
+    ]
+  },
```
simitt: @watson looking at the [error msg on Jenkins](https://apm-ci.elastic.co/job/elastic+apm-agent-nodejs+pull-request+multijob-run-tests/54/NODEJS_VERSION=9,label=linux/console) could it be that this part here in `package.json` causes the issues already when trying to install npm modules?

watson: To me it seems that it's running `npm test` (which in turn will run the `test/lint-commits.sh` script). This happens after npm install. This is all expected, but what I intended was for the script [to discover](https://github.com/elastic/apm-agent-nodejs/pull/82/files#diff-f9ca946ada4c88c2435e8bcb5e555e60R5) that it's running inside Jenkins and not actually do anything - but maybe the `JENKINS` environment variable isn't set?

watson:
As agreed on our last Node.js agent meeting, let's just ship this working on Travis only for now.

I've updated the PR to do just that. So it should be ready for review now.

.travis.yml
```diff
@@ -46,6 +46,12 @@ jobs:
       script: npm run docs
       perl: '5.26'
 
+    # Commit Messages
+    -
+      before_install: git fetch --unshallow
+      env: LINT_COMMIT_MESSAGES=true
+      script: test/lint-commits.sh
```
Qard: If you can run an npm script instead, which triggers the sh file, you could have commitlint in the `PATH`, simplifying the script a bit.

.travis.yml
```diff
@@ -46,6 +46,10 @@ jobs:
       script: npm run docs
       perl: '5.26'
 
+    # Commit Messages
+    -
+      script: commitlint-travis
```
Qard: Seems like this actually does everything the `lint-commits.sh` does. Can we just get rid of that file entirely? 🤔

watson: Was in the process of it... Just wanted to check if it worked first, hence the intermediate commit. Check now

.travis.yml
watson: Updated

.travis.yml
Qard: Ah, I see. 😅

PR #87, opened 2017-12-05
feat(parser): support sourceContext option

### LINKED ISSUE #83: Add agent config option to not collect source code
Currently, whenever a stack trace is being sent to the APM Server, the agent [will include](https://github.com/elastic/apm-agent-nodejs/blob/cc446efafe16cd0413a78d2979373a5708ae6465/lib/parsers.js#L189) the source code around each of the lines in the stack trace.

While this feature does use an in-memory LRU cache, it would probably make sense to add a config option that completely disables this feature. This will reduce both the payload size and the processing time.

Config option naming idea: `sourceContext` (environment variable: `ELASTIC_APM_SOURCE_CONTEXT`)

Example of use:

```js
  appName: 'my-app',
  sourceContext: false
})
```

Qard:
The sourceContext option allows disabling the inclusion of source
code lines with stack trace payloads, reducing payload size and
processing time.

This fixes #83.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -303,6 +303,15 @@ Whether or not the agent should monitor for uncaught exceptions and send them to
 
 Set this option to `false` to disable capture of stack traces for measured traces during instrumentation.
 
+[[source-context]]
+===== `sourceContext`
+
+* *Type:* Boolean
+* *Default:* `true`
+* *Env:* `ELASTIC_APM_SOURCE_CONTEXT`
+
+Set this option to `false` to disable collecting and reporting source code context with stack traces.
+
 [[stack-trace-limit]]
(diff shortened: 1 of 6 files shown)


Qard:
Fixed. I just used a fake agent object here. Would you prefer a proper configured agent?

watson:
@Qard Thanks 👍 I think the fake agent is just fine for this. I've used something similar in other tests

watson:
@Qard Just remembered that we need to add this new config option to the docs as well. Would be nice to have that being part of the PR 😃 I suggest putting it just after [`captureTraceStackTraces`](https://github.com/elastic/apm-agent-nodejs/blob/master/docs/agent-api.asciidoc#capturetracestacktraces).

Qard:
Added. 👍

PR #95, opened 2017-12-08
Send coverage reports to codecov

### LINKED ISSUE #89: Upload code coverage reports to codecov.io
Just [as the Python agent](https://codecov.io/gh/elastic/apm-agent-python), we should upload code coverage reports to [codecov.io](https://codecov.io/) as part of CI (I imagine only when testing master).

Here's their Node.js example: https://github.com/codecov/example-node
Here's the Python agent codecov.yml file: https://github.com/elastic/apm-agent-python/blob/master/codecov.yml

Qard:
This should enable reporting coverage data to codecov

Fixes #89

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,14 @@ jobs:
 
   include:
 
+    # Coverage
+    - stage: 'coverage'
+      script:
+        - npm run coverage
+        - ./node_modules/.bin/nyc report --reporter=lcov > coverage.lcov
+        - npm install -g codecov
+        - codecov
+
     # Docs
     - stage: 'test'
       script: npm run docs


watson:
Wow, that was easier than I thought 😄

To make it easy to read the flow of the build stages, I'd put the `coverage` stage below the docs `test` stage as the `test` stage always executes before any other stage no matter which order they are listed in the .travis.yml file

watson:
Hmm there seems to be an issue with the report though. Not sure what the problem is: https://codecov.io/github/elastic/apm-agent-nodejs/commit/fb0b6806667541b70fa9460f7d0d14ec23f9142c

Qard:
*Should* be working now. It wasn't actually generating the lcov file yet.

watson:
Awesome - [it does!](https://codecov.io/github/elastic/apm-agent-nodejs/commit/743592998e2a5eddcc60c550bcd72450f363cd63)

PR #96, opened 2017-12-11
refactor(*): drop in_app in favor of library_frame

watson:
The new property is the exact opposite of the old so all tests are reversed.


watson:
jenkins test this please

PR #98, opened 2017-12-11
Instrument handlebars

### LINKED ISSUE #85: Add instrumentation of handlebar templates
The agent currently doesn't instrument any templating libraries. The most popular one based on download count that I could find was the [handlebars](https://www.npmjs.com/package/handlebars) module, so it would make sense to support that first.

The trace (soon to be renamed span) type should be `template` to align with the other agents.

Qard:
I went for the simple approach and just instrumented the outer compile function and the render function it returns.

@watson Do you think it's worthwhile going deeper than this?

Fixes #85.

### DIFF
--- a/lib/instrumentation/modules/handlebars.js
+++ b/lib/instrumentation/modules/handlebars.js
@@ -1,0 +1,47 @@
+'use strict'
+
+var debug = require('debug')('elastic-apm')
+var shimmer = require('../shimmer')
+
+module.exports = function (handlebars, agent, version) {
+  debug('shimming handlebars.compile')
+  shimmer.wrap(handlebars, 'compile', wrapCompile)
+
+  return handlebars
+
+  function wrapTemplate (original) {
+    return function wrappedTemplate (data) {
+      var trace = agent.buildTrace()
+      var id = trace && trace.transaction.id
+
+      debug('intercepted call to handlebars render %o', {
+        id: id,
+        data: data
+      })
+
+      if (trace) trace.start('handlebars', 'template.handlebars.render')
+      var ret = original.apply(this, arguments)
+      if (trace) trace.end()
+
+      return ret
+    }
+  }
+
+  function wrapCompile (original) {
+    return function wrappedCompile (input) {
+      var trace = agent.buildTrace()
+      var id = trace && trace.transaction.id
+
+      debug('intercepted call to handlebars compile %o', {
+        id: id,
+        input: input
+      })
+
+      if (trace) trace.start('handlebars', 'template.handlebars.compile')
+      var ret = original.apply(this, arguments)
+      if (trace) trace.end()
+
+      return wrapTemplate(ret)
+    }
+  }
+}
--- a/.tav.yml
+++ b/.tav.yml
@@ -63,3 +63,6 @@ koa-router:
 elasticsearch:
   versions: '>=8.0.0'
   commands: node test/instrumentation/modules/elasticsearch.js
+handlebars:
+  versions: '*'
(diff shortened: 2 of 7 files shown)


lib/instrumentation/modules/handlebars.js
```diff
@@ -0,0 +1,47 @@
...
+      var trace = agent.buildTrace()
+      var id = trace && trace.transaction.id
+
+      debug('intercepted call to handlebars compile %o', {
+        id: id,
+        input: input
+      })
+
+      if (trace) trace.start('handlebars', 'template.handlebars.compile')
+      var ret = original.call(this, input, options, hb)
```
watson: There's a risk that the handlebars module at some point change their method signature for instance adding a 4th argument. If possible I like to just `original.apply(this, arguments)` to ensure that we pass on everything exactly as it was given to us

lib/instrumentation/modules/handlebars.js
```diff
@@ -0,0 +1,47 @@
...
+      var trace = agent.buildTrace()
+      var id = trace && trace.transaction.id
+
+      debug('intercepted call to handlebars render %o', {
+        id: id,
+        data: data
+      })
+
+      if (trace) trace.start('handlebars', 'template.handlebars.render')
+      var ret = original.call(this, data)
```
watson: There's a risk that the handlebars module at some point change their method signature for instance adding a 2nd argument. If possible I like to just `original.apply(this, arguments)` to ensure that we pass on everything exactly as it was given to us

Qard: fixed.

lib/instrumentation/modules/handlebars.js
Qard: fixed.

watson:
I forgot to mention that we need to make sure that we run the new test file you created against all the versions of the handlebars module that we support. This will also ensure that we automatically test new versions that are released in the future.

Here's how to do that:

- Add a handlebars section to [.tav.yml](https://github.com/elastic/apm-agent-nodejs/blob/master/.tav.yml)
- Add `handlebars` to one of the `TAV` environment variables for each of the Node versions in the `dependencies` build stage in [.travis.yml](https://github.com/elastic/apm-agent-nodejs/blob/54685b7d73f967fb16044021452ffc15f2d69f78/.travis.yml#L47-L153). Both `TAV=bluebird` and `TAV=ioredis,pg` currently take a long time to complete, so I don't want to add anything more to those right now. So either append it to `TAV=generic-pool,mysql,redis,koa-router` or `TAV=knex,ws,graphql,express-graphql,elasticsearch`
- Similarly append `handlebars` to the same line in [test/.jenkins_tav.yml](https://github.com/elastic/apm-agent-nodejs/blob/master/test/.jenkins_tav.yml)

If you already know that we wont work with all versions of handlebar earlier than a specific version, it makes sense to add a guard in the file where we patch it skip patching in that case. [Here's how we've done that for the koa-router](https://github.com/elastic/apm-agent-nodejs/blob/54685b7d73f967fb16044021452ffc15f2d69f78/lib/instrumentation/modules/koa-router.js#L8-L11) (though you don't have to set an upper version bound necessarily).

watson:
@Qard just saw you added the TAV tests to the Travis matrix 👍 Remember to add them to the Jenkins matrix:

https://github.com/elastic/apm-agent-nodejs/blob/54685b7d73f967fb16044021452ffc15f2d69f78/test/.jenkins_tav.yml#L2

Qard:
Ah, yep. Missed that. Should be included there now too. 👍

PR #100, opened 2017-12-12
test(*): add tests to test async await instrumentation

watson:
I wanted to get these tests into master to make sure that the switch to async-hooks doesn't break this functionality.

I think there shouldn't be any difference as (to the best of my knowledge) async/await is just syntactic sugar on top of native promises, so if our patching or async-hooks works with promises it should also work with async/await.

But I've never used this my self, and hence my knowledge is very very limited on this subject, so please correct me if I'm wrong or if the tests are actually not testing what I think they are testing 😅 

The idea of the test is to start one async operation and then start another that finishes before the first and hence potentially mess with the currently active transaction.

### DIFF
--- a/test/instrumentation/async-await.js
+++ b/test/instrumentation/async-await.js
@@ -1,0 +1,47 @@
+'use strict'
+
+var agent = require('../..').start({
+  appName: 'test',
+  captureExceptions: false
+})
+var ins = agent._instrumentation
+
+var test = require('tape')
+var semver = require('semver')
+
+// async/await isn't supported in old versions of Node.js
+if (semver.lt(process.version, '7.0.0')) process.exit()
+
+var _async = require('./_async-await')
+
+test('await promise', function (t) {
+  t.plan(4)
+  var t1 = ins.startTransaction()
+  _async.promise(100).then(function (result) {
+    t.equal(result, 'SUCCESS')
+    t.equal(ins.currentTransaction && ins.currentTransaction.id, t1.id)
+  })
+  var t2 = ins.startTransaction()
+  _async.promise(50).then(function (result) {
+    t.equal(result, 'SUCCESS')
+    t.equal(ins.currentTransaction && ins.currentTransaction.id, t2.id)
+  })
+})
+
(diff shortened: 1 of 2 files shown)


test/instrumentation/async-await.js
```diff
@@ -0,0 +1,35 @@
...
+  var t2 = ins.startTransaction()
+  a1(50).then(function (result) {
+    t.equal(result, 'success')
+    t.equal(ins.currentTransaction && ins.currentTransaction.id, t2.id)
+  })
+})
+
+async function a1 (delay) {
+  return a2(delay)
+}
```
sqren: This method seems a little redundant. Why not call straight to `a2`?

We could put `a1` to a bit of use, by testing the `await` keyword:

```js
async function a1 (delay) {
    const res = await a2(delay);
    return res.toUpperCase();
}
```

That will at least make sure there are no gotchas with `async/await`. But might not be necessary.

watson: I wanted to make sure we used the await keyword somewhere. As far as I know you don't need to specify it if you just return right away (in fact my linter complained about it), so I removed it, but I was hoping it was implicit. But you're right it's probably better to have it there and then just do something with the result as you suggest

watson: Fixed 😃

test/instrumentation/async-await.js
```diff
@@ -0,0 +1,29 @@
+'use stirct'
```
Qard: strict*

Qard:
BTW, async/await *is* actually mechanically slightly different from promises. With a promise, the continuation is triggered via a callback given to the `then` method, whereas async/await does not do this--not even internally! The execution of a callback given to `then` is guaranteed to be async, delegating to the microtask queue, while an await is *not*. You can do things like `await 1` and, it might not go to the microtask queue.

test/instrumentation/async-await.js
watson: haha ups - fixed :)

watson:
@Qard just added the tests for synchronous await that we talked about

PR #107, opened 2017-12-15
refactor(agent): make all config properties private

watson:
Previously the agent inadvertently exposed some of the config options set when the user configured the agent, e.g. `agent.appName`. All these properties have now been moved under an `agent._conf` namespace to indicate that they should be considered private.

It was never intended that the user could manipulate these properties, but since they wasn't prefixed with an underscore, this might not have been clear.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -51,75 +51,48 @@ Agent.prototype.buildTrace = function () {
 }
 
 Agent.prototype._config = function (opts) {
-  opts = config(opts)
+  this._conf = config(opts)
 
-  logger.init({ level: opts.logLevel })
+  this._conf.serverHost = this._conf.serverUrl
+    ? parseUrl(this._conf.serverUrl).hostname
+    : 'localhost'
 
-  // TODO: Make as much private as possible
-  this.appName = opts.appName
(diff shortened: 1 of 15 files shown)


Qard:
Rather than `_conf`, we could use a `Symbol` to better protect it from outside access.

watson:
@Qard do you mean something like this?

```js
const conf = Symbol('config')
this[conf] = config(opts)
```

We would still need access our selfs... how would that work, or maybe I misunderstood?

Qard:
Yes. something like that. We can export the config symbol somewhere for our code to share. If someone *really* wants to get at that config, they can require whatever file contains the symbol and use that, but then it's pretty clear they are doing something they probably shouldn't.

watson:
That's not necessarily a bad idea. But we have the same "issue" with every `_*` function or property on the agent object, so why treat this one specially?

Qard:
Well, I'm more implying *anything* private could be behind a symbol. Just a suggestion though. I'm good either way. 👍

watson:
@Qard let's take a look at how we can do this for all private properties/functions on the Node agent as a separate PR 👍

PR #121, opened 2017-12-17
test(ci): don't test GraphQL >=0.12.0

watson:
For the moment we do not support GraphQL 0.12.0 and above. Until support for these versions are added, we'll limit the TAV tests so that our builds don't break.

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -33,7 +33,7 @@ ws:
   commands: node test/instrumentation/modules/ws.js
 graphql:
   preinstall: rm -fr node_modules/express-graphql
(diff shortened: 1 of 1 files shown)


simitt:
jenkins, run the tav tests

simitt:
@watson I wonder if there is a high level overview of which libs are supported in which versions in the docs?

watson:
@simitt Yes there is: https://www.elastic.co/guide/en/apm/agent/nodejs/current/compatibility.html

Related: https://github.com/elastic/apm-agent-nodejs/pull/122

PR #125, opened 2017-12-17
Instrument incoming http2 requests

watson:
For now this PR is only a test to show that http2 instrumentation isn't added yet. More to come...

This addresses #123


watson:
Close in favor of #205

PR #127, opened 2017-12-18
feat(express): auto-track errors

### LINKED ISSUE #84: Automate adding of Express agent middleware
If you're running express with Elastic APM, you today need to add our [custom error catching middleware](https://www.elastic.co/guide/en/apm/agent/nodejs/current/express.html#express-error-handlers) manually.

This have confused quite a few people and is unique to how Express compare to our setup instructions for instrumenting other Node.js frameworks, so it would be nice if middleware was added automatically when starting the agent, similar to how Express is patched in other ways.

Qard:
This enables automatic tracking of express errors, rather than requiring manual use of the middleware.

Fixes #84

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -642,13 +642,7 @@ In which case it will automate this processes for you.
 apm.middleware.connect()
 ----
 
-Alias: `apm.middleware.express()`
-
 Returns a middleware function used to collect and send errors to the APM Server.
-
-The middleware can be used as-is with either Connect or Express in the same way.
-
-The middleware must be added _after_ your router to pick up any errors that may result from handling a request.
 
 [source,js]
 ----
@@ -672,7 +666,7 @@ app.use(apm.middleware.connect())
 app.listen(3000)
 ----
 
-NOTE: `apm.middleware.express` or `apm.middleware.connect` _must_ be added to the middleware stack _before_ any other error handling middleware functions or there's a chance that the error will never get to the agent.
+NOTE: `apm.middleware.connect` _must_ be added to the middleware stack _before_ any other error handling middleware functions or there's a chance that the error will never get to the agent.
 
 [[apm-start-transaction]]
 ==== `apm.startTransaction()`
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -12,12 +12,45 @@ module.exports = function (express, agent, version) {
     return express
   }
 
+  // express 5 moves the router methods onto a prototype
+  var routerProto = semver.satisfies(version, '^5')
+    ? (express.Router && express.Router.prototype)
+    : express.Router
+
+  var layerPatchedSymbol = Symbol('layer-patched')
+  var reportedSymbol = Symbol('reported')
+  function patchLayer (layer) {
+    if (!layer[layerPatchedSymbol]) {
+      layer[layerPatchedSymbol] = true
+      debug('shimming express.Router.Layer.handle function')
+      shimmer.wrap(layer, 'handle', function (orig) {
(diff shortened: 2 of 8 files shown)


lib/instrumentation/modules/express.js
```diff
@@ -12,12 +12,45 @@ module.exports = function (express, agent, version) {
...
+  function patchLayer (layer) {
+    if (layerPatched) return
+    layerPatched = true
+
+    debug('shimming express.Router.Layer.handle_error function')
+
+    shimmer.wrap(layer.constructor.prototype, 'handle_error', function (orig) {
+      return function (err, req) {
+        if (!err[reportedSymbol]) {
+          err[reportedSymbol] = true
```
watson: I really like this approach. I can think of a few places in our code base where we should do that instead 😅

lib/instrumentation/modules/express.js
```diff
@@ -12,12 +12,45 @@ module.exports = function (express, agent, version) {
     return express
   }
 
+  // express 5 moves the router methods onto a prototype
```
watson: Nice that you're thinking of Express 5 already 👍

Qard: I haven't fully tested express 5, but that was an easy part to fix.

watson:
We should probably also add the new Express test file file to .tav.yml 😃

Qard:
Since it's in end-user code, would it be better to have a deprecation warning for a bit?

watson:
Normally I would deprecate yes, but as we're still in beta and since we're making a few other breaking changes when go GA, I think it's ok for us to make this one breaking as well. The more we can clean up now and not carry around as legacy for a while the cleaner/better I think. We'll have to make an upgrade guide for both old Opbeat users and for beta users when we go into GA anyway, so we can have a section mentioning breaking changes like this here. That's at least my current philosophy 😄

Qard:
Made those changes and also, when adding the TAV details discovered the approach I took didn't work before express 4.6 so I made some changes to how the patch gets applied. It's not using the `layer.constructor.prototype` anymore, so each layer instance will get patched separately, but as far as I can tell that should only happen the first time that layer is used.

docs/agent-api.asciidoc
```diff
@@ -642,11 +642,9 @@ In which case it will automate this processes for you.
 apm.middleware.connect()
 ----
 
-Alias: `apm.middleware.express()`
-
 Returns a middleware function used to collect and send errors to the APM Server.
 
-The middleware can be used as-is with either Connect or Express in the same way.
+The middleware can be used as-is with either Connect in the same way.
```
watson: I'd just remove the entire line here

watson:
@Qard if you feel it's better I think we should consider just dropping support for Express <4.6 as that version was released in June 2014. What do you think?

docs/agent-api.asciidoc
Qard: fixed.

watson:
@Qard As you also mentioned in https://github.com/elastic/apm-agent-nodejs/pull/131#issuecomment-352911890, we forgot to add the new tav test to the `TAV` environment variables. That should be the last thing missing 😃

Before merging we need to figure out if we want to publish a new version of the module right after (as it's changing the docs), or if there's a better way to handle this. See #134 for more details.

Tip: I just tested it locally on Node 9 by having `the-all-versions` module installed globally and then running this command:

```
tav express '>=4' -- node test/instrumentation/modules/express.js
```

... and it all worked perfectly 👍

watson:
Weird... I tried to resolve the conflicts using the "Resolve conflicts" button, but even though the commit that resolves them are included in the PR, GitHub still thinks the conflicts exists.

PR #128, opened 2017-12-18
test(ci): add codecov.yml for status indicators

### LINKED ISSUE #113: Switch codecov to report via PR checks instead of a comment
Today codecov will post a pretty length comment on PR's. Takes up a lot of real-estate 😞 

Instead we should do what the Python agent is doing and just have it change the status of the PR:

image

See the [codecov.yml](https://github.com/elastic/apm-agent-python/blob/master/codecov.yml) file from the Python agent for inspiration.

Qard:
Fixes #113

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -8,3 +8,4 @@ docker-compose.yml
 Jenkinsfile
 .nyc_output
 coverage
+codecov.yml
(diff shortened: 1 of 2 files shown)


watson:
I would have expected this PR it self to then have the codecov checks, but maybe it only works properly after it have been checked into master?

image

@beniwohli what's your experience?

beniwohli:
@watson can't remember tbh :(

watson:
We only live once 😎  Let's try and merge it and see if it starts working after it's in master.

Qard:
fixed

PR #129, opened 2017-12-19
chore(travis): improve readability of .travis.yml

watson:
The order of the the different stages are now in the same order as they are executed on Travis and big headlines separate them so it's easier to read

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,18 +31,29 @@ jobs:
 
   include:
 
-    # Coverage
-    - stage: 'coverage'
+    ###########################################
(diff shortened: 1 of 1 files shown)


Qard:
Should the docs job maybe have it's own stage so it appears separately in the results?

watson:
I wish we could separate it in some way, but if we put it in a separate stage it would mean that the entire test suite would take 2 minutes longer as each stage run synchronously.

Qard:
Hmm...yeah, best we leave it in the test group then.

PR #131, opened 2017-12-19
test(hapi): fix hapi tests

### LINKED ISSUE #126: Tests failing for hapi <=15.0.1
According to the [compatibility page](https://www.elastic.co/guide/en/apm/agent/nodejs/current/compatibility.html) we're supposed to support hapi all the way back to version 9. But if running the [hapi tests](https://github.com/elastic/apm-agent-nodejs/blob/master/test/instrumentation/modules/hapi.js) on version 15.0.1 or older of hapi, the test suite fails.

I believe that it used to work for these versions, but the tests were never covered by our [TAV test suite](https://github.com/elastic/apm-agent-nodejs/blob/master/.tav.yml), so the issue might have snug in at some point later when refactoring either the tests or the actual [hapi instrumentation code](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/hapi.js).

Qard:
connectionless tests should not run before 15.0.2.
server.stop(...) requires a stop callback until 10.x.

Fixes #126

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -66,3 +66,6 @@ elasticsearch:
 handlebars:
   versions: '*'
   commands: node test/instrumentation/modules/handlebars.js
+hapi:
+  versions: '>=9.0.1 <17.0.0'
+  commands: node test/instrumentation/modules/hapi.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,7 +59,7 @@ jobs:
       script: 'if ! [[ $TRAVIS_BRANCH == greenkeeper/* ]]; then tav --quiet; fi'
     -
       node_js: '9'
-      env: TAV=knex,ws,graphql,express-graphql,elasticsearch
+      env: TAV=knex,ws,graphql,express-graphql,elasticsearch,hapi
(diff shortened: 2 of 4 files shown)


watson:
Forgot to mention that we need to manually add this new hapi tav test to the `TAV` environment variables in Travis and Jenkins:

- [.travis.yml](https://github.com/elastic/apm-agent-nodejs/blob/d2107a55d535767b28713d843c2c5c84500f1106/.travis.yml#L62)
- [test/.jenkins_tav.yml](https://github.com/elastic/apm-agent-nodejs/blob/d2107a55d535767b28713d843c2c5c84500f1106/test/.jenkins_tav.yml#L5)

Qard:
Ugh, sorry. I keep forgetting those. I need to add the new express test file too. 😞

watson:
I could have done a better job documenting it. By default tav will run all tests in .tav.yml, unless a `TAV` environment variable is detected, in which case that's used as a white-list. But since we always have that defined, we have to manually maintain it

Qard:
Updated to exclude 9.0.0.

PR #132, opened 2017-12-20
test(mongodb): test all supported versions of mongodb-core

watson:
This PR also locks CI to test dependencies against 7.9.x instead of 7.x as the issues related to 7.10.x is now affecting the TAV tests after mongodb-core have been added to the mix.

BREAKING CHANGE: Dropped official support for mongodb-core v1.2.7 to v1.2.18 as testing those old versions are complicated.


Qard:
All Node.js versions appears to fail on mongodb-core 1.3.21, except 7.x which fails at 2.1.17. 😿

watson:
I fixed the tests

PR #139, opened 2017-12-20
fix(graphql): support sync execute

### LINKED ISSUE #119: Add support for GraphQL 0.12
GraphQL 0.12 have just been released and we need to add support for it.

Currently our test-suite fails if run against this version.

Qard:
Fixes #119

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -33,7 +33,7 @@ ws:
   commands: node test/instrumentation/modules/ws.js
 graphql:
   preinstall: rm -fr node_modules/express-graphql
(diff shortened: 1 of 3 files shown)


Qard:
I didn't bump 0.12 in the dev dependencies as it seemed to conflict with the express-graphql tests which restricts to 0.10 and 0.11 via peer dependencies.

watson:
Yeah, until a new version of express-graphql is released which support graphql 0.12, then we'll stick with our current default devDependencies 👍

PR #145, opened 2017-12-27
feat(mysql): support mysql2 module

LewisJEllis:
Context: I work on a project that uses the Sequelize ORM. Sequelize moved from node-mysql to node-mysql2 as of version 4.x about 6 months ago. node-mysql2 is largely drop-in compatible with node-mysql, so much that someone added mysql2 support to that other big APM provider's agent by basically just duplicating what they did to instrument node-mysql, so I tried that here and it experimentally works for me.

I'm guessing we'd like to run the existing mysql tests against mysql2, but I'm not sure what the nicest way to get there would be given your existing test setup and not wanting to just duplicate all the tests, so seeking some guidance on that.


watson:
@LewisJEllis thanks for the contribution 👍 

After I merged in your other PR #144, this one now has a merge conflict. Could I get you to rebase?

For the tests to run for the mysql2 modules as well, you need to make the following changes also:

- Add `mysql2` to the `.tav.yml` file in a similar fasion as to the `mysql` module:
  https://github.com/elastic/apm-agent-nodejs/blob/938f47a7a2aca7dd921a0914d7b14363f0ae5605/.tav.yml#L4-L8
- For Travis CI to actually run the tests, you also need to add `mysql2` to the `TAV` environment variable for each of the Node.js versions here (I suggest you add it next to `mysql`):
  https://github.com/elastic/apm-agent-nodejs/blob/938f47a7a2aca7dd921a0914d7b14363f0ae5605/.travis.yml#L66-L172
- Similarly, we need to tell our other CI service Jenkins about it as well:
  https://github.com/elastic/apm-agent-nodejs/blob/938f47a7a2aca7dd921a0914d7b14363f0ae5605/test/.jenkins_tav.yml#L1-L5

lib/instrumentation/modules/mysql-common.js
```diff
@@ -0,0 +1,161 @@
+'use strict'
```
Qard: It might be better to move this file up a level, like we do with `http-shared.js`, so we aren't conflicting with the possible `mysql-common` package name.

Qard:
@watson This PR doesn't currently include a test file for `mysql2`. I assume it could be adapted to share tests with `mysql` similar to how the instrumentation is shared.

watson:
@Qard the two modules should be almost identical in their API's. So our tests for the mysql module might just work for the mysql2 module as well. But it remains to be seen. So I was just thinking that a first step would be to add the mysql2 module to the test matrix so we could get an overview.

.tav.yml
```diff
@@ -6,6 +6,11 @@ mysql:
   commands:
     - node test/instrumentation/modules/mysql/mysql.js
     - node test/instrumentation/modules/mysql/pool-release-1.js
+mysql2:
+  versions: ^1.0.0
+  commands:
+    - node test/instrumentation/modules/mysql/mysql.js
+    - node test/instrumentation/modules/mysql/pool-release-1.js
```
Qard: Looks like these should be mysql2, not mysql.

LewisJEllis: Yea, I ran it locally and it worked (but was actually just running mysql...) so I pushed and then realized I have quite a bit left to fix :)

Qard: No worries. Thanks for the great work you've done so far! I'll try to get this reviewed as soon as I can, once it's been updated. 👍

LewisJEllis: I was hoping to work toward sharing test cases between node-mysql and node-mysql2, but unfortunately I've run into multiple spots where node-mysql2 isn't a perfect drop-in replacement. The two codebases read very similarly, and many parts of the instrumentation work out of the box, but there are subtle differences that end up breaking some cases.

For example, in node-mysql2 `Connection.query(<a Query>)` uses the Query object without invoking `Connection.createQuery`, but in node-mysql the same call passes the Query object to `Connection.createQuery` which immediately returns it.

The connection factory tests are otherwise mostly working by passing over the case I described above, but the pool tests are dead on arrival - no trace recorded in the basic case.

This is going to take some more work, will revisit.

test/instrumentation/modules/mysql2/mysql.js
```diff
@@ -0,0 +1,566 @@
...
+]
+
+factories.forEach(function (f) {
+  var factory = f[0]
+  var type = f[1]
+
+  // Prior to mysql v2.4 pool.query would not return a Query object.
+  // See: https://github.com/mysqljs/mysql/pull/830
+  // var skipStreamTest = type === 'pool' && semver.satisfies(mysqlVersion, '<2.4.0')
+  var skipStreamTest = false
```
watson: This boolean is probably not needed for mysql2. If that's the case it's better if we just remove it as it makes it easier to read the code 😃

test/instrumentation/modules/mysql2/mysql.js
```diff
@@ -0,0 +1,566 @@
...
+  // Prior to mysql v2.2 pool.query required a callback.
+  // See: https://github.com/mysqljs/mysql/pull/585
+  var skipNoCallbackTest = type === 'pool' && semver.satisfies(mysqlVersion, '<2.2.0')
+  skipNoCallbackTest = true
```
watson: This boolean is probably not needed for mysql2. If that's the case it's better if we just remove it as it makes it easier to read the code 😃

watson:
@LewisJEllis We just updated the CI setup by removing Node.js 5 and 7 from the test matrix, which unfortunately conflicts with this PR. I think the conflict with `lib/instrumentation/modules/mysql.js` is still related to the typo fix you made earlier. Could I get you to rebase and resolve conflicts? Then I can take a more in-depth look

.tav.yml
Qard: Ah, yeah, that sounds tough. I'd be happy to work through this together, if you like. It looks like we're in the same time zone, so we could sync up on a slack somewhere, if we're on any of the same ones. 😉

danieljuhl:
@LewisJEllis great work - and you still working on this PR? If not, then I'm willing to spend some time to get this change merged.

.tav.yml
LewisJEllis: I'm actually not on any Node-related Slacks - maybe invite me to one if you like.

LewisJEllis:
I am not currently working on this - it turned out that the portion I got working here was enough for my immediate need. I might still work some more on it, but I won't have time in the immediate future, so feel free to take over.

watson:
Closing this in favor of #211

@LewisJEllis Thanks for getting the ball rolling 😃

watson:
Support for mysql2 have now been released in v1.7.0. Thanks so much for all your help landing this 😃

PR #146, opened 2017-12-28
Hapi 17

Qard:
This adds hapi v17 support.


Qard:
@watson Looks like it's working here. I'll get it cleaned up tomorrow and perhaps you can review it when you're back to work?

lib/instrumentation/modules/hapi.js
```diff
@@ -14,113 +14,137 @@ module.exports = function (hapi, agent, version) {
...
+    } else if (typeof server.events.on === 'function') {
+      attachEvents(server.events)
+    } else {
+      debug('unable to enable hapi error tracking')
+    }
+
+    // When the hapi server has no connections we don't make connection
+    // lifecycle hooks
+    var conns = server.connections
+    if (conns && conns.length === 0) {
```
AdriVanHoudt: This needs an hapi 17 check since in hapi 17 connections are not a thing any more.

lib/instrumentation/modules/hapi.js
```diff
@@ -14,113 +14,137 @@ module.exports = function (hapi, agent, version) {
...
+      debug('unable to enable hapi instrumentation on connectionless server')
+      return
+    }
+
+    // Hooks that are only allowed when the hapi server has connections
```
AdriVanHoudt: Not a 100% correct statement anymore

AdriVanHoudt: The check is still fine afaik

lib/instrumentation/modules/hapi.js
Qard: The existence check of `server.connections` covers that.

lib/instrumentation/modules/hapi.js
Qard: @watson Any thoughts on wording here?

watson: @AdriVanHoudt Do you mean related to the removal of multi-connections support in hapi 17?
 What would be a more appropriate comment?

lib/instrumentation/modules/hapi.js
watson: I guess what @AdriVanHoudt means is that outputting the debug message "unable to enable hapi instrumentation on connectionless server" isn't really that nice if running hapi 17?

So it should be more like `if (!hapi17plus && conns && cons.length === 0) {` where `hapi17plus` is a boolean that's true only of we're running hapi version 17 or higher.

Qard: We'd never reach that debug message on hapi 17, because `server.connections` would have to *exist*.

watson: Ah my bad 👍

AdriVanHoudt: it was more for correctness ^^

lib/instrumentation/modules/hapi.js
AdriVanHoudt: As far as I can tell hapi 17 will have it always.
In 16 it only becomes available after adding a connecting but in 17 you can't run a server without it. The moment you create the server the ext will be there in 17.
So maybe `When we can hook into the lifecycle to do proper transaction naming` or something?

watson: @AdriVanHoudt how about just `Hooks that are only allowed when the hapi server has connections (with hapi 17+ this is always the case)`

lib/instrumentation/modules/hapi.js
watson: Yeah I see it can be a little confusing to read because of the above comment. How about just updating the comment to something like:

```
// Prior to hapi 17, when the server has no connections we can't make connection
// lifecycle hooks (in hapi 17+ the server always have connections, though the
// `server.connections` property doesn't exists, so this if-statement wont fire)
```

Qard: updated.

lib/instrumentation/modules/hapi.js
Qard: updated.

Issue #88, opened 2017-12-06
Add app.environment to payload sent to APM Server


watson:
To run Elastic APM on multiple environments of the same service, you'd previously normally add a post-fix to the `serviceName`, e.g: `my-service-staging` or `my-service-prod`.

In Elastic APM 6.2 we'll introduce a new `environment` property tailored specifically for this. This means that users should use the same `serviceName` (`my-service`), and instead specify the environment via a separate property whos value should be sent up to the APM Server using `service.environment` in the payload.

We need to settle on the name of the config option, but we'll probably name it either `env` or `environment`.

Example usage:

```js
require('elastic-apm-node').start({
  serviceName: 'my-service',
  env: 'production'
})
```

The associated environment variable would then be either `ELASTIC_APM_ENV` or `ELASTIC_APM_ENVIRONMENT` depending on what name we end up choosing.

Related APM Server PR: elastic/apm-server#366

watson:
We should default its value to be the that of `process.env.NODE_ENV` (hat tip to @Qard)

AdriVanHoudt:
If it is called `environment` in the Elastic APM I would use the same name in the config

watson:
Another unanswered question: Should we default to a value if neither it nor `process.env.NODE_ENV` is set? Or should it then just be left out entirely?

watson:
(just renamed `app` to `service` in the issue description to align with the new naming conventions)

AdriVanHoudt:
> Should we default to a value if neither it nor process.env.NODE_ENV is set?

From a user perspective, leave it blank. If you have use case for it business wise or something you can default it.

Qard:
Most popular web frameworks internally default to 'development' when NODE_ENV is not present.

see: https://github.com/expressjs/express/blob/c0136d8b48dd3526c58b2ad8666fb4b12b55116c/lib/application.js#L71

watson:
@roncohen I think we settled on calling the config variable `environment` as opposed to `env` right?

roncohen:
This is blocked waiting for UI support

watson:
@roncohen what's the status on the API UI support?

sqren:
@alvarolobato I don't think we are tracking this issue in the UI board so created an issue https://github.com/elastic/kibana/issues/21192

alvarolobato:
Thanks @sqren

havber:
[...]

PR #150, opened 2018-01-02
chore(config): add environment name

Qard:
Fixes: #88


watson:
After writing the above comment we just decided to move this feature to Elastic 6.3 instead of 6.2, so it needs to go into a new branch we'll call `elastic-6.3` instead.

Qard:
Oh, oops. My bad. I meant to target elastic-6.2. If we're bumping it to 6.3, I'll just close this and leave it for now. I can redo it later.

PR #154, opened 2018-01-05
feat(instrumentation): support sampling

### LINKED ISSUE #148: Add simple head based transaction sampling
Currently the agent sends up all information about all transactions to the APM Server. In Elastic 6.2 we should allow for "head based random sampling" - i.e. where only a configurable percentage of the total transactions are sampled. The non-sampled transactions will still be sent to the APM Server, but without any spans and without any context.

The config option should be called `transactionSampleRate` and default to 100% - i.e. everything is collected.

It's preferable if we can find a way to disable the [wrapping of callbacks](https://github.com/elastic/apm-agent-nodejs/blob/826efc9f7affca4369fe0dc21070427fcb3879d6/lib/instrumentation/index.js#L110-L125) when a transaction isn't sampled. This means that there's certain things we cannot do as easily as we'll not have an active transaction available during for instance the request/response life cycle. This can mean that setting the name and result for HTTP based transactions can be difficult and might require a bit of refactoring.

Note that this feature will need to go into the [elastic-6.2 branch](https://github.com/elastic/apm-agent-nodejs/tree/elastic-6.2) as it can't be released until Elastic 6.2 version is released.

Qard:
This is a basic random sampling implementation.

Fixes #148

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -36,7 +36,8 @@ var DEFAULTS = {
   asyncHooks: true,
   ff_captureFrame: false,
   sourceContext: true,
-  transactionMaxSpans: Infinity
+  transactionMaxSpans: Infinity,
+  transactionSampleRate: 1.0
 }
 
 var ENV_TABLE = {
@@ -61,7 +62,8 @@ var ENV_TABLE = {
   asyncHooks: 'ELASTIC_APM_ASYNC_HOOKS',
   ff_captureFrame: 'ELASTIC_APM_FF_CAPTURE_FRAME',
   sourceContext: 'ELASTIC_APM_SOURCE_CONTEXT',
-  transactionMaxSpans: 'ELASTIC_APM_TRANSACTION_MAX_SPANS'
+  transactionMaxSpans: 'ELASTIC_APM_TRANSACTION_MAX_SPANS',
+  transactionSampleRate: 'ELASTIC_APM_TRANSACTION_SAMPLE_RATE'
 }
 
 var BOOL_OPTS = [
@@ -78,6 +80,12 @@ var BOOL_OPTS = [
   'sourceContext'
 ]
 
+var NUM_OPTS = [
+  'stackTraceLimit',
+  'abortedErrorThreshold',
+  'transactionSampleRate'
+]
+
 function config (opts) {
   opts = Object.assign(
     {},
@@ -88,6 +96,7 @@ function config (opts) {
   )
 
   normalizeIgnoreOptions(opts)
+  normalizeNumbers(opts)
   normalizeBools(opts)
   truncate(opts)
 
@@ -126,6 +135,12 @@ function normalizeIgnoreOptions (opts) {
     })
     delete opts.ignoreUserAgents
   }
+}
+
+function normalizeNumbers (opts) {
+  NUM_OPTS.forEach(function (key) {
+    if (key in opts) opts[key] = Number(opts[key])
+  })
 }
 
 function normalizeBools (opts) {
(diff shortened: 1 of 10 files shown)


lib/config.js
```diff
@@ -124,6 +133,12 @@ function normalizeIgnoreOptions (opts) {
   }
 }
 
+function normalizeNumbers (opts) {
```
watson: 🎉 nice addition 😃

watson:
@Qard did you look at whether it was easy to disable the wrapping of async callbacks when a transaction shouldn't be sampled?

https://github.com/elastic/apm-agent-nodejs/blob/938f47a7a2aca7dd921a0914d7b14363f0ae5605/lib/instrumentation/index.js#L110-L125

Qard:
I'm still looking into it. I think the sampling *may* cause a transaction leak into other requests.

watson:
We need to make sure that the transaction is still recorded in a very high level so that we know it happened. Otherwise the RPM numbers reported in the Kibana API will be wrong. So we still need either instantiate it or at least keep some kind of record of it, so that that it can be added to the `transactions` array in the JSON payload with a `sampled: false` property, but no spans or context property. All sampled transactions should have `sampled: true`.

Qard:
Updated. Now it can skip the bindFunction wrap for anything that occurs within an unsampled transaction. I also added the necessary components to report the `sampled` field and a test to verify it along with the absence of spans. 👍

watson:
So simple! Nice 😃 Sorry about the late review.

The un-sampled transactions being sent to the intake API should not include the `spans` or the `context` properties. We should probably also include a protocol test for this.

I have been looking into if it would be possible to not store any reference to the HTTP request or response objects on an un-sampled transaction, which is currently done here:

https://github.com/elastic/apm-agent-nodejs/blob/a3506ff59c4b01c14a5d97416ae03edb6e3494df/lib/instrumentation/http-shared.js#L20-L21

The current downside to this would be that we couldn't name the transaction properly if we didn't have access to the request. This in turn would mean that we would not be able to group the transaction on the APM Server as the name is used for grouping, which in turn would mean that the RPM numbers for the sampled groups would be wrong.

It's currently used for naming the transaction in the following places:

- When calling the `end` function on the transaction, it will use the presence of the request to set the name based on the HTTP route:
https://github.com/elastic/apm-agent-nodejs/blob/a3506ff59c4b01c14a5d97416ae03edb6e3494df/lib/instrumentation/transaction.js#L142
- When an HTTP route cannot be determined, we use the presence of the request to set a very high level name based on the HTTP method:
https://github.com/elastic/apm-agent-nodejs/blob/a3506ff59c4b01c14a5d97416ae03edb6e3494df/lib/instrumentation/transaction.js#L20-L24
- It's also used in the GraphQL instrumentation to better name GraphQL transactions:
https://github.com/elastic/apm-agent-nodejs/blob/a3506ff59c4b01c14a5d97416ae03edb6e3494df/lib/instrumentation/modules/graphql.js#L119

watson:
I did some research today on how to avoid adding any transactions to the in-memory queue, but instead queue the payload that should be sent to the APM Server. And I'm almost ready with a PR for that.

This means that having references to the HTTP request and response on the transaction object isn't really a big deal any more as we don't need to hold on to references for the transaction any longer than the http request is active.

So there's no reason to look into optimizing that any longer.

Qard:
I made the changes to omit `spans` and `context` when not sampled, and included a test for that.

watson:
@Qard Awesome! Seems like the tests doesn't like the spans and context to be missing. Probably easy to fix 😃

PR #155, opened 2018-01-05
chore(*): move process fields to namespace

### LINKED ISSUE #149: Move process related properties to own namespace in intake API JSON
Currently we sent up process related data to the APM Server in the `service` JSON object (both for errors and transactions). These are `service.pid`, `service.process_title` and `service.argv`. We should instead move these properties to their own `process` namespace:

```json
"process": {
  "pid": 1234,
  "title": "node",
  "argv": [
    "node",
    "index.js"
  ]
}
```

Related APM Server PR: elastic/apm-server#445

Qard:
The process fields pid, title and argv have been moved into
a process namespace in the message envelopes.

Fixes #149

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -210,9 +210,6 @@ function envelope (agent) {
   var payload = {
     service: {
       name: agent._conf.serviceName,
-      pid: process.pid,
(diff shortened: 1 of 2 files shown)


Qard:
fixed

Issue #106, opened 2017-12-15
Add full url to context.request.url in APM Server payload


watson:
We should find a way to assemble a full URL (https://example.com:8080/path?query=params) from a combination of `req.url`, http headers and socket information.

The full url should be added to the APM Server payload as the property `context.request.url.full`.

If a full URL cannot be assembled because we for instance are missing the domain, the `full` property should not be set.

This is where we currently process the URL from requests:

https://github.com/elastic/apm-agent-nodejs/blob/66f54d115f015412fce7d7c9a0c7de68d45acbec/lib/parsers.js#L88-L132

Qard:
PR is: #166

PR #166, opened 2018-01-09
feat(*): add full to url

Qard:
Fixes: #106

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -1,11 +1,11 @@
 'use strict'
 
-var parseUrl = require('url').parse
 var util = require('util')
 var stringify = require('fast-safe-stringify')
 var objectAssign = require('object-assign')
 var afterAll = require('after-all-results')
 var httpHeaders = require('http-headers')
+var getUrlFromRequest = require('original-url')
 var debug = require('debug')('elastic-apm')
 var stackman = require('./stackman')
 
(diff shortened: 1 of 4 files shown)


watson:
@beniwohli you mentioned at some point something about some HTTP headers that were added by certain proxies, load balancers etc. that would hold information about the original URL or host. Do you remember what those were?

beniwohli:
@watson can't think of any. I did mention that WSGI has environment variables that contain certain information about the host/path: https://www.python.org/dev/peps/pep-0333/#environ-variables

AdriVanHoudt:
Google cloud load balancer adds headers like `X-Forwarded-For` and `X-Forwarded-Proto`

beniwohli:
`X-Forwarded-For` contains the original IP of of the caller (or a list of IPs) AFAIK, so I don't think it'll help in this case

AdriVanHoudt:
@beniwohli hmm true, the `X-Forwarded-Proto` contains the original protocol (http/https) so that would be useful
@watson looking forward to that one

watson:
Finished the original-url module now: https://github.com/watson/original-url

@Qard let me know what you think

Qard:
@watson Are we intending to drop the `raw` field? It seems to not be included in the output of original-url.

watson:
@Qard We'll keep the `raw` field. The only reason why I didn't include it in the returned object from the "original-url" module was because I thought that it was too specific to our use-case and because it's easy to add manually to the returned objected afterwards. But as we talked about in other channels, let's just add it to "original-url" anyway.

watson:
I've released v1.1.0 of original-url which includes your PR adding raw to the result

Qard:
updated.

Qard:
rebased.

PR #169, opened 2018-01-09
chore(ci): remove Node.js 5 and 7 from test matrix

### LINKED ISSUE #158: Drop official support for Node.js v5 and v7?
There have long been an issue with the latest version of Node.js 7 (v7.10.x) that means we cannot instrument it properly as all `net.connect` calls uses an internal function call that cannot be patched. This issue [was fixed](https://github.com/nodejs/node/pull/12861) in 8.x, but the fix was never back-ported to the 7.x branch. So we only support up to version 7.9.0 on the 7.x branch.

Currently we have pinned our CI servers to test against Node.js version 7.9.x instead of 7.x, but that's not really a realistic scenario as people who still run version 7 would most likely run the latest version.

So maybe we should consider completely dropping support for version 7 (and 5 while we're at it) so that we don't waste time testing these versions on CI?

For reference, here's the official end-of-life dates for all the major Node.js versions:

| Release | Initial<br>Release | LTS<br>Start | Maintenance<br>Start | End-of-Life | EoL<br>Status |
|---------|-------------------|---------------|------------------------|------------|-----------------|
| v0.10 | 2013-03-11 | | | 2016-10-31 |  #f03c15 
| v0.12 | 2015-02-06 | | | 2016-12-31 |  #f03c15 
| v4 | 2015-09-08 | 2015-10-12 | 2017-04-01 | 2018-04-30 | #c5f015
| v5 | 2015-10-29 | | 2016-04-30 | 2016-06-30 | #f03c15
| v6 | 2016-04-26 | 2016-10-18 | 2018-04-30 | 2019-04-01 | #c5f015
| v7 | 2016-10-25 | | 2017-04-30 | 2017-06-30 | #f03c15
| v8 | 2017-05-30 | 2017-10-31 | Apr 2019 | Dec 2019 | #c5f015
| v9 | 2017-10-01 | | Apr 2018 | Jun 2018 | #c5f015
| v10 | Apr 2018 | Oct 2018 | Apr 2020 | Apr 2021 | #c5f015

Qard:
Fixes #158


Qard:
updated.

PR #170, opened 2018-01-09
feat(transaction): add transactionMaxSpans config

### LINKED ISSUE #163: Add config option to set a max number of spans a transaction can have
By default all spans that happen during a transaction should be recorded and sent to the APM Server. But the user should be able to configure the agent with a max number of spans that an individual transaction can have. If that number if reached, any subsequent spans should be dropped, but the transaction should still be sent to the APM Server (just without the dropped spans).

We should keep track of how many spans are dropped per transaction and add the count using the following JSON to the transaction in the APM Server intake API payload:

```json
"span_count": {
  "dropped": {
    "total": <number-of-dropped-spans>
  }
}
```

The config option should be called `transactionMaxSpans`.

Target branch: `master`

Related PR on APM Server: elastic/apm-server#448

### To do

- [ ] Implement code
- [ ] Add tests
- [ ] Update docs

Qard:
Fixes #163

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -344,6 +344,17 @@ Setting it to `0` will disable stack trace collection.
 Any finite integer value will be used as the maximum number of frames to collect.
 Setting it to `Infinity` means that all frames will be collected.
 
+[[transaction-max-spans]]
+===== `transactionMaxSpans`
+
+* *Type:* Number
+* *Default:* `Infinity`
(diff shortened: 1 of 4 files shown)


watson:
It isn't strictly necessary for this to go into the elastic-6.2 branch as it's backwards compatible with APM Server v6.1 (with the only caveat that the number of dropped spans will not be stored in ES as that field is just silently ignored by the server as it doesn't know about it), but I'll merge it anyway as it's no big deal.

PR #171, opened 2018-01-11
Set default stack trace limit to 50 frames

### LINKED ISSUE #110: Fix stack trace limit mismatch and lower the default limit
Today there's a bit confusion about our stack trace limit (i.e. max number of frames in stack traces we send up to the APM Server).

On one hand we have the [`stackTraceLimit` config option](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#stack-trace-limit), which defaults to `Infinity`.

On the other hand we manually truncate after 300 frames:

https://github.com/elastic/apm-agent-nodejs/blob/66f54d115f015412fce7d7c9a0c7de68d45acbec/lib/request.js#L167-L170

This needs to be aligned. I imagine we don't even need to manually truncate if we configure V8 to limit the number of frames for us based on the `stackTraceLimit` option (which I believe we currently do).

But it would make sense to lower the default limit from `Infinity` to something more reasonable. Let's set it to `100`.

When fixing this issue, remember to update the `stackTraceLimit` documentation with the new default limit.

watson:
Fix #110 

BREAKING CHANGE: This changes the default stack trace limit. If the user doesn't already overwrite the `stackTraceLimit` config option, this commit will affect their setup.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -337,7 +337,7 @@ Set this option to `false` to disable collecting and reporting source code conte
 ===== `stackTraceLimit`
 
 * *Type:* Number
-* *Default:* `Infinity`
+* *Default:* `50`
 * *Env:* `ELASTIC_APM_STACK_TRACE_LIMIT`
 
 Setting it to `0` will disable stack trace collection.
--- a/lib/config.js
+++ b/lib/config.js
@@ -26,7 +26,7 @@ var DEFAULTS = {
   active: true,
   logLevel: 'info',
   hostname: os.hostname(),
-  stackTraceLimit: Infinity,
+  stackTraceLimit: 50,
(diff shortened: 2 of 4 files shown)


watson:
@Qard what do you think? I just pulled the limit 100 out of a hat. Do you think that's reasonable?

Qard:
I think I'd go somewhat lower, personally. I'm not really sure what the ideal value would be though. At 100 frames though, I'd assume most of those would have to be recursion problem repeats.

watson:
@Qard The default in Node.js is 10 which I think is too low in many cases. 42?

Qard:
Yeah, 10 is too low. Conservatively, 25 might be enough. I definitely wouldn't think it'd need to be more than 50.

watson:
Updated it to 50

PR #181, opened 2018-01-14
feat(errors): add captureError call location stack trace

### LINKED ISSUE #112: Rename ff_captureFrame config option and default it to true
It should no longer be considered behind a feature flag (ff) and should therefore default to being on (`true`). It also used to only add an extra top frame to the existing stack trace, but now it adds a full separate stack trace to the error (currently under the `log` namespace), so the name doesn't make much sense any more either.

What would a good new name be? `captureLocationStackTrace` is a bit wordy and I don't really like it for several reasons, but I can't come up with anything better either

(remember to document this new config option in the docs when adding it).

### LINKED ISSUE #160: Capture stack trace when capturing strings and param objects
If calling `captureError` with either a string a the special parameterized object, the message will be stored as `log.message`, but a stack trace will not be captured along with it. This should be added to `log.stacktrace` just at it happens with `exception.stacktrace` for regular errors.

### Gotcha

As we're not dealing with an actual error object, we'll have to capture the stack trace our selfs. When we do we should pop the top stack frame(s) (as those will point to the agent), so that the new top stack frame is the place from where `captureError` was called.

watson:
~~Depends on #178 + #180~~

Adds `captureErrorLogStackTraces` config option.

Normally only `Error` objects have a stack trace associated with them. This stack trace is stored along with the error message when the error is sent to the APM Server. The stack trace points to the place where the `Error` object was instantiated.

But sometimes its valuable to know, not where the `Error` was instantiated, but where it was detected. For instance, when an error happens deep within a database driver, the location where the error bubbles up to, is sometimes more useful for debugging, than where the error occurred.

Set the `captureErrorLogStackTraces` config option to `always` to &mdash; besides the error stack trace &mdash; also capture a stack trace at the location where `captureError` was called.

By default, the `captureErrorLogStackTraces` config option have the value `messages`, which means that a stack trace of the capture location will be recorded only when `captureError` is called with either a string or the special parameterized error object, in which case a normal stack trace isn't available.

Set the `captureErrorLogStackTraces` config option to `never` to never record a capture location stack trace.

A capture location stack trace is never generated for uncaught exceptions.

Closes #160, closes #112

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -315,12 +315,12 @@ Note that if a custom logger is provided, the `logLevel` option will be ignored.
 
 Whether or not the agent should monitor for uncaught exceptions and send them to the APM Server automatically.
 
-[[capture-location-stack-traces]]
-===== `captureLocationStackTraces`
+[[capture-error-log-stack-traces]]
+===== `captureErrorLogStackTraces`
 
 * *Type:* String
-* *Default:* `non-errors`
-* *Env:* `ELASTIC_APM_CAPTURE_LOCATION_STACK_TRACES`
+* *Default:* `messages`
+* *Env:* `ELASTIC_APM_CAPTURE_ERROR_LOG_STACK_TRACES`
 
 Normally only `Error` objects have a stack trace associated with them.
 This stack trace is stored along with the error message when the error is sent to the APM Server.
@@ -335,16 +335,16 @@ the location where the error bubbles up to,
 is sometimes more useful for debugging,
 than where the error occurred.
 
-Set this config option to `all` to --
+Set this config option to `always` to --
 besides the error stack trace --
 also capture a stack trace at the location where <<apm-capture-error,`captureError`>> was called.
 
 By default,
-this config option has the value `non-errors`,
-which means that a stack trace of the capture location will be recorded only when `captureError` is called with either a <<message-strings,string>> or the <<parameterized-error-object,special parameterized error object>>,
+this config option has the value `messages`,
+which means that a stack trace of the capture location will be recorded only when `captureError` is called with either a <<message-strings,string>> or the <<parameterized-message-object,special parameterized message object>>,
 in which case a normal stack trace isn't available.
 
-Set this config option to `off` to never record a capture location stack trace.
+Set this config option to `never` to never record a capture location stack trace.
 
 A capture location stack trace is never generated for uncaught exceptions.
 
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -340,7 +340,7 @@ besides the error stack trace --
 also capture a stack trace at the location where <<apm-capture-error,`captureError`>> was called.
 
 By default,
-this config option have the value `1`,
+this config option has the value `1`,
 which means that a stack trace of the capture location will be recorded only when `captureError` is called with either a <<message-strings,string>> or the <<parameterized-error-object,special parameterized error object>>,
 in which case a normal stack trace isn't available.
 
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -315,6 +315,39 @@ Note that if a custom logger is provided, the `logLevel` option will be ignored.
 
 Whether or not the agent should monitor for uncaught exceptions and send them to the APM Server automatically.
 
+[[capture-location-stack-traces]]
+===== `captureLocationStackTraces`
+
+* *Type:* Number
+* *Default:* `1`
+* *Env:* `ELASTIC_APM_CAPTURE_LOCATION_STACK_TRACES`
+
+Normally only `Error` objects have a stack trace associated with them.
+This stack trace is stored along with the error message when the error is sent to the APM Server.
+The stack trace points to the place where the `Error` object was instantiated.
+
+But sometimes its valuable to know,
+not where the `Error` was instantiated,
+but where it was detected.
+For instance,
+when an error happens deep within a database driver,
+the location where the error bubbles up to,
+is sometimes more useful for debugging,
+than where the error occurred.
+
+Set this config option to `2` to --
+besides the error stack trace --
+also capture a stack trace at the location where <<apm-capture-error,`captureError`>> was called.
+
+By default,
+this config option have the value `1`,
+which means that a stack trace of the capture location will be recorded only when `captureError` is called with either a <<message-strings,string>> or the <<parameterized-error-object,special parameterized error object>>,
+in which case a normal stack trace isn't available.
+
+Set this config option to `0` to never record a capture location stack trace.
+
+A capture location stack trace is never generated for uncaught exceptions.
+
 [[capture-trace-stack-traces]]
 ===== `captureTraceStackTraces`
 
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -318,8 +318,8 @@ Whether or not the agent should monitor for uncaught exceptions and send them to
 [[capture-location-stack-traces]]
 ===== `captureLocationStackTraces`
 
-* *Type:* Number
-* *Default:* `1`
+* *Type:* String
+* *Default:* `non-errors`
 * *Env:* `ELASTIC_APM_CAPTURE_LOCATION_STACK_TRACES`
 
 Normally only `Error` objects have a stack trace associated with them.
@@ -335,16 +335,16 @@ the location where the error bubbles up to,
 is sometimes more useful for debugging,
 than where the error occurred.
 
-Set this config option to `2` to --
+Set this config option to `all` to --
 besides the error stack trace --
 also capture a stack trace at the location where <<apm-capture-error,`captureError`>> was called.
 
 By default,
-this config option has the value `1`,
+this config option has the value `non-errors`,
 which means that a stack trace of the capture location will be recorded only when `captureError` is called with either a <<message-strings,string>> or the <<parameterized-error-object,special parameterized error object>>,
 in which case a normal stack trace isn't available.
 
-Set this config option to `0` to never record a capture location stack trace.
+Set this config option to `off` to never record a capture location stack trace.
 
 A capture location stack trace is never generated for uncaught exceptions.
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -141,8 +141,8 @@ Agent.prototype.captureError = function (err, opts, cb) {
   var _isError = isError(err)
 
   if (!(opts && opts.uncaught) &&
-      (agent._conf.captureLocationStackTraces === config.CAPTURE_LOCATION_STACK_TRACE_ALL ||
-       (!_isError && agent._conf.captureLocationStackTraces === config.CAPTURE_LOCATION_STACK_TRACE_NON_ERRORS))
+      (agent._conf.captureErrorLogStackTraces === config.CAPTURE_ERROR_LOG_STACK_TRACES_ALWAYS ||
+       (!_isError && agent._conf.captureErrorLogStackTraces === config.CAPTURE_ERROR_LOG_STACK_TRACES_MESSAGES))
      ) {
     var captureLocation = {}
     Error.captureStackTrace(captureLocation, Agent.prototype.captureError)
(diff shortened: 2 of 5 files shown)


watson:
With this PR, when capturing a regular error while having `captureLocationStackTraces: 2`, the agent will add an extra stack trace to `log.stacktrace`, but since there's no message, `log.message` will not be populated. Turns out the APM Server doesn't like that 😬 

```
I[#] S[#] doesn't validate with "error#"
  I[#/errors/0/log] S[#/properties/errors/items/properties/log/required] missing properties: "message"
```

@elastic/apm-server @elastic/apm-ui any way we can make `log.message` optional to support this use-case?

simitt:
The apm-server currently defines that either `log` or `exception` has to be provided, and in both cases `message` is a required field. 
The message is used for calculating the `grouping_key` if nothing else is provided, thus making this field optional could lead to empty grouping keys.

watson:
@simitt the `exception.message` is present. It's only the `log.message` that's not there

simitt:
So it would work in this case, but not generally, except the logic is changed to require a `log.message` only if no `exception` is given, which might be a bit tricky to define with json spec.

watson:
@simitt Hmmm yeah, true. I wonder how I could work around this somehow. I could just set a fake `log.message` - e.g. set et equal to `exception.message`, but that looks kind of weird in the UI unfortunately 😞  Do you know if the grouping key will be based on both `log.message` **and** `exception.message` if they are both there?

simitt:
If the grouping key is empty after checking for several other fields, the `error.exception.message` is used, if this is also empty `error.log.message` is used, so only one of them is used for the key. 
[Calculation of grouping key](https://github.com/elastic/apm-server/blob/master/processor/error/event.go#L193)

watson:
Ok thanks, so I shouldn't set `log.message` to `n/a` or something. That would just group everything together then.

simitt:
I think we have a slight misunderstanding for this special case. 
The `error.log.message` is only used if no `error.exception.message` (therefore no `error.exception`) is given. So in this case you are describing `log.message` would not be used for grouping.

However, I generally agree that using some default value could turn out grouping together unrelated errors.

watson:
After talking to @simitt, I've updated the PR to set the `log.message` to `n/a` in case we don't have one. An alternative approach would be to set it to the value of `exception.message`, but that might not look that good in the UI.

This is more of a hack than a good solution, but I doubt people are in favor up changing the API to support this use-case. At least not in 6.2 🙈

watson:
cc @elastic/apm-ui

Qard:
The docs seem to imply setting the config to 2 would result in an actual error object sending both it's own stack trace *and* the location stack trace, but it seems like only the error stack trace would be sent in that case. 🤔

watson:
@Qard it should add the `log.stacktrace` - where does it seem like it doesn't? There's a passing test for it here: https://github.com/elastic/apm-agent-nodejs/pull/181/files#diff-5f9957336859d2050e3d5e347943e142R283

Qard:
Oh, ignore that. Just misread the diff.

docs/agent-api.asciidoc
```diff
@@ -315,6 +315,39 @@ Note that if a custom logger is provided, the `logLevel` option will be ignored.
...
+the location where the error bubbles up to,
+is sometimes more useful for debugging,
+than where the error occurred.
+
+Set this config option to `2` to --
+besides the error stack trace --
+also capture a stack trace at the location where <<apm-capture-error,`captureError`>> was called.
+
+By default,
+this config option have the value `1`,
```
simitt: `s/have/has/`

lib/agent.js
```diff
@@ -190,21 +193,23 @@ Agent.prototype.captureError = function (err, opts, cb) {
...
         var next = afterAll(function (_, frames) {
           // As of now, parseCallsite suppresses errors internally, but even if
           // they were passed on, we would want to suppress them here anyway
 
           if (frames) {
-            frames.shift() // ignore the first frame as it will be this module
+            // In case there isn't any log object, we'll make a dummy message
+            // as the APM Server requires a message to be present if a
+            // stacktrace also present
+            if (!error.log) error.log = {message: 'n/a'}
```
simitt: As mentioned earlier, this can have undesired side effects in grouping together unrelated errors, if only the `error.log.message` is available for grouping.

watson: I'm only setting the dummy message if there is no message to begin with. So it's not going to make the grouping worse. The only time that this occurs is if we're about to send an exception to the APM Server, in which case there will be an `exception.message`, but no `log.message`. That's the only time I'm forced to set `log.message` to a dummy value as the API doesn't allow it to be empty.

docs/agent-api.asciidoc
watson: My nemesis 🤦‍♂️

nemesis

lib/agent.js
watson: If you have an idea for a better value, or a different approach all together, I'd be happy to hear it though 😃

simitt: I don't know if it might make sense to use some information from the first stack trace frame, e.g. the filename or the function?

Generally I am fine with your solution, just wanted to point out what happens server side with it.

watson: That's a really interesting approach! It could be similar to how the stack frame looks when printed, e.g.:

- `caught at /Users/watson/code/elastic/apm-dev/foo.js:7:9`
- `caught at user (/Users/watson/code/elastic/apm-dev/foo.js:2:9)`

watson: @elastic/apm-ui would you be ok with this?

docs/agent-api.asciidoc
```diff
@@ -315,6 +315,39 @@ Note that if a custom logger is provided, the `logLevel` option will be ignored.
 
 Whether or not the agent should monitor for uncaught exceptions and send them to the APM Server automatically.
 
+[[capture-location-stack-traces]]
+===== `captureLocationStackTraces`
+
+* *Type:* Number
+* *Default:* `1`
```
roncohen: Just an idea, but you have constants like `CAPTURE_LOCATION_STACK_TRACE_ALL` and `CAPTURE_LOCATION_STACK_TRACE_NON_ERRORS`. Instead of using a magic number here, perhaps we should encourage people to use those constants? 

For environment variables, using strings are an option too.

lib/agent.js
watson: After discussing this with the @elastic/apm-ui team, we came to the conclusion that showing either `n/a` or `caught at ...` would not look great on the errors list page as `log.message` would be prioritized and shown instead of `exception.message` if present.

This leaves us with only one option: Copy the `exception.message` into `log.message`. This however unfortunately means that the error specific page will show the message twice - one time for `log` and another time for `exception`. But it's the best solution we can implement given our constraints.

We need to bring this up again for the next version of the APM Server, to ensure this use-case is supported in the intake API so that we don't have to make up a "fake" `log.message`. cc @elastic/apm-server

roncohen:
@watson what does the general case look like when setting `captureLocationStackTraces=2` and the user is using express or similar? Will our express middleware generally catch different exceptions in the same way and thus they will all have the same stack trace?

docs/agent-api.asciidoc
watson: Yeah, I was thinking of this, but came to the conclusion that since these wasn't available when setting environment variables, we had to reveal the numbers anyway, so we couldn't hide them behind constants.

Using strings in the environment variables is an interesting approach that I hadn't thought of. Right now though, we don't have a way in our config logic to handle the case where the environment variables should be translated from strings to numbers when [initially parsed](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/config.js).

So it would make that logic more complicated, so I wonder if it's worth it?

watson:
@roncohen Yea, errors caught by the express middleware are all caught in the same way and are all considered "handled". As such they will all have a capture location stack trace if `captureLocationStackTraces=2`.

I see a few "solutions" to this:

1. Maybe that's fine. Nothing to change
1. We should  consider if errors caught by our Express middleware should be marked as unhandled if that's actually more appropriate? That would mean that they wouldn't get a `log.stacktrace` in that situation

lib/agent.js
roncohen: I'm good with copying the `exception.message` to `log.message` in this case 👍

docs/agent-api.asciidoc
watson: Good point. I'll change it to use strings instead

beniwohli: I have string values for similar settings in the Python agent, e.g.: https://www.elastic.co/guide/en/apm/agent/python/2.x/configuration.html#config-collect-local-variables

Maybe makes sense to align?

watson: @beniwohli Good idea, though looking over it I don't see much we can re-use. I was thinking of going with `off`, `non-errors` and `all` (maybe `log` instead of `non-errors`... can't really decide).

Yours are for a slightly different use-case. Only overlap I see is `all` and potentially `none` instead of `off`, though I think `off` better conveys its meaning in this use-case.

watson: @roncohen would you still suggest that the users should use the constants on the config module or should they just type the strings manually? If it's the former, we have to expose them in a nicer way. Right now they have to do something like this:

```js
var apmConfig = require('elastic-apm-node/lib/config')
var apm = require('elastic-apm-node').start({
  captureLocationStackTraces: apmConfig.CAPTURE_LOCATION_STACK_TRACE_ALL
})
```

docs/agent-api.asciidoc
watson: Fixed

lib/agent.js
watson: Fixed

docs/agent-api.asciidoc
watson: Updated: No uses the possible values of `off`, `non-errors` and `all`. Question still remains if `non-errors` is a good value and if we should expose the config constants

watson:
@roncohen I'll mark errors caught Express middleware as unhandled in another PR

lib/config.js
```diff
@@ -7,6 +7,9 @@ var bool = require('normalize-bool')
 var trunc = require('unicode-byte-truncate')
 
 config.INTAKE_STRING_MAX_SIZE = 1024
+config.CAPTURE_LOCATION_STACK_TRACE_OFF = 'off'
+config.CAPTURE_LOCATION_STACK_TRACE_NON_ERRORS = 'non-errors'
```
watson: @roncohen what do you feel about this value? I don't like that it consists of two words and contains a hyphen. There's a high likelihood that people will mistype it. Though right now it's the default value, which means people probably don't need to type it 🤔

The only alternative I've been able to come up with is `log`.

roncohen: `non-errors` is a bit ambiguous. I thought it was for `spans`. Can we come up with a better name? I don't think the dash is a problem generally.

docs/agent-api.asciidoc
roncohen: using the strings directly is probably fine - people will have to do that anyway if they use the env vars.  However, it would be great to check at configuration time/start-up that the values are set correctly to avoid spelling errors.

lib/config.js
watson: `log.*` and `exception.*` is only allowed on the `/v1/errors` endpoint of the APM Server. This PR is about what happens when you call `agent.captureError()` and is not related to any performance instrumentation

roncohen: > This PR is about what happens when you call agent.captureError() and is not related to any performance instrumentation

@watson I understand. I'm saying that from the choice of wording for `CAPTURE_LOCATION_STACK_TRACE_NON_ERRORS` and `non-errors` its not clear that this is unrelated to performance instrumentation.

watson: It makes sense in context of the `captureError` function as you can pass it both errors and things that are not errors. But I don't like it either. Would like a better name if you have any suggestions?

Qard: Something like `always`, `never` and `when-missing-location` maybe?

watson: After talking with @roncohen we came to the conclusion that the config option should be renamed `captureErrorLogStackTraces` and the options should be `never`, `messages` and `always`

watson:
Updated PR based on all your feedback 👍 Should hopefully be ready now

Issue #182, opened 2018-01-15
Ensure request/response objects can be garbage collected after HTTP request ends


watson:
Currently [transaction objects](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/transaction.js) are added to the queue [when they end](https://github.com/elastic/apm-agent-nodejs/blob/9737fc85117ddb017c4cf263d5734e47ff627ed1/lib/instrumentation/index.js#L74). When the queue is flushed, all queued transaction objects [are encoded](https://github.com/elastic/apm-agent-nodejs/blob/9737fc85117ddb017c4cf263d5734e47ff627ed1/lib/instrumentation/index.js#L35-L41) for transport.

Unfortunately all HTTP transactions [hold references](https://github.com/elastic/apm-agent-nodejs/blob/9737fc85117ddb017c4cf263d5734e47ff627ed1/lib/instrumentation/http-shared.js#L20-L21) to both the HTTP request and response objects, which in turn means that these objects are not garbage collected until the queue is flushed. Worst case this might mean that they are moved into the old heap space, which is generally bad for memory management.

The simplest solution to this problem would be to generate the transaction payload that in the end will be sent to the APM Server already when the transaction ends. So instead of adding the transaction to the queue, we add the resulting payload instead.

This means that we hopefully should lose all references to the transaction object it self and through that any reference to the request and response objects.

The [`currentTransaction`](https://github.com/elastic/apm-agent-nodejs/blob/9737fc85117ddb017c4cf263d5734e47ff627ed1/lib/instrumentation/transaction.js#L64) property might still hold a reference to the transaction though &mdash; in case any patched or async callbacks haven't been called yet &mdash; which further means that we'll still not be able to garbage collect the objects.

To get around this last problem, we could simply set the `req` and `res` properties on the transaction to `null` once we've added its payload to the queue. This have the downside that any error that occurs after the request is finished will not have any request information associated with it.

There's two solutions to that:

1. Either we don't care as users can just look up the associated transaction as it will hold all the request and response information
1. Or we store a copy of the transaction payload on the transaction object it self, so any errors that occur later can copy it

PR #184, opened 2018-01-15
Encode transactions as they are added to the queue

watson:
This is an attempt to solve #182. It doesn't try to remove the references to the request and response objects on the transaction once the request has ended, but it's assumed that `currentTransaction` will not reference the transaction for much longer after the request have ended, in which case it's almost as effective.


lib/instrumentation/index.js
```diff
@@ -70,8 +68,20 @@ Instrumentation.prototype.start = function () {
 
 Instrumentation.prototype.addEndedTransaction = function (transaction) {
   if (this._started) {
+    var queue = this._queue
     debug('adding transaction to queue %o', {id: transaction.id})
-    this._queue.add(transaction)
+    // encode the transaction as early as possible in an attempt to free up
+    // objects for garbage collection
+    queue.add(new Promise(function (resolve, reject) {
```
watson: We haven't used promises anywhere else in the agent. But I had some race conditions in the tests when the transaction was added to the queue asynchronously when calling `addEndedTransaction`. I felt it was easier to reason about if the queue is bumped synchronously instead of waiting till the `transaction._encode` callback is called.

watson:
jenkins test this please

PR #185, opened 2018-01-15
 chore(elasticsearch): Only include db context for _search queries

### LINKED ISSUE #183: Ensure only Elasticsearch _search requests have their body captured
When a request to Elasticsearch is detected, a db trace is created and the request body is stored as the query on the trace:

https://github.com/elastic/apm-agent-nodejs/blob/9737fc85117ddb017c4cf263d5734e47ff627ed1/lib/instrumentation/modules/elasticsearch.js#L25

We need to ensure that the query is only added to the trace if we're dealing with a `_search` request. Creation of indexes etc. should still result in a trace, but their body should not be captured on the trace.

Qard:
Fixes #183

### DIFF
--- a/lib/instrumentation/modules/elasticsearch.js
+++ b/lib/instrumentation/modules/elasticsearch.js
@@ -3,6 +3,8 @@
 var debug = require('debug')('elastic-apm')
 var shimmer = require('../shimmer')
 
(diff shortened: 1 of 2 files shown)


watson:

Qard:
rebased

PR #186, opened 2018-01-16
chore(config): update flushInterval to 10 seconds

### LINKED ISSUE #175: Use lower default value for flushInterval config option
The [`flushInterval`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#flush-interval) config option currently defaults to 60 seconds. This is too long and we should decide on a better default.

Qard:
Fixes #175

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -370,7 +370,7 @@ Setting to `Infinity` means that spans will never be dropped.
 ===== `flushInterval`
 
 * *Type:* Number
-* *Default:* `60`
+* *Default:* `10`
 * *Env:* `ELASTIC_APM_FLUSH_INTERVAL`
 
 The agent maintains an in-memory queue to which recorded transactions are added when they end.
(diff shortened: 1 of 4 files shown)


watson:
The tests fails here:

https://github.com/elastic/apm-agent-nodejs/blob/d3f28c18824b4e81319e5aa406c213ee6ece3de4/test/instrumentation/modules/http/aborted-requests-disabled.js#L95

Not really sure why that is though. Doesn't seem directly related to the `flushInterval` 🤔

watson:

Qard:
Fixed. It was just a timing thing.

watson:
@Qard Travis and Jenkins had several outages last night, and so I merged this too quickly. Turns out the real course of the bug you fixed by changing a `setTimeout` to a `setImmediate` was that `flushInterval` became `NaN` as the mocked agent didn't have the property set, so it was `undefined`. I discovered it in master, amended the commit [with a fix](https://github.com/elastic/apm-agent-nodejs/commit/c3cfc9a41f188fe22e653271c5aff2555599841e#diff-728d83d6dd614acedce1051318b9f7b5R21), and force pushed.

Qard:
Ah, good catch.

PR #187, opened 2018-01-16
feat(agent): add public flush function

### LINKED ISSUE #173: Add public function to flush transaction queue
The agent maintains an in-memory queue of recorded transactions which is sent to the APM server and emptied periodically. Currently [a private function exists](https://github.com/elastic/apm-agent-nodejs/blob/a3506ff59c4b01c14a5d97416ae03edb6e3494df/lib/instrumentation/queue.js#L51-L55) which the agent uses internally to flush the queue.

It would make sense to make this function public, so that users of the agent can call it as well - e.g.  before shutdown of the process (the agent could also hook into the process shutdown procedure to automate this task, but that's a separate issue).

This would also make our test suite better as it's never a good idea to use private functions in a test, which we currently do all over the place to force a flush of the queue during testing.

Suggestions for function name:

- `agent.flush()`
- `agent.flushQueue()`
- `agent.flushTransactionQueue()` - If we later include other things than transactions in the queue, this might be a bad name

### To-do

- [ ] Implement code
- [ ] Make tests
- [ ] Update agent API documentation

Qard:
Fixes #173


docs/agent-api.asciidoc
```diff
@@ -837,3 +837,13 @@ the node process is terminated automatically when an uncaught exception has been
...
+
+[[apm-flush]]
+==== `apm.flush()`
+
+[source,js]
+----
+apm.flush()
+----
+
+Flush the transaction queue immediately.
```
watson: We should probably underline that this means that the transactions are sent to the APM Server and optionally also add a bit about how to configure it, e.g.

```asciidoc
Manually flush the in-memory transaction queue and send all the transactions to the APM Server.
The queue is otherwise flushed automatically,
controlled by the <<flush-interval,`flushInterval`>> and/or <<max-queue-size,`maxQueueSize`>> config options.
```

watson:
How about making `apm.flush()` accept a callback? In a shutdown scenario, it would be nice to know when the flush was done so that you can complete the shutdown.

I just followed the breadcrumbs and it seems simple to implement as the `sendTransactions` function in `lib/request.js` can easily be modified to accept a callback just as `sendErrors`:

https://github.com/elastic/apm-agent-nodejs/blob/d3f28c18824b4e81319e5aa406c213ee6ece3de4/lib/request.js#L49-L64

watson:

Qard:
Added a callback to `apm.flush(...)`, improved the docs and added tests.

test/instrumentation/queue.js
```diff
@@ -28,3 +28,37 @@ test('queue flush isolation', function (t) {
...
+  var flush = 0
+  var queue = new Queue({maxQueueSize: 2}, function (arr, done) {
+    t.equal(arr.length, 1)
+    t.deepEqual(arr, [1])
+    flush++
+    done()
+  })
+
+  queue.add(1)
+  queue._flush(function (err) {
```
watson: Travis seems to be having problems, so it still hasn't run the tests, but according Jenkins, standard is complaining this error object isn't handled

Qard:
fixed.

PR #190, opened 2018-01-17
feat(mongodb): add support for mongodb-core 3.x

watson:
Went through all the functions that we patch and there doesn't seem to be any breaking changes in those between v2 and v3, so we should be good to go


watson:
jenkins, run the tav tests

PR #196, opened 2018-01-19
feat(*): control amount of source context lines

### LINKED ISSUE #192: Add config options to change amount of source code collected for errors and spans
Today the Node.js agent collects 15 lines of source code for each stack frame for both errors and spans. The user should be able to have more fine-grained control over when and how many lines are collected.

For errors, the user should be able to control:

- Number of lines collected for each in-app stack frame (default should be 5)
- Number of lines collected for each library stack frame (default should be 5)

For spans, the user should be able to control:

- Number of lines collected for each in-app stack frame (default should be 5)
- Number of lines collected for each library stack frame (default should be 0)

watson:
Added the following 4 new numeric config options as a replacement for the `sourceContext` boolean:

- `sourceContextErrorAppFrames` - default: `5`
- `sourceContextErrorLibraryFrames` - default: `5`
- `sourceContextTraceAppFrames` - default: `5`
- `sourceContextTraceLibraryFrames` - default: `0`

This can be used to control how many lines of context is included for each stack frame that's sent to the APM Server.

It's possible to have different values for errors vs trace stack traces and for in-app frames vs library frames.

Setting all of these to `0` is equal to setting the old `sourceContext` to `false`.

Fixes #192

BREAKING CHANGE: Remove `sourceContext` config option.


watson:
The reason this PR is still marked as `in progress` is because we still need to discuss whether these config options makes sense like this or if we should change them slightly.

docs/agent-api.asciidoc
```diff
@@ -324,14 +324,82 @@ Whether or not the agent should monitor for uncaught exceptions and send them to
...
+===== `sourceContextErrorAppFrames` + `sourceContextErrorLibraryFrames`
 
-* *Type:* Boolean
-* *Default:* `true`
-* *Env:* `ELASTIC_APM_SOURCE_CONTEXT`
+When an error is captured by the agent,
+its stack trace is stored in Elasticsearch.
+
+By default the agent will also collect a few lines of source code around the lines for each frame in the stack trace.
+This can make it easier to determine the cause of an error as the source code related to the error is visible directly in Kibana.
```
beniwohli: I try to avoid referencing the X-Pack UI, as some users might not have installed it.

Maybe that's the wrong approach and we should remind users of the awesomeness that is the  X-Pack APM UI at every possibility. Worth discussing :)

docs/agent-api.asciidoc
```diff
@@ -324,14 +324,82 @@ Whether or not the agent should monitor for uncaught exceptions and send them to
...
+
+The agent differentiates between so called in-app frames and library frames.
+Library frames are frames belonging to Node core and code inside the applications `node_modules` folder.
+In-app frames are everything else.
+
+Use the following 2 config options to change how many lines of source code to include for the different types of stack frames:
```
beniwohli: I usually spell out small numbers in text paragraphs (http://www.chicagomanualofstyle.org/qanda/data/faq/topics/Numbers.html?page=1)

docs/agent-api.asciidoc
watson: You can see it in Discover as well, so I didn't think that mentioning Kibana is mentioning X-Pack?

docs/agent-api.asciidoc
watson: Good point

watson: Fixed 😃

docs/agent-api.asciidoc
beniwohli: In Discover you see a deeply nested JSON blob that, with a lot of imagination, can give you the source lines...

watson: No it's not ideal, but the data _is_ there and users can extract it in other fashions as well. I don't agree that this is equivalent to mentioning X-Pack (which I agree we shouldn't). But I'd love input on how to formulate this better if you have any suggestions - just couldn't find a better solution my self 😃 

Brainstorming: If the UX really is that terrible as you say for non-X-Pack users, should we consider disabling source code context completely unless you were an X-Pack user? The agent will of course not know this, so this would suggest that our default should be to set all these config options to `0` and then have a "Recommended setup for X-Pack users" document.

test/parsers.js
```diff
@@ -544,139 +544,82 @@ test('#parseError()', function (t) {
...
+      opts.pre = ['  // before 2', '  // before 1']
+      opts.line = '  var err = new Error()'
+      opts.post = ['  // after 1']
+      break
+    case 5:
+      opts.pre = ['  // before 2', '  // before 1']
+      opts.line = '  var err = new Error()'
+      opts.post = ['  // after 1', '  // after 2']
+      break
+  }
```
Qard: An alternative to doing this here would be a case factory like this:

```js
function makeCase (data) {
  var lines = data.lines
  if (lines) {
    data.line = '  var err = new Error()'
    var pre = data.pre = []
    var post = data.post = []
    var before = Math.ceil((lines - 1) / 2)
    var after = Math.floor((lines - 1) / 2)
    for (var i = before; i > 0; i--) {
      pre.push('  // before ' + i)
    }
    for (var i = 1; i <= after; i++) {
      post.push('  // after ' + i)
    }
  }
  return data
}
```

watson: I considered that, but I also wanted the tests to contain as little magic as possible (I already feel that I'm pushing the limit in this case). The more complex this logic becomes the bigger the risk that I introduce an error that in worse case gives us a false positive test.

docs/agent-api.asciidoc
dedemorton: It's hard for me to visualize how this looks but could you be slightly less specific and simply remove "as the source code related to the error is visible directly in Kibana." It sort of makes sense already that having some context is useful.

Another option is that you could show an example of the event (but then if the event structure is likely to change, you'll need to remember to update the example).

watson: @dedemorton removing the last part of the sentence is actually not a bad solution

PR #200, opened 2018-01-23
Test on Windows using AppVeyor

### LINKED ISSUE #198: Run CI on Windows as well

watson:
This PR tests the latest Node.js LTS version on Windows using the AppVeyor service.

Since AppVeyor doesn't allow parallel builds without payment, I'm only running one build, as it would otherwise take so long that we might end up waiting for the Windows tests. If we wanted to run all builds, this is how the `.appvoyer.yml` file would have to be modified:

```diff
   PGUSER: "postgres"
   PGPASSWORD: "Password12!"
   MYSQL_PASSWORD: "Password12!"
+  matrix:
+    - NODEJS_VERSION: "9"
+    - NODEJS_VERSION: "8"
+    - NODEJS_VERSION: "6"
+    - NODEJS_VERSION: "4"
+    -
+      NODEJS_VERSION: "9"
+      ELASTIC_APM_ASYNC_HOOKS: "0"
+    -
+      NODEJS_VERSION: "8"
+      ELASTIC_APM_ASYNC_HOOKS: "0"

 # Install scripts runs after repo cloning
 install:
   - ps: .\test\script\appveyor\install-redis.ps1
   - ps: .\test\script\appveyor\install-elasticsearch.ps1
-  - ps: Install-Product node 8 # Latest LTS
+  - ps: Install-Product node $env:NODEJS_VERSION
   - npm install

 # Services are started after all install scripts have run
```

I've split this PR up into multiple commits in case we ever want to switch to another service than AppVeyor. That way we can easily revert only the AppVeyor commit.

Fixes #198

### DIFF
--- a/test/script/appveyor/install-redis.ps1
+++ b/test/script/appveyor/install-redis.ps1
@@ -1,0 +1,14 @@
+Write-Host "Downloading Redis..." -ForegroundColor Cyan
+$redisRoot = "$env:SYSTEMDRIVE\Redis"
+$zipPath = "$($env:USERPROFILE)\redis-2.8.19.zip"
+(New-Object Net.WebClient).DownloadFile('https://github.com/MSOpenTech/redis/releases/download/win-2.8.19/redis-2.8.19.zip', $zipPath)
+7z x $zipPath -y -o"$redisRoot" | Out-Null
+del $zipPath
+
+Write-Host "Installing Redis as a Windows service..."
+& "$redisRoot\redis-server.exe" --service-install
+
+Write-Host "Starting Redis service..."
+& "$redisRoot\redis-server.exe" --service-start
+
+Write-Host "Redis installed" -ForegroundColor Green
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,26 @@
+environment:
+  PGUSER: "postgres"
+  PGPASSWORD: "Password12!"
+  MYSQL_PASSWORD: "Password12!"
+
+# Install scripts runs after repo cloning
+install:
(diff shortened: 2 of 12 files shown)


test/script/appveyor/install-redis.ps1
```diff
@@ -0,0 +1,14 @@
...
+$zipPath = "$($env:USERPROFILE)\redis-2.8.19.zip"
+(New-Object Net.WebClient).DownloadFile('https://github.com/MSOpenTech/redis/releases/download/win-2.8.19/redis-2.8.19.zip', $zipPath)
+7z x $zipPath -y -o"$redisRoot" | Out-Null
+del $zipPath
+
+Write-Host "Installing Redis as a Windows service..."
+& "$redisRoot\redis-server.exe" --service-install
+
+Write-Host "Starting Redis service..."
+& "$redisRoot\redis-server.exe" --service-start
```
beniwohli: why no wait-loop here, similar to ES? Because redis start time is usually quick enough?

watson: yeah, Redis starts really fast whereas ES might take 30+ seconds. In reality I don't think we need the wait loop for ES either as it takes a while to run `npm install` and boot up the other services after anyway - in which time ES probably have started. But this might be a problem if anybody ever runs this locally on a Windows box where they have already run `npm install` before.

simitt:
@watson I appreciate you mentioning me as I was involved in the jenkins test setup, but as I am lacking of experience with `appveyor` and generally setup on windows I don't feel to be the right person to review this.

PR #201, opened 2018-01-23
chore(*): only list Node versions in 'engines' field that we support

watson:
As of now the old and the new version of this field is the same, but when 10.x is released we don't want that to be covered as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "url": "git://github.com/elastic/apm-agent-nodejs.git"
   },
   "engines": {
-    "node": "4 || 6 || >= 8"
+    "node": "4 || 6 || 8 || 9"
   },
   "keywords": [
     "opbeat",


Qard:
Is there any cases where this would prevent install on 10.x? Is that intended? Last I checked, npm itself only uses engines as an advisory field, but other tools may not. I know heroku uses it to control what version of Node.js to install on a server.

watson:
It should just print out a warning if you run it on a unsupported version. I haven't heard of any scenario where it will block you from running it.

watson:
We should obviously bump this together with the other files we bump when a new version is released, e.g. `.travis.yml` etc

watson:
Btw, I'm just saw that AppVeyor is trying to run even though there's no `appveyer.yml` in the branch. I thought I had disabled this, but seems I forgot to press save. So just ignore that failure.

PR #205, opened 2018-01-24
feat(http2): instrument incoming http2 requests

Qard:
This is a work-in-progress. I need to figure out what to do about differences between http request/response objects and http2 stream objects.

Fixes elastic/apm-agent-nodejs#123


lib/instrumentation/modules/http2.js
```diff
@@ -0,0 +1,69 @@
...
+      }
+
+      return original.apply(this, arguments)
+    }
+  }
+
+  function wrapRespond (original) {
+    return function (headers) {
+      var trans = agent._instrumentation.currentTransaction
+      trans.result = 'HTTP ' + (headers[':status'] || '').toString()[0] + 'xx'
```
watson: Be careful that `trans` might be `null`

lib/instrumentation/modules/http2.js
```diff
@@ -0,0 +1,69 @@
...
+
+  return http2
+
+  // The `createServer` function will unpatch itself after patching
+  // the first server prototype it patches.
+  function wrapCreateServer (original) {
+    return function wrappedCreateServer (options, handler) {
+      var server = original.apply(this, arguments)
+      shimmer.wrap(server.constructor.prototype, 'emit', wrapEmit)
+      wrappedCreateServer.__obUnwrap()
```
watson: What's the reason for un-patching this?

watson: (Fun fact: `ob` is short for `Opbeat` - We should probably rename these at some point)

Qard: It only needs to wrap the emit method once. I could also have just used a Boolean check, but it's marginally better performance if I just restore the original function after the first run.

lib/instrumentation/modules/http2.js
Qard: Ah, yep. I'll fix that when I get more time to work on this.

lib/instrumentation/modules/http2.js
watson: Ah true. Missed that

Qard:
@watson I was able to get everything working sufficiently by imitating http request/response objects and intercepting the `stream.respond(...)` and `stream.end(...)` calls to populate some of the missing bits.

Everything *should* be working now. When you have some time, can you have a look over this and see if this approach seems reasonable to you?

lib/instrumentation/modules/http2.js
```diff
@@ -0,0 +1,86 @@
...
+          socket: stream.session.socket,
+          method: headers[':method'],
+          url: headers[':path'],
+          httpVersion: '2.0'
+        }
+
+        debug('intercepted request event call to http2.Server.prototype.emit')
+
+        var trans = agent.startTransaction()
+        trans.type = 'http2-request'
```
watson: @elastic/apm-server normally a transaction dealing with an HTTP request have the type `request`. Do HTTP2 transactions need to have a special type, for instance like this one, or should we just set the type to `request` for these as well?

lib/instrumentation/modules/http2.js
```diff
@@ -0,0 +1,86 @@
...
+    var patched = false
+    return function wrappedEmit (event, stream, headers) {
+      if (event === 'stream') {
+        if (!patched) {
+          patched = true
+          shimmer.wrap(stream.constructor.prototype, 'respond', wrapRespond)
+          shimmer.wrap(stream.constructor.prototype, 'end', wrapEnd)
+        }
+
+        // TODO: Figure out if we need to build a response object too
```
watson: @Qard Is this TODO still relevant? Seems like you build a response object below

lib/instrumentation/modules/http2.js
```diff
@@ -0,0 +1,86 @@
...
+        trans.type = 'http2-request'
+        trans.req = request
+        trans.res = {
+          statusCode: 200,
+          headersSent: false,
+          finished: false,
+          headers: null
+        }
+
+        eos(stream, function () {
```
watson: @Qard Will the stream not be reused and hence not end across multiple requests and responses, or is this an abstraction on top of the TCP stream?

lib/instrumentation/modules/http2.js
simitt: The server doesn't really care which value you set as type, this might be more interesting to discuss with @elastic/apm-ui  and to align with other agent devs.

The main indicator for the transaction to be a http2 request is the `request.http_version`, which you already set to `2.0` here.

watson:
@Qard nice work! 😃

lib/instrumentation/modules/http2.js
Qard: Each request within a connection gets its own stream.

lib/instrumentation/modules/http2.js
Qard: fixed

lib/instrumentation/modules/http2.js
Qard: I changed it to `request` to match http. We can discuss more if we want it to be something else.

Qard:
@watson Anything we can do here in terms of docs? I know we don't have a "getting started" guide for http. But maybe http/2, being a new thing, might be worth a bit more of an intro?

watson:
I imagine this will just work out of the box if using the http2 core module. In that case I don't see docs as being required, but as we haven't done this for other core modules either, I might just be blind to it. Any suggestions?

Qard:
I was mainly just thinking about the error and custom data stuff, but maybe we should actually just pull that out to separate pages?

Qard:
@elastic/apm-agent-devs Do any other agent devs have any input on this? Particularly on trying to make HTTP2 request and response objects look like HTTP1. Some things like response headers may be difficult to capture so easily in other languages, since they can be interleaved with the response body and are sent in a compressed binary format so it's difficult to capture without intercepting and storing every header write call before serialization.

mikker:
Ruby is only taking baby steps into http2 so I'm just ignoring it for now.

felixbarny:
In Java, the objects look the same. I don't think there is any special handling required. Then again, I didn't test it yet.

axw:
Go's basic HTTP/2 support is transparent, so it should just work. I'm not planning on doing anything special for streaming or server push at the moment.

Qard:
Some things that need some thought: should push-streams be separate transactions? They are meant to represent a request the client doesn't know it wants yet, so it can get sent along side another request.

Also, it seems you can't actually send the same header multiple times, but you *can* send array headers.

alvarolobato:
@watson and @Qard will work on a proposal to solve this situation.

Qard:
Rebased with changes to work with custom logger PR changes.

Qard:
@watson I've fixed push streams now. It will disable the context so they don't try to report anything. We should think about how to get proper visibility into those in the future.

PR #214, opened 2018-01-31
chore(config): replace logBody with captureBody

### LINKED ISSUE #208: Add captureBody config option (replacing logBody)
The `logBody` config option should be replaced with a `captureBody` option.

The `captureBody` config option should take a string that can have the following values: `errors`, `transactions`, `all`, and `off`.

- The default value should be `off`
- If the value is `errors`, the agent should include the HTTP request body for all errors that are sent to the APM Server
- If the value is `transactions`, the agent should include the HTTP request body for all transactions that are sent to the APM Server
- If the value is `all`, the agent should include the HTTP request body for both errors and transactions that are sent to the APM Server

Qard:
Fixes #208


test/config.js
```diff
@@ -167,3 +169,48 @@ test('valid serviceName => active', function (t) {
...
 })
+
+var captureBodyTests = [
+  { value: 'off', errors: '[REDACTED]', transactions: '[REDACTED]' },
+  { value: 'transactions', errors: '[REDACTED]', transactions: 'test' },
+  { value: 'errors', errors: 'test', transactions: '[REDACTED]' },
+  { value: 'all', errors: 'test', transactions: 'test' }
+]
+
+captureBodyTests.forEach(captureBodyTest => {
```
watson: We haven't used arrow functions before in the agent. I think we should either agree to start doing it or keep to the same style just to be consistent. Though I don't have a string opinion either way

watson: There's also the question of maybe only using it for one line functions vs for multi line functions. But again, I don't have a strong opinion at the moment 😄

Qard: Oh, didn't even realize I had used an arrow function there. I can just change it to a regular function.

Qard:
Updated. Do you think anything more needs to be said about the different options in the docs?

watson:
Yeah, I would probably explain what the different values mean. I think people not that familiar with our terminology will know what setting the option to `transactions` mean for instance. I normally just make a bullet list explaining each value

Qard:
I expanded the docs a bit. Does that look better?

watson:
Great! 😃

PR #217, opened 2018-02-01
Reflect default value in flush interval desc

danieljuhl:


watson:
@danieljuhl good catch 😃 Could I get you to split the lines again as before?

danieljuhl:
done

watson:
Thank you again for helping out 😃 The docs will be updated for `master` within the hour and for the `0.x` branch at our next release

PR #227, opened 2018-02-06
docs: improve the filtering config option docs

watson:


docs/agent-api.asciidoc
```diff
@@ -194,8 +194,7 @@ Example usage:
 require('elastic-apm-node').start({
   ignoreUrls: [
     '/ping',
-    /^\/admin\//i,
-    new RegExp('^/api/v1')
```
Qard: Is the removal of this line intentional? I'm not sure why this would be changed.

watson: Yeah, I was never happy with having two regular expressions there. It was a silly attempt on my part a long time ago to show you could add regexes in "different ways". But I just think it's confusing.

Qard: Ah, gotcha.

PR #233, opened 2018-02-07
refactor: use symbols to store values on 3rd party objects

### LINKED ISSUE #221: Rename `wrapped.__ob*` properties
We currently [set two properties](https://github.com/elastic/apm-agent-nodejs/blob/4a1234830d32180ccb00d519427465f6d366b03f/lib/instrumentation/shimmer.js#L53-L59) on functions that we wrap: `__obWrapped` and `__obUnwrap`. The `ob` stands for Opbeat, so it would make sense to rename these. While we're at it, we should consider if using symbols instead would be a better idea.

watson:
Closes #221

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -9,6 +9,7 @@ var afterAll = require('after-all-results')
 var isError = require('core-util-is').isError
 var debug = require('debug')('elastic-apm')
 var config = require('./config')
+var sym = require('./symbols')
 var logger = require('./logger')
 var Filters = require('./filters')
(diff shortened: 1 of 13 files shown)


watson:
@Qard I just realized this might conflict with the new http2 instrumentation where, if I remember correctly, you call the unwrap function right? Suggestions welcome on what to do instead 😃

watson:
@Qard I updated it to make the unwrap symbol available via the shared `lib/symbols.js` file. What do you think of that approach?

PR #236, opened 2018-02-08
fix: don't add req/res to unsampled transactions

### LINKED ISSUE #235: TypeError: Cannot set property 'request' of null
Versions:

- node: 8.9.4
- apm-agent-nodejs: 1.0.0
- elasticsearch, apm-server: 6.2

the nodejs client throws (briefly after start when first sample is apparently processed) with
```
TypeError: Cannot set property 'request' of null
    at /APPDIR/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:160:31
    at /APPDIR/node_modules/after-all-results/index.js:9:18
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickDomainCallback (internal/process/next_tick.js:218:9)
```

corresponding entry in apm-server log:
`2018-02-08T10:55:33.755Z        INFO    [request]       beater/handlers.go:150  handled request {"request_id": "c64c3d8b-e467-4f79-91ce-0af2d185a214", "response_code": 202, "method": "POST", "URL": "/v1/errors", "content_length": 1410, "remote_address": "9.8.7.6", "user_agent": "elastic-apm-node/1.0.0 elastic-apm-http-client/5.1.0"}`

content of elastic-apm-node.js:
```
module.exports = {
	active: process.env.NODE_ENV === 'production',
	serviceName: 'service-name',
	serverUrl: 'http://1.2.3.4:8200',
	flushInterval: 5,
	maxQueueSize: 40,
	stackTraceLimit: 100,
	transactionSampleRate: 0.1
}
```

gladly tell me what other information would be helpful :-)

watson:
If a transaction isn't sampled, it will not have a `context` property. The code tried to set properties on the `context` property if a request or response had been associated with the transaction. That would result in errors like this:

    TypeError: Cannot set property 'request' of null

This commit ensures that we only try to parse the associated `trans.req` and `trans.res` if the transaction is sampled.

Fixes #235

### DIFF
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -127,7 +127,9 @@ Transaction.prototype._encode = function (cb) {
       duration: self.duration(),
       timestamp: new Date(self._timer.start).toISOString(),
       result: String(self.result),
-      sampled: self.sampled
+      sampled: self.sampled,
+      context: null,
+      spans: null
     }
 
     if (self.sampled) {
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -127,9 +127,7 @@ Transaction.prototype._encode = function (cb) {
       duration: self.duration(),
       timestamp: new Date(self._timer.start).toISOString(),
       result: String(self.result),
-      sampled: self.sampled,
-      context: null,
-      spans: null
+      sampled: self.sampled
     }
 
     if (self.sampled) {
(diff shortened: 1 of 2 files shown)


lib/instrumentation/transaction.js
```diff
@@ -127,9 +127,7 @@ Transaction.prototype._encode = function (cb) {
       duration: self.duration(),
       timestamp: new Date(self._timer.start).toISOString(),
       result: String(self.result),
-      sampled: self.sampled,
-      context: null,
-      spans: null
```
watson: @Qard I hope it's ok that I removed the two `null` properties here? As far as I remember you added these at some point. Is there a performance or other reason for them being here?

Qard: Yeah, I put them there to avoid the [hidden class](https://gist.github.com/twokul/9501770) problem.

watson: I thought that didn't have any effect on anonymous objects like this one? If it does, then there's a few places we should add this as well 😃

Qard: It impacts any kind of object. It gets allocated with the initial shape and then any future additions will create a new hidden class linking back to it.

watson: Didn't know that. I wonder how it can do that. I think I'll do a little research into this. For now I've restored it to look like before

PR #238, opened 2018-02-10
feat(config): add serverTimeout

### LINKED ISSUE #210: Add serverTimeout config option
Currently the agent will never time out when making HTTP requests to the APM Server. We should set a default timeout and allow the user to overwrite it using a `serverTimeout` config option.

Qard:
Fixes #210 

This depends on https://github.com/elastic/apm-nodejs-http-client/pull/1 landing and getting released.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -534,6 +534,15 @@ the actual interval is ajusted by +/- 5% between each flush.
 This all helps to ensure that multiple servers started at the same time will not establish connections to the APM Server simultaneously.
 ====
 
+[[server-timeout]]
+===== `serverTimeout`
+
+* *Type:* Number
+* *Default:* `30`
+* *Env:* `ELASTIC_APM_SERVER_TIMEOUT`
+
+Specify the timeout in seconds when reporting transactions to APM Server.
+
 [[max-queue-size]]
 ===== `maxQueueSize`
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -45,7 +45,8 @@ var DEFAULTS = {
   flushInterval: 10,
   transactionMaxSpans: Infinity,
   transactionSampleRate: 1.0,
-  maxQueueSize: 100
+  maxQueueSize: 100,
+  serverTimeout: 30
 }
 
 var ENV_TABLE = {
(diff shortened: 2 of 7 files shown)


watson:
elastic/apm-nodejs-http-client#1 has landed in v5.2.0

Qard:
@watson I added the package.json bump. Do you think we should have a test in here, even though it's already tested in the other module?

watson:
👍 Probably a good idea to add a test... and just remembered that docs needs to be updated as well 😄

Qard:
Oh, yep. Good point.

lib/config.js
```diff
@@ -44,7 +44,8 @@ var DEFAULTS = {
   sourceLinesSpanLibraryFrames: 0,
   flushInterval: 10,
   transactionMaxSpans: Infinity,
-  transactionSampleRate: 1.0
+  transactionSampleRate: 1.0,
+  serverTimeout: Infinity
```
watson: After investigating an issue today with a socket that wasn't timed out properly by the APM Server, which resulted in the agent just opening more and more connections to the APM Server, I think it would make sense to choose a sensible default here, e.g. somewhere between 30 and 60 seconds. What do you think?

Qard: Sounds good to me. I just went with Infinity because that was the *original* behaviour. Having an actually *reasonable* timeout is definitely a good idea though.

watson: Good point regarding original behavior... while on one hand this could be considered a breaking change, as long as we don't choose a default that's too low, then you could also consider it a bug fix 😅

Qard:
@watson I set the default timeout to 30 seconds now.

watson:
Nice! Just missing docs and a simple test I think :)

Qard:
Updated the docs and added a test.

docs/agent-api.asciidoc
```diff
@@ -534,6 +534,15 @@ the actual interval is ajusted by +/- 5% between each flush.
...
 ====
 
+[[server-timeout]]
+===== `serverTimeout`
+
+* *Type:* Number
+* *Default:* `30`
+* *Env:* `ELASTIC_APM_SERVER_TIMEOUT`
+
+Specify the timeout in seconds when reporting transactions to apm-server.
```
watson: I think we at some point decided to standardize on the naming and always write it `APM Server`

PR #242, opened 2018-02-14
Fix: Context URL full property too long.

jroith:


watson:
Thanks for fixing this 😃Your patch have been released in version 1.0.3 of the agent which have just been released.

PR #268, opened 2018-03-08
test(hapi): don't fail if running incompatible node version

watson:


test/instrumentation/modules/hapi.js
```diff
@@ -7,11 +7,15 @@ var agent = require('../../..').start({
   logLevel: 'fatal'
 })
 
+var semver = require('semver')
+var pkg = require('hapi/package.json')
+
+// hapi 17+ requires Node.js 8.9.0 or higher
+if (semver.lt(process.version, '8.9.0') && semver.gte(pkg.version, '17.0.0')) process.exit()
```
Qard: Does the runner isolate the test files enough that this `process.exit()` won't terminate and skip the rest of the tests when running multiple test files?

watson: It runs each file in the test directory as a child process (if not we'd have issues with running more than one instance of the agent in the same process as well)

watson: https://github.com/elastic/apm-agent-nodejs/blob/63469f3f5f5673d670e713a4dd14db560fe507b7/test/test.js#L18-L22

PR #269, opened 2018-03-10
fix(span): Do not pass stack frames into promises 

Qard:
I replaced the code that queues and waits for promises with something that just defers the queue until after the encode. The excessive promise usage had some noticeable overhead.

This may result in transactions entering the queue out of order, but I don't think that matters.

### DIFF
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -5,6 +5,7 @@ var debug = require('debug')('elastic-apm')
 var Timer = require('./timer')
 var stackman = require('../stackman')
 var parsers = require('../parsers')
+var Value = require('async-value-promise')
 
 module.exports = Span
 
@@ -33,7 +34,9 @@ Span.prototype.start = function (name, type) {
   this.name = name || this.name || 'unnamed'
   this.type = type || this.type || 'custom'
 
-  if (!this._stackObj) this._recordStackTrace()
+  if (this._agent._conf.captureSpanStackTraces && !this._stackObj) {
+    this._recordStackTrace()
+  }
 
   this._timer = new Timer()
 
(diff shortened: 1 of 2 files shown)


watson:
I think the reason why the tests are failing is the same as I discovered when I originally moved the encoding to before the transactions were added to the queue (see background below).

Besides fixing the tests, it would be preferable to find a way to deal with the new state a transaction can be in: Ended, but not yet added to the queue. If a user calls `agent.flush()` before shutting down the process, transactions in this state would normally be lost. Solving this, might be the way to also fix the tests.

Alternatively, we move the I/O part of the encoding to flush time. This means that the first part of the encoding that happens before the transactions are added to the queue becomes synchronously. Come to think of it, I actually think this is the most elegant solution, as it doesn't add any extra state to the agent.

### Background

Before we used promises for this, when a transaction ended, it was added to the queue synchronously. When the queue needed flushing, we looped over all the transaction objects in it and called their `_encode` function to generate the simpler intake-API-ready objects that was then serialized to JSON.

The Promise got added when we decided to do the encoding of the transaction up front and add the simpler object to the queue instead. Reason being that the encoding of transactions is an async operation as it does I/O.

If we didn’t add a Promise to the queue in that case, we would have a mid state where a transaction was actually ended, but not yet added to the queue. This would make testing a lot harder as we would not be able to ask questions like, “when 10 transactions have ended, there should be 10 transactions in the queue” etc. It would also mean that we wouldn’t be able to flush the queue on process exit and be sure all transactions safely made it to the server.

lib/instrumentation/span.js
```diff
@@ -112,22 +141,10 @@ Span.prototype._encode = function (cb) {
...
-
-      if (!process.env.ELASTIC_APM_TEST) callsites = callsites.filter(filterCallsite)
-
-      var next = afterAll(done)
-
-      callsites.forEach(function (callsite) {
-        parsers.parseCallsite(callsite, false, self._agent._conf, next())
-      })
-    })
+    this._stackObj.then(
```
watson: Today `this._stackObj` will be `null` if `captureSpanStackTraces` is `false`, in which case this is safe. But as this is a pretty loose relationship, I wonder if we should add a guard to ensure we can't get to here unless `this._stackObj` is set. Just to make it more future-proof. What do you think?

watson: It might just be me that's overly cautious 😅

Qard: Sounds reasonable.

Qard: fixed

PR #270, opened 2018-03-10
fix(config): set default maxQueueSize to 100

Qard:
This will prevent overflowing the queue in high-traffic environments that have not been configured to sample requests.


Qard:
@watson Added the docs change.

PR #274, opened 2018-03-15
chore(*): sort and group require statements

Qard:
I grouped all require statements into core, external and internal buckets and sorted each group alphabetically by variable name.


watson:
Great initiative getting this all aligned 👍 

I don't think we're gaining much by aligning _all_ variable names with their respective node module names. A lot of modules today are named in a very long-winded fashion because it's hard to find good and short module names nowadays. Some even contain implementation details in their name, e.g. `wasm` or `native` or `fast` or `unicode-safe` etc.

There [is](https://github.com/mafintosh/is-my-json-valid) [a](https://github.com/mafintosh/end-of-stream) [trend](https://github.com/watson/unicode-byte-truncate) that the module owner will suggest a variable name to be used in their usage example. In many cases I've tried to just follow those suggestions.

Would you be ok with leaving most of them as is, and only change names that are currently misleading?

Qard:
Yeah, that sounds fine to me. I'd prefer to expand the few partial-word names to the complete word though, like sym and trunc should probably be symbol and truncate. It's easier to understand that way, especially for anyone who might want to contribute that doesn't natively speak English.

watson:
👍 (`normalizeBool` as you suggested might also be better)

Qard:
I changed a few things to be more concise. `unicode-byte-truncate` is now assigned to `truncate`, `require-in-the-middle` has been restored to `hook` and `require-ancestors` has been restored to `ancestors`.

watson:
There's a linter error:

```
/home/travis/build/elastic/apm-agent-nodejs/lib/config.js:170:10: 'truncate' is already defined.
```

Qard:
@watson Fixed. 👍

lib/instrumentation/transaction.js
```diff
@@ -1,16 +1,17 @@
...
 var afterAll = require('after-all-results')
-var trunc = require('unicode-byte-truncate')
-var express = require('./express-utils')
 var debug = require('debug')('elastic-apm')
-var logger = require('../logger')
+var truncate = require('unicode-byte-truncate')
+var uuid = require('uuid')
+
 var config = require('../config')
+var getPathFromRequest = require('../express-utils').getPathFromRequest
```
watson: This should just be `./express-utils`

Qard: fixed.

PR #278, opened 2018-03-17
test(tav): test latest combos of express-graphql and graphql

watson:


.tav.yml
```diff
@@ -41,26 +41,57 @@ graphql:
...
+  name: express-graphql
+  peerDependencies: graphql@^0.12.0
+  versions: '>=0.6.12'
+  commands: node test/instrumentation/modules/express-graphql.js
+express-graphql-0.6.12_13:
+  name: express-graphql
+  peerDependencies: graphql@^0.13.0
+  versions: '>=0.6.12'
   commands: node test/instrumentation/modules/express-graphql.js
+
```
Qard: Needing to manually describe all this multi-dimensional testing for each patch release is not ideal. We should probably see if we can figure out a way to express multi-dimensionality better in tav.

watson: yeah, a matrix config option would be better. Haven't had time to build that yet and the close relationship between `graphql` and `express-graphql` is the only place we have this need, so it hasn't been a big issue yet. I used to actually just test against the latest version of graphql that the given version express-graphql supported.

PR #284, opened 2018-03-20
refactor: don't use promises in the queue

watson:
The overhead of using the Promise API is too great. This reduces memory allocations and is easier on GC.

I took a stab at using your new module @Qard 😄 

As it doesn't have an `.all` function, I've implemented that manually. But it might make sense to build that into the async-value-promise module instead. If so I'll just update this PR.

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -2,6 +2,8 @@
 
 var fs = require('fs')
 var path = require('path')
+var afterAll = require('after-all-results')
+var AsyncValuePromise = require('async-value-promise')
 var semver = require('semver')
 var hook = require('require-in-the-middle')
 var Transaction = require('./transaction')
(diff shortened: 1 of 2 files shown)


Qard:
I'll just add that functionality to `async-value-promise`. Seems like a good idea to be closer to feature parity with native promises.

Qard:
Bump to v1.1.0 and there's an `all` method now. 👍

lib/instrumentation/index.js
```diff
@@ -69,19 +70,25 @@ Instrumentation.prototype.start = function () {
 Instrumentation.prototype.addEndedTransaction = function (transaction) {
   if (this._started) {
     var queue = this._queue
+
     debug('adding transaction to queue %o', {id: transaction.id})
+
+    var payload = new AsyncValuePromise()
+
+    payload.then(null, function (err) {
```
Qard: I also added the `catch` method in 1.1.0, by the way. ;)

watson: Refactored in #288

PR #294, opened 2018-04-10
docs: fix addFilter example in API docs

### LINKED ISSUE #290: addFilter example outdated
Hi,

I am trying to use the addFilter feature to filter out some request.
The example in the doc seems to be outdated:

```js
apm.addFilter(function (payload) {
  if (payload.context.request && payload.context.request.headers) {
    // redact sensitive data
    payload.context.request.headers['x-secret'] = '[REDACTED]'
  }
  // remember to return the modified payload
  return payload
})
```
Indeed, `payload.context` does not exist. Instead, we could make something like that :
```js
apm.addFilter(function (payload) {
  payload.transactions.map((transaction) => {
    if (transaction.type === 'request' &&  transaction.context.request.headers) {
      transaction.context.request.headers['x-secret'] = '[REDACTED]'
    }
  })
  return payload
})
```

watson:
Fixes #290

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -625,10 +625,17 @@ Example usage:
 [source,js]
 ----
 apm.addFilter(function (payload) {
-  if (payload.context.request && payload.context.request.headers) {
-    // redact sensitive data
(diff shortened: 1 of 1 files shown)


watson:
jenkins please test this

PR #211, opened 2018-01-25
WIP: feat(mysql): support mysql2 module

danieljuhl:
Credits to @LewisJEllis elastic/apm-agent-nodejs#145 - I've simply created a new PR to get this merged.

PR #298, opened 2018-04-13
mysql2

Qard:
Fixes #211.


lib/instrumentation/modules/mysql2.js
```diff
@@ -0,0 +1,100 @@
...
+      }
+
+      return result
+
+      function wrapCallback (cb) {
+        hasCallback = true
+        if (span) span.start()
+        return agent._instrumentation.bindFunction(span ? wrappedCallback : cb)
+        function wrappedCallback () {
+          if (span) span.end()
```
watson: `wrappedCallback` will only ever get called if there is a span

lib/instrumentation/modules/mysql2.js
```diff
@@ -0,0 +1,100 @@
...
+        arguments[1] = wrapCallback(values)
+      } else if (typeof cb === 'function') {
+        arguments[2] = wrapCallback(cb)
+      }
+
+      var result = original.apply(this, arguments)
+      if (result && !hasCallback) {
+        ins.bindEmitter(result)
+        if (span) {
+          let started = false
```
watson: Nitpick: we use `var` everywhere else in this file, so it seems odd to suddenly use `let` here (not a merge blocker, just an observation)

Qard: changed to var

lib/instrumentation/modules/mysql2.js
Qard: fixed

Qard:
Updated the Node.js 10 section of the `.travis.yml` TAV settings.

Issue #289, opened 2018-03-22
Adding custom logger is not working


watson:
### Expected behavior

When using the [`logger`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#logger) config option it should be possible to overwrite the default logger.

### Actual behavior

The `logger` config option just seems to be ignored (as pointed out in https://github.com/elastic/apm-agent-nodejs/issues/232#issuecomment-375087000)

### Steps to reproduce

1. `npm install pino`
1. Create a file just containing this:
    ```js
    require('elastic-apm-node').start({
      logger: require('pino')()
    })
    ```
1. The output in the console should be some like this:
    ```
    {"level":50,"time":1521726287091,"msg":"Elastic APM isn't correctly configured: Missing serviceName","pid":7540,"hostname":"watson.local","v":1}
    ```
    But because the custom `logger` isn't picked up, it's just:
    ```
    Elastic APM isn't correctly configured: Missing serviceName
    ```

coolboi567:
Can this logger configuration be used to save all the debug/info logs into a separate file?

watson:
@coolboi567 No, you need to supply a custom logger for that purpose. In the future, could you please ask questions in our [Discuss forum](https://discuss.elastic.co/c/apm) or open a new issue?

coolboi567:
@watson Sure would do from now on. Sorry for this but I saw that forum was not that active, and it took days before anyone reponds, that's why I choose to ask question in related issues like these.

watson:
@coolboi567 We respond just as fast here as in the forums. It was probably just because it was during the Christmas holiday 😃

PR #299, opened 2018-04-17
 fix(config): restore custom logger feature 

Qard:
This PR restores the custom logger feature.
Closes elastic/apm-agent-nodejs#289


lib/agent.js
```diff
@@ -5,7 +5,7 @@ var parseUrl = require('url').parse
 var path = require('path')
 
 var afterAll = require('after-all-results')
-var debug = require('debug')('elastic-apm')
+var ConsoleLogLevel = require('console-log-level')
```
watson: It's kind of a convention that variables that start with an uppercase letter like this needs to be instantiated with the `new` keyword, so I'd probably change it to `consoleLogLevel`

lib/agent.js
```diff
@@ -68,7 +67,11 @@ Agent.prototype._config = function (opts) {
     ? parseUrl(this._conf.serverUrl).hostname
     : 'localhost'
 
-  logger.init({level: this._conf.logLevel})
+  this.logger = this._conf.logger || ConsoleLogLevel
+
+  if (typeof this.logger === 'function') {
+    this.logger = this.logger({ level: this._conf.logLevel })
```
watson: Is it safe to assume that all loggers that are a function will have this API?

lib/parsers.js
```diff
@@ -39,7 +38,8 @@ exports.parseMessage = function (msg) {
 exports.parseError = function (err, agent, cb) {
   stackman.callsites(err, function (_err, callsites) {
     if (_err) {
-      debug('error while getting error callsites: %s', _err.message)
+      console.log(agent)
```
watson: Ups 😄

lib/instrumentation/shimmer.js
```diff
@@ -15,7 +15,7 @@
  * BSD-2-Clause, http://opensource.org/licenses/BSD-2-Clause
  */
 
-var debug = require('debug')('elastic-apm')
+var agent = require('../agent').singleton
```
watson: Why not `require('../../')`?

test/parsers.js
```diff
@@ -381,6 +381,12 @@ test('#parseError()', function (t) {
       _conf: {
         sourceLinesErrorAppFrames: 5,
         sourceLinesErrorLibraryFrames: 5
+      },
+      logger: {
```
watson: Consider DRY'ing this up (is the DRY ruby'ism known in outside of the Ruby world?)

test/queue.js
```diff
@@ -7,7 +7,13 @@ var Queue = require('../lib/queue')
 test('maxQueueSize', function (t) {
   var opts = {
     maxQueueSize: 5,
-    flushInterval: 1e6
+    flushInterval: 1e6,
+    logger: {
```
watson: Consider DRY'ing this up

lib/agent.js
Qard: I meant for this to just be a possible initialization option. If you have a logger that doesn't initialize this way, you can just initialize it externally and provide the constructed object instead.

lib/agent.js
Qard: fixed

lib/parsers.js
Qard: fixed

lib/instrumentation/shimmer.js
Qard: fixed

test/parsers.js
Qard: fixed

test/queue.js
Qard: fixed

lib/agent.js
watson: I'm thinking if a simpler API simply would be to require people to initialize their loggers before hand?

Qard:
I changed it to not have the logger-as-a-function thing, so the level can just be controlled externally. The docs already partially suggested that in the custom logger section, so I just added an extra note to the `logLevel` option.

PR #301, opened 2018-04-24
fix(test): appName -> serviceName in http test

Qard:
Spotted an unconverted `appName` in the tests.


watson:
Good find and thanks for taking care of it 😃 I was sure I found this exact same bug a month ago and fixed it. So strange

PR #313, opened 2018-04-25
feat(transaction): add addTags method

### LINKED ISSUE #105: Add ability to set multiple tags at the same time
Today we have a [`setTag(key, value)`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-set-tag) function which allows you to set a tag on the current transaction.

As [discussed on the APM forum](https://discuss.elastic.co/t/custom-fields-in-index-pattern/111819/3), it would be nice to be able to set more than one tag at a time.

Since the existing `setTag` is singular, I'm thinking it's better to create a new function. We can keep supporting both functions or we can deprecate the old one. I'm not sure what's best.

Proposed new API:

```js
agent.setTags({key1: 'value', key2: 'value'})
```

Questions:

- Should we deprecate `setTag` or keep supporting it?
- Should it be possible to clear the already set tags, or will calling `setTags` multiple times just add to the list of tags?

Qard:
Fixes elastic/apm-agent-nodejs#105

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -729,6 +729,29 @@ Must not contain periods (`.`) as those have special meaning in Elasticsearch
 If a non-string data type is given,
 it's converted to a string before being sent to the APM Server
 
+[[apm-add-tags]]
+==== `apm.addTags()`
+
+[source,js]
+----
+apm.addTags({ [name]: value })
+----
+
+Add several tags on the current transaction.
+You can add tags multiple times.
+If an error happens during the current transaction,
+it will also get tagged with the same tags.
+
+Tags are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via `setCustomContext()`).
+
+Arguments:
+
+* `name` - Any string.
+Must not contain periods (`.`) as those have special meaning in Elasticsearch
+* `value` - Any string.
+If a non-string data type is given,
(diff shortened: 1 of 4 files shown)


docs/agent-api.asciidoc
```diff
@@ -729,6 +729,29 @@ Must not contain periods (`.`) as those have special meaning in Elasticsearch
...
+You can add tags multiple times.
+If an error happens during the current transaction,
+it will also get tagged with the same tags.
+
+Tags are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via `setCustomContext()`).
+
+Arguments:
+
+* `name` - Any string.
+Must not contain periods (`.`) as those have special meaning in Elasticsearch
```
watson: If you indent the lines following the first line in the bullet point, I think it's easier to see the relation with the bullet points (same with the bullet point below)

Qard: It's just copy/pasted from the `setTag` docs above. Should that change too?

watson: Oh, maybe there's something about asciidocs that doesn't work the way I think - or maybe I was't thinking when I wrote the above docs. Lets just leave it then. I can always go over the docs later and clean these things up

Issue #307, opened 2018-04-24
Node 10 support


alvarolobato:
- [ ] - Run compatibility tests with Node 10
- [ ] - Make needed compatibility changes to the agent

alvarolobato:
The problem we are seeing is caused by: https://github.com/nodejs/node/issues/20274

alvarolobato:
Current status is there were two issues with asyncAwait, one of them has been fixed (upstream) and the other one is still pending. We don't see any progress on that one yet.

PR #314, opened 2018-04-25
Add support for Node.js 10

watson:
### Todo

- [x] Add Node.js 10 to CI test matrix
- [ ] Fix errors uncovered by the tests

Closes elastic/apm-agent-nodejs#307


watson:
@Qard Thought I'd just kick of this PR now that Node 10 is out. As our preliminary testing has shown, it will fail, but feel free to push commits to this PR to fix it if you get time.

There's a real chance that the errors are due to Node core using some internals that we can't get access to patch, in which case we have to either see if it can be fixed in Node core or we have do drop support for running it without AsyncHooks in Node 10

watson:
I changed the lint and code coverage tests to use the latest LTS release instead of 10, so that they don't fail just because the latest release have issues. Just so we don't waste time looking in the wrong places.

Qard:
The hapi failure is a result of changes to promise hooks in V8, which result in broken continuation from async_hooks when an `await` occurs. The issue can be tracked here: https://github.com/nodejs/node/issues/20274

Qard:
@watson How do we want to handle landing this? Should we configure Node.js 10 to allow failures for now?

watson:
In any case we should probably update the [Compatibility docs](https://www.elastic.co/guide/en/apm/agent/nodejs/1.x/compatibility.html) to mention the missing support.

How about adding this line:

screen shot 2018-05-07 at 19 54 07

Qard:
It already states that async_hooks is experimental in the note below there. Perhaps we could just expand that with an additional sentence explaining that async/await in Node.js 10 has known issues?

watson:
I think it's better to have it in the list so it's not berried inside a paragraph. But I agree it could be more to the point. What do you think of this instead:

---

- v10.x - Async/Await in Node.js 10 is currently buggy and can therefore not be instrumented properly ([nodejs/node#20274](https://github.com/nodejs/node/issues/20274))

---

Qard:
Maybe replace "not be instrumented properly" with "not be supported" to make the intent clearer.

test/instrumentation/modules/hapi.js
```diff
@@ -13,6 +13,9 @@ var semver = require('semver')
 // hapi 17+ requires Node.js 8.9.0 or higher
 if (semver.lt(process.version, '8.9.0') && semver.gte(pkg.version, '17.0.0')) process.exit()
 
+// hapi does not work on Node.js 10 because of https://github.com/nodejs/node/issues/20274
```
Qard: Same as above.

docs/compatibility.asciidoc
```diff
@@ -15,8 +15,9 @@ image::https://raw.githubusercontent.com/nodejs/Release/master/schedule.png[]
 Some versions of Node.js contain bugs or issues that limit our ability to instrument them correctly.
 The following versions of Node.js are known to not be fully instrumented:
 
-- v8.0.x - Recommended solution: Upgrade to v8.2.0 or higher to get full support
+- v10.x - Async/Await in Node.js 10 is currently buggy and can therefore not be supported (https://github.com/nodejs/node/issues/20274:[nodejs/node#20274])
```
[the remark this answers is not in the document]
watson: 👍 updated

test/instrumentation/modules/hapi.js
watson: 👍 updated

PR #317, opened 2018-05-01
feat(transaction): add toJSON to transaction

Qard:
@watson I made this change while debugging the context loss in Node.js 10 and thought it might actually be a good idea to keep it. It might be good to do a similar thing to spans. 🤔

### DIFF
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -114,6 +114,52 @@ Transaction.prototype.buildSpan = function () {
   return span
 }
 
+Transaction.prototype.toJSON = function () {
+  var payload = {
+    id: this.id,
+    name: this.name,
+    type: this.type,
+    duration: this.duration(),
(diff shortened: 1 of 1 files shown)


watson:
Looks nice, great idea 👍 

I'd not call this a feature though, but more of an internal refactoring. Unless you think the `toJSON` function should be public?

Qard:
I think it could have some value as a public API. Certainly people might try to log transactions or spans to debug their own implementation. It makes sense to be able to serialize them without complaining about circular references.

watson:
True. I'm just extra careful every time we make a new public method. Makes it harder to change later. It's easier to make it private first and then make it public if it's needed than the other way around 😅

Qard:
We could perhaps make it public with messaging that it will return a serializable object but that the contents are implementation-specific and could change at any time. 🤔

Qard:
Yeah. In this case though, it's a language standard thing though, so I think it's fairly safe to expose.

PR #328, opened 2018-05-08
feat(transaction): call end with result

### LINKED ISSUE #272: Allow setting of result when ending transaction
A common pattern today when working with custom transactions are:

```js
var trans = agent.startTransaction('name', 'type')
doSomeWork(function (err) {
  if (err) trans.result = 'error'
  trans.end()
})
```

It might prove useful to be able to pass a result to the `end` function, eg:

```js
var trans = agent.startTransaction('name', 'type')
doSomeWork(function (err) {
  trans.end(err && 'error')
})
```

Qard:
Fixes #272

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -207,10 +207,14 @@ Transaction.prototype.setDefaultNameFromRequest = function () {
   this.setDefaultName(req.method + ' ' + path)
 }
 
-Transaction.prototype.end = function () {
+Transaction.prototype.end = function (result) {
   if (this.ended) {
     this._agent.logger.debug('tried to call transaction.end() on already ended transaction %o', {id: this.id})
     return
+  }
+
+  if (result !== undefined) {
+    this.result = result
   }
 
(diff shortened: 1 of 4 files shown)


lib/instrumentation/transaction.js
```diff
@@ -207,12 +207,16 @@ Transaction.prototype.setDefaultNameFromRequest = function () {
...
 }
 
-Transaction.prototype.end = function () {
+Transaction.prototype.end = function (result) {
   if (this.ended) {
     this._agent.logger.debug('tried to call transaction.end() on already ended transaction %o', {id: this.id})
     return
   }
 
+  if (typeof result !== 'undefined') {
```
watson: I think `result !== undefined` is marginally easier to read and also doesn't need to create a string (though I expect v8 is probably pretty good at optimzing away any difference between the two versions).

Qard: fixed

PR #337, opened 2018-05-09
docs: fix wrong description of agent.flush() arguments

watson:


docs/agent-api.asciidoc
```diff
@@ -1087,7 +1087,7 @@ controlled by the <<flush-interval,`flushInterval`>> and/or <<max-queue-size,`ma
 
 The callback is called *after* the event has been sent to the APM Server with the following arguments:
 
-* `err` - the captured exception
+* `err` - optional error in case something went wrong
```
Qard: To me, the word "optional" implies the user controls if there is in error or not, which is maybe a bit inaccurate.

watson: Yeah, I was thinking the same. But couldn't come up with a better way to write it... I'll just check what the Node docs write

watson: I found one example where they said `The final argument, "callback", is a callback function callback function that is invoked with a possible error argument.`. How about something like that?

Qard: Sounds good. Could just swap the "optional" you have there now with "possible".

watson: So in our case something like this:

```diff
-The callback is called *after* the event has been sent to the APM Server with the following arguments:
-
-* `err` - the captured exception
+The callback is called *after* the event has been sent to the APM Server with a possible error argument
```

Qard: Works for me. 👍

watson: Fixed

PR #338, opened 2018-05-09
test: schema validate payload sent to APM Server

### LINKED ISSUE #325: Add tests to validate payload against APM Server JSON schemas
Download the JSON schemas from the APM Server repo ([inspiration](https://github.com/elastic/apm-agent-python/blob/master/tests/scripts/download_json_schema.sh)) and validate that the payload this agent generates validates against the schemas.

watson:
These tests will be skipped on Windows as the schema download script doesn't work on AppVeyor. It can be done, but I just didn't see any reason to spend time on it.

Closes elastic/apm-agent-nodejs#325


test/integration/api-schema/download-json-schemas.sh
```diff
@@ -0,0 +1,30 @@
...
+  "stacktrace_frame.json" \
+  "system.json" \
+  "user.json" \
+)
+
+mkdir -p ${schemadir}/errors ${schemadir}/transactions ${schemadir}/sourcemaps
+
+for i in "${FILES[@]}"; do
+  output="${schemadir}/${i}"
+  curl -sf --compressed https://raw.githubusercontent.com/elastic/apm-server/master/docs/spec/${i} > ${output} || exit $?
```
watson: Would it make sense to download from the `6.x` or the `6.2` branch instead? Otherwise we might suddenly get breaking tests if the schema is intentionally updated in a breaking way. The downside of that is that we'll not get notified on those updates, so we'll have to remember to update it - or have a separate test build that pulls from master and allows for failures.

We could also just keep it as is for now and update the process once it becomes an issue

Qard:
Is there a reason not to just download the schemas with node? It seems to only be triggered by node, and it looks like it actually gets triggered twice between the two validator thunks. 🤔

watson:
@Qard Only reason was so it was easy to share it with other agents. I got it from the python agent: https://github.com/elastic/apm-agent-python/blob/master/tests/scripts/download_json_schema.sh

PR #353, opened 2018-05-16
feat(config): disable instrumentation

### LINKED ISSUE #273: Allow users to disable instrumentation of specific modules
We should allow the users to blacklist certain modules from being instrumented. This allows users to opt out in case instrumentation of certain modules cause problems or generate too much noisy data.

Suggestion for config option name: `disableInstrumentation`.

Usage example (inline config):

```js
  disableInstrumentation: ['mysql', 'ws']
})
```

Usage example (environment variables):

```
ELASTIC_APM_DISABLE_INSTRUMENTATION=mysql,ws
```

Notice the use of comma separated values in the environment variable. The config logic of the Node.js agent currently doesn't support this, but that should be simple to add.

Qard:
Here's a basic implementation for disabling instrumentation. It still needs docs and such.

@watson Does this approach seem okay?

Fixes elastic/apm-agent-nodejs#273 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -47,7 +47,8 @@ var DEFAULTS = {
   transactionMaxSpans: Infinity,
   transactionSampleRate: 1.0,
   maxQueueSize: 100,
-  serverTimeout: 30
+  serverTimeout: 30,
+  disableInstrumentation: []
 }
 
 var ENV_TABLE = {
@@ -79,7 +80,8 @@ var ENV_TABLE = {
   sourceLinesSpanLibraryFrames: 'ELASTIC_APM_SOURCE_LINES_SPAN_LIBRARY_FRAMES',
   transactionMaxSpans: 'ELASTIC_APM_TRANSACTION_MAX_SPANS',
   transactionSampleRate: 'ELASTIC_APM_TRANSACTION_SAMPLE_RATE',
-  serverTimeout: 'ELASTIC_APM_SERVER_TIMEOUT'
+  serverTimeout: 'ELASTIC_APM_SERVER_TIMEOUT',
+  disableInstrumentation: 'ELASTIC_APM_DISABLE_INSTRUMENTATION'
 }
 
 var BOOL_OPTS = [
@@ -104,6 +106,10 @@ var NUM_OPTS = [
   'serverTimeout'
 ]
 
+var ARRAY_OPTS = [
+  'disableInstrumentation'
+]
+
 function config (opts) {
   opts = Object.assign(
     {},
@@ -115,6 +121,7 @@ function config (opts) {
 
   normalizeIgnoreOptions(opts)
   normalizeNumbers(opts)
+  normalizeArrays(opts)
   normalizeBools(opts)
   truncateOptions(opts)
 
@@ -158,6 +165,16 @@ function normalizeIgnoreOptions (opts) {
 function normalizeNumbers (opts) {
   NUM_OPTS.forEach(function (key) {
     if (key in opts) opts[key] = Number(opts[key])
+  })
+}
+
+function maybeSplit (value) {
+  return typeof value === 'string' ? value.split(',') : value
+}
+
+function normalizeArrays (opts) {
+  ARRAY_OPTS.forEach(function (key) {
+    if (key in opts) opts[key] = maybeSplit(opts[key])
   })
 }
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -48,7 +48,7 @@ var DEFAULTS = {
   transactionSampleRate: 1.0,
   maxQueueSize: 100,
   serverTimeout: 30,
-  disableInstrumentation: []
+  disableInstrumentations: []
 }
 
 var ENV_TABLE = {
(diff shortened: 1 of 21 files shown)


watson:
@beniwohli I think you had something similar in the Python agent right? Just want to make sure we're aligning on the config option names if possible 😃

lib/config.js
```diff
@@ -161,6 +168,16 @@ function normalizeNumbers (opts) {
   })
 }
 
+function maybeSplit (value) {
+  return typeof value === 'string' ? value.split(',') : value
```
watson: Nice! I was thinking that we needed to support this from the command line as well. I like this pragmatic approach 👍

lib/instrumentation/index.js
```diff
@@ -47,8 +47,18 @@ Instrumentation.prototype.start = function () {
...
+  const patches = new Set(MODULES)
+
+  this._agent._conf.disableInstrumentation.forEach(patch => {
+    patches.delete(patch)
+  })
+
+  this._applyHooks(Array.from(patches))
+}
+
+Agent.prototype._applyHooks = function (modules) {
```
watson: s/Agent/Instrumentation/

lib/config.js
felixbarny: Does this handle spaces after the comma?

watson: No, but I don't think that's necessary as long as it's documented

lib/config.js
```diff
@@ -79,7 +80,8 @@ var ENV_TABLE = {
   sourceLinesSpanLibraryFrames: 'ELASTIC_APM_SOURCE_LINES_SPAN_LIBRARY_FRAMES',
   transactionMaxSpans: 'ELASTIC_APM_TRANSACTION_MAX_SPANS',
   transactionSampleRate: 'ELASTIC_APM_TRANSACTION_SAMPLE_RATE',
-  serverTimeout: 'ELASTIC_APM_SERVER_TIMEOUT'
+  serverTimeout: 'ELASTIC_APM_SERVER_TIMEOUT',
+  disableInstrumentation: 'ELASTIC_APM_DISABLE_INSTRUMENTATION'
```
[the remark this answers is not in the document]
watson: We should probably link to the [compatibility docs](https://www.elastic.co/guide/en/apm/agent/nodejs/current/compatibility.html) from the place where this config option is documented. I'd like to only have one place where we list these modules, so if we can keep it to that page then it's the easiest 😃

beniwohli:
Sorry, haven't seen this before. The Python agent does indeed have a similar option, but only by setting certain environment variables, and it's not documented, so I'd say it isn't covered by backwards compat guarantees (meaning I could change it to align with other agents).

The main reason I haven't implemented it as a config setting yet is that the config object is bound to the agent instance (you can theoretically have multiple agents with differing config), while instrumentation is global, so there's a bit of a mismatch there

Qard:
The disabling feature itself is working and tested. I don't know if we want to perhaps use a different name though? I haven't done anything with docs yet--I don't want to put time into that until we're happy with the code.

Also, I haven't done anything about keeping continuation patches when instrumentation is disabled. I'm thinking a good approach might be to add another argument to the patch function signature to indicate if the module is enabled, and then leave it up to each patch function to decide what to do when it should be disabled. 🤔

lib/instrumentation/patch-async.js
```diff
@@ -135,6 +135,8 @@ module.exports = function (ins) {
 
   var http = require('http')
 
+  if (!http) console.trace('wat')
```
watson: 👀

test/config.js
```diff
@@ -220,3 +235,85 @@ captureBodyTests.forEach(function (captureBodyTest) {
     })
   })
 })
+
+test('disableInstrumentation', function (t) {
+  // var hook = require('require-in-the-middle')
```
watson: 🤔

Qard: Yep...a bit work-in-progress still. I broke that test and was trying to figure out why. Seems to be some sort of circular dependency issue. I haven't yet figured out a good way to fix it.

Qard:
@watson I've added a bunch more tests. Do we want to go forward with the approach and naming of the `disableInstrumentation` array/csv-string? I want to get started on docs, if the rest of the PR looks reasonable.

lib/config.js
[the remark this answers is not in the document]
felixbarny: @Qard WDYT?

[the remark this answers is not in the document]
Qard: Yep, did that. 👍

lib/instrumentation/modules/redis.js
```diff
@@ -46,37 +49,28 @@ module.exports = function (redis, agent, version) {
...
   function wrapSendCommand (original) {
     return function wrappedSendCommand (command) {
-      var span = agent.buildSpan()
+      var span = enabled && agent.buildSpan()
       var id = span && span.transaction.id
       var args = Array.prototype.slice.call(arguments)
 
       agent.logger.debug('intercepted call to RedisClient.prototype.internal_send_command %o', { id: id, command: command })
 
-      if (span && args.length > 0) {
```
watson: Why dropping the check for `span` here? Is it to ensure that the callback is bound even if there's no active transaction?

Qard: Yep, that's exactly it. I do `var span = enabled && agent.buildSpan()` above, so the span will only exist when in a sampled transaction *and* the redis instrumentation is enabled.

test/config.js
```diff
@@ -4,12 +4,14 @@ var IncomingMessage = require('http').IncomingMessage
...
 var util = require('util')
 
+var semver = require('semver')
 var test = require('tape')
 var isRegExp = require('core-util-is').isRegExp
 
 var Agent = require('./_agent')
 var config = require('../lib/config')
 var request = require('../lib/request')
+var Instrumentation = require('../lib/instrumentation')
```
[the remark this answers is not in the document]

PR #360, opened 2018-05-16
test(ci): test unreleased versions of Node.js

watson:
Depends on #359 (tests will fail until that PR is merged into `master`).

This PR is not meant to be merged into `master`, instead it's targeted a new `rc` branch.

This PR will allow us to run the test suite against both the Node.js nightly builds and against the official release candidates.

By keeping this functionality in a separate branch, and having Travis CI check out `master` before running the test suite, we can set up a cron job on Travis CI that runs these tests once a day without influencing our regular tests.

The following nightlies are tested:
- v11
- v10
- v9
- v8
- v6

The following release candidates are tested:
- v10
- v9
- v8
- v6

### To do

- [x] Set up daily cron job on Travis CI

Closes #319

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,145 +18,45 @@ services:
 before_script:
   - wait-on tcp:9200
 
-node_js:
-  - '10'
-  - '9'
-  - '8'
-  - '6'
-  - '4'
+git:
+  depth: false # allows us to check out master in the install step
 
-jobs:
-  fast_finish: true
+install:
+  - git checkout master
+  - npm install
 
+matrix:
   include:
-
-    ###########################################
-    #                TEST STAGE               #
-    ###########################################
-
-    # Disable Async Hooks
+    # Nightlies
+    -
+      node_js: '11'
+      env: NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/nightly/
     -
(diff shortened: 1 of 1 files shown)


watson:
There's also a chakracore-nightly, a chakracore-rc, a v8-canary etc. Would it make sense to add some of these to the test suite as well? See https://nodejs.org/download/

watson:
Oh!! When did they add that feature to build stages 💯

I was suggesting that in their github, but didn't see that it had been added. I'll take a look

Qard:
Can we have it as conditional tasks that only run when triggered via cron? That way the failures wouldn't appear on non-cron run because it wouldn't even run. 🤔

watson:
That was my intention, but since it will be running on the master branch it will affect the build badge in the readme. But we could maybe change the build badge to point to the `1.x` branch instead.

Qard:
Ah, I see. Yeah, I'd go for just pointing the badge at 1.x, it should express the "stable" branch anyway.

The master branch is kind of more of a "dev" branch, even if it should, for the most part, be in-line with what is in the stable branch. Things could land in master but sit for a bit while other things get rebased and such.

watson:
Closing in favor of #370

PR #366, opened 2018-05-18
test: test that unsampled http transactions have the expected result

watson:
I created this test while debugging a related issue. It turned out not to be related to sampling, but I thought it would be a good test to have nonetheless, so here it is


Qard:
Good call. Perhaps it should also verify `spans` is empty?

watson:
@Qard the test above already checks that

PR #370, opened 2018-05-18
test(ci): test unreleased versions of Node.js

watson:
Should we add others than `nightly` and `rc`?

See a full list of possibilities here: https://nodejs.org/download/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -178,6 +178,50 @@ jobs:
       env: TAV=knex,ws,graphql,express-graphql,elasticsearch,hapi,express
       script: tav --quiet
 
+    ###########################################
+    #        NODE.JS PRE-RELEASE STAGE        #
+    ###########################################
+
+    # Nightlies
+    - stage: Node.js pre-releases
+      node_js: '11'
+      if: type = cron
+      env: NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/nightly/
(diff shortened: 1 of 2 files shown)


watson:
@Qard currently the release candidates of Node.js 10 and 11 breaks our builds: https://travis-ci.org/elastic/apm-agent-nodejs/builds/380357638

Should we still merge or do we want to check what's going on first?

Qard:
It only gets triggered by cron, so I think it's fine to have failures we are aware of.

watson:
Hmmm yeah, I'm just worried about having a cron job that have certain tests that are continuously failing but also contain tests that - if they fail - should require immediate action (the TAV tests).

Maybe we should just have a look at why the two nightly tests are failing to see if it's a simple fix or if it's something more fundamental.

Qard:
Is there any way to group into different cron tasks so we could have one that's just RCs and one that is just TAV tests? Then we could pull all that out of the per-commit tests and run those two groups periodically, tracking the TAV results more closely.

watson:
The only way that I know of is to make that dummy branch that I did in #360 and have that contain a custom `.travis.yml` file that checks out master before running the tests. That way you can have a separate cron job on that branch, which will even produce slightly different failure mails so you can differentiate directly in the e-mail subject.

Issue #363, opened 2018-05-16
Agent not sending timeline spans for self-calls


jdelman:
**Describe the bug**
Timeline spans do not show up for requests made to an API server from itself.

**To Reproduce**
1. Make a call to Node server in request, i.e.
```
app.get('/test', (req, res) => request('http://localhost:3000/somethingelse').pipe(res));
```

**Expected behavior**
Expected to see the timeline spans in APM/Kibana.

**Environment**
- OS: Amazon Linux
- Node.js version: 8.11.1
- APM Server version: 6.2.1
- Agent version: 1.5.1

**How are you starting the agent? (please tick one of the boxes)**
- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**
We're using an AWS ELB and some other proxies in between the server and the internet. Direct calls to the API don't have this problem.

#### Agent config options 

<details><summary>Click to expand</summary>

```
{
  serviceName: 'antares',
  maxQueueSize: 50,
  flushInterval: 5,
  active: NODE_ENV === 'production' && APM_SERVER_URL,
  serverUrl: APM_SERVER_URL,
  captureErrorLogStackTraces: 'always',
  captureExceptions: true,
}
```
</details>

#### `package.json` dependencies

<details><summary>Click to expand</summary>

```
{
    "@snaps/data-helpers": "^1.7.0",
    "@snaps/eslint-config-snaps": "^1.0.0",
    "@snaps/sails-mongo": "^0.12.4",
    "JSV": "^4.0.2",
    "abbrev": "^1.0.7",
    "accepts": "^1.3.3",
    "after": "^0.8.1",
    "align-text": "^0.1.4",
    "ambi": "^2.5.0",
    "amdefine": "^1.0.0",
    "anchor": "^0.11.2",
    "ansi-regex": "^0.2.1",
    "ansi-styles": "^1.0.0",
    "ap": "^0.2.0",
    "apn": "^2.2.0",
    "argparse": "^0.1.16",
    "array-find-index": "^1.0.1",
    "arraybuffer.slice": "^0.0.6",
    "asap": "^1.0.0",
    "asn1": "^0.2.3",
    "assert-plus": "^0.2.0",
    "async": "^2.1.4",
    "aws-sdk": "^2.5.2",
[...]

PR #371, opened 2018-05-18
fix(instrumentation): add bindEmitter for requests

Qard:
Event emitters need their event handlers bound to the appropriate context. This binds request and response handlers to the context at which they were attached.

Fixes elastic/apm-agent-nodejs#363

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


test/instrumentation/modules/http/request.js
```diff
@@ -0,0 +1,78 @@
+'use strict'
+
+var agent = require('../../_agent')()
+
+var http = require('http')
+
+var test = require('tape')
+var express = require('express')
+var request = require('request')
```
watson: `request` isn't a dependency

lib/instrumentation/http-shared.js
```diff
@@ -90,6 +95,8 @@ exports.traceOutgoingRequest = function (agent, moduleName) {
         agent.logger.debug('request details: %o', {protocol: protocol, host: req._headers.host, id: id})
       }
 
+      ins.bindEmitter(req)
```
watson: Should we bind the client response as well?

lib/instrumentation/index.js
```diff
@@ -152,6 +152,14 @@ Instrumentation.prototype.bindFunction = function (original) {
   }
 }
 
+Instrumentation.prototype.bindEmitter = function (emitter) {
+  var self = this
+  var on = emitter.on
+  emitter.on = function (name, handler) {
```
watson: We should probably patch the `once`, `prependListener` etc alternatives as well

test/instrumentation/modules/http/request.js
Qard: Oops. I guess something else depends on it, so I didn't notice. 😅

lib/instrumentation/http-shared.js
Qard: Probably, yes. I meant to mark this as in-progress. Just wanted to get your input on it first to make sure there wasn't something I was missing in regard to not wrapping emitter handlers before.

lib/instrumentation/index.js
Qard: Yep. I'll make sure those get added too.

lib/instrumentation/index.js
```diff
@@ -152,6 +153,20 @@ Instrumentation.prototype.bindFunction = function (original) {
   }
 }
 
+Instrumentation.prototype.bindEmitter = function (emitter) {
+  var ins = this
+
+  shimmer.massWrap(emitter, [
+    'on',
```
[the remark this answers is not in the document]
Qard: `on` points _directly_ to `addListener`, it doesn't _call_ it, therefore if we only patch one of the functions, the other would remain unwrapped.

As for `once` and `prependOnceListener`, those _do_ call the underlying `on` and `prependListener` functions, so wrapping those _would_ result in double-wrapping.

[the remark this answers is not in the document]
watson: > on points directly to addListener, it doesn't call it, therefore if we only patch one of the functions, the other would remain unwrapped.

Ah yes, I wasn't paying attention.

> As for once and prependOnceListener, those do call the underlying on and prependListener functions, so wrapping those would result in double-wrapping.

As mentioned in my previous comment, we have a guard to make sure double wrapping doesn't happen. So doing this would only be to ensure this would work even if Node core in the future didn't forward the call. But this is highly unlikely to happen, so I think we're safe not do it 👍 

But it might make sense to add a few quick tests to see that it works also for the aliases and the `once` versions.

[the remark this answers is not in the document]
Qard: I've added a bunch more tests proving behaviour of both `bindFunction` and `bindEmitter`.

PR #377, opened 2018-05-27
docs(span): reference apm.startSpan() over span.start()

watson:


Qard:
Is the intent here specifically to _avoid_ suggesting the use of `span.start()` in favor of `agent.startSpan()`? We don't want to say you can do A or B?

watson:
`span.start()` is deprecated

Qard:
Ah, right. Good reason to avoid it then. 😅

PR #382, opened 2018-05-29
test: fix http2 race condition

watson:


watson:
A proper fix have been created in #384 - closing

PR #385, opened 2018-06-01
test: fix silent error in lambda tests

watson:
Previously these tests never ran through the tests for the 2nd request as the server was closed prematurely. The test would output the following error, but would not actual fail:

    Error: connect ECONNREFUSED 127.0.0.1:52177
        at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1161:14)

### DIFF
--- a/test/_apm_server.js
+++ b/test/_apm_server.js
@@ -24,7 +24,7 @@ util.inherits(APMServer, EventEmitter)
 function APMServer (agentOpts, mockOpts) {
   if (!(this instanceof APMServer)) return new APMServer(agentOpts, mockOpts)
   var self = this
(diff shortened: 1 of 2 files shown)


watson:
I noticed that we don't use the `forwardTo` option either:

https://github.com/elastic/apm-agent-nodejs/blob/8530debf704ea099323e30042ff0adee3cea1d19/test/_apm_server.js#L55-L61

Can we just remove that?

Qard:
I put the `forwardTo` in there just to make debugging easier. It lets us forward anything sent to the mocked server to a real apm-server so we can look at the data in the UI. I'd just leave it there, personally. But I'm good either way.

PR #396, opened 2018-06-13
feat(express-queue): retain continuity through express-queue

### LINKED ISSUE #373: Add support for express-queue module
Related issue: elastic/apm-agent-nodejs#367

The [express-queue](https://github.com/alykoshin/express-queue) module adds a user-land callback queue and therefore breaks the async context propagation.

Steps to reproduce:

1. Start the following server
    ```js
    const agent = require('./').start({
      serviceName: 'test',
      captureExceptions: false
    })
    
    // hijack the request to the APM Server to validate the payload
    agent._httpClient.request = function (endpoint, headers, data, cb) {
      data.transactions.forEach(function (trans) {
        assert.equal(trans.result, 'HTTP 2xx')
      })
    }
    
    const queue = require('express-queue')
    const app = express()
    
    app.use(queue({activeLimit: 2, queuedLimit: -1}))
    
    app.get('/', function (req, res) {
      setImmediate(function () {
        res.end()
      })
    })
    
    app.listen(3000)
    ```
1. Send a lot of simultainious requests to the server using autocannon:
    ```sh
    autocannon localhost:3000
    ```

### Todo

- [ ] Patch the module
- [ ] Add tests
- [ ] Document it in the [Continuity section](https://github.com/elastic/apm-agent-nodejs/blob/master/docs/compatibility.asciidoc#continuity) in compatibility.asciidoc

Qard:
This fixes the continuity issue with express-queue.

Fixes #373

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -97,6 +97,10 @@ express-graphql-0.6.12_13:
   versions: '>=0.6.12'
   commands: node test/instrumentation/modules/express-graphql.js
 
+express-queue:
+  versions: '>=0.0.11'
+  commands: node test/instrumentation/modules/express-queue.js
+
 koa-router:
   node: '>=6.0.0'
   peerDependencies: koa@2
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,7 @@ jobs:
     -
       node_js: '10'
       if: type IN (cron, pull_request) AND NOT branch =~ ^greenkeeper/.*
-      env: TAV=knex,ws,graphql,express-graphql,elasticsearch,hapi,express
+      env: TAV=knex,ws,graphql,express-graphql,elasticsearch,hapi,express,express-queue
       script: tav --quiet
 
     # Node.js 9
@@ -109,7 +109,7 @@ jobs:
     -
       node_js: '9'
       if: type IN (cron, pull_request) AND NOT branch =~ ^greenkeeper/.*
-      env: TAV=knex,ws,graphql,express-graphql,elasticsearch,hapi,express
+      env: TAV=knex,ws,graphql,express-graphql,elasticsearch,hapi,express,express-queue
       script: tav --quiet
 
     # Node.js 8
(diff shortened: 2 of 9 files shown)


watson:
Seems like the express-queue v0.0.10 and below isn't supported by this patch. If it's gonna be complicated to add support for it, I say we just don't support older versions and call it a day.

Qard:
Strange. I did a TAV test locally, before pushing the code up, and had zero failures. 🤔

Qard:
Oh...I did it on v4, which exits early. 😅

Qard:
@watson It looks like pre-0.0.11 will be somewhat more difficult to patch, but is functionally identical, from the user perspective. I've restricted this PR to only support 0.0.11+. I'd recommend not bothering with earlier versions unless someone asks for it.

watson:
jenkins test this please

watson:
jenkins test this please

watson:
jenkins test this please

watson:
Finally Jenkins is passing (flimsy tests). But now I have to figure out why codecov is failing. So far I'm getting a 401 when trying to access the codecov site for some reason.

Qard:
Works for me. Probably just a minor blip in uptime for codecov.

watson:
@Qard I would to some degree consider this a bug fix rather than a new feature. It's of course not black and white, and I can probably go either way. What do you think?

Qard:
I can see a case for either.

On one hand, it doesn't really add a "feature" from a user-facing perspective. On the other hand, calling it a bug fix could be interpreted as a statement that _every_ occurrence of continuity-loss in userland modules should be accounted for, which is unrealistic. Considering support for continuity through a given module to be a feature might make more sense.

I'm good either way though. What do you think?

watson:
Yeah, it's tricky - up there with cache invalidation and naming things 😉

A user using `express-queue` without this patch would see what they would characterize as bugs. So from their point of view - without knowing what we know - this is a "bug fix".

But I agree we can't take the responsibility for these sort of errors as we have no control over user land code - and we might even have situations where we simply say that we can't add support for a given module. I'm personally not afraid that people would start assuming stuff like this is our fault even if we called it a `fix` in this case.

I'll go ahead and rename the commit to say `fix` when merging #yolo

PR #399, opened 2018-06-15
feat(graphql): add support for the upcoming GraphQL v14.x

watson:


watson:
There's really no big reason to support pre-releases, but adding it already now means that we know in advance if this is going to be an issue

watson:
Jenkins test this please

PR #419, opened 2018-06-27
doc: add performance tuning doc

Qard:
@watson I'm not sure this is quite done yet, but I don't have any more time to work on this tonight. Feel free to make your own edits, otherwise I'll get back to it next Tuesday.

### DIFF
--- a/docs/performance-tuning.asciidoc
+++ b/docs/performance-tuning.asciidoc
@@ -122,7 +122,7 @@ will capture five lines of code around each stack frame.
 Source lines are cached in-process.
 In memory-constrained environments,
 the source line cache may use more memory than desired.
-Turning the limits down will help prevent excessive memory use
+Turning the limits down will help prevent excessive memory use.
 
 [float]
 [[performance-stack-frame-limit]]
--- a/docs/performance-tuning.asciidoc
+++ b/docs/performance-tuning.asciidoc
@@ -2,7 +2,7 @@
 
 ifdef::env-github[]
 NOTE: For the best reading experience,
-please view this documentation at https://www.elastic.co/guide/en/apm/agent/nodejs/current/performance.html[elastic.co]
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/nodejs/current/performance-tuning.html[elastic.co]
 endif::[]
 
 == Performance Tuning
@@ -37,34 +37,37 @@ require('elastic-apm-node').start({
 [[performance-transaction-queue]]
 === Transaction Queue
 
-The agent uses a queue to send to the apm server in batches.
-This queue uses both a time interval and a size limit to balance memory usage against socket descriptor saturation.
-
-[float]
-[[performance-max-queue-size]]
-==== Max Queue Size
-
-The <<max-queue-size,`maxQueueSize`>> controls the maximum number of request transactions that may remain in the queue before they must be sent.
-
-Lowering this will reduce memory consumption,
-however it will increase the number of requests made to the apm server.
+The agent buffers the collected data using an in-memory queue before sending it to the APM Server.
+The queue is flushed either after a specific <<performance-flush-interval,amount of time>> or when it reaches <<performance-max-queue-size,a certain size>> -
+whichever comes first.
+Lowering these defaults can reduce memory usage,
+but will increase the number of requests to the APM Server.
 
 [float]
 [[performance-flush-interval]]
 ==== Flush Interval
 
 To prevent items from staying in the queue for a long time during low activity,
-a flush interval is used to ensure the queue empties if anything in it is too old.
+the <<flush-interval,`flushInterval`>> setting is used to ensure the queue empties if anything in it is too old.
 
-Lowering the flush interval will ensure transactions are sent to the apm server faster,
+Lowering the flush interval will ensure transactions are sent to the APM Server faster,
 but may also result in increased HTTP traffic and cpu usage.
 
 [float]
-[[performance-server-timeout]]
-=== Server Timeout
+[[performance-max-queue-size]]
+==== Max Queue Size
 
-In the event that the connection to the apm server is slow or unstable,
-a server timeout can be set to ensure connections don't stay open too long.
+The <<max-queue-size,`maxQueueSize`>> controls the maximum number of transactions that may remain in the queue before they must be sent.
+
+Lowering this will reduce memory consumption,
+however it will increase the number of requests made to the APM Server.
+
+[float]
+[[performance-server-timeout]]
+=== APM Server Timeout
+
+In the event that the connection to the APM Server is slow or unstable,
+the <<server-timeout,`serverTimeout`>> setting can be set to ensure connections don't stay open too long.
 If the server timeout is too high,
 it may result in too many socket descriptors being held open.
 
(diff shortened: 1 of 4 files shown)


watson:
I added a link from the index page to the performance doc and fixed the commit message linter bug. With the new link, the index page will look like this:

image

I'm not sure if that's the best location for it?

I also added line breaks after punctuation marks. I always feels that it's a little excessive when ever I break on all punctuation marks, but at least it saves me from considering each case individually 😅

I also went over the docs to see what other config options could be used to tune the performance. We should consider adding some of these:

- [`flushInterval`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#flush-interval)
- [`maxQueueSize`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#max-queue-size)
- [`serverTimeout`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#server-timeout) - Only really an issue if your APM Server or the connection to it is unstable
- [`transactionMaxSpans`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#transaction-max-spans)
- [`ignoreUrls`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#ignore-urls) - Not really that performance related
- [`ignoreUserAgents`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#ignore-user-agents) - Not really that performance related

watson:
Oh just remembered that we should add a link to this page from the README ToC as well

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
+[[performance-tuning]]
+
+ifdef::env-github[]
+NOTE: For the best reading experience,
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/nodejs/current/performance.html[elastic.co]
```
watson: I messed up when I moved some things around and now the actual generated html file is called `performance-tuning.html` because of the new anchor tag 😅 

**Option 1:** Rename the link to point to `performance-tuning.html`
**Option 2:** Keep the link as it is and rename the generated file back to just `performance.html`

I vote for option 1. But in that case we should also consider renaming the asciidoc file to match the generated html file. Just to make it easier to know what comes from what. All other doc files have a one-to-one filename correlation (`setup.asciidoc` being the only exception, as it generates `advanced-setup.html` - but we might consider "fixing" that as well).

What do you think?

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+require('elastic-apm-node').start({
+  transactionSampleRate: 0.2
+})
+----
+
+[float]
+[[performance-transaction-queue]]
+=== Transaction Queue
+
+The agent uses a queue to send to the apm server in batches.
```
watson: I find this paragraph a little hard to parse for some reason. How about something along the lines of this instead:

```asciidoc
The agent buffers the collected data using an in-memory queue before sending it to the APM Server.
The queue is flushed either after a specific <<performance-flush-interval,amount of time>> or when it reaches <<performance-max-queue-size,a certain size>> -
whichever comes first.
Lowering these defaults can reduce memory usage,
but will increase the number of requests to the APM Server.
```

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+This queue uses both a time interval and a size limit to balance memory usage against socket descriptor saturation.
+
+[float]
+[[performance-max-queue-size]]
```
watson: I would move this section below the `performance-flush-interval` section, so that they are listed in the same order as described above

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+The <<max-queue-size,`maxQueueSize`>> controls the maximum number of request transactions that may remain in the queue before they must be sent.
+
+Lowering this will reduce memory consumption,
+however it will increase the number of requests made to the apm server.
+
+[float]
+[[performance-flush-interval]]
+==== Flush Interval
+
+To prevent items from staying in the queue for a long time during low activity,
```
watson: Missing link to the config option

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+=== Transaction Queue
+
+The agent uses a queue to send to the apm server in batches.
+This queue uses both a time interval and a size limit to balance memory usage against socket descriptor saturation.
+
+[float]
+[[performance-max-queue-size]]
+==== Max Queue Size
+
+The <<max-queue-size,`maxQueueSize`>> controls the maximum number of request transactions that may remain in the queue before they must be sent.
```
watson: s/request transactions/transactions/

(all transactions are added to the queue, not only request transactions)

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+
+Lowering this will reduce memory consumption,
+however it will increase the number of requests made to the apm server.
```
watson: s/apm server/APM Server/

watson: Same goes for any other location in this doc where it says `apm server`

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+a flush interval is used to ensure the queue empties if anything in it is too old.
+
+Lowering the flush interval will ensure transactions are sent to the apm server faster,
+but may also result in increased HTTP traffic and cpu usage.
+
+[float]
+[[performance-server-timeout]]
+=== Server Timeout
+
+In the event that the connection to the apm server is slow or unstable,
```
watson: Missing link to the config option

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+
+To prevent items from staying in the queue for a long time during low activity,
+a flush interval is used to ensure the queue empties if anything in it is too old.
+
+Lowering the flush interval will ensure transactions are sent to the apm server faster,
+but may also result in increased HTTP traffic and cpu usage.
+
+[float]
+[[performance-server-timeout]]
+=== Server Timeout
```
watson: Consider renaming this headline to "APM Server Timeout"

docs/performance.asciidoc
```diff
@@ -0,0 +1,153 @@
...
+
+This will mainly impact memory usage,
+but may also have a noticeable impact on speed too.
+The cpu time required to capture and convert a stack frame to something JavaScript can understand is not insignificant.
+
+[float]
+[[performance-error-log-stack-traces]]
+==== Error Log Stack Traces
+
+Most stack traces recorded by the agent will point to where the error was instantiated,
```
watson: Missing link to the config option

docs/performance.asciidoc
Qard: Option 1 sounds good to me.

docs/performance-tuning.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+
+The first knob to reach for when tuning the performance of the agent is <<transaction-sample-rate,`transactionSampleRate`>>.
+Adjusting the sampling rate controls what percent of requests are traced.
+By default,
+the sample rate is set at `1.0`,
+meaning _all_ requests are traced.
+
+The sample rate will impact all four performance categories,
+so simply turning down the sample rate is an easy way to improve performance.
+
```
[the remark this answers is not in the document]
Qard: fixed.

PR #429, opened 2018-07-02
fix: ensure correct streaming when using mimic-response

### LINKED ISSUE #423: Instrumentation sometimes breaks outgoing HTTP requests
**Describe the bug**

We've been hunting down multiple weird bugs recently, all of them related to outgoing HTTP requests. After some trial-and-error I found it's most likely caused by APM's instrumentation as disabling it fixed all issues.

Example of our code:

```js
got('https://registry.npmjs.org/uikit').then((response) => {
    console.log(JSON.parse(response.body));
});
```

This sometimes throws an error:

```js
Unexpected token  in JSON at position 0 in "https://registry.npmjs.org/uikit": 
    ��T�n�6���MQw=hh��/���H��\K�JR�E���d;N�h���I<�3�����@
```

The response, which looks completely broken at first, is actually correct, except it's gzipped - npm responds with `content-encoding: gzip` header and the got module is supposed to gunzip the response and pass the result to our callback. For some reason, *sometimes* it passes the raw response instead.

At first I suspected this was an issue with got, so I tried downgrading to an older version which we've been using for a while without any problems but that didn't help. Then I tried to disable instrumentation and as I already said, that fixed all issues. I went through the code of this module but didn't find any got-specific code, so I'm not sure if this only affects that module or all outgoing requests.

Unfortunately, I was not able to reliably reproduce this locally. In production, we get the error roughly for one in a hundred requests and usually when a request for the same URL is made a few seconds later, it succeeds. It is also not specific to the host being `npmjs.org` or response format being `json` - we fetch data and files from multiple different services and it affects all of them.

Let me know if there's anything I can do to get this problem identified and fixed.
 
- OS: Linux
- Node.js version: 8.11.3
- APM Server version: 6.3.0
- Agent version: 1.8.1

**Additional context**

- Agent config options 
  <details>

    ```
        active: process.env.NODE_ENV === 'production',
        serviceName: 'jsdelivr-origin',
        serviceVersion: require('../package.json').version,
        logLevel: 'fatal',
        captureSpanStackTraces: false,
        captureErrorLogStackTraces: 'always',
        errorOnAbortedRequests: true,
        abortedErrorThreshold: 30000,
        transactionSampleRate: .2,
        asyncHooks: false, // had this enabled but tried to disable - didn't help
    ```
  </details>
- `package.json` dependencies:
  <details>

    ```
		"@octokit/rest": "^15.9.3",
		"aws-sdk": "^2.265.1",
		"bluebird": "^3.5.1",
		"bunyan": "^1.8.12",
		"bunyan-prettystream": "0.1.3",
		"bytes": "^3.0.0",
		"clean-css": "^4.1.11",
		"combine-source-map": "0.8.0",
		"compression": "^1.7.2",
		"config": "^1.30.0",
		"config-mapper-env": "^1.0.1",
		"connect-modrewrite": "^0.10.2",
		"convert-source-map": "^1.5.1",
		"dedent-js": "~1.0",
		"elastic-apm-node": "^1.8.1",
		"elastic-apm-utils": "^1.1.0",
		"elasticsearch": "^15.0.0",
		"express": "^4.16.3",
		"fs-extra": "^6.0.1",
		"got": "^8.3.1",
		"gunzip-maybe": "^1.4.1",
		"h-logger2": "^1.0.2",
		"h-logger2-elastic": "^2.1.3",
		"handlebars": "^4.0.11",
		"is-hexdigest": "^1.0.1",
		"is-minified-code": "^1.1.0",
		"is-safe-path": "~1.0",
		"lodash": "^4.17.10",
		"lru-cache": "^4.1.3",
		"mime": "^2.3.1",
		"morgan": "^1.9.0",
		"natural-compare": "~1.4",
		"promisepipe": "^2.1.3",
		"queue": "^4.4.2",
		"recursive-readdir": "^2.2.2",
		"semver": "^5.5.0",
		"serve-favicon": "^2.5.0",
		"tar-fs": "^1.16.3",
		"through2-spy": "~2.0",
		"uglify-js": "^3.3.23",
		"write-file-atomic": "^2.3.0"
    ```
  </details>

watson:
The `mimic-response` module exposes a single function with the following API:

```js
mimicResponse(fromStream, toStream)
```

If someone is calling `mimicResponse` where the `fromStream` have had its emitter functions bound using `bindEmitter(fromStream)`, the `toStream` will wrongly bind all its event listeners to the `fromStream`.

This becomes especially problematic in case the data is transformed either in the `toStream` or in between `fromStream` and `toStream`, eg:

```js
fromStream.pipe(zlib.createGunzip()).pipe(toStream)
```

In those cases, the data emitted by `toStream` will actually be the un-transformed data coming from the `fromStream`.

This commit patches `mimic-response` so that the event emitters of `toStream` runs in the expected context.

Fixes #423

### DIFF
--- a/test/instrumentation/modules/http/github-423.js
+++ b/test/instrumentation/modules/http/github-423.js
@@ -11,6 +11,8 @@ var fs = require('fs')
 var got = require('got')
 var test = require('tape')
 
+var fileSize = fs.readFileSync(__filename, 'utf8').length
+
 test('https://github.com/elastic/apm-agent-nodejs/issues/423', function (t) {
   // Start dummy remote server to fetch gzip'ed data from
   var remote = http.createServer(function (req, res) {
@@ -25,7 +27,8 @@ test('https://github.com/elastic/apm-agent-nodejs/issues/423', function (t) {
     // Start simple server that performs got-request on every request
     var server = http.createServer(function (req, res) {
       got(url).then(function (response) {
-        t.equal(response.body[0], '\'', 'body should be uncompressed')
+        t.equal(response.body.length, fileSize, 'body should be expected size')
+        t.equal(response.body.slice(0, 12), '\'use strict\'', 'body should be uncompressed')
         res.end()
       })
     })
--- a/test/instrumentation/modules/http/github-423.js
+++ b/test/instrumentation/modules/http/github-423.js
@@ -1,0 +1,45 @@
+'use strict'
+
+require('../../../..').start({
+  serviceName: 'test',
+  captureExceptions: false
+})
+
+var http = require('http')
+var zlib = require('zlib')
+var fs = require('fs')
+var got = require('got')
+var test = require('tape')
+
+test('https://github.com/elastic/apm-agent-nodejs/issues/423', function (t) {
+  // Start dummy remote server to fetch gzip'ed data from
+  var remote = http.createServer(function (req, res) {
+    res.setHeader('Content-Encoding', 'gzip')
+    fs.createReadStream(__filename).pipe(zlib.createGzip()).pipe(res)
+  })
+
+  remote.listen(function () {
+    var port = remote.address().port
+    var url = 'http://localhost:' + port
+
+    // Start simple server that performs got-request on every request
+    var server = http.createServer(function (req, res) {
+      got(url).then(function (response) {
+        t.equal(response.body[0], '\'', 'body should be uncompressed')
+        res.end()
+      })
+    })
+
+    server.listen(function () {
+      var port = server.address().port
+      var url = 'http://localhost:' + port
+
+      http.get(url, function (res) {
+        res.resume()
+        server.close()
+        remote.close()
+        t.end()
+      })
+    })
+  })
+})
--- a/.tav.yml
+++ b/.tav.yml
@@ -1,6 +1,9 @@
 generic-pool:
   versions: ^2.0.0 || ^3.1.0
   commands: node test/instrumentation/modules/generic-pool.js
+mimic-response:
+  versions: ^1.0.0
+  commands: node test/instrumentation/modules/mimic-response.js
 mysql:
   versions: ^2.0.0
   commands:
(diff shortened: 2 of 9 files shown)


watson:
Btw, I discovered that this isn't the first time we've had issues with `mimic-response`. We already added some tests to make sure we were compatible in https://github.com/elastic/apm-agent-nodejs/commit/4f2b9ebc4827ca787ba0b335ed58bccf2bf2a568 (which fixes https://github.com/opbeat/opbeat-node/issues/179), but we unfortunately hadn't thought of this use-case back then.

test/instrumentation/modules/http/github-423.js
```diff
@@ -0,0 +1,45 @@
...
+  })
+
+  remote.listen(function () {
+    var port = remote.address().port
+    var url = 'http://localhost:' + port
+
+    // Start simple server that performs got-request on every request
+    var server = http.createServer(function (req, res) {
+      got(url).then(function (response) {
+        t.equal(response.body[0], '\'', 'body should be uncompressed')
```
Qard: I'd verify more than just the first character.

watson: If it's gzipped, it will always start with a few magic gzip bytes, so it can never be `'`. But I could just read the entire file into memory up top and do a complete comparison - that way it's good for sure 😃

watson: Hmm actually the downside of that is that the test output will display the binary gzipped data if the test fails when mimic-response isn't correctly patched. And that might mess with your shell. The first magic byte in the gzip header is `\x1f` and will always be displayed as such, so there's no danger of messing up the terminal in that case.

Here's how it looks if I display it all:

image

watson: Continuing the conversation with my self: Actually the dangerous bytes are probably all escaped - just as `\x1f`, so I guess my concern is invalid 😅

watson: Ok, ended up testing just the first line and added a test for the total length as well. That should hopefully cover everything 😃

PR #430, opened 2018-07-02
test: don't patch Module.prototype.require more than once

watson:
Ever since require-in-the-middle allowed multiple hooks to be added in the same app, the `Module.prototype.require` function was patched multiple times if the agent was initialized more than once in the same test file.

This resulted in a lot over overhead every time `require()` was called by any module. This was especially visible in the `disableInstrumentations` tests that would take about a minute to complete because of this.

With this commit, the `disableInstrumentations` tests only take a few seconds.


watson:
Depends on https://github.com/elastic/require-in-the-middle/pull/15 being published

watson:
Updated with new version of require-in-the-middle

PR #435, opened 2018-07-04
chore: ensure rogue node_modules folders aren't published

watson:


watson:
Backstory: I had a `benchmarks` folder in the root of the module locally that I hadn't checked in which contained a lot of benchmark scripts. Any node module that the benchmark scripts used was installed in `benchmarks/node_modules`. Before doing `npm publish` I would first `git add -A && git stash` to ensure none of  the benchmark scripts was included in the published package. But since the `benchmarks/node_modules` folder was excluded by `.gitignore` those wouldn't be stashed. And you'd think that all was stashed because it wouldn't show up when running `npm status`. But npm would still include the entire `benchmarks/node_modules` in the published package (I caught this in time because the new npm version lists the content of the package just as it's about to publish it, so I could abort the publish before it completed)

PR #437, opened 2018-07-06
feat(cassandra): instrumentation

### LINKED ISSUE #303: Cassandra instrumentation

Qard:
This adds instrumentation for the cassandra-driver module. It currently lacks support for the `client.batch(...)` method, and I have some more work to do on supporting older versions of the module.

Fixes elastic/apm-agent-nodejs#303 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


lib/instrumentation/index.js
```diff
@@ -12,7 +12,31 @@ var request = require('../request')
 var Transaction = require('./transaction')
 var shimmer = require('./shimmer')
 
-var MODULES = ['http', 'https', 'http2', 'generic-pool', 'mongodb-core', 'pg', 'mysql', 'mysql2', 'express', 'express-queue', 'hapi', 'redis', 'ioredis', 'bluebird', 'knex', 'koa-router', 'ws', 'graphql', 'express-graphql', 'elasticsearch', 'handlebars', 'mimic-response']
+var MODULES = [
```
watson: 👍👍👍👍👍👍👍👍👍

.travis.yml
```diff
@@ -72,7 +73,7 @@ jobs:
     - stage: dependencies
       node_js: '10'
       if: type IN (cron, pull_request) AND NOT branch =~ ^greenkeeper/.*
-      env: TAV=generic-pool,mysql,mysql2,redis,koa-router,handlebars,mongodb-core
+      env: TAV=generic-pool,mysql,mysql2,redis,koa-router,handlebars,mongodb-core,cassandra-driver
```
watson: We should probably add this to the `ioredis,pg` build as that's currently the one that takes the shortest. This one seems to on average take the longest.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,50 @@
...
+ * TODO: Support batch queries
+ */
+
+var semver = require('semver')
+var sqlSummary = require('sql-summary')
+
+var shimmer = require('../shimmer')
+
+module.exports = function (cassandra, agent, version, enabled) {
+  if (!enabled) return cassandra
```
watson: Does cassandra have a callback queue we need to patch even if disabled?

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,50 @@
...
+
+  function wrapInnerExecute (original) {
+    return function wrappedInnerExecute (...args) {
+      var span = agent.buildSpan()
+      var id = span && span.transaction.id
+
+      if (span) {
+        var query = args[0]
+        span.type = 'db.cassandra.query'
+        span.setDbContext({ statement: query, type: 'sql' })
```
watson: I've never used Cassandra, so excuse my ignorance here... The query language looks like SQL, but they seem to call it CQL. But is it just an alias?

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,50 @@
...
+    return cassandra
+  }
+
+  agent.logger.debug('shimming cassandra.Client.prototype._innerExecute')
+  shimmer.wrap(cassandra.Client.prototype, '_innerExecute', wrapInnerExecute)
+
+  return cassandra
+
+  function wrapInnerExecute (original) {
+    return function wrappedInnerExecute (...args) {
```
watson: The spread operator isn't isn't supported in Node 4 without a flag.

We can drop support for v4, but should bump major if we do. If not too much of a hassle I'd like to wait bumping major until v2 of the intake API. What do you think?

lib/instrumentation/modules/cassandra-driver.js
watson: I just checked with the Python agent and they call it `sql` as well, but I think that might be wrong. I just checked what done in OpenTracing. I found one semi official Cassandra instrumentation and they use type `cassandra`:

https://github.com/opentracing-contrib/java-cassandra-driver/blob/a3d068c8601146dfdaf46557422a55379505d9fb/src/main/java/io/opentracing/contrib/cassandra/TracingSession.java#L319

lib/instrumentation/modules/cassandra-driver.js
Qard: Yeah, I don't need to use the spread operator. I just keep forgetting it didn't exist yet in 4.x. 😅

lib/instrumentation/modules/cassandra-driver.js
Qard: Nope. There's no pooling or callback queue in the driver.

.travis.yml
Qard: Will do. 👍

lib/instrumentation/modules/cassandra-driver.js
Qard: I changed it to `cassandra`.

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed.

.travis.yml
Qard: fixed.

Qard:
Seems a ps1 script is needed to install cassandra on appveyor. I'm not too familiar with how that stuff works. Jenkins also seems to be missing cassandra...or it's somewhere else? 🤔

watson:
@Qard I've updated the PR to install Cassandra on both AppVeyor and Jenkins

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,82 @@
+'use strict'
+
+/**
+ * TODO: Support batch queries
```
watson: Do you want to release without support for batch queries?

.tav.yml
```diff
@@ -128,3 +128,8 @@ hapi-async-await:
   node: '>=8.2'
   versions: '^17.0.0'
   commands: node test/instrumentation/modules/hapi.js
+
+cassandra-driver:
+  versions: '>3.1.0 <4'
```
watson: From the tav tests it looks like 3.1.x isn't supported.

You can run `./node_modules/.bin/tav cassandra-driver ^3.1.0 --compat -- node test/instrumentation/modules/cassandra/cassandra.js` locally to get an overview of what's supported.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+const semver = require('semver')
+const sqlSummary = require('sql-summary')
+
+const shimmer = require('../shimmer')
+
+module.exports = function (cassandra, agent, version, enabled) {
+  if (!enabled) return cassandra
+  if (!semver.satisfies(version, '^3.0.0')) {
```
watson: This semver range doesn't match the one listed in `compatibility.asciidoc`

lib/instrumentation/modules/cassandra-driver.js
watson: Looking further down it seems like support for batch queries have been added. Maybe just old comment?

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,82 @@
...
+    agent.logger.debug('cassandra version %s not supported - aborting...', version)
+    return cassandra
+  }
+
+  if (cassandra.Client) {
+    agent.logger.debug('shimming cassandra.Client.prototype._innerExecute')
+    shimmer.wrap(cassandra.Client.prototype, '_innerExecute', wrapInnerExecute)
+    shimmer.wrap(cassandra.Client.prototype, '_batchCb', wrapBatch)
+  }
```
watson: Consider adding an `else` block with something like:

```js
agent.logger.debug('could not find cassandra.Client in cassandra-driver@%s - aborting...', version)
```

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,82 @@
...
+
+const semver = require('semver')
+const sqlSummary = require('sql-summary')
+
+const shimmer = require('../shimmer')
+
+module.exports = function (cassandra, agent, version, enabled) {
+  if (!enabled) return cassandra
+  if (!semver.satisfies(version, '^3.0.0')) {
+    agent.logger.debug('cassandra version %s not supported - aborting...', version)
```
watson: I would probably write the full module name here (s/cassandra/cassandra-driver/) just to be consistent 😃

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,82 @@
...
+    shimmer.wrap(cassandra.Client.prototype, '_batchCb', wrapBatch)
+  }
+
+  return cassandra
+
+  function wrapBatch (original) {
+    return function wrappedBatch (queries, options, callback) {
+      const spans = []
+
+      for (let query of queries) {
```
watson: Seems kind of wasteful to create an array of queries that are all ended an started at the same time. Does the driver send the queries to the database all in the same request, or is it just an abstraction layer where each query is actually performed independently?

watson:
Besides the `db.cassandra.query` span type that we have here, the Python agent also has a `db.cassandra.connect` span type which is used to indicate when a new socket is opened before a query can be made. Would it make sense for us to create a similar span?

lib/instrumentation/modules/cassandra-driver.js
Qard: Yep. Old comment...removing...

lib/instrumentation/modules/cassandra-driver.js
Qard: done.

lib/instrumentation/modules/cassandra-driver.js
Qard: I'm not really sure how to express the multiple queries in a single span. Ideas?

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,160 @@
...
+      const ret = original.call(this, wrapCallback(callback))
+
+      if (typeof callback !== 'function') {
+        ret.then(resolve, resolve)
+      }
+
+      return ret
+
+      function wrapCallback (cb) {
+        if (!cb) return
```
watson: It might be safest to expand this check to `typeof cb !== 'function'`

watson: Also, it should probably then also return the callback, so it's as if you called `original` with that argument just as if we hadn't patched it.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,160 @@
...
+      span.start()
+
+      function resolve () {
+        span.end()
+      }
+
+      // Wrap the callback
+      const ret = original.call(this, wrapCallback(callback))
+
+      if (typeof callback !== 'function') {
```
watson: It might be safest if we checked if `ret.then` was a function as well. Just in case the API changes. In that case we can't end the span though, so we should consider what to do instead.

The best solution is probably to simply bail in that case and abort the span. I don't think we have a way to do that at the moment, but it should be too hard to add a `span.[cancel|abort|destroy]` function. That function needs to either remove the span from the `transaction._builtSpans` array or update the logic that loops over this array when the transaction ends to ensure that it's ignored instead of being truncated.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,160 @@
...
+      if (!isPromise && spans.length) {
+        arguments[index] = function wrappedCallback () {
+          resolve()
+          return cb.apply(this, arguments)
+        }
+      }
+
+      const ret = original.apply(this, arguments)
+
+      if (isPromise && spans.length) {
```
watson: Just as in the `wrapConnect` function above we should make sure that `ret.then` is actually a function and maybe find ways to cancel the span in case it's not.

lib/instrumentation/modules/cassandra-driver.js
watson: I guess we have the same issue with not ending the span in these edge cases with all the other instrumentations we're doing. So maybe it's out of scope of this instrumentation and we should simply make another issue to improve this.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,160 @@
...
+      }
+
+      span.type = 'db.cassandra.query'
+      span.setDbContext({ statement: query, type: 'cassandra' })
+      span.name = sqlSummary(query)
+      span.start()
+
+      // Wrap the callback
+      const index = arguments.length - 1
+      const hasCallback = typeof arguments[index - 1] === 'function'
```
watson: It took me a while to understand this logic. I think it would make it easier to read if you renamed this variable to `hasRowCallback`.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,160 @@
...
+      const cb = hasCallback ? arguments[index] : () => {}
```
watson: There's check for if `arguments[index]` actually isn't a function in case people use the cassandra-driver API wrong. Should we check for that you think and if so how can we deal with it?

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,160 @@
...
+      if (!isPromise) {
+        arguments[index] = function wrappedCallback () {
+          resolve()
+          return cb.apply(this, arguments)
+        }
+      }
+
+      const ret = original.apply(this, arguments)
+
+      if (isPromise) {
```
watson: Just as in the `wrapConnect` function above we should make sure that `ret.then` is actually a function and maybe find ways to cancel the span in case it's not.

lib/instrumentation/modules/cassandra-driver.js
watson: Do you know if the batch of queries are sent to the database in one request? If so why not simply have a single span? Splitting up the spans seem to indicate multiple requests to the database.

I guess your question is how we then record the query string. This is a bit of a hack, but how about joining all query strings with `;\n` and reporting the resulting string on the batch span? I don't like that solution, but I think I like it more than creating one span per query in the batch. But I might be viewing this from the wrong angle. So I can probably be convinced otherwise.

Qard: Well, they kind of come back off a stream, internally. So it _theoretically_ "starts" and "finishes" the query at a different time, but that's not really usefully visible to the user in any way. Just joining on `;` should be fine. I just wasn't sure if there was any UI implications of doing that, or if we wanted to keep them logically separate for any reason.

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed.

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed.

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed.

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed.

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed.

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,186 @@
...
+      // Wrap the callback
+      const index = arguments.length - 1
+      const hasRowCallback = typeof arguments[index - 1] === 'function'
+
+      function resolve () {
+        span.end()
+      }
+
+      if (hasRowCallback) {
+        const cb = arguments[index]
```
watson: It's technically possible for the user to call the batch function with a proper rowCallback in the 2nd to last position and a non-function argument in the last position. I was wondering if we could deal with that in some way... what do you think?

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,186 @@
...
+    return function wrappedBatch (queries, options, callback) {
+      const span = agent.buildSpan()
+      if (!span) {
+        return original.apply(this, arguments)
+      }
+
+      const queryStrings = queries.map(toQueryString)
+      const querySummaries = queryStrings.map(sqlSummary)
+      const query = queryStrings.join(';\n')
+      const querySummary = querySummaries.join('; ')
```
watson: I'm worried this might become a really long name in case there's a lot of queries.

Possible solutions:

1. Only show the first query summary. This is how we do it for SQL query batches in mysql and postgresql today. But this is actually just out of laziness and should probably be reconsidered anyway
1. Make sure all query summaries are unique, e.g. `querySummaries.filter(unique).join('; ')`
1. Don't show any summary at all and just name the span `Batch Query`

cc @beniwohli

lib/instrumentation/modules/cassandra-driver.js
```diff
@@ -0,0 +1,186 @@
...
+
+  function wrapConnect (original) {
+    return function wrappedConnect (callback) {
+      const span = agent.buildSpan()
+      if (!span) {
+        return original.apply(this, arguments)
+      }
+
+      span.type = 'db.cassandra.connect'
+      span.start()
```
watson: This span isn't named (it will fall back to the default name `unnamed`). We should coordinate with the Python agent and agree on a name.

Btw, you can just give the name and type as arguments to `start()` instead of setting them as properties, eg:

```js
span.start('Cassandra: Connect', 'db.cassandra.connect')
```

cc @beniwohli

test/instrumentation/modules/cassandra/cassandra.js
```diff
@@ -0,0 +1,239 @@
...
+  const trans = data.transactions[0]
+  t.equal(trans.name, 'foo', 'transaction name')
+  t.equal(trans.spans.length, 2, 'span count')
+
+  assertConnectSpan(t, trans.spans[0])
+  assertSpan(t, sql, summary, trans.spans[1])
+}
+
+function assertConnectSpan (t, span) {
+  t.equal(span.type, 'db.cassandra.connect', 'span type')
```
watson: Add a check for the `span.name` as well

lib/instrumentation/modules/cassandra-driver.js
Qard: What would we do in that situation? It's an invalid input, and I'm not sure how we could recover from that other than just leaving the span un-ended. 🤔

lib/instrumentation/modules/cassandra-driver.js
Qard: Yeah, I think something like just `BATCH` is better for the name than joining all the queries. It gets long quickly, even summarized.

lib/instrumentation/modules/cassandra-driver.js
Qard: I wasn't sure what to call it, since there's no query to summarize for the name. I'm good with whatever. 🤷‍♂️

test/instrumentation/modules/cassandra/cassandra.js
Qard: Will do, when we decide what the name should be. 👍

lib/instrumentation/modules/cassandra-driver.js
watson: No we can't recover, and in we probably need a general way to "destroy" a span if we find that we can't detect when it ends. But adding that ability to spans is probably out of scope for this PR.

In regards to invalid input in general, I try to just forward that invalid input to the module being patched so that it can use its normal error handling.

lib/instrumentation/modules/cassandra-driver.js
watson: Any reason why you want to uppercase it?

Qard: Not really. Sometimes people capitalize keywords in SQL, and I was thinking of it kind of like that, but whatever works. 🤷‍♂️

Qard: Makes sense. Maybe just `batch query`?

watson: We have previously prefixed certain span names with the module that created them, e.g. Elasticsearch span names are prefixed `Elasticsearch: `. For SQL queries we don't normally do this as it's usually pretty obvious where they are coming from, but just as with Elasticsearch we might consider prefixing Cassandra spans as well. If so it would be something like `Cassandra: Batch query`. What do you think?

lib/instrumentation/modules/cassandra-driver.js
beniwohli: in this case, we fall back to the class/method name that we wrap, e.g. `Cluster.connect`

watson: The Python agent prefixes spans with the name of the class and function that was used to start the span. So in the case of here the span would be called `Cluster.connect` in Python. I think we for now just should call it `Cassandra: Connect`. That aligns with how we name the Elasticsearch queries. Aligning across agents will be a bigger task and should probably not hold back this PR.

Qard: fixed

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed

lib/instrumentation/modules/cassandra-driver.js
Qard: fixed

test/instrumentation/modules/cassandra/cassandra.js
Qard: fixed

Qard:
Rebased to deal with conflicts from MSSQL PR.

PR #442, opened 2018-07-11
perf: don't patch newer versions of mimic-response

watson:
Version 1.0.1 of mimic-response have just been released with the following fix:

https://github.com/sindresorhus/mimic-response/pull/1

This fix makes our own patch redundant, so it's no longer necessary for us to patch this or newer versions. We should still continue to patch version 1.0.0 however.

### DIFF
--- a/lib/instrumentation/modules/mimic-response.js
+++ b/lib/instrumentation/modules/mimic-response.js
@@ -1,9 +1,16 @@
 'use strict'
 
+var semver = require('semver')
+
 var shimmer = require('../shimmer')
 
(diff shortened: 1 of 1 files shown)


watson:
I guess one could argue that we shouldn't patch version 1.0.0 at all now that 1.0.1 is out as we normally don't patch other peoples modules just because they contain bugs. But on the other hand, this bug actually ment that users apps would crash because of the agent, so in that case I think we can make an exception. This is also why I think it's in our interest to keep mimic-response in our TAV test matrix in case new versions of the module reintroduces the issue.

PR #449, opened 2018-07-17
docs: remove reference to Node 10 incompatibility

### LINKED ISSUE #448: Status for Node 10 ?
The Compatibility Overview says that node v10.x is buggy and not supported. There is a reference to an issue in the node repo, but that issues seems to be closed.

Are we tracking it?, and is node v10.x still not supported?

watson:
The bug in V8 that meant we couldn't instrument Node 10 was fixed and Node 10.4.0 shipped with a newer version of V8.

This is the original bug report in Node: https://github.com/nodejs/node/issues/20516
This is the PR that fixed it in Node: https://github.com/nodejs/node/pull/19989

Closes #448

### DIFF
--- a/test/instrumentation/modules/hapi.js
+++ b/test/instrumentation/modules/hapi.js
@@ -13,8 +13,8 @@ var semver = require('semver')
 // hapi 17+ requires Node.js 8.9.0 or higher
 if (semver.lt(process.version, '8.9.0') && semver.gte(pkg.version, '17.0.0')) process.exit()
 
-// hapi does not work on early versions of Node.js 10 because of https://github.com/nodejs/node/issues/20516
-if (semver.satisfies(process.version, '>=10.0.0-rc <10.4.0')) process.exit()
+// hapi does not work on Node.js 10 because of https://github.com/nodejs/node/issues/20516
+if (semver.gte(process.version, '10.0.0-rc')) process.exit()
 
 var http = require('http')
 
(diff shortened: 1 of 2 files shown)


test/instrumentation/modules/hapi.js
```diff
@@ -13,9 +13,6 @@ var semver = require('semver')
 // hapi 17+ requires Node.js 8.9.0 or higher
 if (semver.lt(process.version, '8.9.0') && semver.gte(pkg.version, '17.0.0')) process.exit()
 
-// hapi does not work on Node.js 10 because of https://github.com/nodejs/node/issues/20516
-if (semver.gte(process.version, '10.0.0-rc')) process.exit()
```
Qard: Rather than fully removing this, can we change it to a range to indicate that certain early versions of 10.x _are_ broken?

watson: Good idea. Fixed

Qard:
Looks like the tests are failing on 10.x. 🤔

watson:
I think it's "just" hapi behaving different under Node 10 for some reason. As far as I can see hapi is emitting some tags now that it wasn't doing under Node 9 when running our tests. Those are caught as errors and that causes our tests to fail. At least that's what I get from looking at the first failure. The other failures might be different.

I will have to look into if this is expected behavior and if so update the tests to expect it. This can take a while unfortunately.

watson:
I'll merge this even though Jenkins is acting up. It's unrelated.

PR #462, opened 2018-07-20
feat(agent): add currentTransaction getter

### LINKED ISSUE #436: Add API to get current transaction
Today it's only possible to get the current transaction if you start it your self using [`agent.startTransaction()`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-start-transaction). But transactions started automatically by the agent it self (because of an incoming HTTP request) aren't accessible to the user.

Normally the user doesn't need access to these transactions, so it has never been an issue. But in case the user wants to for instance access the current transaction ID for logging purposes, this would be impossible.

I therefore suggest we add an API to make this possible.

It could either be a function:

```js
agent.currentTransaction()
```

Or simply a getter created using `Object.defineProperty`:

```js
agent.currentTransaction
```

The user should still take care to guard against `null` values as there might not always be a current transaction.

Qard:
Fixes #436 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -1032,6 +1032,20 @@ Alternatively you can call <<transaction-end,`end()`>> directly on an active tra
 
 A value may be provided to set as a result.
 
+[[apm-current-transaction]]
+==== `apm.currentTransaction`
+
+[[source,js]]
+----
+var transaction = apm.currentTransaction
+----
+
+Get the currently active transaction,
+if used within the context of a transaction.
+
+NOTE: If there's no active transaction available,
+`null` will be returned.
+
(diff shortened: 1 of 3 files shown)


watson:
I'm always in two minds about these getters (and setters). With a function call you know that it involves computation, but with a getter, you think that it's just a property and the cost of getting its value it is "zero".

What's the reason you went with a getter in this case?

watson:
I like that way of thinking of it. What do you think of rewording the documentation to not use "function language"? E.g. currently you say that it "returns" something and that "if executed" it does something.

Qard:
Fixed.

PR #465, opened 2018-07-23
feat: support APM Server intake API version 2

### LINKED ISSUE #356: Intake Protocol V2 POC
POC using the new Intake Protocol V2 from https://github.com/elastic/apm-server/pull/930

watson:
Closes elastic/apm-agent-nodejs#356

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Fix TAV tests


watson:
This PR should technically be ready for testing with the POC. The checklist above is meant for making it production ready

roncohen:
would be great to get some benchmarks to see if this implementation is better, for example from a  memory usage perspective.

lib/agent.js
```diff
@@ -281,8 +306,24 @@ Agent.prototype.captureError = function (err, opts, cb) {
...
+
+    truncate.error(error, agent._conf)
+
+    agent.logger.info(`Sending error ${id} to Elastic APM`)
+    if (agent._apmServer) {
+      agent._apmServer.sendError(error, function () {
+        agent._apmServer.flush(cb)
+      })
+    } else {
+      process.nextTick(cb)
```
Qard: Can you log something here so we can tell when someone tries to send something without `_apmServer` set?

lib/agent.js
```diff
@@ -308,7 +349,11 @@ Agent.prototype.handleUncaughtExceptions = function (cb) {
 }
 
 Agent.prototype.flush = function (cb) {
-  this._instrumentation.flush(cb)
+  if (this._apmServer) {
+    this._apmServer.flush(cb)
+  } else {
+    process.nextTick(cb)
```
Qard: Again, log something so we know when `_apmServer` is not set.

lib/instrumentation/transaction.js
```diff
@@ -50,6 +49,15 @@ function Transaction (agent, name, type) {
     }
   })
 
+  Object.defineProperty(this, 'timestamp', {
+    get: function () {
```
Qard: Could use shorthand here. `get () {`

lib/instrumentation/transaction.js
```diff
@@ -112,17 +121,16 @@ Transaction.prototype.buildSpan = function () {
   }
 
   if (this.ended) {
-    this._agent.logger.debug('transaction already ended - cannot build new span %o', {id: this.id})
+    this._agent.logger.debug('transaction already ended - cannot build new span %o', {id: this.id}) // TODO: Should this be supported in the new API?
```
Qard: When everything is fully streaming, I don't think there's any _reason_ to prevent further spans. It's only useful _without_ streaming to ensure the transaction doesn't stay open forever.

watson: The reason why I'm hesitant with this is that we might have a transaction that kicks off some form of background job that gets contextually attached to the transaction that started it. Any spans that's created as a result of this would get associated with the transaction. This will make the span waterfall almost useless as it can suddenly cover a period of days or weeks. Am I being too paranoid?

lib/agent.js
watson: fixed

lib/agent.js
watson: fixed

lib/instrumentation/transaction.js
watson: fixed

watson:
To make the scope of this PR as small as possible, I've made it go to a different branch than `master`. This way we can merge it when we're ok with the current scope of it and add new v2 features in new PR's to the same branch. Then when we're ready to land support for v2 in `master`, we make a PR that merges the v2 branch into master

watson:
To help this PR getting merge, I've created the meta issue #471 and moved all the non-essential todo items from this PR to that

lib/instrumentation/transaction.js
Qard: That'd be user mode queuing though. We won't actually form that link unless we explicitly bind the callback. The logical continuation to the next job would be out-of-band from the request, it'd be triggered by the completion of another job in the queue.

I could see _maybe_ time-based jobs naively using `setTimeout` for everything, and _that_ maybe being an issue. But that's just not great design and we probably _should_ be drawing attention to it in some way. 🤔

watson: Yeah `setTimeout` and friends is probably the most common way to trigger this.

Question is however if the spans happening after the transaction ends are relevant to the user at all?

@elastic/apm-agent-devs @roncohen @makwarth In the new intake API, we get the ability to record spans that start after the parent transaction ends. Is this something we want to do or should those just be ignored? I.e. are they adding any value to the user or are they just noise?

felixbarny: > I.e. are they adding any value to the user or are they just noise?

I think that depends on how the application is making use of that. I'd rather default to not ignore than to ignore these spans.

watson: If we don't ignore them by default, would it be a good idea to have a config option to ignore them?

It could even have 3 settings:

- Ignore spans started after the transaction ends
- Ignore spans started before the transaction ends, but ended after
- Allow all spans

roncohen: i understand you're worried about the case where people would set up a recurring periodic job from inside a transaction, because it would mean spans continue to happen in an endless stream. Two ideas:

* We could experiment in making the assumption that there are two categories of delayed execution `process.nextTick()/setImmediate()` and `setTimeout()/setInterval()`. The first one could indicate that the developer intended this to be part of the transaction, while the second category will start a new transaction.
* We could set a numerical limit to the number of spans that we will consider, after a transaction has ended - or a wall clock timeout which would be the limit to how much time we'll continue to "monitor" a transaction after is has ended.

Qard: We could also include a special escape hatch that specifically calls to `setTimeout` or `setInterval` _after_ the transaction ends would _not_ continue the context. That way, anything _else_ in the code path will continue, but anything that could potentially be significantly delayed will _not_.

watson:
jenkins, run tav tests please

lib/instrumentation/index.js
```diff
@@ -106,27 +90,44 @@ Instrumentation.prototype._patchModule = function (exports, name, version, enabl
...
 Instrumentation.prototype.addEndedTransaction = function (transaction) {
+  var agent = this._agent
+
   if (this._started) {
-    var queue = this._queue
+    var payload = agent._filters.process(transaction._encode()) // TODO: Update filter to expect this format
+    if (!payload) return agent.logger.debug('transaction ignored by filter %o', {id: transaction.id})
+    truncate.transaction(payload)
+    agent.logger.debug('sending transaction %o', {id: transaction.id})
+    if (agent._apmServer) agent._apmServer.sendTransaction(payload)
```
Qard: Is there any way for `agent._apmServer` to be null when `this._started` is true? 🤔

lib/request.js
```diff
@@ -1,254 +0,0 @@
...
-      ppid: process.ppid,
-      title: truncate(String(process.title), config.INTAKE_STRING_MAX_SIZE),
-      argv: process.argv
-    },
-    system: {
-      hostname: agent._conf.hostname,
-      architecture: process.arch,
-      platform: process.platform
-    }
-  }
```
Qard: It seems like process, system and runtime stuff no longer exists? I don't see it in the http client PR either. Is that correct?

test/agent.js
```diff
@@ -186,138 +186,121 @@ test('#addTags', function (t) {
...
+          t.pass('called callback')
         })
       })
-      .on('request', validateErrorRequest(t))
-      .on('body', function (body) {
-        t.equal(body.errors.length, 1)
-        t.equal(body.errors[0].exception.message, 'with callback')
+      .on('data-error', function (data) {
+        t.equal(data.exception.message, 'with callback')
+        t.end()
```
Qard: Does tape work correctly when you explicitly `t.end()` while using a `t.plan(...)`?

test/instrumentation/_agent.js
```diff
@@ -54,7 +49,7 @@ module.exports = function mockAgent (cb) {
     sharedInstrumentation._agent = agent
     agent._instrumentation = sharedInstrumentation
     agent._instrumentation.currentTransaction = null
-    agent._instrumentation._queue._clear()
+    agent._apmServer = mockClient(expected, cb || noop) // TODO: Expected will not work here
```
Qard: Is this comment still accurate? What's the issue?

test/instrumentation/span.js
```diff
@@ -137,19 +135,16 @@ test('#_encode() - ended named', function myTest2 (t) {
...
+  span.end()
   span._encode(function (err, payload) {
     t.error(err)
-    t.deepEqual(Object.keys(payload), ['name', 'type', 'start', 'duration', 'stacktrace'])
-    t.equal(payload.name, 'foo')
-    t.equal(payload.type, 'bar.truncated')
-    t.ok(payload.start > 0)
-    t.ok(payload.duration > 0)
-    assert.stacktrace(t, 'myTest3', __filename, payload.stacktrace, agent)
+    t.deepEqual(Object.keys(payload), ['transactionId', 'timestamp', 'name', 'type', 'start', 'duration'])
```
Qard: You don't think we need to validate the contents of the object here?

test/integration/no-sampling.js
```diff
@@ -16,17 +17,23 @@ getPort().then(function (port) {
...
+      })
+      req.on('end', function () {
+        t.equal(received.metadata, 1, 'expected 1 metadata to be sent')
+        t.equal(received.transaction, 20, 'expected 20 transactions to be sent')
         res.end()
         server.close()
-        var data = JSON.parse(Buffer.concat(buffers))
-        t.equal(data.transactions.length, 20, 'expect 20 transactions to be sent')
         t.end()
+        process.exit()
```
Qard: Does this `process.exit()` need to be here? I'd rather not have these hiding in the tests as they could cause confusion when adding more tests to the file in the future, if you don't notice them.

lib/instrumentation/index.js
watson: No you're right - if `this._started === true`, then we'll also have an `agent._apmServer`. Would that be a better check? The current check is guaranteed to always work, whereas the conditions of the other might change in the future right?

lib/request.js
watson: Yeah, I can see it's a bit confusing given the current state of the all the in-progress dependencies.

As you can see here, this PR depends on a custom branch of the http client called `v2-1` that currently lives on my fork: https://github.com/elastic/apm-agent-nodejs/pull/465/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R74

In that branch the metadata responsibilities have been moved to the http client: https://github.com/watson/apm-nodejs-http-client/blob/4f02d7eda173b738a309fdca2e3464a1bbae5652/index.js#L293-L332

But there's no PR for this branch yet as I don't want to make the current http client PR too complicated, so once the the current PR is merged, I'll open a new one for the `v2-1` branch 😅

test/agent.js
watson: Yes, all it does when reaching the `t.end` is just fail if the number of assertions doesn't match what's planned. This is a way to to check our exceptions of async execution order. I.e. we reach this point, we should have made the planned number of assertions.

test/instrumentation/_agent.js
watson: Hmm I think you're right. I think this line is completely irrelevant and can just be removed. I'll update it

lib/agent.js
```diff
@@ -314,7 +355,12 @@ Agent.prototype.handleUncaughtExceptions = function (cb) {
 }
 
 Agent.prototype.flush = function (cb) {
-  this._instrumentation.flush(cb)
+  if (this._apmServer) {
+    this._apmServer.flush(cb)
+  } else {
+    this.logger.warn(new Error('cannot flush agent before it is started'))
+    process.nextTick(cb)
```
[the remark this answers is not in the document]
watson: I left it out as this only occurs if the agent isn't started. So from the outside I want the user to use the module in the same way no matter if the agent is started or not. And normally errors coming from the `_apmServer` is just logged, so I think it's best to maintain that behavior here as well.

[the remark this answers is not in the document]
watson: Actually I just realized that the callback given to `captureError` will pass a similar error along in the callback: https://github.com/elastic/apm-agent-nodejs/pull/465/files#diff-7403da6ce2244c65d641fdfcc095be93R331

But I think maybe that should be avoided in `captureError`. If so the `captureError` callback will never be called with an error.

test/integration/no-sampling.js
watson: Fixed

test/instrumentation/span.js
watson: Fixed

test/instrumentation/_agent.js
watson: Removed

test/_apm_server.js
```diff
@@ -40,26 +53,25 @@ function APMServer (agentOpts, mockOpts) {
...
-        var body = JSON.parse(Buffer.concat(buffers))
-        if (mockOpts.forwardTo) {
-          var client = new HttpClient({
-            serverUrl: mockOpts.forwardTo,
-            userAgent: 'elastic-apm-node/' + pkg.version
-          })
-          client.request('transactions', {}, body, () => {})
+      req.pipe(zlib.createGunzip()).pipe(ndjson.parse()).on('data', function (data) {
+        if (req.method !== 'POST') throw new Error(`Unexpected HTTP method: ${req.method}`)
+        if (req.url !== '/v2/intake') throw new Error(`Unexpected HTTP url: ${req.url}`)
```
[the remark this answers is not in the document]
watson: Fixed

lib/config.js
```diff
@@ -28,7 +28,9 @@ var DEFAULTS = {
   verifyServerCert: true,
   active: true,
   logLevel: 'info',
-  hostname: os.hostname(),
+  hostname: os.hostname(), // TODO: Should we just let the http client default to this?
```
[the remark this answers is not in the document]
watson: Removed

lib/request.js
watson: Just merge the first http client PR and created the next based on my `v2-1` branch: https://github.com/elastic/apm-nodejs-http-client/pull/6 - soon it should hopefully be less confusing

watson:
I think I fixed the tav tests now. The remaining tav failures are because of a bug in the version of npm that currently ships with node 8+9, where running `npm install` in a location where there's already node modules installed but one of those installed node modules is defined in the package.json as a github link, will simply delete that module.

It's been fixed in newer npm releases and [there's a PR](https://github.com/nodejs/node/pull/21593) to upgrade the npm version in Node 8, but Node 9 will obviously never get fixed.

I could manually upgrade npm in `.travis.yml` in this PR, but we don't want that as soon as the the http client (which we currently depend upon using such a github link in package.json) is published to npm.

We could alternatively "fix" this by simply releasing a tagged version to npm, but that will make it harder to iterate and might be overkill.

watson:
jenkins, run tav tests please

lib/instrumentation/index.js
Qard: I just made that comment, because this check is already in an outer `if` block checking `this._started`, so it seemed redundant. Github cuts off the preview just after that line though. 😅

lib/request.js
Qard: Ah. Gotcha. I was trying to match up the two PRs to make sure the functionality lined up between the two. 😅

test/agent.js
Qard: Ah, cool. 👍

lib/instrumentation/index.js
watson: Oh yes... I'll remove it

watson: Fixed

Qard:
Approved for v2 branch. Any remaining issues can be fixed in another PR, before merging v2 to master.

PR #466, opened 2018-07-23
test: ensure APM Server 503's are handled

watson:


Qard:
Wouldn't this end the test before the error is actually handled by the agent? 🤔

watson:
AFAIK `server.close()` just means that the server stops accepting new requests and is unreffed. But any remaining requests will finish. So the entire script should run to completion - including the handling the response on the client side. But I'll just investigate a little further...

watson:
Just did a little testing and my hypophysis was correct. The agent will output the expected error from `lib/request.js` no matter if I keep the code as is, or close the server and end the test after a delay:

https://github.com/elastic/apm-agent-nodejs/blob/25901682771bff7b765a9092bb4d056146d398cb/lib/request.js#L41

I tried both this change:

```diff
-      server.close()
-      t.end()
+      req.on('end', function () {
+        server.close()
+        t.end()
+      })
```

And this one:

```diff
-      server.close()
-      t.end()
+      req.on('end', function () {
+        setTimeout(function () {
+          server.close()
+          t.end()
+        }, 100)
+      })
```

All 3 produce the same result.

watson:
Hmm come to think of it, this might actually make it more readable:

```diff
-      server.close()
-      t.end()
      })
 
     server.listen(port, function () {
-      agent.captureError(new Error('foo'))
+      agent.captureError(new Error('foo'), function (err) {
+        t.error(err)
+        t.end()
+        server.close()
+      })
```

Do you like that better?

Qard:
Yeah, I was just worried about the `t.end()` running before the request actually finishes on the client end.

PR #467, opened 2018-07-23
test: ensure APM Server socket close is handled

watson:


Qard:
Same as my question in #466 -- wouldn't this end the test before the agent gets a chance to recover? 🤔

watson:
As I commented on #466 this would technically work, but I agree it could be more readable. I've updated it 😃

PR #472, opened 2018-07-25
test(docker): move docker files into test directory

### LINKED ISSUE #438: Move docker files from root into test directory
There's no reason for the `docker-compose.yml` file to be in the root directory as it's only used for testing (the Python agent already have it in its test directory).

Qard:
Fixes #438


Qard:
Actually, I don't think the `.dockerignore` _does_ anything anymore, since it no longer uses `ADD`, it mounts a volume instead. (Because `ADD` doesn't allow paths above the location of the `Dockerfile`)

watson:
If so it should just be removed 😃 @simitt does that sound right?

Qard:
Updated `.npmignore` and removed `.dockerignore`. If we still want that file though, I can restore it.

simitt:
From how I understood `.dockerignore` files, completely agree with @Qard. I only added that because of the `ADD`, thus no need for it anymore.

watson:
jenkins, run the tav tests please

PR #483, opened 2018-07-28
test: refactor how to run tests locally

### LINKED ISSUE #482: Make it easy to run start test dependencies in Docker locally
Today the `test/docker-compose.yml` file will start both all the test dependencies (databases) + run the test suite it self inside its own container.

It would be much quicker and in may ways easier if the tests were run locally on the dev machine and only the databases were run inside docker containers

watson:
Closes elastic/apm-agent-nodejs#482


watson:
@Qard I got frustrated about how long it would take me to start the test suite when using Docker, so I made this 😄

This change makes it super easy for us to just run SQL Server in Docker and everything else locally (if you like to get the added performance).

1. First start all databases locally (excluding SQL Server):
    ```
    npm run local-start
    ```
1. Then run the tests signaling that you want to run start SQL Server in Docker:
    ```
    npm test mssql
    ```

Alternative approach:

1. First start all databases locally (excluding SQL Server):
    ```
    npm run local-start
    ```
1. Then start SQL Server in Docker:
    ```
    npm run docker-start mssql
    ```
1. Run test suite without any dependencies:
    ```
    npm test --no-deps
    ```

If you want to run all the databases in Docker, but still want to run the tests locally, simply run:

```
npm test
```

If you want to pre-load all the databases in Docker:

```
npm run docker-start
```

After this you can just run `npm test` and it will discover that the databases are already started in Docker and not try to start them again. It will not try to shut them down either.

.npmignore
```diff
@@ -5,6 +5,7 @@
 .travis.yml
 CONTRIBUTING.md
 CODE_OF_CONDUCT.md
+TESTING.md
```
Qard: Missing a T.

package.json
```diff
@@ -9,16 +9,21 @@
     "lint-fix": "standard --fix",
     "lint-commit": "test/lint-commits.sh",
     "check": "standard && npm run test-deps && npm run lint-commit",
-    "precoverage": "test/pretest.sh",
-    "coverage": "npm run check && nyc node test/test.js",
-    "postcoverage": "test/posttest.sh",
+    "coverage": "test ! -z $CI && npm run test-suite-coverage || COVERAGE=true ./test/script/docker/run_tests_dev.sh",
```
Qard: I'd prefer if we could encode this sort of logic in the shell scripts rather than having this much complexity in npm scripts.

TESTING.md
```diff
@@ -0,0 +1,111 @@
...
+
+| Database      | Service Name |
+|---------------|--------------|
+| PostgreSQL    | `postgres` |
+| MySQL         | `mysql` |
+| MongoDB       | `mongodb` |
+| Elasticsearch | `elasticsearch` |
+| Cassandra     | `cassandra` |
+| SQL Server    | `mssql` |
+| Redis         | `redis` |
```
Qard: Do you think it's important to include this list here? It's another place we need to remember to keep in-sync with reality, as we add new instrumentation.

TESTING.md
```diff
@@ -0,0 +1,111 @@
...
+```
+npm test -- --no-deps
+```
+
+If you only have a subset of the required databases installed and running locally,
+run `npm test` with a list of the datases you want it to start for you,
+e.g:
+
+```
+npm test mssql cassandra
```
Qard: Shouldn't this be `npm test -- mssql cassandra`?

TESTING.md
watson: Good point. We could just say that you have to choose one or more of the services you get by running:

```
docker-compose -f test/docker-compose.yml config --services
```

That will generate this output:

```
postgres
mongodb
mssql
mysql
redis
elasticsearch
cassandra
node_tests
```

TESTING.md
watson: No, the only reason why the `--` is needed in the other commands is because `--docker` and `--no-deps` starts with `--` it self. Then npm thinks that it's a config option for npm and will not pass it on to the script. Therefore I was actually thinking about changing `--docker` to `all` and `--no-deps` to just `none`. That way you'd not have to remember to add the extra `--`. What do you think?

test/script/docker/run_tests.sh
```diff
@@ -17,9 +17,7 @@ docker_nyc_report_output="/app/.nyc_output"
 NODE_VERSION=$1
 if [[ ! -z $2  ]]; then
   TAV_VERSIONS=`echo "$2" | sed -e 's/\+/,/g'`
-  CMD='npm i test-all-versions &&
-       export PATH=./node_modules/.bin:$PATH &&
-       tav --quiet'
+  CMD='npm run tav --quiet'
```
Qard: Shouldn't this be `test-tav`?

TESTING.md
Qard: Sounds good to me. 👍

TESTING.md
Qard: Ah, I didn't know npm did that. 😅 

I like the idea of using `all` and `none` instead of the flags.

package.json
watson: I actually had this logic inside the script originally, but moved it out here because it felt wrong to run a script meant for development in CI, only to have it relay the call to the CI script and relay the exit code. But maybe I could make a "routing" script to figure out which sub-script to run based on the environment? Or I could just merge the two scripts together? What do you think?

test/script/docker/run_tests.sh
watson: Fixed

TESTING.md
watson: Updated to use `all` and `none`

TESTING.md
watson: Fixed

watson:
jenkins, run tav tests please

package.json
watson: Fixed

watson:
jenkins, run tav tests please

Qard:
This breaks appveyor. 😿

watson:
jenkins, run tav tests please

PR #484, opened 2018-07-31
fix(graphql): handle execute args object

bausmeier:
This PR fixes handling of arguments passed to the graphql execute function by adding support for passing the arguments as an object. This is [how the latest version of apollo does it](https://github.com/apollographql/apollo-server/blob/e77701c77253db007171a2437fe361c5bb2c2312/packages/apollo-server-core/src/runQuery.ts#L232) and the implementation is based on [how this is handled in the graphql module itself](https://github.com/graphql/graphql-js/blob/3124fa32dc77c5844bf74ebc07e38239ad78e856/src/execution/execute.js#L164-L183).

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/modules/graphql.js
+++ b/lib/instrumentation/modules/graphql.js
@@ -95,7 +95,7 @@ module.exports = function (graphql, agent, version, enabled) {
   }
 
   function wrapExecute (orig) {
-    return function wrappedExecute (schema, document, rootValue, contextValue, variableValues, operationName) {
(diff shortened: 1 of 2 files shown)


watson:
Thanks a lot for the fix 💯

bausmeier:
You're welcome. Thanks for the module!

watson:
Published to npm as version 1.10.1

PR #485, opened 2018-08-01
feat(filters): filter set-cookie headers

### LINKED ISSUE #468: Automatic filtering of Set-Cookie
Today we automatically filter sensitive data from the `Cookie` header. We should do the same with the `Set-Cookie` header

Qard:
Fixes elastic/apm-agent-nodejs#468

Depends on Qard/set-cookie-serde/pull/1

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/filters.js
+++ b/lib/filters.js
@@ -2,6 +2,7 @@
 
 var cookie = require('cookie')
 var redact = require('redact-secrets')('[REDACTED]')
+var SetCookie = require('set-cookie-serde')
 
 var REDACTED = '[REDACTED]'
 
@@ -32,6 +33,12 @@ Filters.prototype.process = function (payload) {
   return result
 }
 
+function stringify (value) {
+  return Array.isArray(value)
+    ? value.map(value => value.toString())
+    : value.toString()
+}
+
 function httpHeaders (payload) {
   var arr = payload.transactions || payload.errors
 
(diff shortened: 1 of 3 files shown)


watson:
Note: The Node 8+9 tav tests are currently failing because of a bug in npm triggered by the use of a github dependency link.

lib/filters.js
```diff
@@ -51,6 +58,17 @@ function httpHeaders (payload) {
         .map(function (k) { return k + '=' + cookies[k] })
         .join('; ')
     }
+
+    if (typeof headers['set-cookie'] !== 'undefined') {
```
watson: I don't think we need to protect our selfs from people overwriting the `undefined` property right? If so you can just do a simple comparison instead. Unless there's an edge case I'm not considering?

lib/filters.js
```diff
@@ -51,6 +58,17 @@ function httpHeaders (payload) {
...
+      try {
+        var setCookies = new SetCookie(headers['set-cookie'])
+        redact.forEach(setCookies)
+        headers['set-cookie'] = stringify(setCookies)
+      } catch (err) {
+        // Ignore error
+        headers['set-cookie'] = '[malformed set-cookie header]'
```
watson: Are we sure we want to overwrite it in this case? I guess it's to prevent leaking secrets, but it does make it hard to fix it if the user never sees how it looked 🤔

lib/filters.js
Qard: You mean do `if (headers['set-cookie'] !== undefined)` instead?

watson: yep

lib/filters.js
watson: @elastic/apm-agent-devs @roncohen: Should we redact the entire `Set-Cookie` header if we can't parse it or just report it as is?

felixbarny: In the Java agent, we always redact the whole value. I’m not sure if the overhead of parsing the value is worth it.

lib/filters.js
Qard: fixed

package.json
```diff
@@ -84,6 +84,7 @@
     "require-ancestors": "^1.0.0",
     "require-in-the-middle": "^3.1.0",
     "semver": "^5.5.0",
+    "set-cookie-serde": "github:qard/set-cookie-serde#implementation",
```
watson: This PR is marked as "In Review", but I guess it should still be "In Progress" until this dependency have been published right?

PR #490, opened 2018-08-03
test(api-schema): fix to support paths with spaces

Qard:
The api-schema tests break if the path to the repo contains a directory with a space in the name.

### DIFF
--- a/test/instrumentation/modules/http/_echo_server_util.js
+++ b/test/instrumentation/modules/http/_echo_server_util.js
@@ -7,7 +7,8 @@ exports.echoServer = echoServer
 
 function echoServer (type, cb) {
   if (typeof type === 'function') return echoServer('http', type)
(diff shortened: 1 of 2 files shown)


test/instrumentation/modules/http/_echo_server_util.js
```diff
@@ -7,7 +7,8 @@ exports.echoServer = echoServer
 
 function echoServer (type, cb) {
   if (typeof type === 'function') return echoServer('http', type)
-  var cp = exec('node ' + path.join(__dirname, '/_echo_server.js ' + type))
+  var script = path.join(__dirname, '/_echo_server.js')
```
watson: Just noticed an unrelated bug here: The leading `/` in `/_echo_server.js`. It's unrelated to your change, but since you were changing things related to the path anyway, lets just get this fixed as well

PR #493, opened 2018-08-04
test(hapi): reenable for 10.8+

### LINKED ISSUE #455: Re-enable hapi tests in Node.js 10
We disabled hapi tests in Node 10 because of a bug in Node core/V8. This should have been fixed since Node 10.4.0, but if we reenable the tests, they start failing. As far as I can see this is for different reasons and might just be because hapi behaves slightly different on Node 10 (and our test suite expects it to behave like normal). So we'll have to go over each failing test and see if it's expected behavior and if so just fix the tests.

Qard:
Fixes elastic/apm-agent-nodejs#455

### DIFF
--- a/test/instrumentation/modules/hapi.js
+++ b/test/instrumentation/modules/hapi.js
@@ -14,7 +14,7 @@ var semver = require('semver')
 if (semver.lt(process.version, '8.9.0') && semver.gte(pkg.version, '17.0.0')) process.exit()
 
 // hapi does not work on Node.js 10 because of https://github.com/nodejs/node/issues/20516
-if (semver.gte(process.version, '10.0.0-rc')) process.exit()
+if (semver.satisfies(process.version, '>=10.0.0-rc <10.8')) process.exit()
 
 var http = require('http')
 


watson:
What's the reason the versions between 10.4 and 10.8 were still failing?

test/instrumentation/modules/hapi.js
```diff
@@ -14,7 +14,7 @@ var semver = require('semver')
 if (semver.lt(process.version, '8.9.0') && semver.gte(pkg.version, '17.0.0')) process.exit()
 
 // hapi does not work on Node.js 10 because of https://github.com/nodejs/node/issues/20516
-if (semver.gte(process.version, '10.0.0-rc')) process.exit()
+if (semver.satisfies(process.version, '>= 10.0.0-rc < 10.8')) process.exit()
```
watson: I think it's more readable if it's written in the "official" way without the whitespace, i.e. `>=10.0.0-rc <10.8`

Qard: fixed

PR #497, opened 2018-08-06
docs: use relative links in markdown files

watson:


Qard:
An interesting side effect of this is that the links point to docs from the same commit hash.

PR #498, opened 2018-08-06
refactor: remove built in payload truncation logic

watson:
Changes for intake API v2.

Depends on https://github.com/elastic/apm-nodejs-http-client/pull/8

This will instead be handled by the elastic-apm-http-client module when the above PR have been merged.


test/instrumentation/modules/hapi.js
```diff
@@ -37,7 +37,7 @@ test('extract URL from request', function (t) {
     t.equal(request.url.search, '?foo=bar')
     t.equal(request.url.raw, '/captureError?foo=bar')
     t.equal(request.url.hostname, 'localhost')
-    t.equal(request.url.port, String(server.info.port))
```
watson: 🤔 Previously, the act of truncating this property would convert it to a string. Now that truncation isn't part of this module any more, it's a number. But that's technically not a legal type for this property according to the intake API JSON spec:

https://github.com/elastic/apm-server/blob/27acff4c0ba45efdbbe20de3ca403d3b5c481416/docs/spec/request.json#L78-L82

Should the JSON spec allow for a number here?

watson:
The CI failures are because of the npm bug with the github link in the dependencies

test/instrumentation/modules/hapi.js
Qard: Why do we even truncate ports? They can't ever be longer than 5 digits. 🤔

watson: When implementing the truncation logic, I just went over all strings that had a limit on them and added the truncation.

Using a filter the user can overwrite it with anything they want. You could argue though that if they did the server JSON schema validation would just reject it, so it might not be that important.

watson: In related news, I realized the other day that `process.title` [cannot be more than 16 bytes on Linux](https://stackoverflow.com/questions/23534263/what-is-the-maximum-allowed-limit-on-the-length-of-a-process-name)

PR #506, opened 2018-08-10
perf: only instrument knex if necessary

watson:
If `captureSpanStackTraces` is disabled, there's no reason to capture the intermediate knex stack trace which is only used to improve the span stack traces of pg and mysql spans.


watson:
@Qard I haven't written any tests for this as I wasn't sure exactly how I could test this in a reasonable way. If you have any suggestions that aren't too complicated let me know 😃

Qard:
You could just check for existence of the unwrap function on `Knex.Client`.

trentm:
Obsoleted by https://github.com/elastic/apm-agent-nodejs/pull/2879

Issue #308, opened 2018-04-24
Drop support for Node 4


alvarolobato:
As Node.js v4 reaches EOL (currently in maintenance) and a new LTS is released we will drop support for Node 4 allowing us to use language features available after Node 4

The following should be done when upgrading:

- [x] Bump read-pkg-up dependency to newer version
- [x] Bump wait-on dev-dependency to newer version
- [x] Drop support for Node 9

watson:
@alvarolobato FYI: I just updated the title and description of this issue to use the word "drop" instead of "deprecate" as the latter can mean that we still support it, but just warn people that the next version will no longer support it. Using the word "drop" is less ambiguous IMO 😃

watson:
When dropping Node 4 support, consider if it's worth it to set the minimum supported Node 6 version to v6.5.0 as that has 99% support for ES2015, as opposed to only 95% for <6.5.0. The we don't need extra 4% for anything, there's no need for it.

Qard:
The remaining 4% is fairly obscure, so it's not really that important.

watson:
Support for Node 4 (and 9) have been removed from the `api-v2` branch, so I'll close this issue now.

PR #508, opened 2018-08-10
feat(config): default serviceName to package name

### LINKED ISSUE #405: Consider setting serviceName based on name in package.json
Currently the only required config option is `serviceName`. But almost all Node.js apps have a `package.json` file which contains a `name` property. It would be simple to use this name as the default value.

The only downside of this is that there's no guarantee that the `name` property is set or that there's even a `package.json` file available. In that case we would have to abort just like we do today if the `serviceName` isn't set manually.

Qard:
Fixes #405

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


test/config.js
```diff
@@ -172,13 +158,6 @@ test('should separate strings and regexes into their own blacklist arrays', func
   t.end()
 })
 
-test('missing serviceName => inactive', function (t) {
```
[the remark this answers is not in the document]
Qard: I've added a commit with a test that generates basic test app and runs it in a child process, dumping the `apm._conf` as JSON and doing some basic validation on the other end. It's a bit complicated. Thoughts?

test/config.js
```diff
@@ -193,6 +176,114 @@ test('valid serviceName => active', function (t) {
...
+        to: path.join(tmp, 'node_modules/elastic-apm-node')
+      }
+    ]
+
+    // NOTE: Reduce the sequence to a promise chain rather
+    // than using Promise.all(), as the tasks are dependent.
+    let promise = files.reduce((p, file) => {
+      return p.then(() => {
+        switch (file.action) {
+          case 'create': {
```
watson: TIL: I don't think I've ever seen blocks used together with `case` 😄It's not technically needed here, but I guess it does no harm

test/config.js
```diff
@@ -193,6 +176,114 @@ test('valid serviceName => active', function (t) {
...
+    })
+  }
+
+  t.test('should be active when valid', function (t) {
+    var pkg = {
+      name: 'valid'
+    }
+
+    return testServiceConfig(pkg).then(conf => {
+      t.equal(conf.active, true)
```
watson: I would check that `serviceName` is correctly set to `valid` as well

test/config.js
```diff
@@ -193,6 +176,114 @@ test('valid serviceName => active', function (t) {
...
+  t.test('should be inactive when missing', function (t) {
+    var pkg = {}
+
+    return testServiceConfig(pkg).then(conf => {
+      t.equal(conf.active, false)
+      t.end()
+    })
+  })
+
+  t.end()
```
watson: I don't think this `t.end()` is needed - at least I've never used Tape this way. But maybe I've been using it wrong?

test/config.js
Qard: Yeah, I do that all the time. It reads better to me, and scopes stuff better.

test/config.js

test/config.js
Qard: Yep, you're right. It's required in `tap`, but not `tape`. I use `tap` for my personal stuff, so I mix that up sometimes. 😅

test/config.js
```diff
@@ -193,6 +176,117 @@ test('valid serviceName => active', function (t) {
   t.end()
 })
 
+test('serviceName defaults to package name', function (t) {
+  // TODO:
```
watson: Are these TODO's still relevant?

Qard: Yeah, do you have any thoughts on it? Should those requires be moved up? Should I use already promisified modules?

watson: No I think it's fine to just use the normal modules. This also means we don't have to depend on two different versions of mkdirp.

It might be cleaner to move the require's up top. I get why you put them here, but to follow our thing about how to order the require statements, it's probably better to be consistent unless there's a specific reason.

Qard:
@watson This should be ready to land, assuming everything passes review. Minor note: this uses a slightly older version of `read-pkg-up`, as the latest is drops compatibility with 4.x, which we are still supporting for now. We should probably bump that when we release 2.x.

watson:
I've added bumping the `read-pkg-up` dependency to a todo list on the "drop 4.x" issue: #308

PR #510, opened 2018-08-11
Fixes #509

### LINKED ISSUE #509: TypeError: Cannot create property 'Symbol(reported)' on string '...'
in lib/instrumentation/modules/express.js there is following error handler check:
```
if ((isError(err) || typeof err === 'string') && !err[reportedSymbol]) {
            err[reportedSymbol] = true
            agent.captureError(err, { request: req })
}
```
there is a check if it is a string, in strict mode a string cannot have a property though. So "err[reportedSymbol] = true" does fail in this case with the error "Cannot create property 'Symbol(reported)' on string '...'."

Not quite sure how to deal with this, I'm not totally aware why this flag is needed at all? Why would it be called twice?
Alternatively it would be also great to deactivate this error handling when captureExceptions is set to false.

Thanks
Simon

simllll:
In strict mode you cannot set a property on a string, this causes a fatal error on my side. This proposed change does still report the error, but does not set the reported symbol, furthermore it prints out a warning that a string in strict mode cannot be correctly tracked.
Besides this it allows to disable the error handler at all (we are using redis/sentry for this anyway), and do not need this error reporting feature. 
Let me know what you think.

### DIFF
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -25,15 +25,24 @@ module.exports = function (express, agent, version, enabled) {
 
   var layerPatchedSymbol = Symbol('layer-patched')
   var reportedSymbol = Symbol('reported')
+
+  function shouldReport (err) {
+    if (typeof err === 'string') return true
+    if (isError(err) && !err[reportedSymbol]) {
+      err[reportedSymbol] = true
+      return true
+    }
+    return false
+  }
+
   function patchLayer (layer) {
     if (!layer[layerPatchedSymbol]) {
       layer[layerPatchedSymbol] = true
       agent.logger.debug('shimming express.Router.Layer.handle function')
       shimmer.wrap(layer, 'handle', function (orig) {
-        if (orig.length !== 4) return orig
+        if (!agent._conf.captureExceptions || orig.length !== 4) return orig
         return function (err, req, res, next) {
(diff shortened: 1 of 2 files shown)


lib/instrumentation/modules/express.js
```diff
@@ -30,13 +32,17 @@ module.exports = function (express, agent, version, enabled) {
       layer[layerPatchedSymbol] = true
       agent.logger.debug('shimming express.Router.Layer.handle function')
       shimmer.wrap(layer, 'handle', function (orig) {
-        if (orig.length !== 4) return orig
+        if (!agent._conf.captureExceptions || orig.length !== 4) return orig
```
watson: Great addition, thanks 👍

Qard:
You could also make a helper function like:

```js
function shouldReport (err) {
  if (typeof err === 'string') return true
  if (isError(err) && !err[reportedSymbol]) {
    err[reportedSymbol] = true
    return true
  }
  return false
}
```

That way the core logic only needs to be:

```js
if (shouldReport(err)) {
  agent.captureError(err, { request: req })
}
```

That abstracts away the tracking mechanism from the code that needs it.

simllll:
I like the idea of a helper function @Qard , just commited it and fixed the code indentation.

Thanks guys

simllll:
Fixed test and commit messages.

PR #512, opened 2018-08-13
fix(express): don't set property on error if it's a string

### LINKED ISSUE #509: TypeError: Cannot create property 'Symbol(reported)' on string '...'

watson:
Fixes #509

### DIFF
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -32,8 +32,9 @@ module.exports = function (express, agent, version, enabled) {
       shimmer.wrap(layer, 'handle', function (orig) {
         if (orig.length !== 4) return orig
         return function (err, req, res, next) {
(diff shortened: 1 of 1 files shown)


watson:
Oh hang on. Overlooked that there already were a PR open to fix this bug: #510

PR #516, opened 2018-08-15
feat(ioredis): add support for ioredis version 4.x

watson:


watson:
Hmm something seems to be broken with the tests on Node 9 and below for version 4.0.0

PR #524, opened 2018-08-28
feat: distributed tracing

Qard:
This is an in-progress PR for distributed tracing.

Depends on the api-v2 branch being merged. (No PR number yet)


Qard:
BTW, this currently just uses uuid as the span id, but that will change later. This is just to have an ID of _some_ sort to work with, for now.

watson:
@Qard I think it's best if we made this PR against the `api-v2` branch right?

Qard:
Closing in favor of #538.

PR #526, opened 2018-08-29
test: bump standard to version 12

watson:
It's gotta take me some time to get used to include the extra spaces 😓


watson:
This fails on Node 4. Let's wait with this until we release 2.0.0. I'll add this to the todo list

PR #532, opened 2018-08-30
test: do not capture exceptions in disableInstrumentations test

Qard:
This should make the `disableInstrumentations` test stop failing due to the agent being unable to connect to apm-server.


watson:
I've opened another issue to fix the underlying error: #535

PR #534, opened 2018-08-30
test: bump standard to version ^12.0.1

watson:
Someone beat me to the punch of making a fix for v12 that wouldn't fail on Node 4. So with version 12.0.1 we can now use it with our CI setup 👍 

It's still gotta take me some time to get used to include the extra spaces 😓


watson:
jenkins, test this please

Qard:
I've always put spaces around object contents, so it's no trouble for me. 😅

Issue #535, opened 2018-08-30
Fix flaky `disableInstrumentations` test


watson:
The tests for `disableInstrumentations` are flaky. The often fail with the following error (here captured on AppVeyor, but the error occurs on Travis and Jenkins as well):

```
# multiple modules by csv string -> bluebird,elasticsearch,express-graphql,generic-pool,handlebars,http,http2,knex,mimic-response,mysql,pg,restify,ws
ok 140 disabled all selected modules
(node:3872) ExperimentalWarning: The http2 module is an experimental API.
C:\projects\apm-agent-nodejs\lib\request.js:70
  payload = agent._filters.process(payload)
                           ^
TypeError: agent._filters.process is not a function
    at Object.sendTransactions [as transactions] (C:\projects\apm-agent-nodejs\lib\request.js:70:28)
    at C:\projects\apm-agent-nodejs\lib\instrumentation\index.js:68:17
    at value.get.value (C:\projects\apm-agent-nodejs\node_modules\async-value-promise\index.js:30:27)
    at AsyncValue.get (C:\projects\apm-agent-nodejs\node_modules\async-value\index.js:17:7)
    at AsyncValuePromise.then (C:\projects\apm-agent-nodejs\node_modules\async-value-promise\index.js:29:20)
    at Queue.onFlush [as _onFlush] (C:\projects\apm-agent-nodejs\lib\instrumentation\index.js:66:41)
    at Queue.flush (C:\projects\apm-agent-nodejs\lib\queue.js:39:8)
    at Timeout.<anonymous> (C:\projects\apm-agent-nodejs\lib\queue.js:53:10)
    at Timeout.elasticAPMCallbackWrapper [as _onTimeout] (C:\projects\apm-agent-nodejs\lib\instrumentation\index.js:187:27)
    at ontimeout (timers.js:498:11)
    at Timer.unrefdHandle (timers.js:611:5)
C:\projects\apm-agent-nodejs\test\test.js:112
    if (err) throw err
```

PR #536, opened 2018-08-30
perf: declare instance properties in constructor

watson:
Probably this isn't going to change much as the code path is cold. I also looked over `Transaction` and `Span` as those are much hotter, but they all conformed to best practices.


Qard:
Jenkins failed on #535.

watson:
jenkins, test this please

PR #537, opened 2018-08-30
fix: always call agent.flush() callback

watson:
If there wasn't any transactions in the queue, the flush callback would never be called


watson:
jenkins, test this please

watson:
jenkins, test this please

Issue #421, opened 2018-06-28
Add support for distributed tracing


alvarolobato:
Work is underway to add support for distributed tracing to the API/server: https://github.com/elastic/apm-server/issues/913 / https://github.com/elastic/apm-server/issues/968.
Support needs to be added to this agent too.

alvarolobato:
Moved "[test that responses from the server during chunked writes](https://github.com/elastic/apm-server/issues/1298#issuecomment-415704054) work with the client you’re using (if your agent uses chunked writes) " to v2 issue

PR #538, opened 2018-08-30
feat: distributed tracing

Qard:
Retargetting elastic/apm-agent-nodejs#524 to api-v2...

Fixes elastic/apm-agent-nodejs#421


Qard:
Rebased with #557 to get tests passing.

lib/instrumentation/context.js
```diff
@@ -0,0 +1,97 @@
...
+    // 1 byte
+    this.version = version
+    // 16 bytes
+    this.traceId = traceId
+    // 8 bytes
+    this.id = id
+    // 1 byte
+    this.flags = flags
+    // 8 bytes
+    this.parentId = parentId
```
watson: This doesn't seem to be used for anything at the moment?

lib/instrumentation/transaction.js
```diff
@@ -1,9 +1,9 @@
 'use strict'
 
 var truncate = require('unicode-byte-truncate')
-var uuid = require('uuid')
 
 var config = require('../config')
+var TraceContext = require('./context')
```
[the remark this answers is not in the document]
watson: Also not important... just a nit 😅

lib/instrumentation/context.js
```diff
@@ -0,0 +1,97 @@
...
+      this.version,
+      this.traceId,
+      this.id,
+      this.flags
+    ]
+
+    return parts.map(toHex).join('-')
+  }
+
+  toJSON () {
```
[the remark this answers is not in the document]
Qard: `toJSON` was originally intended just for debug purposes just because the buffer form is not really readable. I ended up using it in a couple spots internally though. Might be better to just have a private buffer and string getters though? 🤔

lib/instrumentation/context.js
Qard: It's just a utility to verify structural integrity of context inheritance. Java does it this way too. We could do without it though. It'll be helpful if we eventually try to inherit spans from other spans though, rather than inheriting everything from the transaction context.

lib/instrumentation/transaction.js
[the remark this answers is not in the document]
Qard: I thought about renaming the file to `trace-context.js`. Might be at least _a little_ better. 🤔

lib/instrumentation/index.js
```diff
@@ -128,8 +128,8 @@ Instrumentation.prototype.addEndedSpan = function (span) {
   }
 }
 
-Instrumentation.prototype.startTransaction = function (name, type) {
-  return new Transaction(this._agent, name, type)
+Instrumentation.prototype.startTransaction = function (name, type, header) {
```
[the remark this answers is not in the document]
Qard: Good call. 👍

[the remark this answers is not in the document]

Qard:
@watson I made some changes to try having a shared buffer and string getters.

lib/instrumentation/context.js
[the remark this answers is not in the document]
Qard: I tried a new implementation with a shared buffer and string getters.

lib/instrumentation/context.js
```diff
@@ -0,0 +1,97 @@
...
+    this.id = id
+    // 1 byte
+    this.flags = flags
+    // 8 bytes
+    this.parentId = parentId
+  }
+
+  static create (sampled = false) {
+    const bytes = crypto.randomBytes(24)
+    const flags = Buffer.alloc(1)
```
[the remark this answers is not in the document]
Qard: I just simplified it to assigning `0x01` for now. Probably don't need anything more than that until we start trying to encode other things into the flags.

lib/instrumentation/transaction.js
```diff
@@ -58,7 +58,19 @@ function Transaction (agent, name, type) {
     }
   })
 
-  this.id = uuid.v4()
+  Object.defineProperty(this, 'id', {
+    get () {
+      return this.context.id.toString('hex')
```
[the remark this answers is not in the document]
Qard: I made some changes to use a shared buffer. I could maybe stringify the values out of the buffer in the constructor and store them, rather than use the getters? The downside is that it'd then be storing fields even when they aren't used. Might also be a good idea to `Object.freeze(this)` in the constructor, since the object itself _should_ be basically immutable in normal circumstances. 🤔

lib/instrumentation/span.js
```diff
@@ -22,6 +23,12 @@ function Span (transaction) {
   this._stackObj = null
   this._agent = transaction._agent
 
+  Object.defineProperty(this, 'id', {
+    get () {
+      return this.context.id.toString('hex')
```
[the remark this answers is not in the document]
Qard: Should probably actually just log the context object, in the future, so the trace id is there too. 🤔

lib/instrumentation/transaction.js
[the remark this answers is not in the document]
Qard: I ended up changing to a getter-based lazy properties thing.

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,84 @@
...
+
+  static create (sampled = false) {
+    const buffer = Buffer.alloc(SIZES.all)
+    crypto.randomFillSync(buffer, OFFSETS.traceId, SIZES.ids)
+    if (sampled) buffer[OFFSETS.flags] = 0x01
+    return new TraceContext(buffer)
+  }
+
+  static fromString (header) {
+    return new TraceContext(Buffer.from(header.replace(/-/g, ''), 'hex'))
```
watson: We should probably add a comment about how we're currently not following all the rules about validating the version byte, the length of the header etc as specified here: https://w3c.github.io/distributed-tracing/report-trace-context.html#versioning-of-traceparent

It's fine now as we're using our own custom header name - i.e. it's not a blocker - so just a comment would suffice.

lib/instrumentation/span.js
```diff
@@ -143,17 +174,9 @@ Span.prototype._encode = function (cb) {
...
-      transactionId: self.transaction.id,
-      timestamp: self.transaction.timestamp,
-      name: self.name,
-      type: self.type,
-      start: self.offsetTime(),
-      duration: self.duration()
-    }
+    var payload = self.toJSON()
 
     if (frames) payload.stacktrace = frames
```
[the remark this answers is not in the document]
Qard: I split it up just for debugging. I can restore it though, if you prefer it like that. It needs to be sync to work properly as a `toJSON` function though, so not really anything I can do about stack frames being separate.

lib/agent.js
```diff
@@ -345,13 +355,13 @@ Agent.prototype.captureError = function (err, opts, cb) {
     error = agent._errorFilters.process(error)
 
     if (!error) {
-      agent.logger.debug('error ignored by filter %o', { id: id })
+      agent.logger.debug('error ignored by filter %o', { context })
```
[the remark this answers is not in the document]
Qard: Oh, that's just left over from when errors inherited the trace context id rather than generating its own. I'll just restore that to `id: id`.

lib/instrumentation/async-hooks.js
```diff
@@ -17,6 +17,18 @@ module.exports = function (ins) {
     }
   })
 
+  const activeSpans = new Map()
```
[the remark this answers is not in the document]
Qard: I called it that just to differentiate from `bindingSpan`, since those have different meanings and can point to different things at the same time.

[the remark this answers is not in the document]
watson: Then maybe rename `transactions` to `activeTransactions`?

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,99 @@
...
+  all: 34
+}
+
+const OFFSETS = {
+  version: 0,
+  traceId: SIZES.version,
+  id: SIZES.version + SIZES.traceId,
+  flags: SIZES.version + SIZES.ids,
+
+  // Additional parentId is stored after the header content
```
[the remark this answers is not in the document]
Qard: The parent ID concept isn't part of the header, but it's needed from the perspective of our UI to nest properly.

The parent ID is not included in the header string form. The remote service doesn't actually need to be aware of the parent of the span that is given in the header, it only needs that span itself to use as a parent of a new transaction.

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,99 @@
...
+
+function hexSliceFn (buffer, offset, length) {
+  return () => buffer.slice(offset, length).toString('hex')
+}
+
+function maybeHexSliceFn (buffer, offset, length) {
+  const fn = hexSliceFn(buffer, offset, length)
+  return () => {
+    const value = fn()
+    if (/[1-9af]/.test(value)) {
```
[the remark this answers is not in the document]
Qard: I didn't do `\d` because it needs to be `1-9`, not `0-9`. It is identifying an empty id as all digits are zero.

Good catch on the a-f though. 😅

lib/instrumentation/trace-context.js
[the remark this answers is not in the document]
watson: Ah, so we just store it in the buffer to make it easier to copy bits around?

lib/instrumentation/async-hooks.js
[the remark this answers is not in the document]
Qard: done.

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,109 @@
...
+    defineLazyProp(this, 'traceId', hexSliceFn(buffer, OFFSETS.traceId, OFFSETS.id))
+    defineLazyProp(this, 'id', hexSliceFn(buffer, OFFSETS.id, OFFSETS.flags))
+    defineLazyProp(this, 'flags', hexSliceFn(buffer, OFFSETS.flags, OFFSETS.parentId))
+    defineLazyProp(this, 'parentId', maybeHexSliceFn(buffer, OFFSETS.parentId))
+  }
+
+  static create (sampled = false) {
+    const buffer = Buffer.alloc(SIZES.all)
+    crypto.randomFillSync(buffer, OFFSETS.traceId, SIZES.ids)
+    if (sampled) buffer[OFFSETS.flags] = 0x01
```
watson: I think the flags in the spec changed, so we should update to reflect that: https://github.com/w3c/distributed-tracing/blob/master/trace_context/HTTP_HEADER_FORMAT.md#trace-flags

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,109 @@
...
+
+function hexSliceFn (buffer, offset, length) {
+  return () => buffer.slice(offset, length).toString('hex')
+}
+
+function maybeHexSliceFn (buffer, offset, length) {
+  const fn = hexSliceFn(buffer, offset, length)
+  return () => {
+    const value = fn()
+    if (/[1-9a-f]/.test(value)) {
```
watson: Consider adding a comment saying that this is to check that the value isn't all zeros. I got confused and thought that you wanted to check that it was valid hex

lib/instrumentation/trace-context.js
Qard: So "sampled" should mean both requested _and_ recorded?

lib/instrumentation/trace-context.js
Qard: Will do. 👍

lib/instrumentation/trace-context.js
watson: I'm not entirely sure. I think only `0b00000010` is sampled, but maybe @felixbarny can verify?

felixbarny: The sampling probability is determined by the transaction_sample_rate configuration option. If a trace should be sampled, both the requested and the recorded flags are set. I.e. the sampler `requested` the trace to be sampled and that's why we are` record`ing it.

watson: @felixbarny Ok so in this case the flag should be `0b11`. And in case of an incoming request, what flag should we rely on to turn on sampling?

lib/instrumentation/http-shared.js
```diff
@@ -79,14 +80,35 @@ exports.traceOutgoingRequest = function (agent, moduleName) {
...
+        if (!(reqUrl instanceof URL)) {
+          reqUrl = new URL(reqUrl)
+        }
+        options = Object.assign({}, options, reqUrl)
+      }
+
+      if (!options.headers) options.headers = {}
+
+      // Use span context as traceparent header
+      options.headers['Elastic-APM-traceparent'] = span.context.toString()
```
watson: According to the design doc this needs to be `Elastic-APM-Traceparent` - it probably doesn't make a difference, but just in case

lib/instrumentation/trace-context.js
felixbarny: For now, that would be the requested flag. If the parent propagated the requested flag, we should also set recorded to true. In the future, we might want to add rate limiting here.

lib/instrumentation/http-shared.js
Qard: Headers are technically _supposed_ to be case-insensitive, and Node.js _will_ actually lower-case it internally. I'm not sure how well-supported case-insensitivity of headers is across other platforms though. 🤔

watson: I agree it probably will not be an issue. I just mentioned it as it's technically says to use `Elastic-APM-Traceparent` in the design doc, so I thought it's better to just follow it by the letter.

When we remove our header prefix once TraceContext is finalized, the spec actually say to use `traceparent` (all lowercase) as there is certain issues with casing in the systems. So I guess sometimes it's important 🤷‍♂️

Qard: fixed.

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  get requested () {
+    return !!(this[bufferSymbol][OFFSETS.flags] & FLAGS.requested)
+  }
+  get reported () {
+    return !!(this[bufferSymbol][OFFSETS.flags] & FLAGS.reported)
+  }
+  get sampled () {
+    return this.requested && this.reported
```
watson: I think technically `this.reported` is all that's required for it to be considered sampled.

And we should make sure (in this naive implementation) to turn on the reported flag when we instantiate a `TraceContext` based on a parent whos `requested` flag is set.

lib/instrumentation/transaction.js
```diff
@@ -13,77 +12,97 @@ var Timer = require('./timer')
...
-      if (!this._timestamp) {
-        this._timestamp = new Date(this._timer.start).toISOString()
-      }
-      return this._timestamp
-    }
-  })
+function Transaction (agent, name, type, traceparent) {
+  this.context = TraceContext.startOrResume(traceparent, agent._conf)
+  const verb = this.context.parentId ? 'continue' : 'start'
+  agent.logger.debug('%s transaction %o', verb, { id: this.id, name: name, type: type })
```
watson: Would it be more correct to say "start/continue trace" instead of "start/continue transaction"?

test/instrumentation/index.js
```diff
@@ -534,6 +534,125 @@ test('bind', function (t) {
...
+  resetAgent(6, function (data) {
+    t.equal(data.transactions.length, 1)
+    t.equal(data.spans.length, 5)
+
+    const trans = data.transactions[0]
+    t.equal(trans.name, 'foo')
+    t.equal(trans.type, 'custom')
+    t.equal(trans.span_count.started, 5)
+
+    const t0 = findObjInArray(data.spans, 'name', 't0')
```
watson: Nit: `s/t(\d+)/s\1/` (here and below)

test/instrumentation/trace-context.js
```diff
@@ -0,0 +1,114 @@
...
+    id,
+    flags
+  }, 'trace context serializes fields to hex strings, in JSON form')
+
+  t.end()
+})
+
+test('startOrResume', t => {
+  t.test('resume from header', t => {
+    const context = TraceContext.startOrResume(header)
```
watson: I would add tests for different combinations of flags from parent (`0b00`, `0b01`, and `0b11`) to see how the child flags and the sampled property are affected

lib/instrumentation/transaction.js
Qard: fixed.

test/instrumentation/trace-context.js
Qard: done.

lib/instrumentation/trace-context.js
Qard: I added something to take requested flag from the parent to add reported flag when making a child.

As for the `sampled` getter: the `requested` flag would indicate that children of that context should also be reported, so I think using both makes sense here. What are your thoughts on that?

Qard: So there's no scenario internally in which the context could ever have the reported flag without the requested flag. It's _technically_ possible to receive that from a header though.

I added a commit to negate the reported flag on a child of a parent with `reported` and _not_ `requested`. Do you think that's correct behaviour?

test/instrumentation/trace-context.js
```diff
@@ -0,0 +1,143 @@
...
+
+  t.test('reported', t => {
+    const header = `${version}-${traceId}-${id}-02`
+    const context = TraceContext.fromString(header).child()
+
+    isValid(t, context)
+    t.equal(context.version, version, 'version matches')
+    t.equal(context.traceId, traceId, 'traceId matches')
+    t.notEqual(context.id, id, 'has new id')
+    t.equal(context.flags, '00', 'reported remains reported')
```
watson: The assertions message should probably be `reported without requested gets reset` or something like that

lib/instrumentation/trace-context.js
watson: > As for the `sampled` getter: the `requested` flag would indicate that children of that context should also be reported, so I think using both makes sense here. What are your thoughts on that?

You're correct that this is true for children. But as far as I can see, this getter is only used by `Transaction.prototype.sampled`, which to my knowledge only concerns it self with the transaction it self right?

Also, technically if this was the root transaction in the trace - and it decided it should be sampled - it would not be requested, but only reported right?

> I added a commit to negate the reported flag on a child of a parent with `reported` and _not_ `requested`. Do you think that's correct behaviour?

Yes I think that's correct. @felixbarny do you agree?

lib/instrumentation/span.js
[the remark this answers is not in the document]
watson: Hmm... I think it is a bit weird that we have a "public" method called `toJSON` which doesn't return the full payload, so if it's only for debugging, I think it's better to join it two again.

Also, I don't really like the "JSON" part of the name as it doesn't actually return JSON (same goes for the `toJSON` function on transactions). That's why I opted to use the word "encode" for the "private" function.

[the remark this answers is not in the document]
Qard: I just called it that because that's the standard name to support `JSON.stringify(...)`.

test/instrumentation/trace-context.js
Qard: fixed.

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,155 @@
...
+    defineLazyProp(this, 'flags', hexSliceFn(buffer, OFFSETS.flags, OFFSETS.parentId))
+    defineLazyProp(this, 'parentId', maybeHexSliceFn(buffer, OFFSETS.parentId))
+
+    defineLazyProp(this, 'requested', () => {
+      return !!(this[bufferSymbol][OFFSETS.flags] & FLAGS.requested)
+    })
+    defineLazyProp(this, 'reported', () => {
+      return !!(this[bufferSymbol][OFFSETS.flags] & FLAGS.requested)
+    })
+    defineLazyProp(this, 'sampled', () => {
```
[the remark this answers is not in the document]
Qard: I changed them to just be set in the constructor.

Qard:
@watson I did a bunch more squashing/rebasing. Other than a bit of uncertainty about the correctness of the flags behaviour, it should be in a good state to review and (hopefully) land now.

lib/instrumentation/trace-context.js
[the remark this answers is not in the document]
watson: Maybe it was also just me being tired and not understanding what was going on 😅

lib/instrumentation/trace-context.js
watson: Hmm yeah you might be right. I was thinking that `requested` meant that the parent requested it. But I guess in our case the `TraceContext` actually carry two meanings: What were we asked to do by our parent and what will we ask our children to do. And in the case of a root transaction, we were technically not asked to do anything as we're the root, but we will ask our children to do this. I guess I was just thinking of this from one side

Qard:
I made some changes to ensure http.request always propagates the traceparent header and that it handles the arguments better.

lib/instrumentation/http-shared.js
```diff
@@ -74,19 +75,52 @@ function isRequestBlacklisted (agent, req) {
...
       var id = span && span.transaction.id
 
       agent.logger.debug('intercepted call to %s.request %o', moduleName, { id: id })
 
-      var req = orig.apply(this, arguments)
+      var options = {}
+      var newArgs = [ options ]
+      for (let arg of args) {
+        if (typeof arg === 'function') {
+          newArgs.push(arg)
```
watson: hopefully there will not be more than one callback 😬😄

test/instrumentation/trace-context.js
```diff
@@ -0,0 +1,146 @@
...
+    t.equal(context.version, version, 'version matches')
+    t.notEqual(context.traceId, traceId, 'has new traceId')
+    t.notEqual(context.id, id, 'has new id')
+    t.equal(context.sampled, false, 'is sampled')
+
+    t.end()
+  })
+})
+
+test('child', t => {
```
watson: Probably best to add a test for `00` flag as well

Qard: done.

lib/instrumentation/http-shared.js
Qard: If there _was_, it'd find them just fine. I'm a bit wary of the argument folding thing for everything else though.

I need to do the folding to be able add the header to the inputs. In 10.9+, it supports `(urlString, options, ...)`, but less than that and it will only allow `(urlString, ...)` or `(options, ...)`, so I need to ensure the URL is converted to an object. There's also a problem with the `URL` type being not directly compatible with the `options` structure, so it needs to be stringified and converted back to an object with `url.parse(...)`, to ensure it has a valid structure. 🤔

lib/instrumentation/trace-context.js
```diff
@@ -0,0 +1,162 @@
...
+      },
+      sampled: {
+        enumerable: true,
+        value: reported
+      }
+    })
+  }
+
+  static startOrResume (traceparent, conf) {
+    return isValidHeader(traceparent)
```
felixbarny: A potential performance improvement would be to return a boolean from resume which indicates whether the header could be parsed without errors. If it's false, start a new trace. That way, you can get rid of the regex check.

Just an idea, feel free to ignore :)

Qard: The parsing can't fail. It just copies whatever is in the header into the buffer memory, regardless of what is there. The regex check needs to be made _somewhere_ to validate the contents of the data to be copied into the buffer, might as well double as a switch between resuming from a header or starting a new context.

felixbarny: See also https://github.com/elastic/apm-agent-java/blob/d8b583f30c6b406ba8f81e63f0ce71af22dc2469/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/Transaction.java#L85-L87

felixbarny: Sorry, didn't see your previous comment. I implemented the hex parsing in a way that it throws an exception if there are invalid chars: https://github.com/felixbarny/apm-agent-java/blob/e78d187659ea77899c6f928de93f52ec0028ced6/apm-agent-core/src/main/java/co/elastic/apm/util/HexUtils.java#L67.
That's why with this approach, there is no need to validate upfront. In case I did not overlook something, that is.

lib/instrumentation/http-shared.js
watson: Yeah, it's probably fine - I didn't mean that I wanted you to change anything. I was just wondering...

watson:
@Qard the merge conflicts look a little bit complicated, so I didn't want to spend too much time on it. I assume you could do it very quickly 😃

watson:
@Qard never mind. I managed to fix the merge conflicts

PR #542, opened 2018-08-30
test: fix linter error

watson:


watson:
AppVeyor builds fails because of #535. That's unrelated to this PR.

PR #553, opened 2018-09-04
feat(transaction): span_count.total

### LINKED ISSUE #552: Add `transaction.span_count.total`
Implement https://github.com/elastic/apm-server/issues/1241

Qard:
Fixes elastic/apm-agent-nodejs#552
Depends on elastic/apm-agent-nodejs#557

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -161,9 +161,7 @@ Transaction.prototype.toJSON = function () {
 
     // Only include dropped count when spans have been dropped.
     if (this._droppedSpans > 0) {
-      payload.span_count.dropped = {
-        total: this._droppedSpans
-      }
+      payload.span_count.dropped = this._droppedSpans
     }
 
     if (this.req) {
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -142,7 +142,10 @@ Transaction.prototype.toJSON = function () {
     timestamp: this.timestamp,
     result: String(this.result),
     sampled: this.sampled,
-    context: null
+    context: undefined,
+    span_count: {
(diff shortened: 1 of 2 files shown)


Qard:
One as-yet unanswered question is if span_count should be null on unsampled transactions.

watson:
The tests fail because of #554

I don't like `span_count === null` when the transaction is upsampled to be honest. I know about the performance implications, but it also just feels "unclean" to send up these JSON objects containing `null` objects. But I'm not sure if it's just vanity or if there's something deeper 🤔

Qard:
By that, do you mean you'd prefer the `span_count` is _always_ present, or that it should be `undefined`?

watson:
I mean that I prefer the JSON document not to include it for un-sampled transactions. There's a lot of optional fields in our payload that I feel would be weird if they were reported as `null`. But the APM Server will just ignore them (I assume), so I guess there's no technical problem in keeping it as `null`.

Qard:
I _think_ explicitly defining them as `undefined` should avoid the hidden-class problem the same as defining as `null` does. We could try that.

Qard:
Rebased with #557 to get tests passing.

Qard:
Updated to reflect the current state of the decision in https://github.com/elastic/apm-server/issues/1241.

Qard:
I've also changed the `dropped.total` to `dropped`, as suggested in https://github.com/elastic/apm-server/issues/1241#issuecomment-419977863.

watson:
Should this PR be rebased or squashed and merged?

Qard:
Could squash and merge, if you edit the commit title/message to describe all the changes. Rebase is fine too though.

PR #557, opened 2018-09-05
ci: fix tav install failures

Qard:
This should fix the issues with `tav` accidentally removing `elastic-apm-http-client`.

NOTE: Because of the npm upgrade, Node.js 4 no longer works. Since this is targeting the api-v2 branch, I've just removed the v4 stuff from travis.


.travis.yml
```diff
@@ -17,14 +17,14 @@ services:
   - cassandra
 
 before_script:
+  - npm install -g npm@latest
```
watson: Am I correct in assuming that this will not update npm before after the normal `npm install` have run? If so that's 👍

watson:
Maybe we should first make a separate PR to `api-v2` that just removes v4 and v9 from both Travis and Jenkins, and then make this one that adds the npm upgrade?

Qard:
I changed this PR to only remove v4 and v9, but that results in the tav failure still being present. Do we want to land a v4+v9 removal even though the tests would be failing on v8 temporarily?

Qard:
I've switched back to doing both in the same PR, but in separate commits now, just to verify it passes. I can split part to another PR after, if you like, or we could just land as the two commits.

watson:
👍 doing it in two commits as you suggested

We should update Jenkins to not test v4+v9 as well in the same commit

Qard:
Oh, oops. I meant to squash it into the other commit. I'll fix it. 😅

Qard:
Fixed. 😸

PR #558, opened 2018-09-06
Stalling `npm install` workaround + cache node modules

cotsog:
Suggestions from Travis CI to make your builds more stable.

1) Retry `npm install` up to 3 times and prevent it from stalling using the `travis_wait` utility function.

2) Cache the `node_modules` directory to decrease the number of network calls made when calling `npm install`.


watson:
👍 The retry logic looks good to me. Nice and simple 😄 

It was originally on purpose that I didn't cache the `node_modules` folder. This might have been misguided though.

My understanding was that Travis would then re-use the installed modules just as how it works locally on my dev machine were I also have a persistent `node_modules` folder. The reason why I didn't want to persist the `node_modules` folder on Travis was to ensure that we'd always test with the latest modules, as there might be a new release of a sub-dependency that had introduced a bug.

cotsog:
@watson Sorry for the delay of my reply.

Re: caching, my understanding is that `npm install` should respect what's in your package.json file. If a new module version is available, it will be installed as we are always running `npm install` even with the cache feature enabled.

It does happen sometimes that a cache archive becomes stale/invalid but I think it's less likely to happen then the behavior you are seeing now. Following on from that, it's ok if you want to just try the "retry + timeout" suggestion in this PR at first.

cotsog:
Do you have insights on why the "Lint Commit Message" job is failing?

https://travis-ci.org/elastic/apm-agent-nodejs/jobs/425079188

watson:
I can fix the commit message failure once I merge so don't worry about it 😃 It fails here because we want to prefix our commits with a type and optionally a scope formatted like `type: message` or `type(scope): message`. So in this case the commit message could be `ci(travis): cache node modules + npm install workaround`

Regarding caching the `node_modules` folder, I'm still not sure I fully understand how the mechanism works. I have two questions:

1. Say `package.json` have the following dependency: `"foo": "^1.0.0"` and version `foo@1.0.0` is the latest version. The tests have been run at least once, so `foo@1.0.0` is cached inside of `node_modules` on Travis. Then `foo@1.0.1` is released. On my own development machine npm would not upgrade to `foo@1.0.1` unless I manually deleted `foo` from the `node_modules` folder. Isn't that also the case on Travis?

1. Say `package.json` have the following dependency: `"foo": "^1.0.0"`, and that in turn have this sub-dependency: `"bar": "^1.0.0"`. The latest version of `bar` is `1.0.0`. The tests have been run at least once, so `bar@1.0.0` is cached inside of `node_modules` on Travis. Then `bar@1.0.1` is released but `foo` doesn't bump its dependency. On my local development machine, npm will not install `bar@1.0.1` unless I delete the old version from the `node_modules` folder. Isn't that also the case on Travis?

If the answer is yes to either of the two questions, then caching of `node_modules` will leave the project untested on those newly released dependencies and sub-dependencies, and users installing our package from npm might get errors that we didn't catch because of this.

If this project was an app and not an npm package, I'd use a `package-lock.json` and then caching would make sense. But if I understood how caching works on Travis, I don't think it makes sense for npm packages. But I might completely have misunderstood this 😅

cotsog:
@watson Thank you for your insights on how `npm install` works with already installed modules. I was unfortunately wrong and for your specific use case, it would be preferable to not use our caching feature. I've removed it and hopefully fixed the comment format.

watson:
@cotsog thank you so much for all your help. Really appreciated 😃

cotsog:
@watson My pleasure!

Please keep us posted whether this stabilizes your builds or not.

Happy building!

watson:
I just noticed that we had a build where [it timed out after 5 minutes](https://travis-ci.org/elastic/apm-agent-nodejs/jobs/428604515#L510) and restarted the `npm install`. Worked like a charm 👍

PR #564, opened 2018-09-07
test(docker): use in-memory volumes

watson:


watson:
For some reason I cannot get this to work. But if we can get Docker to use in-memory volumes it should speed up the tests

Qard:
Closing because, as far as I can tell, the in-memory storage driver is only for the docker registry server and not docker engine. https://docs.docker.com/registry/storage-drivers/inmemory/

PR #565, opened 2018-09-08
docs: show correct type for disableInstrumentations options

watson:


Qard:
Both strings _and_ an array of strings are valid, should it list both?

watson:
While you technically can supply a string programmatically, we don't mention it anywhere, so I think it's fine to leave that detail out. That you use a string for the environment variable I just see as a necessary hack. It's not because we treat it as a string - we treat it as an array when we parse it.

In a way it's not different from all the number types as they can be strings as well, but we're just converting them to numbers when we parse them.

Qard:
That's fair.

PR #570, opened 2018-09-08
chore(docs): move build_docs.sh script under docs folder

watson:
This just keeps the root of the repo cleaner.


watson:
@dedemorton Does the central docs build need this file to be in its original location or can we put it under another subdirectory like this without problems?

watson:
@Qard I was just checking out if the change to use `npm run docs` in the Docker test worked. On the surface it seemed to have worked, but no where can I see that it actually ran the npm command. It looks like it just ran the old command, but that wouldn't make sense. Can you just verify?

Here's the log: https://apm-ci.elastic.co/job/elastic+apm-agent-nodejs+pull-request+multijob-docs/46/console

PR #571, opened 2018-09-08
chore: rename test/script to test/scripts

watson:
#OCD 😅


watson:
Jenkins except files to be at a certain location... no need to bother with this for now then

PR #575, opened 2018-09-10
chore: use mixins to simplify .travis.yml

Qard:
Just an idea to simplify the `.travis.yml` a bit.


watson:
Though it contained a fair bit of repetition before, it was easier to read. I guess the question is if we want to optimize for reading or writing.

But I don't feel strongly for this, so if you think it's better this way then I'm 👍

Qard:
I have no strong opinion either way. I just tried it to see what it would look like. It's only a difference of 10 lines though. I'm cool with just closing it though, if you feel it reduces readability. 👍

PR #578, opened 2018-09-12
chore(package): pin dependency-check to 2.x version range

watson:
Version 3.2.1 was just released which no longer works on Node.js 4. Version 3.x never officially supported Node.js 4, but did run fine until now.

This commit and be reverted once the following PR is released:
https://github.com/maxogden/dependency-check/pull/99


watson:
Merged this my self as it was blocking tests from passing

PR #579, opened 2018-09-12
feat: improved filtering function API

watson:
Changes the `agent.addFilter()` function to be called once per event
instead of with an array of events.

Besides the existing filtering function, three new custom filtering
functions have been added:

- `agent.addErrorFilter()`
- `agent.addTransactionFilter()`
- `agent.addSpanFilter()`

BREAKING CHANGE: The `agent.addFilter()` API have changed.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


Qard:
LGTM other than failing tests. Seems to be because the abort tests expect no connection to be made, but should it not be that a connection is made, but no data is sent?

watson:
I'll rebase once #580 is merged which fixes the tests

PR #581, opened 2018-09-12
test: update intake API endpoint path

watson:
After I merged the update to the HTTP client these tests have started to fail as we haven't pinned the new API v2 HTTP client to any particular commit/version yet.


watson:
Had to merge this without review so I could get the test suite back in working order

PR #584, opened 2018-09-14
fix: update ids to correct v2 api format

watson:
All the new id's are hex encoded random numbers. Their sizes are:

- error id: 128 bit
- transaction id: 64 bit
- span id: 64 bit
- trace id: 128 bit

The implemented id generator is temporary (except error id's) and will be replaced by the OpenTracing id engine once that lands.

### Checklist

- [x] Implement code
- [x] Add tests


watson:
My original idea was to wait with this until #538 was merged, but I needed the trace-id's in order to make #585

lib/agent.js
```diff
@@ -243,7 +243,7 @@ Agent.prototype.captureError = function (err, opts, cb) {
   var agent = this
   var trans = this.currentTransaction
   var timestamp = new Date().toISOString()
-  var id = uuid.v4()
+  var id = crypto.randomBytes(128 / 8).toString('hex')
```
Qard: Actually, it seems the spec implies this should be 64 bits. https://github.com/elastic/apm-server/blob/v2/docs/spec/errors/v2_error.json#L12

PR #591, opened 2018-09-25
Added initial type definitions for usage in Typescript

### LINKED ISSUE #586: Add @types for typescript
Add @types for typescript

Shahaed:
Resolves elastic/apm-agent-nodejs#586 

Adding initial type declarations for typescript and updating documentation. Not everything is perfectly typed (there are a decent amount of `any`). More explicit typing suggestions are welcome!

### Checklist

- [x] Implement code
- [x] Update documentation

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,78 @@
+// Type definitions for Elastic APM Node.js Agent 1.x
+// Project: https://www.elastic.co/solutions/apm
+// Definitions by: Shahaed Hasan <https://github.com/shahaed>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.7
+
+export = a;
+
+declare const a: a.Agent;
+
+declare namespace a {
+
+    interface Agent {
+
+        currentTransaction: any;
+        logger: any;
+
+        start(opts?: AgentOptions): Agent;
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
+        setUserContext(context: Context): boolean;
+        setCustomContext(context: any): boolean;
+        setTag(name: string, value: string): any;
+        addTags(tags: any): any;
+        captureError(error: Error | string | object, options?: object, callback?: (err?: any) => any): void;
+        startTransaction(name?: string, type?: string): any;
+        endTransaction(result?: any): any;
+        setTransactionName(name: string): any;
+        startSpan(name?: string, type?: any): any;
+        handleUncaughtExceptions(callback?: (err: Error) => any): any;
+        flush(callback?: any): any;
+        lambda(handler: any): any;
+        lambda(type: any, handler: any): any;
+    }
+
+    interface AgentOptions {
+        serviceName?: string;
+        secretToken?: string;
+        serverUrl?: string;
+        verifyServerCert?: boolean;
+        serviceVersion?: string;
+        active?: boolean;
+        instrument?: boolean;
+        asyncHooks?: boolean;
+        ignoreUrls?: (RegExp | string)[];
+        ignoreUserAgents?: (RegExp | string)[];
+        captureBody?: string;
+        errorOnAbortedRequests?: boolean;
+        abortedErrorThreshold?: number;
+        transactionSampleRate?: number;
+        hostname?: string;
+        frameworkName?: string;
+        frameworkVersion?: string;
+        logLevel?: string;
+        logger?: any;
+        captureExceptions?: boolean;
+        captureErrorLogStackTraces?: string;
+        captureSpanStackTraces?: boolean;
+        sourceLinesErrorAppFrames?: number;
+        sourceLinesErrorLibraryFrames?: number;
+        sourceLinesSpanAppFrames?: number;
+        sourceLinesSpanLibraryFrames?: number;
+        errorMessageMaxLength?: number;
+        stackTraceLimit?: number;
+        transactionMaxSpans?: number;
+        flushInterval?: number;
+        serverTimeout?: number;
+        maxQueueSize?: number;
+        filterHttpHeaders?: boolean;
+        disableInstrumentations?: string;
+    }
+
+    interface Context {
+        id?: string;
+        username?: string;
+        email?: string;
+    }
+}
--- a/docs/setup.asciidoc
+++ b/docs/setup.asciidoc
@@ -105,3 +105,30 @@ import apm from 'elastic-apm-node/start'
 
 Then either configure the agent using the environment variables associated with each <<apm-start,config option>>,
 or use the optional <<agent-configuration-file,agent configuration file>>.
+
+[float]
+[[typescript]]
+=== Typescript support
+
+If you are using Typescript, type definitions are provided. Importing the `elastic-apm-node` module:
+
+[source,js]
+----
+import * as apm from 'elastic-apm-node';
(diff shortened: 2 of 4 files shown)


watson:
Thanks for taking this one on 💯 

Two questions:

I don't know much about TypeScript, so this might be a stupid question, but `index.js` is exporting an instance of the `Agent` class. In your `index.d.ts` file, you export both each function by it self _and_ the `Agent` class.

Wouldn't it be possible to define the `Agent` class internally in `index.d.ts` (similar to how defined the `AgentOptions` interface) and then export an instance of that class instead?

I was also wondering if it would be feasible to use the specific types for the function arguments instead of the `any` type. That would probably help TypeScript users quite a lot as well 😃

Shahaed:
@watson I simplified the `index.d.ts` file (following the example from [express](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/express/index.d.ts) ).

I typed the function best I could from the docs. If you don't mind, could you do a code review and point out which `any` types could be better (and tell me the types lol)? :smile:

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        start(opts?: AgentOptions): Agent;
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
+        setUserContext(context: Context): boolean;
+        setCustomContext(context: any): boolean;
+        setTag(name: string, value: string): any;
+        addTags(tags: any): any;
+        captureError(error: Error | string | object, options?: object, callback?: (err?: any) => any): void;
+        startTransaction(name?: string, type?: string): any;
```
Qard: It would be good to implement and return a `Transaction` type here.

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        endTransaction(result?: any): any;
+        setTransactionName(name: string): any;
+        startSpan(name?: string, type?: any): any;
```
Qard: It would be good to implement and return a `Span` type here. Also, `type` should be `string`.

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
+        setUserContext(context: Context): boolean;
+        setCustomContext(context: any): boolean;
+        setTag(name: string, value: string): any;
+        addTags(tags: any): any;
+        captureError(error: Error | string | object, options?: object, callback?: (err?: any) => any): void;
+        startTransaction(name?: string, type?: string): any;
+        endTransaction(result?: any): any;
+        setTransactionName(name: string): any;
```
Qard:
```suggestion
        setTransactionName(name: string): void;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        start(opts?: AgentOptions): Agent;
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
+        setUserContext(context: Context): boolean;
+        setCustomContext(context: any): boolean;
+        setTag(name: string, value: string): any;
+        addTags(tags: any): any;
+        captureError(error: Error | string | object, options?: object, callback?: (err?: any) => any): void;
+        startTransaction(name?: string, type?: string): any;
+        endTransaction(result?: any): any;
```
Qard:
```suggestion
        endTransaction(result?: any): void;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        currentTransaction: any;
+        logger: any;
+
+        start(opts?: AgentOptions): Agent;
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
+        setUserContext(context: Context): boolean;
+        setCustomContext(context: any): boolean;
+        setTag(name: string, value: string): any;
+        addTags(tags: any): any;
```
Qard:
```suggestion
        addTags(tags: any): boolean;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        captureError(error: Error | string | object, options?: object, callback?: (err?: any) => any): void;
```
Qard: The `err` argument of the callback should only ever be an `Error` instance, and the callback return value is not used, so you could make that stricter:

```
captureError(error: Error | string | object, options?: object, callback?: (err?: Error) => void): void;
```

Also, `options` is optional, so another valid signature is:

```
captureError(error: Error | string | object, callback?: (err?: Error) => void): void;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        startTransaction(name?: string, type?: string): any;
+        endTransaction(result?: any): any;
+        setTransactionName(name: string): any;
+        startSpan(name?: string, type?: any): any;
+        handleUncaughtExceptions(callback?: (err: Error) => any): any;
```
Qard:
```suggestion
        handleUncaughtExceptions(callback?: (err: Error) => void): void;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        flush(callback?: any): any;
```
Qard:
```suggestion
        flush(callback?: (err?: Error) => void): void;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+
+        currentTransaction: any;
+        logger: any;
+
+        start(opts?: AgentOptions): Agent;
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
+        setUserContext(context: Context): boolean;
+        setCustomContext(context: any): boolean;
+        setTag(name: string, value: string): any;
```
Qard:
```suggestion
        setTag(name: string, value: string): boolean;
```

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+declare namespace a {
+
+    interface Agent {
+
+        currentTransaction: any;
+        logger: any;
+
+        start(opts?: AgentOptions): Agent;
+        isStarted(): boolean;
+        addFilter(filter: (payload: any) => any): void;
```
Qard: Might be worthwhile for us to describe the payload types here too.

index.d.ts
```diff
@@ -0,0 +1,78 @@
...
+        addTags(tags: any): any;
+        captureError(error: Error | string | object, options?: object, callback?: (err?: any) => any): void;
+        startTransaction(name?: string, type?: string): any;
+        endTransaction(result?: any): any;
+        setTransactionName(name: string): any;
+        startSpan(name?: string, type?: any): any;
+        handleUncaughtExceptions(callback?: (err: Error) => any): any;
+        flush(callback?: any): any;
+        lambda(handler: any): any;
+        lambda(type: any, handler: any): any;
```
Qard:
```suggestion
        lambda(type: string, handler: any): any;
```

watson:
@Shahaed would you still be interested in talking on this PR? If not, that's totally fine of course. If so, we'll probably take a look later when we have more bandwidth 😃

Shahaed:
@watson Yeah, I can finish this PR! I've been AFK for a bit, sorry for the late response

watson:
I just discovered that the definitions on DefinitelyTyped are very out of date (#917), so it would be good to land this ASAP. If you like, I can take over the PR :)

watson:
I'm landing this in #926 instead. Hopefully that will land soon which should make this obsolete. So I'll close this for now. Thanks so much for you help to get this started 😊

PR #594, opened 2018-09-26
feat: add support for Fastify

watson:
### Checklist

- [x] Implement code
- [x] Test correctly capturing of transaction names (both normal and async/await)
- [x] Test correctly capturing of errors handled by Fastify (both normal and async/await)
- [x] Update documentation


watson:
This is still work in progress... I just had a play around with it this evening and wanted to put it up here for now

lib/instrumentation/modules/fastify.js
```diff
@@ -0,0 +1,31 @@
...
+  return wrappedBuild
+
+  function wrappedBuild () {
+    const _fastify = fastify.apply(null, arguments)
+
+    agent.logger.debug('adding onRequest hook fo fastify')
+
+    if (semver.gte(version, '2.0.0')) {
+      _fastify.addHook('onRequest', (req, reply, next) => {
+        const name = req.method + ' ' + reply.context.config.url
```
Qard: Could simplify by just doing `(reply.context || reply._context)` to support all versions with one handler.

watson: Doh, great idea 😄

watson:
@delvedor when using async/await, Fastify will automatically catch the error and report a 500 back to the client. Is there a similar way to response with an error object back to the client without using async/await? I want to make a test that checks that we capture that.

P.S. As you can see right now, the [async/await error test](https://github.com/elastic/apm-agent-nodejs/pull/594/files#diff-3c323d397ad6036c317aae4a46c308f9R46) actually fails because we don't get to send the error to Elastic APM. Is there a central way, similar to the `onRequest` hook, where the agent can be notified on captured/reported errors?

delvedor:
@watson you are correct, when using async await Fastify will automatically catch errors, but it will not if you are using plain old callbacks.

If you want to handle an error, you should use the [`setErrorHandler`](https://www.fastify.io/docs/latest/Server/#seterrorhandler) API, but I would not recommend it since a user could override your handler.
If you just need to be notified, you can take advantage of the `onResponse` hook, and there you will be able to access the status code of the response. [Here](https://www.fastify.io/docs/latest/Lifecycle/) you can see the lifecycle of every request. (You can also use the `onSend` hook, but a user might want to change the response there)

docs/fastify.asciidoc
```diff
@@ -0,0 +1,204 @@
...
+It's important that the agent is started before you require *any* other modules in your Node.js application - i.e. before `fastify`, `http`, etc.
+
+This means that you should probably require and start the agent in your application's main file (usually `index.js`, `server.js` or `app.js`).
+
+Here's a simple Fastify example with the Elastic APM agent installed:
+
+[source,js]
+----
+// Add this to the VERY top of the first file loaded in your app
+var apm = require('elastic-apm-node').start({
```
delvedor: I would also suggest to add `apm` to the Fastify instance by using the [`decorate`](https://www.fastify.io/docs/latest/Decorators/) API, in this way you can make `apm` accessible in every part of the application, which is very useful if you need to catch custom errors (as you have documented below).

```js
const apm = require('elastic-apm-node').start({ ... })

const fastify = require('fastify')()
fastify.decorate('apm', apm) // now you can do `fastify.apm` to access it
...

```

delvedor:
> P.S. As you can see right now, the async/await error test actually fails because we don't get to send the error to Elastic APM. Is there a central way, similar to the onRequest hook, where the agent can be notified on captured/reported errors?

Internally Fastify catches errors generated by async await functions and sends the error to the user with [this utility](https://github.com/fastify/fastify/blob/master/lib/wrapThenable.js), [here](https://github.com/fastify/fastify/blob/master/lib/handleRequest.js#L92-L95) you can see how we are using it.
Doing `reply.send(new Error('kaboom'))` in a classic handler or `throw new Error('kaboom')` is an async handler should return the same things since internally we are calling `reply.send` for the user.

watson:
@delvedor I need access to the original error instance unfortunately as the agent needs to send it up to the APM Server. Each of our framework instrumentations take the errors picked up by the framework and report them to the APM Server. In Express for instance, we automatically [attach a middleware error handler](https://github.com/elastic/apm-agent-nodejs/blob/019705e83e24b112084fb752ea94987d3fab0d1d/lib/instrumentation/modules/express.js#L43-L49) so we can snatch up any errors caught by Express. So it would be preferable if we could do something similar for Fastify.

It sounds like that if I patch `reply.send` then I can get access to the error object.

docs/fastify.asciidoc
watson: The node agent is a singleton, and every time you `require('elastic-apm-node)` we return the agent to you. You of course only need to call `start()` once, which you usually do in you index file. So you can just require the agent again in another file to get access to it. That's what you meant right?

delvedor: > That's what you meant right?

Yes! But since the agent is a singleton, you don't need to use the `decorate` API :)

watson:
Will `setErrorHandler` be called for all errors given to `reply.send` - both those thrown inside async/await functions and those given manually to `reply.send`?

If we were to use the `setErrorHandler` approach we would have to find a way around the limit that there can only be _one_ error handler at any given time - so if someone called `setErrorHandler` after we did, ours would be overwritten (if I understand you correctly). If so, I think that would require a lot of patching. So maybe simply overwriting `reply.send` would be safer and simpler?

watson:
Looks like [Fastify is now more popular than Restify](https://hueniverse.com/on-node-framework-popularity-355fbde5cbe4) 🎉 - just another reason to add support for this framework 😄

KidkArolis:
I've put together a little feature request on Fastify repo for exposing route patterns so that data could be grouped by route pattern in tools such as Prometheus or Elastic APM: https://github.com/fastify/fastify/issues/1275

Let me know if you think that makes sense.

KidkArolis:
Actually, it was a red herring, the router pattern is in fact exposed to plugins.

watson:
@delvedor updated the code use use the new `onError` event and fixed an incompatibility introduced in version 2.0.0-rc by nesting the request under `req.raw`.

As you can see I've left support for pre v2 in, though that will not support automatic error capturing as also stated by the warning log output. Still undecided if we should remove support for pre v2 completely.

We're not in the habit of releasing support for pre-release versions, so for now I think we should wait merging this until version 2.0.0 is released and that we have updated `package.json` to match.

If you have time, could you go over the included Fastify documentation and validate that the code examples are ok in relation to version 2 of Fastify?

.tav.yml
```diff
@@ -221,6 +221,12 @@ restify:
   commands:
     - node test/instrumentation/modules/restify.js
 
+fastify:
+  versions: '>=0.27.0 <0.29.0 || >0.29.0'
```
delvedor: I would recommend to use Fastify ≥1, pre v1 versions where highly unstable and nobody should use them :)

lib/instrumentation/modules/fastify.js
```diff
@@ -0,0 +1,47 @@
+'use strict'
+
+const semver = require('semver')
+
+module.exports = function (fastify, agent, version, enabled) {
+  if (!enabled) return fastify
+
+  agent.logger.debug('wrapping fastify build function')
+
+  return semver.gte(version, '2.0.0-rc') ? wrappedBuild2Plus : wrappedBuildPre2
```
delvedor: Pay attention that `semver` does not handle `-rc`, sometimes this can cause some unexpected behaviors.

lib/instrumentation/modules/fastify.js
```diff
@@ -0,0 +1,47 @@
...
+
+  function wrappedBuild2Plus () {
+    const _fastify = fastify.apply(null, arguments)
+
+    agent.logger.debug('adding onRequest hook to fastify')
+    _fastify.addHook('onRequest', (req, reply, next) => {
```
delvedor: I don't know how your internal implementation works, but if needed you can register also the `onResponse` hook, that will be fired once the Node.js's response fires the `finish` or the `error` events.
The signature of the hook in Fastify v2 is the same of the `onRequest` hook, so you can access the exact same data if needed :)

delvedor:
> We're not in the habit of releasing support for pre-release versions, so for now I think we should wait merging this until version 2.0.0 is released and that we have updated package.json to match.

Agree entirely, furthermore the rc period will not last long :)

lib/instrumentation/modules/fastify.js
watson: True, pre-release versions can be a little hard to reason about when it comes to semver. But in this case it should be ok. Normally I wouldn't measure against a pre-release version, but to get this working with the current pre-release version I had to do it this way.

```js
semver.gte('1.0.0', '2.0.0-rc') // => false
semver.gte('1.0.0-rc.0', '2.0.0-rc') // => false
semver.gte('2.0.0-rc.0', '2.0.0-rc') // => true
semver.gte('2.0.0', '2.0.0-rc') // => true
semver.gte('3.0.0', '2.0.0-rc') // => true
```

lib/instrumentation/modules/fastify.js
watson: I automatically get notified when the request starts and ends without a hook in Fastify because the agent hooks into the `http` module it self. What I need this hook for is knowing what to name the transaction - here I need access to the router. That's all I use this for.

.tav.yml
watson: Yeah, maybe there's no need to test this. And indeed we might even bump it to v2+ as that's the only range where we also support error handling. But normally I just go as low as we can without any modifications to our code and it turned out that in this case the `onRequest` hook was supported in the current configuration all the way down to v0.27 - so that's why I added that as a minimum - i.e. we got it basically for free.

watson:
Status: Waiting for v2.0.0 of Fastify

watson:
v2.0.0 shipped: https://twitter.com/fastifyjs/status/1099953954983682053

lib/instrumentation/modules/fastify.js
delvedor: You can remove the `rc` if you want 🚀

Qard:
Should this be moved to in-review, or is there more that needs to be done still?

docs/supported-technologies.asciidoc
```diff
@@ -50,6 +50,7 @@ These are the frameworks that we officially support:
 so we can't support Koa directly since we rely on router information for full support.
 We currently support the most popular Koa router called https://github.com/alexmingoia/koa-router[koa-router]
 |<<restify,Restify>> |>=5.2.0 |
+|<<fastify,Fastify>> |>=0.27.0 |
```
Qard: Do we _want_ to advertise support for pre-1.x versions, given the stated instability of that range?

watson: 🤷‍♂️ our previous approach have been to just show support for the lowest possible versions where our tests still pass without us having to make any crazy changes. If you like I can change it to say 1.0.0 or 2.0.0 though?

Qard: I'm good either way. Feel free to land, if you are happy with the support range.

watson:
This PR have now been released in version 2.6.0

PR #600, opened 2018-09-28
chore(agent): remove deprecated buildSpan

### LINKED ISSUE #410: Remove deprecated buildSpan API
The `buildSpan` API is deprecated and have been replaced by the `startSpan` API.

Qard:
Fixes #410


watson:
When I deprecated this back in the days, it was a documentation-only deprecation as I was in a hurry and didn't have time change all the places where we used it in the code.

But I think it's just a matter of searching for all places where `buildSpan` is referenced and manually updating the code to use `startSpan`. It means just reorganizing how the span name and span type is set, but if I recall my previous research correctly it should be a fairly straight forward refactor 😃

P.S. The reason why this API became deprecated was because the OpenTracing Node.js agent only relies on a `startSpan` function, and so I wanted to make sure that we wouldn't become incompatible and make it easier to switch between the two API's.

Since then however I have found that other OpenTracing language implementations do have a `buildSpan` concept. So if we decide that this is actually required for some specific use-case, I'm ok with rethinking this.

Qard:
If it exists in other OpenTracing implementations, do we even _want_ to remove it? We could just leave it as-is. 🤔

PR #601, opened 2018-09-28
feat(config): support byte suffixes

### LINKED ISSUE #572: Implement byte config suffixes
Just as with #545, the config options used to configure byte sizes should be written with a suffix so its clear if the value is to be interpreted as bytes, kilobytes, megabytes or gigabytes.

Here's the options that are durations:

Config option | Release | Default
--------------|---------|---------
`apiRequestSize` | 2.x | 750kb

We need to decide what should happen in 2.x when a suffix isn't given:

1. ~~Should we ignore the given config value, output a warning, and fall back to the default value?~~
1. ~~Should we output an error and make the agent inactive?~~
1. ~~Should we throw an error?~~
1. Assume the user meant to use the unit that was the default in 1.x, but output a deprecation warning that it will be required in v3.0.0

Qard:
Fixes #572

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -500,11 +500,16 @@ The default value `0` means that no source code will be collected for span libra
 [[error-message-max-length]]
 ===== `errorMessageMaxLength`
 
-* *Type:* Number
-* *Default:* `2048`
+* *Type:* String
+* *Default:* `2kb`
 * *Env:* `ELASTIC_APM_ERROR_MESSAGE_MAX_LENGTH`
 
-The maximum length allowed for error messages in bytes.
+The maximum length allowed for error messages.
+It is expressed in bytes or includes a size suffix such as `2kb`.
+Size suffixes are case-insensitive and include `b`,
+`kb`,
+`mb`,
+and `gb`.
 Messages above this length will be truncated before being sent to the APM Server.
 
 Set to `-1` do disable truncation.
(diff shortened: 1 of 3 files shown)


docs/agent-api.asciidoc
```diff
@@ -560,8 +561,8 @@ Use the <<api-request-size,`apiRequestSize`>> config option to control the byte
 [[api-request-size]]
 ===== `apiRequestSize`
 
-* *Type:* Number
-* *Default:* `1048576` (1 MiB)
+* *Type:* String
+* *Default:* `1mb` (1 MiB)
```
watson: No reason to write the MB size in brackets anymore

Qard:
I've added docs for the valid suffixes. Technically `tb` is also valid, with the use of the `bytes` module, but is not included in our spec. Is it fine for it to be supported and just not documented, or should we do another layer of regex validation like I did for times?

watson:
This wasn't clear in the original issue, but the parsing rules should actually be quite strict:

> - we accept values matching the regular expression `^(\d+)(B|KB|MB|GB)$` (with case insensitivity enabled). The multipliers are base-2.
> - parsing will be strict, to ensure compatibility between agents: no white space, no additional suffixes, no floating point values

The `bytes` module is "unfortunately" cleaver and will parse both whitespace and floats. While this is nice for Node.js users, it can introduce incompatibilities with the other agents if config values is shared.

Qard:
Switched to a regex parser, like I did for the time suffix parser.

PR #602, opened 2018-09-28
feat(config): support time suffixes

### LINKED ISSUE #545: Implement duration config suffixes
For duration configs (e.g. ELASTIC_APM_API_REQUEST_TIME), we will support time units like 1ms, 1s, 1m. For pre-v2 config, we should allow suffixes, but also continue to allow no suffix and assume the previous unit (e.g. "ms" for min span duration). For v2+ config, we should require suffixes.

Qard:
Fixes #545

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


docs/agent-api.asciidoc
```diff
@@ -581,11 +584,13 @@ The HTTP request is otherwise ended at regular intervals controlled by the <<api
...
 
-* *Type:* Number
-* *Default:* `30`
+* *Type:* String
+* *Default:* `30s`
 * *Env:* `ELASTIC_APM_SERVER_TIMEOUT`
 
-Specify the response timeout in seconds when reporting transactions to APM Server.
+Specify the response timeout when reporting transactions to APM Server.
+The timeout should be in milliseconds,
```
watson: Previously the default unit was seconds. I think the other agents who have implemented this also use seconds as the default unit. Would it make sense to keep this default unit until v3.0.0?

package.json
```diff
@@ -78,6 +78,7 @@
     "fast-safe-stringify": "^2.0.4",
     "http-headers": "^3.0.2",
     "is-native": "^1.0.1",
+    "ms": "^2.1.1",
```
watson: The ms package uses a [much looser time suffix parser](https://www.npmjs.com/package/ms#examples) than what we have decided to support. This theoretically means that a user can write `2.5 hrs` instead of `150m`. This will work fine with the Node.js agent, but if they try to use the same value for the RUM JavaScript agent or the Java agent, it might fail.

If I recall correctly from when we originally discussed this, we wanted to only allow a strict subset of suffixes. Is that still the case?

cc @elastic/apm-agent-devs

felixbarny: Yes, the intention was to be rather strict in order to align across agents to support the case when different agents read from the same environment variable or, going forward, a centralized configuration.

watson: Thanks @felixbarny 

@Qard Instead of reinventing the well, we could continue to use the `ms` package, but add a regex to validate the string before passing it along to the `ms` package. That would be a quick and easy solution. What do you think about that?

Qard: If we want to limit it that much, it's probably better to just skip ms altogether. It's a fairly basic conversation map when there are only three suffixes. 🤔

docs/agent-api.asciidoc
Qard: Well, we also have `abortedErrorThreshold`, which is currently counted in milliseconds, so I think we should at least unify them all on _something_.

Qard: Updated to default to seconds.

package.json
Qard: I've replaced ms with a simple `toSeconds` function that uses the regex with only `s`, `m`, and `ms` support.

docs/agent-api.asciidoc
watson: I do think that we should wait until the unit is required (v3.0.0) to make any changes to the default units.

Qard: Yeah, I switched the default back to seconds, so it should be backwards compatible. 👍

watson:
@Qard I merged the size suffix PR, so there's a bunch of conflicts now

watson:
Adding breaking change label because we're changing the default unit of `ELASTIC_APM_ABORTED_ERROR_THRESHOLD` from ms to s.

This is ok as this config option is not shared with any of the other agents

watson:
@Qard please rebase with api-v2 - all the errors have been fixed now 😃

watson:
If you merge your self, please append the following 3 lines to the squashed merge commit:

```

BREAKING CHANGE: `abortedErrorThreshold` expects its value to be in
seconds.
```

This will make it easier for us when we need to make the upgrade guide from v1 to v2

docs/agent-api.asciidoc
```diff
@@ -591,11 +594,13 @@ The HTTP request is otherwise ended at regular intervals controlled by the <<api
...
```
watson:
```suggestion
The timeout should be in seconds,
```

lib/config.js
```diff
@@ -34,16 +34,21 @@ var DEFAULTS = {
   verifyServerCert: true,
   active: true,
   logLevel: 'info',
+<<<<<<< HEAD
```
watson: 🤔

Qard: 🤦‍♂️ fixed.

PR #604, opened 2018-10-01
feat(config): change config order

### LINKED ISSUE #287: Change order in which config options overrule each other for all agents
Currently the Node.js agent can be configured using a combination of environment variables, a config file, and in-line config options.

Right now config options set in a config file will overrule the same config options set in-line. Likewise, config options set using environment variables will overrule the same config options set in a config file.

It would make sense to align this order across all backend Elastic APM agents (currently Python, Ruby, Go, and Java). I'm leaving out the frontend JavaScript agent as I think the methods of configuration here are restricted to only in-line configuration.

It has been proposed to change the order to the following (higher overrule lower):

1. Environment variables
1. In-line configuration
1. Config file
1. Default values

Reason being that it should be easy for people deploying the application to overrule config options set by the developer in either the config file or in-line. It might also be a use-case that a config file is shared among different apps, so it should be possible to overwrite config options in-line. That's at least our suggestion.

But we'd love input on if this aligns with real world use-cases or if we missed something obvious 😃 

P.S. Changing this order should of course be considered a breaking change, so for the Node.js agent we'll have to bump the major for this to be changed.

/cc @elastic/apm-agent-devs

Qard:
Fixes #287

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -130,9 +130,9 @@ function config (opts) {
   opts = Object.assign(
     {},
     DEFAULTS, // default options
-    readEnv(), // options read from environment variables
     confFile, // options read from elastic-apm-node.js config file
-    opts // options passed in to agent.start()
+    opts, // options passed in to agent.start()
+    readEnv() // options read from environment variables
   )
 
   // Custom logic for setting serviceName so that an empty string in the config
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -52,7 +52,7 @@ by using environment variables,
 or via the <<agent-configuration-file,agent configuration file>>.
 Their equivalent environment variable name is listed together with each option.
 
-Properties on the `options` object will always take precedence over environment variables.
+Environment variables will always take precedence over properties of the `options` object.
 
 The only required parameter is <<service-name,`serviceName`>>.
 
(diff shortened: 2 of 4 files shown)


Qard:
Updated the docs to indicate the new precedence order.

lib/config.js
```diff
@@ -130,9 +130,9 @@ function config (opts) {
   opts = Object.assign(
     {},
     DEFAULTS, // default options
-    readEnv(), // options read from environment variables
     confFile, // options read from elastic-apm-node.js config file
-    opts // options passed in to agent.start()
+    opts, // options passed in to agent.start()
+    readEnv() // options read from environment variables
```
watson: @felixbarny just added you as reviewer as well as I wanted to confirm that this is the same order that you have implemented in the Java agent (with `Object.assign` later arguments override earlier)

felixbarny: The order in the Java agent is defaults, env, config file

felixbarny: So yes, it's pretty much the same in the java agent

watson: java system properties are command line flags right?

felixbarny: correct

PR #613, opened 2018-10-15
feat(config): stricter boolean parsing

### LINKED ISSUE #606: Change boolean config options to only use true/false
Today, when reading environment variables, our boolean options (e.g. `ELASTIC_APM_ACTIVE`) allow both several string values to mean `true`, e.g. `"true"`, `"1"`, `"yes"`. Similar for `false`.

To align with the other agents we should limit this so that `"true"` means `true` and only `"false"` means `false`.

Qard:
Fixes elastic/apm-agent-nodejs#606

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -4,7 +4,6 @@ var fs = require('fs')
 var path = require('path')
 
 var consoleLogLevel = require('console-log-level')
-var normalizeBool = require('normalize-bool')
 var readPkgUp = require('read-pkg-up')
 var truncate = require('unicode-byte-truncate')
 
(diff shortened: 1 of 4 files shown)


Qard:
@watson Should we leave `captureBody` with `off`, `all`, `errors`, `transactions`? That's what this does currently.

lib/config.js
```diff
@@ -211,11 +210,22 @@ function normalizeArrays (opts) {
...
 function truncateOptions (opts) {
   if (opts.serviceVersion) opts.serviceVersion = truncate(String(opts.serviceVersion), config.INTAKE_STRING_MAX_SIZE)
   if (opts.hostname) opts.hostname = truncate(String(opts.hostname), config.INTAKE_STRING_MAX_SIZE)
 }
+
+function strictBool (value) {
+  if (typeof value === 'boolean') {
+    return value
+  }
+  // This will return undefined for unknown inputs, resulting in them being skipped.
```
watson: We should probably log a warning that the value was invalid and that we fell back to the default

watson:
True, access to the logger here isn't easy. Could we use STDERR as a pragmatic solution for this?

The other config options will simply let the invalid value pass through and you'll get an exception or undefined behavior later on when the value is used - so you'll most likely discover you did something wrong. I think that's at least a consistant behavior that is to be expected.

Simply ignoring the input will probably be tricky for the user to reason about and can be very hard to debug.

But I agree that it would be even better if we validated all the input. But that has a lower priority IMO.

lib/config.js
Qard: fixed.

Qard:
I reorganized things slightly to be able to initialize the logger before normalizing bools. I've now added a warning message when an unknown format is found.

watson:
@Qard please rebase with api-v2 - all the errors have been fixed now 😃

watson:
I assume there's no documentation updates? If so just remove the checkboxes you don't need, otherwise it looks like the PR isn't finished

watson:
Seems like there's some linter errors as well according to the tests. Running `standard --fix` should fix it

Issue #473, opened 2018-07-25
Move configuration docs to separate page


watson:
Today the different config options are listed together with the API docs. This makes them hard to navigate to as the right hand menu doesn't show the headlines in that depth. If we had a separate configurations page it would render more nicely. I also think that it would be more logical to link to instead of having it buried under the `apm.start()` docs.

watson:
We need to check with the website team if we can do redirects easily from the old to the new documentation before implementing it

watson:
@dedemorton do we have any process of redirecting anchor links in our doc pages? We're considering moving a whole group of headlines to it's own new page.

dedemorton:
@watson We **cannot** redirect from anchored sections, like https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-start.

To set up redirects from one _page_ to another, we create a request in the website repo (here's [an example](https://github.com/elastic/website-www.elastic.co/issues/2915)).

PR #626, opened 2018-10-18
docs: split agent-api configuration options to a new doc

Qard:
Closes elastic/apm-agent-nodejs#473


docs/configuration.asciidoc
```diff
@@ -0,0 +1,623 @@
+[[configuration]]
+
+ifdef::env-github[]
+NOTE: For the best reading experience,
+please view this documentation at https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html[elastic.co]
+endif::[]
+
+== Configuration
```
bmorelli25:
```suggestion
=== Configuration
```

docs/configuration.asciidoc
```diff
@@ -0,0 +1,623 @@
...
+  // Set custom APM Server URL (default: http://localhost:8200)
+  serverUrl: '',
+
+  // Only activate the agent if it's running in production
+  active: process.env.NODE_ENV === 'production'
+})
+----
+
+[[service-name]]
+=== `serviceName`
```
bmorelli25:
```suggestion
==== `serviceName`
```

docs/configuration.asciidoc
Qard: fixed.

docs/configuration.asciidoc
Qard: fixed.

docs/configuration.asciidoc
```diff
@@ -0,0 +1,623 @@
...
+
+The available configuration options are listed below.
+Most configuration options can be set either in the optional `options` object,
+by using environment variables,
+or via the <<agent-configuration-file,agent configuration file>>.
+Their equivalent environment variable name is listed together with each option.
+
+Environment variables will always take precedence over properties of the `options` object.
+
+The only required parameter is <<service-name,`serviceName`>>.
```
watson: While not strictly related to this PR, I just stumbled upon this sentence that we probably overlooked when we implemented to fallback to `package.json`. So while we remember, we might as well update this text to say:

```suggestion
The only required parameter is <<service-name,`serviceName`>>.
However,
the agent will use the `name` from `package.json` by default if available.
```

Qard: fixed.

Qard:
Rebased.

docs/configuration.asciidoc
watson: 🤔 I don't see it anywhere. Maybe something went wrong during the rebase? 😬

docs/configuration.asciidoc
```diff
@@ -0,0 +1,623 @@
...
+
+Example usage configuring the agent to only be active in production:
+
+[source,js]
+----
+// Add this to the VERY top of the first file loaded in your app
+require('elastic-apm-node').start({
+  // Overwrite service name from package.json
```
watson: I think the change got lost in the rebase:

```suggestion
  // Override service name from package.json
```

Qard:
It's under the API reference section, so it's already covered by the link to that. We aren't currently linking to the sub-pages, should we?

watson:
Ah yes, good point 🤔 By default we shouldn't link to sub-pages, but I'm a little on the fence now because this is such an important page...

watson:
Let's just agree not to do it as part of this PR. If we later feel we need to link to it, it can go in a PR about reformatting the ToC 😃

watson:
Not sure why codecov is failing. I think if the commit msg CI is fixed then codecov will probably fix it self

PR #628, opened 2018-10-19
feat(http): split disabling of spans from transactions

### LINKED ISSUE #434: Allow more fine grained disabling of http module features
It's not possible to switch off the automatic creation of HTTP request transactions when an incoming  HTTP request is detected without also switching off tracing of outgoing HTTP requests. This is because they are both controlled by this cone setting: [`disableInstrumentations: ['http']`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#disable-instrumentations).

It would be very useful - both to users, but especially also during debugging - if it was possible to control the two features separately.

### Considerations

Today there's a one to one relationship between npm package names and the content of the `disableInstrumentations` config option. I don't think it would be a good idea to introduce an exception just to solve this case.

As an alternative we could consider if the config option instead should focus on span types, so instead of disabling `pg`, you would disable `db.postgresql` (the full type is `db.postgresql.query`, but I guess we can just match against the start of the string, which also means that we could write `db` to disable _all_ db spans).

There's pros and cons to this, so it should be considered thoroughly.

Qard:
Fixes elastic/apm-agent-nodejs#434

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


Qard:
I might wait for #626 to land before writing the docs for this... 🤔

lib/config.js
```diff
@@ -222,9 +225,9 @@ function maybeSplit (value) {
   return typeof value === 'string' ? value.split(',') : value
 }
 
-function normalizeArrays (opts) {
-  ARRAY_OPTS.forEach(function (key) {
-    if (key in opts) opts[key] = maybeSplit(opts[key])
+function normalizeSets (opts) {
```
watson: Good idea! I like it 👍

alvarolobato:
It was discussed that this needs a different approach to make it a non breaking change. Will create a new PR.

Qard:
Go home, Zube. You're drunk.

PR #636, opened 2018-10-25
feat(agent): return uuid in captureError callback

### LINKED ISSUE #573: Make apm.captureError() return error uuid
**The problem**
We pass uuids of internal errors through our client API, so that clients can trace and report them back to us. Because there is currently no way of getting the uuid that the agent assigns to errors, we have to add a tag with another uuid, that we can also pass with the error response.

**Preferred solution**
Having `apm.captureError()` return the assigned uuid through callback or directly.

**Alternatives considered**
Using tags with a different uuid.

Qard:
Fixes #573

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -845,7 +845,7 @@ For details see the <<metadata,metadata section>>.
 
 The optional `callback` will be called after the error has been sent to the APM Server.
 It will receive an `Error` instance if the agent failed to send the error,
-and the uuid of the captured error.
+and the id of the captured error.
 
 [[message-strings]]
 ===== Message strings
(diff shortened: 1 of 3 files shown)


docs/agent-api.asciidoc
```diff
@@ -844,6 +844,8 @@ The optional `options` object can be used to log additional metadata with the er
 For details see the <<metadata,metadata section>>.
 
 The optional `callback` will be called after the error has been sent to the APM Server.
+It will receive an `Error` instance if the agent failed to send the error,
+and the uuid of the captured error.
```
watson: Better change this to just "id" as we're going away from uuid in v2 (just so we don't forget)

```suggestion
and the id of the captured error.
```

PR #639, opened 2018-10-26
docs: improve custom transactions example

### LINKED ISSUE #276: Example in Custom Transaction documentation is wrong
The example used to showcase [custom transactions](https://www.elastic.co/guide/en/apm/agent/nodejs/1.x/custom-transactions.html) have a bug as it calls the sqs callback more than once (once per job in the message).

The correct way would be something like this:

```diff
 var apm = require('elastic-apm-node').start()
 var sqs = require('simple-sqs')()
+var afterAll = require('after-all')

 // listen for jobs on the queue
 var queue = sqs(queueUrl, function (msg, cb) {
+  // Notify simple-sqs once all jobs in the current batch have
+  // been processed
+  var next = afterAll(cb)
+
   // The SQS queue will send multiple messages using an array
   // of records
   msg.Body.Records.forEach(function (job) {
     // start one new transaction for each job record received
     // on the queue
     var name = 'Job ' + job.type
     var type = 'job'
     var trans = apm.startTransaction(name, type)

-    // call the function that actually processes the job
+    // Prepare callback to be called asynchronous
+    var done = next()
+
+    // call the function that actually processes the job
     processJob(job, function (err) {
       // if the job could not be processes, set the result to
       // a 5xx error code. Here 500 indicates error, 200 is ok
       trans.result = err ? 500 : 200

       // end the transaction
       trans.end()

-      cb(err)
+      done(err)
     })
   })
 })

 queue.on('error', function (err) {
   // in case the queue encounters an error, report it to Elastic APM
   apm.captureError(err)
 })
```

But this is not a good teaching example as it requires you to think about what `after-all` is and just makes the flow harder to understand.

We should come up with a better example.

Qard:
Fixes #276

### DIFF
--- a/docs/custom-transactions.asciidoc
+++ b/docs/custom-transactions.asciidoc
@@ -42,7 +42,7 @@ var queue = sqs(queueUrl, function (msg, cb) {
       processJob(job, function (err) {
         // if the job could not be processes, set the result to
         // a 5xx error code. Here 500 indicates error, 200 is ok
-        trans.result = err ? 500 : 200
+        trans.result = err ? 'error' : 'success'
 
         // end the transaction
         trans.end()
--- a/docs/custom-transactions.asciidoc
+++ b/docs/custom-transactions.asciidoc
@@ -30,25 +30,35 @@ var sqs = require('simple-sqs')()
 var queue = sqs(queueUrl, function (msg, cb) {
   // The SQS queue will send multiple messages using an array
   // of records
-  msg.Body.Records.forEach(function (job) {
-    // start one new transaction for each job record received
-    // on the queue
-    var name = 'Job ' + job.type
-    var type = 'job'
-    var trans = apm.startTransaction(name, type)
+  var tasks = msg.Body.Records.map(function (job) {
+    return new Promise(function (resolve, reject) {
+      // start one new transaction for each job record received
+      // on the queue
+      var name = 'Job ' + job.type
+      var type = 'job'
+      var trans = apm.startTransaction(name, type)
 
-    // call the function that actually processes the job
-    processJob(job, function (err) {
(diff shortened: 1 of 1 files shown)


docs/custom-transactions.asciidoc
```diff
@@ -30,25 +30,35 @@ var sqs = require('simple-sqs')()
...
+      // on the queue
+      var name = 'Job ' + job.type
+      var type = 'job'
+      var trans = apm.startTransaction(name, type)
+
+      // call the function that actually processes the job
+      processJob(job, function (err) {
+        // if the job could not be processes, set the result to
+        // a 5xx error code. Here 500 indicates error, 200 is ok
+        trans.result = err ? 500 : 200
```
watson: Previously only numeric result values was supported, but we now support strings as well, so we should update this so it doesn't look so HTTP like 😄 

```suggestion
        trans.result = err ? 'error' : 'success'
```

docs/custom-transactions.asciidoc
```diff
@@ -30,25 +30,35 @@ var sqs = require('simple-sqs')()
...
+        if (err) {
+          reject(err)
+        } else {
+          resolve()
+        }
+      })
     })
   })
+
+  Promise.all(tasks).then(function () {
```
watson: Any reason to not just use `cb` directly stead of the nested `function`?

Qard: Just to avoid passing values into the error argument. I suppose we resolve without an argument though, so maybe it'd _technically_ be fine. It's just not good practice to pass error-first callbacks directly into promise resolve callbacks though. 🤔

watson: Ah ok, makes sense.

PR #644, opened 2018-10-27
feat(config): custom transport

### LINKED ISSUE #643: Custom transport
It would be nice to be able to provide a custom transport to the agent to use in-place of the http-client, allowing data to be shipped anywhere the user wants. It would also help for modularirty reasons to clearly defined the borders of the API interface module.

Related: #627

Qard:
Fixes elastic/apm-agent-nodejs#643

~~Waiting for elastic/apm-agent-nodejs#626 to merge before adding docs.~~

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -6,7 +6,6 @@ var parseUrl = require('url').parse
 var path = require('path')
 
 var afterAll = require('after-all-results')
-var ElasticAPMHttpClient = require('elastic-apm-http-client')
 var isError = require('core-util-is').isError
 var ancestors = require('require-ancestors')
 
@@ -22,7 +21,6 @@ var IncomingMessage = http.IncomingMessage
 var ServerResponse = http.ServerResponse
 
 var version = require('../package').version
-var userAgent = 'elastic-apm-node/' + version
 
 module.exports = Agent
 
@@ -33,7 +31,7 @@ function Agent () {
   this._errorFilters = new Filters()
   this._transactionFilters = new Filters()
   this._spanFilters = new Filters()
-  this._apmServer = null
+  this._transport = null
 
   this._conf = null
   this._httpClient = null
@@ -61,7 +59,7 @@ Object.defineProperty(Agent.prototype, 'currentSpan', {
 })
 
 Agent.prototype.destroy = function () {
-  if (this._apmServer) this._apmServer.destroy()
+  if (this._transport) this._transport.destroy()
 }
 
 Agent.prototype.startTransaction = function () {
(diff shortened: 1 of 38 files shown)


lib/agent.js
```diff
@@ -144,37 +144,41 @@ Agent.prototype.start = function (opts) {
...
-    time: this._conf.apiRequestTime * 1000,
-
-    // Debugging
-    payloadLogFile: this._conf.payloadLogFile
-  })
-  this._apmServer.on('error', err => {
-    this.logger.error('An error occrued while communicating with the APM Server:', err.message)
-  })
+  if (this._conf.transport) {
+    this._apmServer = this._conf.transport(this._conf)
```
watson: Thinking if we should rename `_apmServer` to `_transport` 🤔

lib/agent.js
```diff
@@ -144,37 +144,41 @@ Agent.prototype.start = function (opts) {
...
+      // metadata
+      agentName: 'nodejs',
+      agentVersion: version,
+      serviceName: this._conf.serviceName,
+      serviceVersion: this._conf.serviceVersion,
+      frameworkName: this._conf.frameworkName,
+      frameworkVersion: this._conf.frameworkVersion,
+      hostname: this._conf.hostname,
+
+      // Sanitize conf
```
watson: It would be nice if we could put the sanitation logic in a generic `Transport` object that all transports inherited from

test/config.js
```diff
@@ -535,6 +535,92 @@ test('disableInstrumentations', function (t) {
...
+          list.push(item)
+          if (callback) {
+            setImmediate(callback)
+          }
+        }
+      }
+      var first = true
+      return {
+        sendTransaction: makeSenderFor(transactions),
+        sendSpan: makeSenderFor (spans),
```
watson:
```suggestion
        sendSpan: makeSenderFor(spans),
```

test/config.js
```diff
@@ -535,6 +535,92 @@ test('disableInstrumentations', function (t) {
...
+function assertEncodedError (t, error, result, trans, parent) {
+  t.comment('error')
+  t.ok(result.id, 'has a valid id')
+  t.equal(result.trace_id, trans.traceId, 'trace id matches')
+  t.equal(result.transaction_id, trans.id, 'transaction id matches')
+  t.equal(result.parent_id, parent.id, 'parent id matches')
+  t.ok(result.exception, 'has an exception object')
+  t.equal(result.exception.message, error.message, 'exception message matches')
+  t.equal(result.exception.type, error.constructor.name, 'exception type matches')
+  t.ok(result.culprit, 'has a valid error')
```
watson:
```suggestion
  t.ok(result.culprit, 'has a valid culprit')
```

lib/agent.js
Qard: Renamed to `_transport`.

lib/agent.js
Qard: Do we want this PR to depend on that, or should build and land that independently? 🤔

watson: As long as we don't document the new `transport` config option, then I'm ok with landing it independently

PR #648, opened 2018-10-31
feat(apollo-server-express): add support for apollo-server-express

### LINKED ISSUE #641: instrumentation of apollo-server-*
When using `express-graphql` requests/transactions are grouped by query/mutation name, but this is not the case when using `apollo-server-express`, where transactions is simply grouped by the route.

It would be nice if transactions through `apollo-server-*` was grouped by query/mutation name.

I'm happy to help, but a little guidance on where to start would be nice.

danieljuhl:
Added instrumentation for apollo-server-express. As the actual instrumentation happens in apollo-server-core it might also work with all other apollo-server-*, but hasn't been tested yet.

Fixes #641 

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/modules/apollo-server-core.js
+++ b/lib/instrumentation/modules/apollo-server-core.js
@@ -1,0 +1,23 @@
+'use strict'
+
+const semver = require('semver')
+const shimmer = require('../shimmer')
+
+module.exports = function (apolloServerCore, agent, version, enabled) {
+  if (!enabled) return apolloServerCore
+
+  if (!semver.satisfies(version, '^2.0.0')) {
+    agent.logger.debug('apollo-server-core version %s not supported - aborting...', version)
+    return apolloServerCore
+  }
+
+  shimmer.wrap(apolloServerCore, 'runHttpQuery', function (runHttpQuery) {
+    return function wrappedRunHttpQuery () {
+      var trans = agent._instrumentation.currentTransaction
+      if (trans) trans._graphqlRoute = true
+      return runHttpQuery.apply(this, arguments)
+    }
+  })
+
+  return apolloServerCore
+}
--- a/lib/instrumentation/modules/apollo-server-core.js
+++ b/lib/instrumentation/modules/apollo-server-core.js
@@ -6,7 +6,7 @@ const shimmer = require('../shimmer')
 module.exports = function (apolloServerCore, agent, version, enabled) {
   if (!enabled) return apolloServerCore
 
-  if (!semver.satisfies(version, '^2.0.0')) {
+  if (!semver.satisfies(version, '^2.0.2')) {
     agent.logger.debug('apollo-server-core version %s not supported - aborting...', version)
     return apolloServerCore
   }
--- a/.tav.yml
+++ b/.tav.yml
@@ -59,7 +59,7 @@ ws:
   versions: '>=1 <7'
   commands: node test/instrumentation/modules/ws.js
 graphql:
-  preinstall: rm -fr node_modules/express-graphql
+  preinstall: npm uninstall express-graphql
   versions: '>=0.7.0 <0.11.0 || >=0.11.1 <15.0.0 || ^14.0.0-rc'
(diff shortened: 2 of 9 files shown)


danieljuhl:
@Qard @watson can any of you guide me to fixing/handling the failing test with node v4. apollo-server-express requires node >=6.

watson:
@danieljuhl looks like you found the answer your self by adding the `if (semver.lt(process.version, '6.0.0')) process.exit()` line and moving the require statement below that. Now there's just a few double quotes that's making our linter angry. When those are fixed I think the tests should pass fine 👍

danieljuhl:
@watson yes, I think I found the solution, and waiting for the tests to finally pass :)

danieljuhl:
@watson can you please guide me. Does it look like my code is breaking CI or is it related to the PR regarding express-graphql. At least it looks like the issue is with express-graphql - can you confirm?

watson:
Sorry about that. I just merged #646 which should fix the issue you experienced. If you rebase with master, it should be all green now 😄

watson:
I actually haven't reviewed the PR yet btw, so we'll just do that as well. So if you want you can just wait rebasing until afterwards

lib/instrumentation/modules/apollo-server-core.js
```diff
@@ -0,0 +1,23 @@
...
+module.exports = function (apolloServerCore, agent, version, enabled) {
+  if (!enabled) return apolloServerCore
+
+  if (!semver.satisfies(version, '^2.0.2') || typeof apolloServerCore.runHttpQuery !== 'function') {
+    agent.logger.debug('apollo-server-core version %s not supported - aborting...', version)
+    return apolloServerCore
+  }
+
+  var runHttpQuery = apolloServerCore.runHttpQuery
+  apolloServerCore.runHttpQuery = wrappedRunHttpQuery
```
Qard: This could also be written like this:

```js
shimmer.wrap(apolloServerCore, 'runHttpQuery', function (runHttpQuery) {
  return function wrappedRunHttpQuery () {
    var trans = agent._instrumentation.currentTransaction
    if (trans) trans._graphqlRoute = true
    return runHttpQuery.apply(this, arguments)
  }
})
```

Just add `const shimmer = require('../shimmer')` to the requires at the top.

The benefit is that shimmer provides the checking so you don't have to make sure `runHttpQuery` is a function before trying to patch it.

danieljuhl:
@watson @Qard it still looks like there is some conflicts with the dependency tests. Can you guide me, if there is something I can do.

danieljuhl:
> I've just tried to fix it. Let's see what the tests say...

Still not a happy CI. Maybe we need to remove `apollo-server-express` before `express-graphql`?

watson:
I wish npm would remove the concept of peer dependencies. It's nothing but trouble 😞 

I'm trying now to split the tests for `apollo-server-express` and `express-graphql` into two separate builds. That way they should not interfere with each other.

watson:
Didn't fix it... I'll have another look later

watson:
Finally got the tests to pass 😅 

@Qard have your review comments been addressed?

lib/instrumentation/modules/apollo-server-core.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
+
+const semver = require('semver')
+const shimmer = require('../shimmer')
+
+module.exports = function (apolloServerCore, agent, version, enabled) {
+  if (!enabled) return apolloServerCore
+
+  if (!semver.satisfies(version, '^2.0.0')) {
```
Qard: The docs and `.tav.yml` use the a `^2.0.2` range, while here it uses `^2.0.0`. This should be `^2.0.2` also, so we don't patch broken/unsupported versions accidentally.

Qard:
```suggestion
  if (!semver.satisfies(version, '^2.0.2')) {
```

PR #656, opened 2018-11-06
test(mysql2): un-pin mysql2 dev dependency

watson:
The bug in mysql2 v1.6.2 that caused us to pin it have now been resolved and released in v1.6.3.

Bug report:
https://github.com/sidorares/node-mysql2/issues/869

Pinned in:
https://github.com/elastic/apm-agent-nodejs/pull/652

PR #658, opened 2018-11-06
test(hapi): fix hapi tav tests for node 4

watson:


Qard:
Looks like there's syntax errors in mysql2 now too...

Qard:
Actually, I think this and #656 are just cross-dependent. Maybe put them together in one PR with multiple commits? 🤔

watson:
I've tested this PR together with #656 in #660 and the tests pass. I'll merge this one as is now

PR #661, opened 2018-11-06
feat(transaction): add ensureParentId

### LINKED ISSUE #650: Add transaction.ensureParentId() function
Add the function `ensureParentId()` to the Transaction API.

If called, the following should happen:

- If the parent ID of the transaction is empty, the agent generates a new random ID, sets that as the parent ID of the transaction and returns the value as a hex-encoded string
- if the parent ID of the transaction is not empty, returns the parent ID as a hex-encoded string

This will allow users to seed a RUM agent with a new span ID upon its initial request when requesting the HTML page. Users can use this within their favorite template language in this way:

```js
var apm = initApm({
  pageLoadTraceId: "${transaction.traceId}",
  pageLoadSpanId: "${transaction.ensureParentId()}",
  pageLoadSampled: ${transaction.sampled}
})
```

Qard:
Fixes #650

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/transaction-api.asciidoc
+++ b/docs/transaction-api.asciidoc
@@ -49,6 +49,35 @@ The transaction result.
 
 Think of the transaction result as equivalent to the status code of an HTTP response.
 
+[[transaction-ensure-parent-id]]
+==== `transaction.ensureParentId()`
+
+* *Type:* String
+
+If the transaction does not already have a parent id,
+calling this method generates a new parent id,
+sets it as the parent id of this transaction,
+and returns it as a `String`.
+
+This enables the correlation of the spans the JavaScript Real User Monitoring (RUM) agent creates for the initial page load with the transaction of the backend service.
+If your backend service generates the HTML page dynamically,
+initializing the JavaScript RUM agent with the value of this method allows analyzing the time spent in the browser vs in the backend services.
+
+To enable the JavaScript RUM agent,
+add a snippet similar to this to the body of your HTML page,
+preferably before other JavaScript libraries:
+
+[source,js]
+----
+var apm = initApm({
+  pageLoadTraceId: "${transaction.traceId}",
+  pageLoadSpanId: "${transaction.ensureParentId()}",
+  pageLoadSampled: ${transaction.sampled}
+})
+----
+
+See the {apm-rum-ref}[JavaScript RUM agent documentation] for more information.
+
 [[transaction-end]]
 ==== `transaction.end([result])`
 
--- a/lib/instrumentation/trace-context.js
+++ b/lib/instrumentation/trace-context.js
@@ -34,10 +34,13 @@ function defineLazyProp (obj, prop, fn) {
     enumerable: true,
     get () {
       const value = fn()
-      Object.defineProperty(obj, prop, {
-        enumerable: true,
-        value
-      })
+      if (value !== undefined) {
+        Object.defineProperty(obj, prop, {
+          configurable: true,
+          enumerable: true,
+          value
+        })
+      }
       return value
     }
   })
@@ -131,6 +134,15 @@ class TraceContext {
 
   static fromString (header) {
     return new TraceContext(headerToBuffer(header))
+  }
+
+  ensureParentId () {
+    let id = this.parentId
+    if (!id) {
+      crypto.randomFillSync(this[bufferSymbol], OFFSETS.parentId, SIZES.id)
+      id = this.parentId
+    }
+    return id
   }
 
   child () {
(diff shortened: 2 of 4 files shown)


watson:
Jenkins, retest this please

lib/instrumentation/trace-context.js
```diff
@@ -34,10 +34,13 @@ function defineLazyProp (obj, prop, fn) {
     enumerable: true,
     get () {
       const value = fn()
-      Object.defineProperty(obj, prop, {
-        enumerable: true,
-        value
-      })
+      if (typeof value !== 'undefined') {
```
watson: We don't use `undefined` overwrite protection anywhere else. I don't think this is something to worry much about anymore. For now let's just align with the rest of the application, unless you have a specific concern 😃 

```suggestion
      if (value !== undefined) {
```

test/instrumentation/trace-context.js
```diff
@@ -130,3 +130,23 @@ test('child', t => {
...
+  isValid(t, context)
+  t.equal(context.version, version, 'version matches')
+  t.equal(context.traceId, traceId, 'traceId matches')
+  t.equal(context.id, id, 'id matches')
+  t.equal(context.flags, flags, 'flags matches')
+  t.notOk(context.parentId, 'no parent id before')
+
+  const first = context.ensureParentId()
+  t.ok(first, 'returns parent id')
+  t.equal(context.parentId, first, 'contenxt parent id matches returned parent id')
```
watson:
```suggestion
  t.equal(context.parentId, first, 'context parent id matches returned parent id')
```

watson:
The tests are failing I think because we need to rebase `api-v2` with master again. I didn't do it because we had a few open PR's and I didn't want to force you to rebase them. But now that the tests are failing maybe we should go ahead and do it anyway

lib/instrumentation/trace-context.js
Qard: No reason. That's just how I've always written undefined checks. 🤷‍♂️

watson:
> I kind of like how the Java agent [explains in a little more detail](https://www.elastic.co/guide/en/apm/agent/java/master/public-api.html#api-ensure-parent-id) why you'd use this new function. Did you consider stealing some of that doc?

What do you think about this?

Qard:
I updated the docs, what do you think?

docs/transaction-api.asciidoc
```diff
@@ -49,6 +49,36 @@ The transaction result.
 
 Think of the transaction result as equivalent to the status code of an HTTP response.
 
+[[transaction-ensure-parent-id]]
+==== `transaction.ensureParentId()`
+
+* *Type:* String
+
+If the transaction does not already have a parent id,
+calling this methid generates a new parent id,
```
watson:
```suggestion
calling this method generates a new parent id,
```

docs/transaction-api.asciidoc
```diff
@@ -49,6 +49,36 @@ The transaction result.
...
+sets it as the parent id of this transaction,
+and returns it as a `String`.
+
+This enables the correlation of the spans the JavaScript Real User Monitoring (RUM) agent creates for the initial page load
+with the transaction of the backend service.
```
watson: Tiny nit: Unnecessary line break

docs/transaction-api.asciidoc
```diff
@@ -49,6 +49,36 @@ The transaction result.
...
+If your backend service generates the HTML page dynamically,
+initializing the JavaScript RUM agent with the value of this method allows analyzing the time spent in the browser vs in the backend services.
+
+To enable the JavaScript RUM agent,
+add a snippet similar to this to the body of your HTML page,
+preferably before other JS libraries:
```
watson: Not sure if we want to abbreviate JavaScript as JS?

docs/transaction-api.asciidoc
```diff
@@ -49,6 +49,36 @@ The transaction result.
...
+[source,js]
+----
+var apm = initApm({
+  pageLoadTraceId: "${transaction.traceId}",
+  pageLoadSpanId: "${transaction.ensureParentId()}",
+  pageLoadSampled: ${transaction.sampled}
+})
+----
+
+See the {apm-rum-ref}[JavaScript RUM agent documentation] for more information.
```
watson: At some point we might want to update this link to point to https://www.elastic.co/guide/en/apm/agent/js-base/master/distributed-tracing-guide.html - but that page hasn't yet been published outside of master, so I don't think we can link to it yet without the doc build complaining

docs/transaction-api.asciidoc

docs/transaction-api.asciidoc

PR #668, opened 2018-11-10
docs: bump minimum supported version of APM Server

watson:
Depends on ~~elastic/docs#457~~

### DIFF
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -28,7 +28,9 @@ that not all core Node.js API's can be instrumented without the use of Async Hoo
 [[elastic-stack-compatibility]]
 === Elastic Stack Compatibility
 
-This agent is compatible with {apm-server-ref}[APM Server] v6.2 and above.
+This agent is compatible with {apm-server-ref}[APM Server] v6.5 and above.
+For support for previous releases of the APM Server,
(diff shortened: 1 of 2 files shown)


docs/supported-technologies.asciidoc
```diff
@@ -28,7 +28,9 @@ that not all core Node.js API's can be instrumented without the use of Async Hoo
 [[elastic-stack-compatibility]]
 === Elastic Stack Compatibility
 
-This agent is compatible with {apm-server-ref}[APM Server] v6.2 and above.
+This agent is compatible with {apm-server-ref}[APM Server] v6.5 and above.
+For support for previous releases of the APM Server,
+use version {apm-node-ref-1x}[1.x] of the agent.
```
watson: I'm surprised this didn't break the docs build 🤔

bmorelli25: Are you surprised because it didn't break locally?

watson: I would have thought it would check if `apm-node-ref-1x` was a valid attribute when doing the docs build

bmorelli25: Link checking only happens when you build _all_ of the documentation. Which is what I learned when I broke the docs for six hours yesterday :(

PR #678, opened 2018-11-13
test: fix possible race condition

watson:
Sometimes the error would arrive before the span and the tests would fail.

Since we don't actually care about the arrival of the span or the transaction, we can just not end those and the test will not have a race condition anymore.

### DIFF
--- a/test/agent.js
+++ b/test/agent.js
@@ -861,8 +861,6 @@ test('#captureError()', function (t) {
     let span = null
     const expect = [
       'metadata',
-      'transaction',
-      'span',
       'error'
     ]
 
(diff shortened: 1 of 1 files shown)


test/agent.js
```diff
@@ -873,8 +871,6 @@ test('#captureError()', function (t) {
         this.agent.captureError(new Error('with callback'), function () {
           t.pass('called callback')
         })
-        span.end()
-        trans.end()
```
Qard: Maybe put this in a `t.on('end', () => { ... })` just to ensure they are still ended before the test exits.

watson: What's your concern about it? We start transactions/spans in other tests that we never end either, so if there's a problem with this, we should probably look at this in general.

Qard: Mainly just avoiding transactions leaking into other tests and causing strange behaviour. I suppose I'm fine landing this as-is though.

PR #680, opened 2018-11-13
docs: update transaction.ensureParentId() code example

### LINKED ISSUE #677: Update the RUM configuration in docs for correlating backend page load transaction
The RUM configuration should be:

```
elasticApm.init({
    serviceName: <service-name>,
    pageLoadTraceId: <trace-id>, // String
    pageLoadSpanId: <span-id>, // String
    pageLoadSampled: <is-sampled> // Boolean
})
```

watson:
Closes #677

### DIFF
--- a/docs/transaction-api.asciidoc
+++ b/docs/transaction-api.asciidoc
@@ -94,10 +94,11 @@ preferably before other JavaScript libraries:
 [source,js]
 ----
 elasticApm.init({
-  serviceName: <service-name>,
-  pageLoadTraceId: <trace-id>, // String
-  pageLoadSpanId: <span-id>, // String
-  pageLoadSampled: <is-sampled> // Boolean
+  serviceName: 'my-frontend-app', // Name of your frontend app
+  serverUrl: 'https://example.com:8200', // APM Server host
+  pageLoadTraceId: '${transaction.traceId}',
+  pageLoadSpanId: '${transaction.ensureParentId()}',
+  pageLoadSampled: ${transaction.sampled}
 })
 ----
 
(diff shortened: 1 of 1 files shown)


docs/transaction-api.asciidoc
```diff
@@ -93,10 +93,11 @@ preferably before other JavaScript libraries:
...
 ----
-var apm = initApm({
-  pageLoadTraceId: "${transaction.traceId}",
-  pageLoadSpanId: "${transaction.ensureParentId()}",
-  pageLoadSampled: ${transaction.sampled}
+elasticApm.init({
+  serviceName: <service-name>,
+  pageLoadTraceId: <trace-id>, // String
+  pageLoadSpanId: <span-id>, // String
+  pageLoadSampled: <is-sampled> // Boolean
```
Qard: Shouldn't these `<...>` bits be replaced with the actual code from before? In this form, it doesn't actually show where `ensureParentId()` should be used, which is kind of the point of this code sample. 😅

watson: @Qard Hmm good point

@jahtalab Was this what you had in mind with #677?

jahtalab: @watson , The parts should be replaced with the agent specific API calls to get the required data, those are just place holders. Might be useful to provide those calls in the context of a templating engine see [this PR from python agent]: https://github.com/elastic/apm-agent-python/pull/322/files#diff-845588bd4e49dbf10d88a73c846b2b5fR146

watson: Ah ok, that wasn't clear from your issue. There's unfortunately no clear winner in regards to templating engines in the JavaScript world, so I don't know which one to highlight in the example. So I think I'll just write it in plain JavaScript just as in the previous example - unless you have any recommendations

watson: I've now updated the example. LMKWYT 😃

PR #687, opened 2018-11-14
docs: fix APM Server links to 6.4

watson:
These changes are not critical as none of the links are technically broken, but it would just be nicer if the links to the APM Server for version 1.x of the agent pointed to version 6.4 of the APM Server.

@Qard This is the first time we do a PR into an old branch, but I guess this is how we would do it right?

### DIFF
--- a/docs/compatibility.asciidoc
+++ b/docs/compatibility.asciidoc
@@ -28,7 +28,9 @@ that not all core Node.js API's can be instrumented without the use of Async Hoo
 [[elastic-stack-compatibility]]
 === Elastic Stack Compatibility
 
-This agent is compatible with {apm-server-ref}[APM Server] v6.2 and above.
+Version 1.x of the agent is compatible with {apm-server-ref-64}[APM Server] v6.2 to v6.4.
+For support for APM Server v6.5 and above,
+use version 2.0.0 or higher of the Node.js agent.
 
 [float]
 [[compatibility-frameworks]]
--- a/docs/compatibility.asciidoc
+++ b/docs/compatibility.asciidoc
@@ -30,7 +30,7 @@ that not all core Node.js API's can be instrumented without the use of Async Hoo
 
 Version 1.x of the agent is compatible with {apm-server-ref-64}[APM Server] v6.2 to v6.4.
 For support for APM Server v6.5 and above,
-use version 2.0.0 or higher of the Node.js agent.
+use {apm-node-ref-index}[version 2.0.0 or higher] of the Node.js agent.
 
 [float]
 [[compatibility-frameworks]]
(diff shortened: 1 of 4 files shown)


docs/compatibility.asciidoc
```diff
@@ -28,7 +28,9 @@ that not all core Node.js API's can be instrumented without the use of Async Hoo
 [[elastic-stack-compatibility]]
 === Elastic Stack Compatibility
 
-This agent is compatible with {apm-server-ref}[APM Server] v6.2 and above.
+Version 1.x of the agent is compatible with {apm-server-ref-64}[APM Server] v6.2 to v6.4.
+For support for APM Server v6.5 and above,
+use version 2.0.0 or higher of the Node.js agent.
```
watson: @bmorelli25 it might be a good idea to link to the 2.x version of this page. What's the best way to do that? (keep in mind this page have been renamed in 2.x and is now called "supported technologies".

bmorelli25: One way is to create another shared attribute for `:apm-node-ref-2x`: and link to `{apm-node-ref-2x}/supported-technologies.html`. I think a better way would be to use `{branch}`, but I'm still working through the details of how to implement that with Lisa.

watson: New idea:

How about that I create a link like this:

```suggestion
use {apm-node-ref-index}[version 2.0.0 or higher] of the Node.js agent.
```

Where `apm-node-ref-index` is defined as this:

```
:apm-node-ref-index:   https://www.elastic.co/guide/en/apm/agent/nodejs
```

That will take users to the version overview page and they can choose the version they want

bmorelli25: That's super cool. I like that idea. Are you thinking you'll define `apm-node-ref-index` in the `shared/attributes` file (docs repo), or in the `index.asciidoc` file within the node.js repo?

watson: I was thinking to define it in the shared attributes file next to the other `apm-node-ref*` attributes. Is that ok?

bmorelli25: I was thinking since it most likely won't be used outside of the node documentation it might make more sense to just define it locally. I have a call with Lisa today - I'll discuss this with her and get back to you!

watson: Ah ok, let me know what you think is best then after the call 👍

bmorelli25: `:apm-node-ref-index:` in `shared/attributes` is a 👍 
In addtion, elastic/apm#19 will resolve the other problems we're seeing in this PR.

watson: PR created: https://github.com/elastic/docs/pull/479

PR #691, opened 2018-11-24
feat: add support for OpenTracing

### LINKED ISSUE #431: Create OpenTracing API compatibility bridge
Implement lightweight OpenTracing API compatibility bridge.

watson:
~~Depends on #757~~

This is a work-in-progress PR for implementing OpenTracing support to the Elastic APM Node.js Agent.

If you check out the branch and take it for a spin, your feedback would be highly appreciated 😃 

Closes #431

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


lib/instrumentation/transaction.js
```diff
@@ -12,8 +12,11 @@ var Timer = require('./timer')
 
 module.exports = Transaction
 
-function Transaction (agent, name, type, traceparent) {
-  this.context = TraceContext.startOrResume(traceparent, agent._conf)
+function Transaction (agent, name, type, traceContext) {
+  this.context = traceContext instanceof TraceContext
+    ? traceContext.child()
+    : TraceContext.startOrResume(traceContext, agent._conf) // string | null expected
```
Qard: Maybe just make `TraceContext.startOrResume(...)` accept another `TraceContext` instance to call `traceContext.child()` on? 🤔

lib/instrumentation/trace-context.js
```diff
@@ -102,18 +102,17 @@ function start (sampled = false) {
   return new TraceContext(buffer)
 }
 
-const bufferSymbol = Symbol('trace-context-buffer')
-
 class TraceContext {
   constructor (buffer) {
-    this[bufferSymbol] = buffer
+    this.buffer = buffer
```
Qard: If you need to get at the buffer, should we maybe do a getter instead which just returns a copy of the buffer? That way we can continue to prevent tampering.

lib/instrumentation/span.js
```diff
@@ -11,15 +11,20 @@ const TEST = process.env.ELASTIC_APM_TEST
...
 
-  var current = this._agent._instrumentation.activeSpan || transaction
-  this.context = current.context.child()
+  // TODO: Should it be possible to give a traceparent string here? We probably don't need it though
+  if (traceContext) {
+    this.context = traceContext.child()
+  } else {
+    var current = this._agent._instrumentation.activeSpan || transaction
+    this.context = current.context.child()
+  }
```
Qard: Maybe make spans _always_ expect a `TraceContext` instance to descend from, and move the `activeSpan` tracking stuff into the transaction instead?

package.json
```diff
@@ -78,6 +78,7 @@
     "fast-safe-stringify": "^2.0.6",
     "http-headers": "^3.0.2",
     "is-native": "^1.0.1",
+    "opentracing": "^0.14.3",
```
watson: @Qard I'm not a big fan of having a dependency to opentracing here as most people probably don't need it. Alternatively the entire opentracing logic could be moved to a separate module - or something else. What do you think?

Qard: Depends how much opentracing adds to the tree. If it's a minimal delta, it's probably fine to just have it as a dependency. If you feel we'd be better off maintaining it as a separate module, I'd be fine with that too though.

watson: The `opentracing` package doesn't it self have any dependencies. Here's [its stats](https://bundlephobia.com/result?p=opentracing@0.14.3):

image

So I guess it's ok for us to add it

opentracing/index.js
```diff
@@ -0,0 +1,164 @@
...
+    const span = this._agent.currentTransaction === null || this._agent.currentTransaction.ended // TODO: Is this right?
+      ? this._agent.startTransaction(name, null, context)
+      : this._agent.startSpan(name, null, context)
+
+    return span ? new Span(this, span) : Noop.span // TODO: What happens when a user tries to use the context of a Noop span? Maybe the Noop span context should be the context of the transaction?
+  }
+
+  // TODO: Align with other agents about TEXT/BINARY format
+  // TODO: Maybe wait implementing the binary format until: https://github.com/w3c/trace-context/tree/master/trace_context#binary-format
+  // - On the other hand, as we're not putting this is a special location, we might not have to wait on Trace Context - similar to how we don't wait for the HTTP definition when using a custom header prefix
```
axw: we should at least discuss a format within the team, so that it'll work with the other agents

watson:
Rebased against #757 and force-pushed

opentracing/index.js
```diff
@@ -0,0 +1,165 @@
...
+      this._agent.setUserContext({ id, username, email })
+      // TODO: Should these be deleted?
+      delete tags['user.id']
+      delete tags['user.username']
+      delete tags['user.email']
+    }
+
+    if (Object.keys(tags).length === 0) return
+
+    if (!this._isTransaction) return this._agent.logger.info('Elastic APM OpenTracing: Could not set tags on span') // The Elastic APM agent currently doesn't support span tags
```
vigneshshanmugam: nit: this.__tracer.agent.logger

opentracing/index.js
```diff
@@ -0,0 +1,165 @@
...
+        carrier.buffer = spanContext._elasticContext.buffer
+        break
+    }
+  }
+
+  _extract (format, carrier) {
+    let ctx
+
+    switch (format) {
+      case Tracer.FORMAT_TEXT_MAP:
```
vigneshshanmugam: all the OT constants seems to be exported from wrong place, Shouldn't this be `ot.FORMAT_TEXT_MAP` instead of `Tracer.FORMAT_TEXT_MAP`

vigneshshanmugam:
@watson I tried out this branch on one of the Node.js library and got it working with couple of minor changes(changes that i have commented here + some null handling in context and carrier extraction). The spans created through opentracing API are somehow missing the parent_id of the transaction and I couldn't work around it for now. 

Would be happy to discuss once you are available. Here is the repo - https://github.com/vigneshshanmugam/tailor-apm-demo/

I will try to debug and post details if I am able to find the issue.

opentracing/index.js
```diff
@@ -0,0 +1,165 @@
...
+    this.__tracer = tracer
+    this.__context = null
+    this._span = span
+    this._isTransaction = span instanceof Transaction
+  }
+
+  _context () {
+    return this.__context
+      ? this.__context
+      : (this.__context = new SpanContext(this._span.context))
```
vigneshshanmugam: this was really hard to find while doing the test.. should be `this._span._context`

opentracing/index.js
```diff
@@ -0,0 +1,165 @@
...
+
+    if (Object.keys(tags).length === 0) return
+
+    if (!this._isTransaction) return this._agent.logger.info('Elastic APM OpenTracing: Could not set tags on span') // The Elastic APM agent currently doesn't support span tags
+    this._span.addTags(tags)
+  }
+
+  _log (logs, timestamp) {
+    if (logs.event === 'error') {
+      if (logs.error && logs.error.object) {
```
jahtalab: `error.object` seems to be a string constant not an object hierarchy, see the example in https://github.com/opentracing/opentracing-javascript#code-snippet

opentracing/index.js
watson: We decided not to support the binary format until it's agreed upon in the TraceContext spec

opentracing/index.js
watson: You're right 👍

opentracing/index.js
watson: We since added support for tags on spans, so I've removed this line entirely 😄

opentracing/index.js
watson: Good catch!

opentracing/index.js
watson: Oh, you're right 🤦‍♂️

opentracing/index.js
```diff
@@ -0,0 +1,165 @@
...
+        const type = ref.type()
+        if (!context && type === Tracer.REFERENCE_FOLLOWS_FROM) {
+          context = ref
+        } else if (type === Tracer.REFERENCE_CHILD_OF) {
+          context = ref
+          return true
+        }
+      })
+
+      context = context.referencedContext()._elasticContext
```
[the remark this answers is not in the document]
watson: I don't think I understand the issue?

lib/instrumentation/trace-context.js
watson: Rolled it back

lib/instrumentation/trace-context.js
```diff
@@ -107,6 +106,7 @@ const bufferSymbol = Symbol('trace-context-buffer')
 class TraceContext {
   constructor (buffer) {
     this[bufferSymbol] = buffer
+
```
Qard: Unnecessary diff in this file.

lib/instrumentation/transaction.js
watson: Created separate PR: #826

lib/instrumentation/span.js
watson: Not part of this PR anymore. Let's create a separate PR for that if we want to refactor it though.

opentracing/index.js
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  _extract (format, carrier) {
+    let ctx
+
+    switch (format) {
+      case opentracing.FORMAT_TEXT_MAP:
+      case opentracing.FORMAT_HTTP_HEADERS:
+        ctx = ElasticTraceContext.fromString(carrier['elastic-apm-traceparent'])
```
vigneshshanmugam: throws when the elastic-apm-traceparent header is undefined. 

```js
Elastic APM caught unhandled exception: Cannot read property 'replace' of undefined
TypeError: Cannot read property 'replace' of undefined
   at headerToBuffer (/Users/vigneshh/opensource/tailor-apm-demo/node_modules/elastic-apm-node/lib/instrumentation/trace-context.js:83:23)
    at Function.fromString (/Users/vigneshh/opensource/tailor-apm-demo/node_modules/elastic-apm-node/lib/instrumentation/trace-context.js:136:29)
    at Tracer._extract (/Users/vigneshh/opensource/tailor-apm-demo/node_modules/elastic-apm-node/opentracing/index.js:75:35)
    at Tracer.extract (/Users/vigneshh/opensource/tailor-apm-demo/node_modules/opentracing/lib/tracer.js:122:21)
```

 The usage is done like this 

https://github.com/zalando/tailor/blob/52812a3ebaff911524d683377872e23389bb9a69/lib/request-handler.js#L38-L41

watson:
Moved development of the OpenTracing bridge to a separate repo: https://github.com/elastic/apm-agent-nodejs-opentracing

opentracing/index.js
watson: Good catch 😅 

We just moved the opentracing bridge to it's own module: https://github.com/elastic/apm-agent-nodejs-opentracing

I'll fix it there 👍

PR #692, opened 2018-11-26
fix: log APM Server API errors correctly

watson:
APM Server API errors are emitted by the HTTP client as 'request-error' events.

In case of more serious errors, the 'error' event is still used, but indicates that the client could not recover from the error.

The `request-error` event is emitted here: https://github.com/elastic/apm-nodejs-http-client/blob/5d364e1fe9150bde74b6fde8732187f1e2c00e48/index.js#L59

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -195,6 +195,10 @@ function config (opts) {
       })
 
       transport.on('error', err => {
+        agent.logger.error('APM Server transport error:', err.stack)
+      })
+
+      transport.on('request-error', err => {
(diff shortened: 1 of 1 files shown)


watson:
@Qard merging this without a review as it - amongst other things - fixes a pretty annoying bug affecting users who accidentally run the newest agent against an older version of the APM Server (which is easy to accidentally do if you don't pay attention the changelog or upgrade guide)

Qard:
LGTM anyway. 👍

PR #694, opened 2018-11-26
test(finalhandler): add finalhandler to tav test-matrix

watson:
We forgot to do this in #629


watson:
Merging without review as it's just test-related

PR #699, opened 2018-11-28
docs: add redirects file

bmorelli25:
Add hidden `redirects.asciidoc` file to handle "redirect" on current (as we're not supposed to redirect current). Should also backport to 2.x. Will set up website redirect for 2.x as well.


watson:
Will this not show up in the ToC?

bmorelli25:
Correct. It's a hidden file that is excluded from the TOC. Can only be accessed directly

bmorelli25:
Per zoom with Watson - edited comments above to better explain what this file does. Also changed the redirect request elastic/website-www.elastic.co/issues/3284 from `2.x` to `1.x`

Issue #701, opened 2018-12-03
crypto.randomFillSync fails on Node.js 6.12.3 and below


watson:
Calls to `crypto.randomFillSync` was added in Node.js v6.13.0, so we have to add a polyfill if users use Node.js 6.12.3 or below.

PR #705, opened 2018-12-03
test(ci): add early supported node versions to test matrix

watson:
We [had an issue](https://github.com/elastic/apm-agent-nodejs/issues/701) where an early version of Node.js v6 didn't have support for a core API we were using. This wasn't caught by our tests because our tests ran on the latest 6.x version.

By adding the first version in each major version range we support to the test matrix, we enhance our ability to catch these issues in the future.


watson:
@Qard As we discussed yesterday, do you want me to remove `11.0.0`, `10.0.0` and `8.0.0` so we only run these tests on `6.0.0`?

Qard:
I'm good either way.

watson:
I should have checked before this last commit - there's no patch releases on the first release of each major version. So this doesn't change a thing 😂

watson:
@Qard the bug I fixed in `test/instrumentation/modules/restify.js` was only in early versions of Node 8. Though this wasn't an issue as it was in the tests, it could have been inside the core code. So running these tests on early intermediate versions does have a little bit of value. I vote for leaving them in for now.

If we later feel that it's too much, we can revisit it and for instance change them to run as a cron job.

What do you think about that?

Issue #706, opened 2018-12-04
APM agent (and so app) dies when using bodyParser in restify application


waswrongassembled:
**Describe the bug**
When using ES APM Agent in an restify app and the bodyParser Plugin at the same time, the app dies with the first request/an exception is thrown by APM which cannot be handled. APM agent writes his own error ```wrappedHandler (node_modules/elastic-apm-node/lib/instrumentation/modules/restify.js)``` but afterwards the request dies.

**To Reproduce**
1. Create a npm project (elastic-apm-node version doesn't matter - bug exists in 1.x and 2.x)
```
{
  "name": "apm-test",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "elastic-apm-node": "^1.14.3",
    "restify": "^7.2.3"
  }
}
``` 
2. Create a simple server 
```
// // Add this to the VERY top of the first file loaded in your app
var apm = require('elastic-apm-node').start({
    serviceName: 'development_apm-test',
  
    // Set custom APM Server URL (default: http://localhost:8200)
    serverUrl: 'http://XXXX:8200',
  })

var restify = require('restify');

function respond(req, res, next) {
  res.send('hello ' + req.params.name);
  next();
}

var server = restify.createServer();

server.use(restify.plugins.jsonBodyParser());

server.get('/hello/:name', respond);
server.head('/hello/:name', respond);

server.listen(8080, function() {
  console.log('%s listening at %s', server.name, server.url);
});
```

2. Call the app's endpoint in the browser..
```
http://localhost:8080/hello/waldi
```
3. See in the console ```logging error XXXX-XXX-XXX with Elastic APM``` and the request is dead.
5. Output in Kibana/ES:
```
{
  "library_frame": true,
  "exclude_from_grouping": false,
  "filename": "node_modules/elastic-apm-node/lib/instrumentation/modules/restify.js",
  "abs_path": "~/apm-test/node_modules/elastic-apm-node/lib/instrumentation/modules/restify.js",
  "line": {
    "number": 37,
    "context": "        return handler.apply(this, arguments)"
  },
  "function": "wrappedHandler",
  "context": {
    "pre": [
      "          }",
      "        })"
    ],
    "post": [
      "      }",
      "    }"
    ]
  }
},
{
  "exclude_from_grouping": false,
[...]

PR #709, opened 2018-12-04
fix(restify): support an array of handlers

watson:
If an array of handlers is given as an argument to `server.use` instead of just a single function, the agent would cause an exception to be raised.

This commit changes the Restify instrumentation logic so that the handlers no longer needs to be wrapped, and so that the arguments to the `server.use` function no longer needs to be parsed by the agent.

Instead the agent now patches a central (private) error handling function to be notified when ever an error is passed to the `next` function.

The route name is instead gathered by the same logic that also gathers the route name for Express - from `req.route.path` - as Restify just so happens to use the same API for storing the route name on the request object as Express in this case.

This way of getting the route name is actually more stable as the previous implementation didn't work if the request never made it to the actual route handler (e.g. if an error occurred in a middleware handler). In that case the transaction would have been named `use` (the function name used to register the middleware handler).

Fixes: #706

BREAKING CHANGE: Restify <5.2.0 is no longer supported

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -211,13 +211,7 @@ cassandra-driver:
     - node test/instrumentation/modules/cassandra/cassandra.js
 
 restify:
-  # All versions before 4.4 are unsupported. Additionally,
-  # there are several broken releases that need to be skipped:
-  # - 6.2.0
-  # - 6.0.1
-  # - 6.0.0
-  # - 5.0.0
-  versions: '>=4.4 <5 || >5 <6 || >6.0.1 <6.2 || >6.2'
+  versions: '>=5.2.0'
   commands:
     - node test/instrumentation/modules/restify.js
 
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -49,7 +49,7 @@ These are the frameworks that we officially support:
 |<<koa,Koa>> via koa-router |>=5.2.0 <8.0.0 |Koa doesn't have a built in router,
 so we can't support Koa directly since we rely on router information for full support.
 We currently support the most popular Koa router called https://github.com/alexmingoia/koa-router[koa-router]
-|<<restify,Restify>> |>=4.4 <8 |
+|<<restify,Restify>> |>=5.2.0 |
 |=======================================================================
 
 [float]
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -2,6 +2,8 @@
 
 var symbols = require('../symbols')
 
+// This function is also able to extract the path from a Restify request as
+// it's storing the route name on req.route.path as well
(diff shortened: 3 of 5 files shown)


.tav.yml
```diff
@@ -211,13 +211,7 @@ cassandra-driver:
     - node test/instrumentation/modules/cassandra/cassandra.js
 
 restify:
-  # All versions before 4.4 are unsupported. Additionally,
-  # there are several broken releases that need to be skipped:
-  # - 6.2.0
-  # - 6.0.1
-  # - 6.0.0
-  # - 5.0.0
-  versions: '>=4.4 <5 || >5 <6 || >6.0.1 <6.2 || >6.2'
```
watson: FYI: During research on this bug I also found that this version string does't produce the expected range:

```js
semver.validRange('>=4.4 <5 || >5 <6 || >6.0.1 <6.2 || >6.2')
// '>=4.4.0 <5.0.0||>=6.0.0 <6.0.0||>6.0.1 <6.2.0||>=6.3.0'
```

As you can see, `>5` is not translated into `>5.0.0` but into `>=6.0.0`. So our TAV tests have been ignoring a big part of the version range when testing.

watson:
We need to backport this to 1.x as well in which case it can't really be a breaking change anyway. So we have to decide if we can drop support for early versions of Restify 5.x and not consider it a breaking change, or if we need to find another solution that's fixing this bug without loosing support for early versions.

Btw, though the docs say we currently support version 4.4+, this is misleading. The highest 4.x version is 4.3.4, so there is no such thing as 4.4. In reality this means that we currently support 5.0.0 and above.

.tav.yml
watson: Hmm this might actually explain why I can't get the TAV tests to run now. They refuse to install v5.2.0 due to a weird bug. This might have been a problem all along and we just never knew because we didn't run the tests with 5.x

Qard: Weird. I could have sworn it was testing 5.x properly. 🤔

Qard:
I think it's reasonable to consider it a bug fix. It never _truly_ supported the versions being dropped anyway, due to the bug, and it's an older version too, so _probably_ okay. 🤔

watson:
True, and you could also choose to see it this way: It's not a change that will break anything in running apps. It might collect less data, but the app will continue to run as normal. So in that sense it's not a breaking change.

.tav.yml
watson: I found the bug: restify@6.2.0 had a (sub)dependency to npm@2 which then took over when tav tried to continue to test v6.0.1. That's why 6.2.0, 6.0.1, and 6.0.0 failed. I updated test-all-versions to not fail even if one of the modules it installs depends on npm. So now all Restify tests pass without us having to skip any tests.

watson:
@Qard now all tests should be fixed. Could I get you to do another review?

PR #727, opened 2018-12-07
chore: update engines field in package.json

watson:
It wrongly stated support for Node.js 4 which was dropped in v2.0.0 and it also didn't include version 11.


watson:
Gonna merge this without a review so we can release 2.0.4 with this fix

Issue #422, opened 2018-06-28
Add support for Metrics / Stats API


alvarolobato:
Work is underway to add support for Metrics to the API/server: https://github.com/elastic/apm-server/issues/964
Support needs to be added to this agent too.
The default minimum set of metrics should be supported. TBD which additional metrics will be supported for this agent.

PR #731, opened 2018-12-10
feat(metrics): added basic metrics gathering

Qard:
This adds a metrics gatherer based on the `measured` module family.

Depends on #769
Fixes elastic/apm-agent-nodejs#422

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


lib/agent.js
```diff
@@ -324,6 +327,7 @@ Agent.prototype.captureError = function (err, opts, cb) {
 
     if (agent._transport) {
       agent.logger.info(`Sending error to Elastic APM`, { id: error.id })
+      agent._metrics.getOrCreateCounter('agent.errors.sent').inc()
```
watson: We decided to delay the introduction of `agent.*` metrics, so this line should just be removed

lib/instrumentation/index.js
```diff
@@ -110,6 +110,7 @@ Instrumentation.prototype.addEndedTransaction = function (transaction) {
     var payload = agent._transactionFilters.process(transaction._encode())
     if (!payload) return agent.logger.debug('transaction ignored by filter %o', { trans: transaction.id, trace: transaction.traceId })
     agent.logger.debug('sending transaction %o', { trans: transaction.id, trace: transaction.traceId })
+    agent._metrics.getOrCreateCounter('agent.transactions.sent').inc()
```
watson: We decided to delay the introduction of `agent.*` metrics, so this line should just be removed

lib/instrumentation/index.js
```diff
@@ -135,7 +136,10 @@ Instrumentation.prototype.addEndedSpan = function (span) {
       }
 
       agent.logger.debug('sending span %o', { span: span.id, parent: span.parentId, trace: span.traceId, name: span.name, type: span.type })
-      if (agent._transport) agent._transport.sendSpan(payload)
+      if (agent._transport) {
+        agent._metrics.getOrCreateCounter('agent.spans.sent').inc()
```
watson: We decided to delay the introduction of `agent.*` metrics, so this can just be reverted

lib/instrumentation/transaction.js
```diff
@@ -143,6 +143,7 @@ Transaction.prototype.startSpan = function (name, type) {
     return null
   }
   if (this._builtSpans >= this._agent._conf.transactionMaxSpans) {
+    this._agent._metrics.getOrCreateCounter('agent.spans.dropped').inc()
```
watson: We decided to delay the introduction of `agent.*` metrics, so this line should just be removed

lib/metrics/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+const MetricsRegistry = require('./registry')
+
+const registrySymbol = Symbol('metrics-registry')
+const agentSymbol = Symbol('metrics-agent')
+
+class Metrics {
+  constructor (agent) {
+    this[agentSymbol] = agent
+    this[registrySymbol] = undefined
```
watson: I normally use `null` instead of `undefined` in these situations, but that's based on the idea that it's better for the V8 optimizer. Not sure if that's true any more?

test/metrics/index.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    for (const name of Object.keys(metrics)) {
+      const metric = metricset.samples[name]
+      t.comment(name)
+      t.ok(metric, `should be present`)
+      metrics[name](metric.value)
+    }
+
+    t.end()
```
watson: Instead of ending after the first interval, it might be a good test to see that it's called twice, the 2nd time roughly one second after the first.

Qard: fixed

watson:
I think if this is rebased with master the tests will pass

watson:
GitHub really needs a "rebase with master" button 😃

watson:
jenkins, retest this please

package.json
```diff
@@ -73,11 +73,13 @@
     "console-log-level": "^1.4.0",
     "cookie": "^0.3.1",
     "core-util-is": "^1.0.2",
-    "elastic-apm-http-client": "^6.0.1",
+    "cpu-gauge": "^1.1.0",
```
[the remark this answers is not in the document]
watson: Agreed on call: Let's vendor https://github.com/mafintosh/process-top/blob/master/index.js and comment out the event-loop-delay sampler for now until we need it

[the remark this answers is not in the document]
Qard: fixed.

mafintosh:
Alternatively we can add a simple switch to process-top that disables the loop sampler if that’s an issue

watson:
@mafintosh we just vendored the file, so we're good for now. But if you think it could be useful for users in general, then we'd definitely start using the module instead if you implement it like that :)

Qard:
I also disabled the timer as the `measured` gatherer has its own timer, so it might actually be better to just have a simplified module that just averages between two call points without layering any timer over it and then just using that inside process-top. 🤔

lib/metrics/process-cpu.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+
+const os = require('os')
+
+const processTop = require('./process-top')()
+
+const cpus = os.cpus()
+
+module.exports = function processCPUUsage () {
+  return processTop.cpu() / cpus.length
```
watson:
```suggestion
  return processTop.cpu().percent / cpus.length
```

PR #735, opened 2018-12-10
Closes #732

### LINKED ISSUE #732: Requests to APM Server are sometimes logged by the agent
Normally HTTP requests to the APM Server should be ignored by the agent so that it doesn't create spans for the request to the APM Server it sends it self.

But it's seems that sometimes these requests are not ignored: https://discuss.elastic.co/t/how-to-exclude-apm-self-sent-events-capturing/160134

uzarsalan:
changed hostname to host

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -82,7 +82,7 @@ Agent.prototype._config = function (opts) {
   this._conf = config(opts)
 
   this._conf.serverHost = this._conf.serverUrl
-    ? parseUrl(this._conf.serverUrl).hostname
+    ? parseUrl(this._conf.serverUrl).host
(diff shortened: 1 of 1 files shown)


elasticcla:
Hi @uzarsalan, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in yout Git [commit](https://github.com/elastic/apm-agent-nodejs/pull/735.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

watson:
@uzarsalan thank you so much for fixing this 💯 

Your patch will be part of the next release 😃

PR #743, opened 2018-12-14
test(ci): fix apollo-server-express tests on older Node versions

watson:
When apollo-server-express version 2.3.0 was released it contained a
breaking change, requiring Node.js 8.5.0 or newer:

https://github.com/apollographql/apollo-server/issues/2099

If the above issue is ever fixed we should update our tests again.

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -152,47 +152,26 @@ express-graphql-0.7.1_14:
   versions: '^0.7.1'
   commands: node test/instrumentation/modules/express-graphql.js
 
-apollo-server-express-2_12-old:
+apollo-server-express-2_12:
   name: apollo-server-express
   preinstall: npm uninstall express-graphql
   peerDependencies: graphql@^0.12.0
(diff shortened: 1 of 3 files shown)


watson:
jenkins, retest this please

abernix:
> If the above issue is ever fixed we should update our tests again.

Sorry for any inconvenience this caused!

Best I can tell, since this doesn't appear to be actually _utilizing_ file uploads, you shouldn't have to re-visit this.  That is to say: even though uploads "on" is the Apollo Server default, it's not really clear if this is an ideal default since only fraction of users actually utilize file uploads within GraphQL itself and the enabling of uploads introduces unnecessary complexity to the request path.  I'd probably recommend keeping it disabled and we may reconsider this default in the future.

(Also, there's only ~four months of Node.js 6 LTS remaining so most users should be on their way to Node.js v8.9.0+ / v10.13.0+ anyway.)

PR #744, opened 2018-12-14
test(ci): remove forced upgrade of npm to latest version

watson:
This was added in elastic/apm-agent-nodejs#557 to fix a TAV test failure. But the issue requiring us to upgrade have since been fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ install:
   - for i in {1..3}; do travis_wait 5 npm install && break; rm -rf node_modules; done
 
 before_script:
-  - npm install -g npm@latest
   - wait-on tcp:9200
 
 node_js:


watson:
jenkins, reset this please

watson:
Depends on #743

watson:
jenkins, retest this please

Qard:
BTW, I'm able to reproduce the hanging `tav` run on 6.x locally. I left it running last night and it sat for an hour after reaching ok before I just killed the process. As we expected previously, it seems to be an issue with old versions of npm, as it doesn't happen at all if I upgrade npm. I think we need to just leave the upgrade in there until we deprecate 6.x support.

watson:
This should be part of the "drop support for Node.js 6" PR

PR #749, opened 2018-12-17
fix(koa-router): support more complex routes

### LINKED ISSUE #742: Requests come as GET (.*), what could be causing it?
**Describe the bug**
I'm using version `2.0.4` of the agent with Koa, everything works ok, but the requests transactions are being sent as `GET (.*)`. What would cause them to be called with that name?

Everything seems correct, just the name that seems wrong

**To Reproduce**
1. I'm using a simple koa app with a route `router.get('/healthcheck' ctx => { ctx.body = 'ok' })`

**Expected behavior**
It should send the transaction with the name of the route, `/healthcheck`.

- OS: Linux on GKE
- Node.js version: 10
- APM Server version: 6.5.2
- Agent version: 2.0.4

I've enabled debug logs on the apm agent:
image

watson:
This adds support for correctly naming transactions when using middlewares and/or nested routes.

Fixes #742

### DIFF
--- a/lib/instrumentation/modules/koa-router.js
+++ b/lib/instrumentation/modules/koa-router.js
@@ -21,9 +21,12 @@ module.exports = function (Router, agent, version, enabled) {
         return matched
       }
 
-      if (matched && matched.pathAndMethod && matched.pathAndMethod.length) {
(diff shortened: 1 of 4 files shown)


watson:
Released as part of v2.0.6

PR #751, opened 2018-12-17
docs: update code samples

### LINKED ISSUE #640: Use const, let etc in all documentation
We still use `var` in quite a few places in the source code for historical reasons, and that's ok. But it would be nicer if we updated our documentation to show more modern code examples.

bmorelli25:
Fixes elastic/apm-agent-nodejs#640


bmorelli25:
Ah whoops, thanks @watson

PR #757, opened 2018-12-18
feat(span): add tags to spans

watson:
This PR adds tag support to spans by adding the following two functions to the Span object:

- `span.setTag()`
- `span.addTags()`

As a side-effect of the refactor, the following TraceContext related API's will also be exposed on the Span object:

- `span.parentId`
- `span.traceId`
- `span.sampled` (I guess this one isn't really useful as spans are always sampled, but I felt that since it was TraceContext related it still made sense to put in in the base class)

Since version 2.0.0, the TraceContext was exposed on spans and transactions via a `context` property. This was never documented, and this PR renames the property to `_context`. But I don't think this is a breaking change.

### Checklist

- [x] Decide what should happen when a user calls `agent.setTag()` or `agent.addTags()`:
  - Should the tags always be set on the transaction as today? 
  - ~~Should the tags instead be set on the span if one is available?~~
- [x] Implement code
- [x] Update tests
- [x] Update documentation


lib/instrumentation/span.js
```diff
@@ -129,10 +113,10 @@ Span.prototype._encode = function (cb) {
     }
 
     var payload = {
```
Qard: I don't see tags added to the _encode. 🤔

Qard:
Needs tests for `span.setTag(...)` and `span.addTags(...)`.

watson:
Ha, good point. I was so focused on making sure the tests still passed after the refactor that I forgot testing that the new functionality actually worked 😄

docs/agent-api.asciidoc
```diff
@@ -477,6 +480,20 @@ if used within the context of a transaction.
 NOTE: If there's no active transaction available,
 `null` will be returned.
 
+[[apm-current-span]]
+==== `apm.currentSpan`
+
+[[source,js]]
```
bmorelli25:
```suggestion
[source,js]
```

bmorelli25: Also update line 472 to `[source,js]`

docs/transaction-api.asciidoc
```diff
@@ -73,6 +73,55 @@ When a span is started it will measure the time until <<span-end,`span.end()`>>
...
+* `value` - Any string.
+If a non-string data type is given,
+it's converted to a string before being sent to the APM Server
+
+[[transaction-add-tags]]
+==== `transaction.addTags()`
+
+[source,js]
+----
+apm.addTags({ [name]: value })
```
bmorelli25: Is the extra whitespace in this line by design? Looks kinda goofy to me. Your call:
screen shot 2018-12-18 at 12 29 00 pm

docs/agent-api.asciidoc
bmorelli25: @watson ^

watson: Thanks. I'll update line 472 in a separate PR as I saw that this isn't the only place this mistake snug in 😅

docs/transaction-api.asciidoc
watson: Technically yes. But it's not super intuitive. In general I'd like to discuss in general how to better show the method signature as it's currently done in several different ways throughout the documentation with no common logic. I'll create a separate issue for it and mark this review comment as resolved.

watson: Here: #760

bmorelli25: Perfect! Thanks

PR #762, opened 2018-12-18
refactor: improve span payload encoding

watson:
Just a tiny nit that I found. Would normally have done this as part of the refactor in the span tags PR, but I missed it.


watson:
jenkins, retest this please

PR #763, opened 2018-12-18
[1.x] test(ci): fix apollo-server-express tests on older Node versions (#743)

watson:
Backports the following commits to 1.x:
 - test(ci): fix apollo-server-express tests on older Node versions  (#743)

### DIFF
--- a/test/instrumentation/modules/apollo-server-express.js
+++ b/test/instrumentation/modules/apollo-server-express.js
@@ -37,7 +37,7 @@ test('POST /graphql', function (t) {
   var query = '{"query":"{ hello }"}'
 
   var app = express()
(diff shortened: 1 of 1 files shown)


watson:
The test failure seems unrelated to this PR, but before merging we need to figure out why and fix the tests in a separate PR

Qard:
Is the stall happening there consistently? I've seen npm cause tav stalls a couple times.

watson:
I've restarted the test once and got the same result. Just restarted it again, but I don't think it will fix it unfortunately. I think it's somehow related to the version of npm being used. Or at least upgrading npm seems to have fixed this problem in `master`

PR #766, opened 2018-12-21
feat(config): add container info options

### LINKED ISSUE #693: Add Docker container ID to metadata
Implement elastic/apm#22

The task is to see if the `/proc/self/cgroup` file exists, if so, get the Docker container ID from it and add it to `system.container.id` in the `metadata` event. It should read the ID on startup in a blocking way so that we have it available when the first `metadata` event is generated.

~~There's already [an npm module](https://www.npmjs.com/package/docker-container-id) that gets this ID, but it requires Node.js 7.8+ and it doesn't allow for synchronous reading. The reasons for the Node 7.8+ requirement is because of a [unnecessary feature](https://github.com/courajs/docker-container-id/blob/7ebf8d1e4300533831df202472afdfe914293c7e/index.js#L29-L34) whos use-case escapes me. So unless we can get the author to change a few things, we're probably better of forking this and releasing a revised version.~~

Update: There's another module called [docker-id](https://www.npmjs.com/package/docker-id) which does support sync loading and supports earlier versions of Node.js. We should probably just use that.

### LINKED ISSUE #737: Read k8s metadata from the /proc filesystem
See https://github.com/elastic/apm/issues/27

Qard:
Fixes elastic/apm-agent-nodejs#693 
Fixes elastic/apm-agent-nodejs#737
Depends on elastic/apm-nodejs-http-client#42

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -623,3 +623,47 @@ For an always up-to-date list of modules for which instrumentation can be disabl
 see the https://github.com/elastic/apm-agent-nodejs/tree/master/lib/instrumentation/modules[lib/instrumentation/modules] folder in the agent repository.
 Note that not all modules represented in this directory will generate spans,
 and adding those to this array has no effect.
+
+[[container-id]]
+==== `containerId`
+
+* *Type:* String
+* *Env:* `ELASTIC_APM_CONTAINER_ID`
+
+Specify the docker container id to associate with all reported events.
+If absent,
+it will be parsed out of the `/proc/self/cgroup` file.
+
+[[kubernetes-node-name]]
+==== `kubernetesNodeName`
+
+* *Type:* String
+* *Env:* `ELASTIC_APM_KUBERNETES_NODE_NAME`
+
+Specify the kubernetes node name to associate with all reported events.
+
+[[kubernetes-namespace]]
+==== `kubernetesNamespace`
+
+* *Type:* String
+* *Env:* `ELASTIC_APM_KUBERNETES_NAMESPACE`
+
+Specify the kubernetes namespace to associate with all reported events.
+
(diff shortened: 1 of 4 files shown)


Qard:
done.

felixbarny:
Out of curiosity, why did you add configuration options fo that? Is it in case we fail to parse future formats we don't recognize yet? Should all agents add those options? Does the beats kubernetes metadata processor also offer these options?

watson:
@felixbarny Yes, this is in case we can't parse it automatically. According to https://github.com/elastic/apm/issues/21 we were supposed to make this configurable. So we didn't really give it any thought besides just following the implementation issue.

eyalkoren:
@watson @felixbarny the main rationale of these env variables considered [k8 Downwards API](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-pod-fields-as-values-for-environment-variables), which allows setting ENV variables from pod fields, but of course people can set them manually in any other way. Still, I assume the documentation can direct to that (I actually didn't add anything to the documentation 🤦‍♂️...)

watson:
Does this mean that we shouldn't have prefixed them with `ELASTIC_APM_`?

eyalkoren:
I didn't - https://github.com/elastic/apm-agent-java/blob/d83d37b33ea4ef1aa0904bb35394c21779db54bb/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/payload/SystemInfo.java#L129 and looks like Go agent doesn't as well https://github.com/elastic/apm-agent-go/blob/e510527b3a3b5be41fdea807e4b776931bd76689/utils.go#L107.

They can be used either way, but I think the idea was align on how they are listed in elastic/apm#21

PR #767, opened 2018-12-21
feat: include sampled flag on errors

### LINKED ISSUE #753: Add transaction.sampled to errors
Implement elastic/apm#30

Qard:
### Checklist

- [x] Implement code
- [x] Add tests

Closes #753

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -263,7 +263,10 @@ Agent.prototype.captureError = function (err, opts, cb) {
       )
     }
 
-    if (trans) error.transaction_id = trans.id
+    if (trans) {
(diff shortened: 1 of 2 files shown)


lib/agent.js
```diff
@@ -244,6 +244,7 @@ Agent.prototype.captureError = function (err, opts, cb) {
     error.parent_id = context.id
     error.trace_id = context.traceId
     error.timestamp = timestamp
+    error.sampled = trans && trans.sampled
```
watson: I think `trans` will be `null` if there's no active transaction, in which case the `sampled` property wrongly will be set to `null` as well

PR #774, opened 2019-01-14
[docs] Restructure configuration

### LINKED ISSUE #758: Restructure Configuration docs
It would probably make sense to move the Configuration page above the Agent API page:

image

The [Configuration page](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html) should also be updated to better stand on its own. It seems a little out of context currently - as it technically also is as it has been moved from being a sub-section under `apm.start()` - I.e. it doesn't currently say that you can pass the options to the `start()` function.

bmorelli25:
Fixes #758, fixes configuration precedence, updates some links from `agent-api` to `configuration`. 

After looking at this a little deeper, it doesn't make sense to nest the `configuration` options within the `Agent API` - it's simply too big.

I'm wondering what you think of this solution. 

In this PR, I've moved the configuration options out of the API reference and into the `Advanced Setup and Configuration` section. Alternatively, I could move the config out of Advanced setup and into the API reference. Either way, the content was being duplicated and it'd be best if it only lived in one place.

### DIFF
--- a/docs/setup.asciidoc
+++ b/docs/setup.asciidoc
@@ -11,8 +11,8 @@ The following pages will help you set up and configure your APM Agent.
 
 * <<starting-the-agent>>
 * <<configuring-the-agent>>
-* <<agent-configuration-object>>
-* <<agent-configuration-file>>
+** <<agent-configuration-object>>
+** <<agent-configuration-file>>
 * <<es-modules>>
 * <<configuration>>
 
--- a/docs/setup.asciidoc
+++ b/docs/setup.asciidoc
@@ -7,6 +7,13 @@ endif::[]
 
 == Advanced Setup and Configuration
 
+* <<starting-the-agent>>
+* <<configuring-the-agent>>
+* <<agent-configuration-object>>
+* <<agent-configuration-file>>
+* <<es-modules>>
+* <<configuration>>
+
 [float]
 [[starting-the-agent]]
 === Starting the agent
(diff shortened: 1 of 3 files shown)


bmorelli25:
Updated with correct commit messages

docs/setup.asciidoc
```diff
@@ -7,6 +7,13 @@ endif::[]
 
 == Advanced Setup and Configuration
 
+* <<starting-the-agent>>
```
watson: Will this not also be in the right hand side menu?

bmorelli25:
@watson - we have to pop the float tags off in order to get the headings to show up in the TOC. Check out my latest update to this PR to see what that looks like. Only downside is a somewhat small "setup and config" landing page, but I can live with it.

docs/setup.asciidoc
```diff
@@ -5,9 +5,17 @@ NOTE: For the best reading experience,
...
 endif::[]
 
-== Advanced Setup and Configuration
+== Setup and Configuration
+
+The following pages will help you set up and configure your APM Agent.
+
+* <<starting-the-agent>>
+* <<configuring-the-agent>>
+* <<agent-configuration-object>>
```
watson: This bullet point (and the next) is technically sub headlines on the "Configuring the agent" page. What do you think of indenting these two bullets one level?

PR #769, opened 2019-01-08
test(redis): fix timing issue by ignoring order

Qard:
This is just a test fix to make Travis CI play nice with the metrics tests.

PR #779, opened 2019-01-15
[1.x] test(redis): fix timing issue by ignoring order (#769)

watson:
Backports the following commits to 1.x:
 - test(redis): fix timing issue by ignoring order  (#769)

### DIFF
--- a/test/instrumentation/modules/redis.js
+++ b/test/instrumentation/modules/redis.js
@@ -9,6 +9,8 @@ var agent = require('../../..').start({
 var redis = require('redis')
 var test = require('tape')
 
+var findObjInArray = require('../../_utils').findObjInArray
+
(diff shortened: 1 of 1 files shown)


test/instrumentation/modules/redis.js
```diff
@@ -9,6 +9,8 @@ var agent = require('../../..').start({
 var redis = require('redis')
 var test = require('tape')
 
+var findObjInArray = require('../../_utils').findObjInArray
```
Qard: Seems the `_util` helper module doesn't exist in 1.x. 🤔

watson:
I think I'll just not backport this one. It wasn't important, but I just thought "why not". But as long as we're not getting redis errors I don't want to spend much time on it.

PR #788, opened 2019-01-15
feat(config): add captureHeaders config

### LINKED ISSUE #739: Add captureHeaders config option
The `captureHeaders` should default to `true` (which is also the current behavior). If set to `false`, HTTP headers (and potentially other headers) should not be logged on transactions, spans or errors.

Besides being documented on the configuration page, this config option should also be documented on the performance optimization page.

Qard:
Still needs more tests and documentation.

Fixes #739

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -218,6 +218,16 @@ the body needs to be available as a property on the incoming HTTP https://nodejs
 The agent will look for the body on the following properties:
 `req.json || req.body || req.payload`
 
+[[capture-headers]]
+==== `captureHeaders`
+
+* *Type:* Boolean
+* *Default:* true
+* *Env:* `ELASTIC_APM_CAPTURE_HEADERS`
+
+The HTTP headers of incoming HTTP requests are recorded and sent to the APM Server by default.
+This can be disabled by setting this option to `false`.
+
 [[error-on-aborted-requests]]
 ==== `errorOnAbortedRequests`
 
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -225,7 +225,9 @@ The agent will look for the body on the following properties:
 * *Default:* true
 * *Env:* `ELASTIC_APM_CAPTURE_HEADERS`
 
-The HTTP headers of incoming HTTP requests are recorded and sent to the APM Server by default.
+The HTTP headers of incoming HTTP requests,
+and its resulting response headers,
+are recorded and sent to the APM Server by default.
 This can be disabled by setting this option to `false`.
 
 [[error-on-aborted-requests]]
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -88,7 +88,9 @@ exports.parseError = function (err, agent, cb) {
   })
 }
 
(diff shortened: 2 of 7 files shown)


lib/parsers.js
```diff
@@ -88,16 +88,21 @@ exports.parseError = function (err, agent, cb) {
...
+
   var context = {
     http_version: req.httpVersion,
     method: req.method,
     url: getUrlFromRequest(req),
     socket: {
       remote_address: req.socket.remoteAddress,
       encrypted: !!req.socket.encrypted
-    },
-    headers: Object.assign({}, req.headers)
```
watson: If we set `headers: undefined` here, we'll reduce the number of hidden classes by one right?

lib/parsers.js
```diff
@@ -123,10 +128,13 @@ exports.getContextFromRequest = function (req, captureBody) {
   return context
 }
 
-exports.getContextFromResponse = function (res, isError) {
+exports.getContextFromResponse = function (res, conf, isError) {
   var context = {
-    status_code: res.statusCode,
-    headers: res.headers || httpHeaders(res, true)
```
watson: Same here

test/parsers.js
```diff
@@ -146,9 +147,10 @@ test('#getContextFromRequest()', function (t) {
   })
 
   t.test('full URI', function (t) {
+    var conf = { captureHeaders: true, captureBody: 'off' }
     var req = getMockReq()
     req.url = 'https://www.example.com:8080/some/path?key=value'
-    var parsed = parsers.getContextFromRequest(req)
+    var parsed = parsers.getContextFromRequest(req, conf)
```
watson: Since we're not actually testing the body or the headers, I'd just pass in an empty object here. Same goes for the other tests below where we don't need these. What do you think?

lib/parsers.js
Qard: fixed.

test/parsers.js
Qard: fixed.

lib/parsers.js
Qard: fixed.

Qard:
Added docs.

docs/configuration.asciidoc
```diff
@@ -218,6 +218,16 @@ the body needs to be available as a property on the incoming HTTP https://nodejs
...
 `req.json || req.body || req.payload`
 
+[[capture-headers]]
+==== `captureHeaders`
+
+* *Type:* Boolean
+* *Default:* true
+* *Env:* `ELASTIC_APM_CAPTURE_HEADERS`
+
+The HTTP headers of incoming HTTP requests are recorded and sent to the APM Server by default.
```
watson: Adding info about response headers... There's probably a less convoluted way of writing this, but you get the point. Feel free to rewrite it:

```suggestion
The HTTP headers of incoming HTTP requests,
and its resulting response headers,
are recorded and sent to the APM Server by default.
```

watson:
@Qard I just realized that this was marked in-progress and not in-review. I thought it was finished though?

PR #793, opened 2019-01-16
fix(mongo): support 3.1.10+

Qard:
This is a possible fix for the issues with mongo 3.1.10+ no longer having a `cursor._find(...)` method.


Qard:
In the tests, I've added an extra `cursor.next(...)` call to prove that subsequent calls won't trigger more find spans unless `cursor._getmore(...)` is called. I also added a remove to clean up the data after, to ensure the cursor finds the records it expects.

PR #794, opened 2019-01-16
test: verify internal http requests not traced

### LINKED ISSUE #738: Add test to verify that APM Server requests are not traced
Until this was fixed in elastic/apm-agent-nodejs#735 this has apparently been an issue for a long time.

Our test suite didn't catch it because the HTTP client that sends data to the APM Server was mocked, so it never actually sent an HTTP request and so never triggered the bug.

It would be a good improvement to our test suite to add a test that checks this doesn't happen.

Qard:
Fixes elastic/apm-agent-nodejs#738

### Checklist

- [x] Add tests

### DIFF
--- a/test/instrumentation/index.js
+++ b/test/instrumentation/index.js
@@ -653,6 +653,39 @@ test('nested transactions', function (t) {
   t0.end()
 })
 
+test('does not record requests from internals', function (t) {
+  resetAgent(function (data) {
+    t.equal(data.transactions.length, 1, 'has one transaction')
+    t.equal(data.spans.length, 2, 'has two spans')
+
+    const trans = findObjInArray(data.transactions, 'name', 't0')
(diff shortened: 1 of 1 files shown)


test/instrumentation/index.js
```diff
@@ -653,6 +653,27 @@ test('nested transactions', function (t) {
   t0.end()
 })
 
+test('does not record requests from internals', function (t) {
+  resetAgent(3, function (data) {
```
watson: Instead of waiting for exactly 3 events, this will make the mock agent wait about 100ms after the last received event and then call the callback if there have been no other events:

```suggestion
  resetAgent(function (data) {
```

Then we should just check that `data` contains the 1 metadata object, 2 spans, and 1 transaction that we expected. I think that should work.

Qard: fixed

Issue #31, opened 2017-09-18
Countless retries and log spam for misconfigure port.


simitt:
I misconfigured the apm server url by accident `serverUrl: 'https://localhost:82000'`. 
This lead to countless retries of sending data to the APM Server, see logs:
```
logging error 6a9ff729-cec0-45d5-bc5e-a57e27f2a25d with Elastic APM
logging error 54c4eb0f-5a98-4b41-b07e-5b4f60a2ba5f with Elastic APM
logging error 8cbe19af-37fb-468b-b9bf-8070b5531ca4 with Elastic APM
logging error f9d805e8-488f-4a41-b4f6-20154d8101fe with Elastic APM
logging error 53df0b4b-2cd0-419c-8af4-27be7f36d716 with Elastic APM
logging error 80fd91b6-c09f-48cb-8787-b8ac0d1d1775 with Elastic APM
logging error c9858407-17a3-4968-8cce-297e0bd31f3d with Elastic APM
logging error b3a69e7b-8eec-4580-a219-e72eb4e5b637 with Elastic APM
logging error 6935b29d-9629-4dc2-8b62-9954e6559693 with Elastic APM
logging error 6e31e7cb-4a2d-44e1-bb1c-279bef550de5 with Elastic APM
logging error cf647462-65e1-48e3-a3c4-15a3779bce6f with Elastic APM
logging error 79e1590c-bdad-4c41-b97e-ccc43189bdbb with Elastic APM
...
```

When I just misconfigure to another wrong port, e.g. `8201` I get a proper `Connection Refused` error.

watson:
For reference, this is the actual error that occurs. It's thrown by the agent and then caught again by the agent and so on:

```
net.js:1023
      throw new RangeError('"port" option should be >= 0 and < 65536: ' + port);
      ^

RangeError: "port" option should be >= 0 and < 65536: 82000
    at lookupAndConnect (net.js:1023:13)
    at TLSSocket.Socket.connect (net.js:995:5)
    at TLSSocket.connect (/Users/watson/code/node_modules/elastic-apm/lib/instrumentation/async-hooks.js:125:29)
    at Object.exports.connect (_tls_wrap.js:1070:12)
    at Agent.createConnection (https.js:111:22)
    at Agent.createSocket (_http_agent.js:214:26)
    at Agent.addRequest (_http_agent.js:184:10)
[...]

PR #795, opened 2019-01-17
feat(config): validate serverUrl port

Qard:
This should help a bit with avoiding startup log spam mentioned in #31.


watson:
The Jenkins CI is currently broken but can be safely ignored as long as the Travis CI passes.

PR #796, opened 2019-01-17
feat: ignore internal errors

### LINKED ISSUE #31: Countless retries and log spam for misconfigure port.

Qard:
This should prevent internal errors from being reported, prevent infinite error reporting loops.

Fixes elastic/apm-agent-nodejs#31

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -205,6 +205,11 @@ Agent.prototype.addSpanFilter = function (fn) {
   this._spanFilters.add(fn)
 }
 
+const internalFrameRegex = /^node_modules\/elastic-apm/
+function isInternalFrame (frame) {
+  return internalFrameRegex.test(frame.filename)
+}
+
 Agent.prototype.captureError = function (err, opts, cb) {
   if (typeof opts === 'function') return this.captureError(err, null, opts)
 
@@ -233,6 +238,11 @@ Agent.prototype.captureError = function (err, opts, cb) {
     prepareError(parsers.parseMessage(err))
   } else {
     parsers.parseError(err, agent, function (_, error) {
+      const hasInternal = error.exception.stacktrace
+        .some(isInternalFrame)
+
+      if (hasInternal) return
+
       // As of now, parseError suppresses errors internally, but even if they
       // were passed on, we would want to suppress them here anyway
       prepareError(error)


lib/agent.js
```diff
@@ -205,6 +205,10 @@ Agent.prototype.addSpanFilter = function (fn) {
   this._spanFilters.add(fn)
 }
 
+function isInternalFrame (frame) {
+  return /^node_modules\/elastic-apm/.test(frame.filename)
```
watson: While this works, shouldn't it technically be checking for the complete module name: `elastic-apm-node`?

To support Windows we should also use `path.join()`.

And I'm not sure if this is still the case or just an old habit of mine, but I usually define RegExes outside functions so they can be compiled once and reused many times. It used to be that V8 couldn't optimize this part. But maybe that's no longer an issue

lib/agent.js
```diff
@@ -233,6 +237,12 @@ Agent.prototype.captureError = function (err, opts, cb) {
     prepareError(parsers.parseMessage(err))
   } else {
     parsers.parseError(err, agent, function (_, error) {
+      const hasInternal = error.exception.stacktrace
+        .filter(isInternalFrame)
```
watson: You can optimize this by using `some` instead of `filter`. That way it will stop when it finds the first occurrence of an internal frame, e.g:

```js
const hasInternal = error.exception.stacktrace.some(internalFrame)
```

lib/agent.js
Qard: By using only part of the name, it can catch `elastic-apm-http-client` too. I can move the regex out too. 👍

lib/agent.js
Qard: Good call.

lib/agent.js
watson: Good point. Didn't think of `elastic-apm-http-client`

lib/agent.js
watson: At second thought, do we really want to look for this frame _any_ where in the stack? We do a lot of monkey patching and we're probably going to be present in a bunch of stacks even though the error didn't originate from within the agent. Maybe we should just look for the stop stack 🤔 Hmmm this quickly gets tricky

Qard: What do we do if the failure originates from one of our dependencies then? 🤔

lib/agent.js
Qard: I've externalized the regex. Do you think for Windows paths it's sufficient to just check for _either_ `/` or `\`?

lib/agent.js
Qard: Switched to `some(...)`. It's an _improvement_ of the current logic, but seems like it should maybe be different somehow...

lib/agent.js
watson: You could also just change the regex to:

```js
const internalFrameRegex = new RegExp(path.join('^node_modules', ' elastic-apm'))
```

watson: Hmm just ran a test and it seems that the `\` in the windows path is being interpreted as an escape char in the regex

watson: ... so I guess we'll either need to double escape them or do the thing you suggested - which might be easier at this point

Qard: Oh, sorry if it was unclear: I wasn't implying it _had_ support for Windows paths, I was asking if changing it to just understand both was what it needed. 😅

watson: Yeah, got that. I just then tried to investigate if just using `path.join` would be a simpler way to do it, but turns out it wasn't. So allowing it to accept both is probably easiest.

v1v:
jenkins run the tests please

PR #801, opened 2019-01-17
[APM-CI] Jenkins Pipeline for Agent Node.js

kuisathaverat:
Jenkins file to build the APM Agent Node.js with the multibranch pipeline job https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/

image

### DIFF
--- a/test/script/docker/run_tests.sh
+++ b/test/script/docker/run_tests.sh
@@ -24,17 +24,16 @@ else
   CMD='npm test'
 fi
 
-docker build --pull --force-rm \
-  --build-arg NODE_VERSION=$1 \
-  -t apm-agent-nodejs:node-${1} \
-  ./test
-
 NODE_VERSION=${1} docker-compose -f ./test/docker-compose.yml -f ./test/docker-compose.ci.yml run \
-  -e NODE_VERSION=${NODE_VERSION} -e TAV=${TAV_VERSIONS} -e CI=true \
+  -e NODE_VERSION=${NODE_VERSION} \
+  -e TAV=${TAV_VERSIONS} \
+  -e JUNIT=${JUNIT} \
+  -e CI=true \
   -v ${npm_cache}:${docker_npm_cache} \
   -v ${nyc_output}:${docker_nyc_output} \
   -v ${nyc_report_output}:${docker_nyc_report_output} \
   -v "$(pwd)":/app \
+  -w /app \
   --rm node_tests \
   /bin/bash \
   -c "npm config set cache ${docker_npm_cache} --global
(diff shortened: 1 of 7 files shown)


kuisathaverat:
Tests results are loaded

screenshot 2019-01-22 at 16 47 47

test/script/docker/run_tests.sh
```diff
@@ -24,17 +24,17 @@ else
...
-  -t apm-agent-nodejs:node-${1} \
-  ./test
-
 NODE_VERSION=${1} docker-compose -f ./test/docker-compose.yml -f ./test/docker-compose.ci.yml run \
-  -e NODE_VERSION=${NODE_VERSION} -e TAV=${TAV_VERSIONS} -e CI=true \
+  -e NODE_VERSION=${NODE_VERSION} \
+  -e TAV=${TAV_VERSIONS} \
+  -e JUNIT=${JUNIT} \
+  -e CI=true \
+  -e ELASTIC_APM_ASYNC_HOOKS=false \
```
Qard: Why are we disabling async_hooks here?

kuisathaverat: C&P from Travis file

Qard: Ah, we _should_ actually be testing both states. On Travis, some jobs include that and some don't. Specifically, every version of Node.js from 8 and above should test both states.

Qard: The current setup tests with it enabled, so should we maybe just remove this line for now and land, then we can figure out a matrix test to cover both states later?

kuisathaverat: yep, we can make it on the matrix test

Qard:
I hope you don't mind, I've squashed and rebased so I can get a new test run that passes the commit lint test.

kuisathaverat:
@Qard it is fine, seems like Node.js test does not end ever, so the job is killed after an hour, we have a timeout of one hour set.

Qard:
Yeah, I've been looking into the pause. Not sure what the issue is yet. I've so far been unable to reproduce locally, with the same Node.js and module versions. 🤔

kuisathaverat:
> Yeah, I've been looking into the pause. Not sure what the issue is yet. I've so far been unable to reproduce locally, with the same Node.js and module versions. 🤔

Could it be related to the memory amount on Jenkins agents? IIRC Jenkins Agents has about 4G of memory.

kuisathaverat:
@alvarolobato I tested it yesterday, it seems something that it is fixed right now, not related to the CI, I have launched it again to be sure, if the CI pass the PR is ready to merge.

kuisathaverat:
There is something on the test that sometimes fails with Node.js 6/6.0, the behavior it is that is stuck in some place, it is not related to CI.

PR #803, opened 2019-01-17
feat(instrumentation): add patch registry

Qard:
I still need to write tests and docs for this, but this is a possible way to support custom module patches.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

Closes elastic/apm-agent-nodejs#878

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -655,6 +655,23 @@ Read more lambda support in the <<lambda,Lambda>> article.
 [[apm-add-patch]]
 ==== `apm.addPatch(name, handler)`
 
+Register a module patch to apply on intercepted `require` calls.
+
+Arguments:
+
+* `name` - string
+Name of module to apply the patch to, when required.
+* `handler` - function or string
+Must be a patch function or a path to a module exporting a patch function
+** `exports` - The original export object of the module
+** `agent` - The agent instance to use in the patch function
+** `options` - The following options are supported:
+*** `version` - The module version, if applicable.
+*** `enabled` - A flag indicating if the instrumentation is enabled.
+Any module patch can be disabled, by module name, with <<disable-instrumentations,`disableInstrumentations`>>.
+
+A module can have any number of patches and will be applied in the order they are added.
+
 [source,js]
 ----
 apm.addPatch('timers', (exports, agent, { version, enabled }) => {
@@ -675,25 +692,12 @@ apm.addPatch('timers', (exports, agent, { version, enabled }) => {
 apm.addPatch('timers', './timer-patch')
 ----
 
-Register a module patch to apply on intercepted `require` calls.
-
-Arguments:
-
-* `name` - string
-Name of module to apply the patch to, when required.
-* `handler` - function or string
-Must be a patch function or a path to a module exporting a patch function
-** `exports` - The original export object of the module
-** `agent` - The agent instance to use in the patch function
-** `options` - The following options are supported:
-*** `version` - The module version, if applicable.
-*** `enabled` - A flag indicating if the instrumentation is enabled.
-Any module patch can be disabled, by module name, with <<disable-instrumentations,`disableInstrumentations`>>.
-
-A module can have any number of patches and will be applied in the order they are added.
-
 [[apm-remove-patch]]
 ==== `apm.removePatch(name, handler)`
+
+Removes a module patch.
+This will generally only be needed when replacing an existing patch.
+To _disable_ instrumentation while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
 
 [source,js]
 ----
@@ -704,18 +708,14 @@ apm.removePatch('timers', './timers-patch')
 apm.removePatch('timers', timerPatchFunction)
 ----
 
-Removes a module patch.
-This will generally only be needed when replacing an existing patch.
-To _disable_ instrumentation while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
-
 [[apm-clear-patches]]
 ==== `apm.clearPatches(name)`
+
+Clear all patches for the given module.
+This will generally only be needed when replacing an existing patch.
+To _disable_ instrumentation while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
 
 [source,js]
 ----
 apm.clearPatches('timers')
 ----
-
-Clear all patches for the given module.
-This will generally only be needed when replacing an existing patch.
-To _disable_ instrumentation while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -651,3 +651,64 @@ Optionally, a type may also be provided to group lambdas together. By default,
 "lambda" will be used as the type name.
 
 Read more lambda support in the <<lambda,Lambda>> article.
+
+[[apm-add-patch]]
+==== `apm.addPatch(name, handler])`
+
+[source,js]
+----
+apm.addPatch('timers', (exports, agent, version, enabled) => {
+  const setTimeout = exports.setTimeout
+  exports.setTimeout = (fn, ms) => {
+    const span = agent.createSpan('set-timeout')
+    return setTimeout(() => {
+      span.end()
+      fn()
+    }, ms)
+  }
+
+  return exports
+})
+
+// or ...
+
+apm.addPatch('timers', './timer-patch')
+----
+
+Register a module patch to apply on intercepted `require` calls.
+The handler must be a patch function or a path to a module exporting a patch function.
+The patch function receives the original exported object,
+the agent instance,
+the module version string,
+and a flag to indicated if it is enabled.
+Any module patch can be disabled with <<disable-instrumentations,`disableInstrumentations`>>.
+
+A module can have any number of patches and will be applied in the order they are added.
+
+[[apm-remove-patch]]
+==== `apm.removePatch(name, handler)`
+
+[source,js]
+----
+apm.removePatch('timers', './timers-patch')
+
+// or ...
+
+apm.removePatch('timers', timerPatchFunction)
+----
+
+Removes a module patch.
+This will generally only needed when replacing an existing patch.
+To _disable_ instrumentation from while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
+
+[[apm-clear-patches]]
+==== `apm.clearPatches(name)`
+
+[source,js]
+----
+apm.clearPatches('timers')
+----
+
+Clear all patches for the given module.
+This will generally only needed when replacing an existing patch.
+To _disable_ instrumentation from while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -653,11 +653,11 @@ Optionally, a type may also be provided to group lambdas together. By default,
 Read more lambda support in the <<lambda,Lambda>> article.
 
 [[apm-add-patch]]
-==== `apm.addPatch(name, handler])`
+==== `apm.addPatch(name, handler)`
 
 [source,js]
 ----
-apm.addPatch('timers', (exports, agent, version, enabled) => {
+apm.addPatch('timers', (exports, agent, { version, enabled }) => {
   const setTimeout = exports.setTimeout
   exports.setTimeout = (fn, ms) => {
     const span = agent.createSpan('set-timeout')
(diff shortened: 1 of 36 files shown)


lib/instrumentation/index.js
```diff
@@ -58,6 +58,20 @@ function Instrumentation (agent) {
...
+  this._patches = {}
+
+  for (let mod of MODULES) {
+    this.registerPatch(mod, () => require(`./modules/${mod}`))
+  }
+}
+
+Instrumentation.prototype.registerPatch = function (name, handler) {
+  this._modules.push(name)
+  this._patches[name] = typeof handler !== 'function'
```
watson: 🤔 I think I would probably change this check to a `typeof handler === 'string'` - but it's just a gut feeling... maybe it doesn't matter

lib/instrumentation/index.js
```diff
@@ -100,7 +114,9 @@ Instrumentation.prototype.start = function () {
 
 Instrumentation.prototype._patchModule = function (exports, name, version, enabled) {
   this._agent.logger.debug('shimming %s@%s module', name, version)
-  return require('./modules/' + name)(exports, this._agent, version, enabled)
+  var patch = this._patches[name]
+  if (!patch) return exports
```
watson: What is it that you're trying to guard against here?

watson:
Forgot to say in my review: Really great that we're getting this started 😃

lib/instrumentation/index.js
Qard: Should be impossible, but just making sure the corresponding patch function was found. Might be better to explicitly check the type, or we could just skip the check altogether, since it _shouldn't_ ever get there if there's no patch in the first place. 🤔

lib/instrumentation/index.js
Qard: Actually, probably better to explicitly check for both strings and functions, then also log an error if it is neither.

Qard: I changed it to check both and log if it's neither.

lib/instrumentation/index.js
```diff
@@ -59,6 +59,50 @@ function Instrumentation (agent) {
       return this.bindingSpan || this.activeSpan
     }
   })
+
+  // NOTE: we need to track module names for patches
+  // in a separate array rather than using Object.keys()
+  // because the array is given to the hook(...) call.
+  this._modules = []
```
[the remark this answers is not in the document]
Qard: Because that will only be the _original_ set of module names. This PR works by _kind of_ a hack that pushing new items into the array already given to `hook(...)` will allow new patches to be added after the hook has already been attached.

Qard:
As for the replacement behaviour: I did it this way because I felt it's a bit weird having it replace existing instrumentation or fail silently. The multiple wrap thing is maybe possible, but more complicated and I'm not sure how I feel about being able to apply multiple layers of instrumentation that could potentially interfere with each other. 🤔

lib/instrumentation/index.js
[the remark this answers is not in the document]
watson: But we don't allow the user to call `registerPatch` after `hook()` have been called anyway right?

watson:
> and I'm not sure how I feel about being able to apply multiple layers of instrumentation that could potentially interfere with each other. 🤔

Good point.

The main use-case I saw for this was if two different patches did two different things and the user wanted both of them. But this is all theoretical. I don't have a specific thing in mind - just included it for completeness.

watson:
Meeting update:
- Let's not add support for `elastic-apm-patch-` in this PR as we can always add it later (if it was added now and we later decided to change the prefix it would be almost impossible)
- We talked about if we should add a config option in this PR to control which modules are patched. We decided that it would be quicker to merge this PR without it and instead add it in a later PR as we first need to align with the other agents on how to parse a key/value list config option (possible suggestion: `CONFIG_OPTION=key1=value1,key2=value2`)

watson:
There's a discussion about the format of environment variable key/value lists here: https://github.com/elastic/apm/issues/48

watson:
In elastic/apm#48 we decided on the `CONFIG_OPTION=key1=value1,key2=value2` format

lib/agent.js
```diff
@@ -67,6 +67,18 @@ Agent.prototype.destroy = function () {
   if (this._transport) this._transport.destroy()
 }
 
+Agent.prototype.addPatch = function () {
```
watson: If you agree with #918, we should update these functions in a similar fashion

docs/agent-api.asciidoc
```diff
@@ -646,3 +646,64 @@ Optionally, a type may also be provided to group lambdas together. By default,
 "lambda" will be used as the type name.
 
 Read more lambda support in the <<lambda,Lambda>> article.
+
+[[apm-add-patch]]
+==== `apm.addPatch(name, handler])`
+
+[source,js]
+----
+apm.addPatch('timers', (exports, agent, version, enabled) => {
```
watson: Let's just discuss the arguments that this callback is being called with. The function signature have evolved randomly over time and this hasn't been a big concern as it wasn't a public API. Now that we're committing to it, we should just talk about if this is the right API

docs/agent-api.asciidoc
```diff
@@ -646,3 +646,64 @@ Optionally, a type may also be provided to group lambdas together. By default,
 "lambda" will be used as the type name.
 
 Read more lambda support in the <<lambda,Lambda>> article.
+
+[[apm-add-patch]]
+==== `apm.addPatch(name, handler])`
```
watson:
```suggestion
==== `apm.addPatch(name, handler)`
```

lib/agent.js
Qard: Will do. 👍

docs/agent-api.asciidoc
Qard: Well, we will _always_ need the exports, agent and version. I think `enabled` could maybe fit into a config object though. 🤔

lib/agent.js

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+
+Arguments:
+
+* `name` - string
+Name of module to apply the patch to, when required.
+* `handler` - function or string
+Must be a patch function or a path to a module exporting a patch function
+** `exports` - The original export object of the module
+** `agent` - The agent instance to use in the patch function
+** `options` - The following options are supported:\
```
bmorelli25:
```suggestion
** `options` - The following options are supported:
```

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+----
+apm.removePatch('timers', './timers-patch')
+
+// or ...
+
+apm.removePatch('timers', timerPatchFunction)
+----
+
+Removes a module patch.
+This will generally only needed when replacing an existing patch.
```
bmorelli25:
```suggestion
This will generally only be needed when replacing an existing patch.
```

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+To _disable_ instrumentation from while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
```
bmorelli25:
```suggestion
To _disable_ instrumentation while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
```

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+[[apm-clear-patches]]
+==== `apm.clearPatches(name)`
+
+[source,js]
+----
+apm.clearPatches('timers')
+----
+
+Clear all patches for the given module.
+This will generally only needed when replacing an existing patch.
```
bmorelli25:
```suggestion
This will generally only be needed when replacing an existing patch.
```

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+To _disable_ instrumentation from while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
```
bmorelli25:
```suggestion
To _disable_ instrumentation while keeping context propagation support, see <<disable-instrumentations,`disableInstrumentations`>>.
```

lib/instrumentation/index.js
```diff
@@ -60,14 +61,40 @@ function Instrumentation (agent) {
...
+  this._patches = new NamedArray()
+
+  for (let mod of MODULES) {
+    this.addPatch(mod, require.resolve(`./modules/${mod}`))
+  }
+}
+
+Instrumentation.prototype.addPatch = function (name, handler) {
+  if (typeof handler !== 'function' && typeof handler !== 'string') {
+    this._agent.logger.error('Invalid patch handler type')
```
watson: To ease future debugging, I'd probably output the wrong type as well:

```suggestion
    this._agent.logger.error('Invalid patch handler type:', typeof handler)
```

Not sure how clever the V8 optimizer is, but you might also consider to only run the `typeof` once and store the result in a `const`

lib/instrumentation/index.js
```diff
@@ -76,11 +103,20 @@ Instrumentation.prototype.start = function () {
     require('./patch-async')(this)
   }
 
+  this._startHook()
+}
+
+Instrumentation.prototype._startHook = function () {
+  if (this._hook) {
+    this._hook.unhook()
```
watson: What's the reason for this check?

lib/instrumentation/named-array.js
```diff
@@ -0,0 +1,47 @@
+// TODO: Move to a separate module
```
watson: Add `'use strict'` at the top of the file

lib/instrumentation/index.js
```diff
@@ -76,11 +103,20 @@ Instrumentation.prototype.start = function () {
...
+  }
+
+  var self = this
   var disabled = new Set(this._agent._conf.disableInstrumentations)
 
   this._agent.logger.debug('adding hook to Node.js module loader')
 
-  this._hook = hook(MODULES, function (exports, name, basedir) {
+  this._hook = hook(this._patches.keys, function (exports, name, basedir) {
```
watson: Where is this `keys` property coming from?

lib/instrumentation/index.js
```diff
@@ -102,7 +138,25 @@ Instrumentation.prototype.start = function () {
...
+    for (let patch of patches) {
+      if (typeof patch === 'string') {
+        if (patch[0] === '.') {
+          patch = path.resolve(process.cwd(), patch)
+        }
+        patch = require(patch)
+      }
+
+      if (typeof patch !== 'function') {
+        this._agent.logger.error('Invalid patch handler type')
```
watson: Similar to above...

```suggestion
        this._agent.logger.error('Invalid patch handler type:', typeof patch)
```

lib/instrumentation/index.js
Qard: Ah, right...I started working on a way to rebuild the hook rather than relying on the edge-case of `require-in-the-middle` _just working_ when we push new elements into the array of module paths given to it. Forgot about that. 😅

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+
+  return exports
+})
+
+// or ...
+
+apm.addPatch('timers', './timer-patch')
+----
+
+Register a module patch to apply on intercepted `require` calls.
```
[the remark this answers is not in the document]

docs/agent-api.asciidoc
```diff
@@ -651,3 +651,71 @@ Optionally, a type may also be provided to group lambdas together. By default,
...
+Any module patch can be disabled, by module name, with <<disable-instrumentations,`disableInstrumentations`>>.
+
+A module can have any number of patches and will be applied in the order they are added.
+
+[[apm-remove-patch]]
+==== `apm.removePatch(name, handler)`
+
+[source,js]
+----
+apm.removePatch('timers', './timers-patch')
```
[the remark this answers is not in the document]

lib/instrumentation/index.js

lib/instrumentation/named-array.js
Qard: fixed

lib/instrumentation/index.js

lib/instrumentation/index.js

lib/instrumentation/index.js

lib/instrumentation/index.js
```diff
@@ -76,11 +110,29 @@ Instrumentation.prototype.start = function () {
...
+  }
+
+  this._startHook()
+}
+
+Instrumentation.prototype._startHook = function () {
+  if (!this._started) return
+  if (this._hook) {
+    this._agent.logger.debug('removing hook to Node.js module loader')
+    this._hook.unhook()
```
watson: I'm not sure that we can support unhooking and the adding hooks again 🤔 I think that all calls to `require()` that have already been called, would not get updated with the new hooks.

lib/instrumentation/index.js
```diff
@@ -102,7 +154,26 @@ Instrumentation.prototype.start = function () {
...
+      if (typeof patch === 'string') {
+        if (patch[0] === '.') {
+          patch = path.resolve(process.cwd(), patch)
+        }
+        patch = require(patch)
+      }
+
+      const type = typeof patch
+      if (type !== 'function') {
+        this._agent.logger.error('Invalid patch handler type:', type)
```
watson: Nit: If we really wanted to give a good error message here we should include `name`. But feel free to merge without this.

lib/instrumentation/index.js
Qard: It wouldn't impact past requires, but we already have that with the timing of the `agent.start()` call, so I don't think that really matters here. I think it's fine that a patch only applies to future require calls. The target use case for the patch registry interface would be to use it immediately after `agent.start()`, before anything else is loaded.

lib/instrumentation/index.js
Qard: I improve the message a bit.

PR #804, opened 2019-01-17
chore(knex): don't patch when stack traces disabled

Qard:
Fixes #506


watson:
@Qard I'm not sure I understand what you mean by "Fixes #506"? That's a PR that does the same thing as this PR

Qard:
Oh, oops. Didn't notice it was a PR in zube. Just saw the content. 🤦‍♂️

PR #808, opened 2019-01-18
Fix typo

posquit0:
### Checklist

- [x] Implement code


watson:
@posquit0 good catch 😄

Issue #806, opened 2019-01-18
Add `koa` module to instrument to support framework information


posquit0:
**Is your feature request related to a problem? Please describe.**
When I use `koa` for http framework, I cannot get the framework metadata. This is because `elastic-apm-node` only instrument `koa-router` module.

**Describe the solution you'd like**
If `elastic-apm-node` instrument `koa` module to only get the framework information, it will be solved.

**Describe alternatives you've considered**
No

**Additional context**
No

watson:
Good idea 👍

PR #810, opened 2019-01-18
Add koa instrumentation code to get framework info

posquit0:
Fixes elastic/apm-agent-nodejs#806 

I implemented to get Koa framework name and version. When I set `DEBUG=trace`, I can see `elastic-apm-node` can gather this information, however, I cannot see the framework information in Kibana UI. I don't know why. Could you give me some idea?

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/test/config.js
+++ b/test/config.js
@@ -63,7 +63,7 @@ optionFixtures.forEach(function (fixture) {
     var number = typeof fixture[2] === 'number'
     var array = Array.isArray(fixture[2])
 
-    test('should be configurable by envrionment variable ELASTIC_APM_' + fixture[1], function (t) {
+    test('should be configurable by environment variable ELASTIC_APM_' + fixture[1], function (t) {
       var agent = Agent()
       var value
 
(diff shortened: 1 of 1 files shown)


lib/instrumentation/modules/koa.js
```diff
@@ -0,0 +1,16 @@
...
+
+var semver = require('semver')
+
+module.exports = function (koa, agent, version, enabled) {
+  if (!enabled) return koa
+
+  if (!agent._conf.frameworkName) agent._conf.frameworkName = 'koa'
+  if (!agent._conf.frameworkVersion) agent._conf.frameworkVersion = version
+
+  if (!semver.satisfies(version, '^2.0.0')) {
```
watson: Since we don't need to do anything with the Koa module it self, there's no reason to have this if-block validating the version. Similarly you can also just remove the `require('semver')` at the top.

posquit0: I did! thanks :)

posquit0:
Should I add `test` and `docs` for this change? Let me know :)

watson:
Released in v2.2.0. Thanks again for contributing ❤️

PR #820, opened 2019-01-22
fix(parsers): move port fix into parser

### LINKED ISSUE #819: Problem validating JSON document against schema - nodejs agent
Hi,

I don't know if this is a bug but I was encountering this error :

```
Problem validating JSON document against schema: I[#] S[#]doesn't validate with \"transaction#\"\n  I[#] S[#/allOf/0] allOf failed\n    I[#/context/request/url/port] S[#/allOf/0/properties/context/properties/request/properties/url/properties/port/type] expected string or null, but got number
```

I did some investigation and changed inside the function getUrlFromRequest in the elastic-apm-node/lib/parsers.js from:

``` javascript
// port
  if (url.port) result.port = Number(url.port)
  else if (req.headers['x-forwarded-port']) result.port = Number(getFirstHeader(req, 'x-forwarded-port'))
  else if (host.port) result.port = Number(host.port)
```
to:

``` javascript
  // port
  if (url.port) result.port = String(url.port)
  else if (req.headers['x-forwarded-port']) result.port = String(getFirstHeader(req, 'x-forwarded-port'))
  else if (host.port) result.port = String(host.port)
```
And now my APM agent is working fine...

Before I dove in the elastic-apm files, I tried to change my port to string and it didn't work.

I'm using:

* Elasticsearch 6.5.4
* Elasticsearch APM 6.5.4
* APM Agent 2.2.1

In my project I'm using typescript and only express...

Qard:
The request context parser is actually used in multiple places, so the port fix should be applied centrally.

Fixes #819

### DIFF
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -132,11 +132,6 @@ Transaction.prototype.toJSON = function () {
     var conf = this._agent._conf
     if (this.req) {
       payload.context.request = parsers.getContextFromRequest(this.req, conf, 'transactions')
-
-      // TODO: Tempoary fix for https://github.com/elastic/apm-agent-nodejs/issues/813
-      if (payload.context.request && payload.context.request.url && payload.context.request.url.port) {
-        payload.context.request.url.port = String(payload.context.request.url.port)
-      }
     }
     if (this.res) {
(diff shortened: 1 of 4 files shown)


Qard:
I'm going to just land this so we can get the fix out right away.

PR #821, opened 2019-01-23
refactor: remove sampled property alias from trace context

watson:
This just makes the code less complex.

### DIFF
--- a/lib/instrumentation/trace-context.js
+++ b/lib/instrumentation/trace-context.js
@@ -94,7 +94,6 @@ function start (sampled = false) {
   // Generate new ids
   randomFillSync(buffer, OFFSETS.traceId, SIZES.ids)
 
-  // Apply reported and requested flags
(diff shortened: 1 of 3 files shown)


lib/instrumentation/trace-context.js
```diff
@@ -107,23 +107,13 @@ const bufferSymbol = Symbol('trace-context-buffer')
 class TraceContext {
   constructor (buffer) {
     this[bufferSymbol] = buffer
+    this.recorded = !!(buffer[OFFSETS.flags] & FLAGS.recorded)
```
Qard: Should we continue using `Object.defineProperty(...)` to keep this immutable?

```suggestion
    Object.defineProperty(this, 'recorded', {
      value: !!(buffer[OFFSETS.flags] & FLAGS.recorded),
      enumerable: true
    })
```

PR #823, opened 2019-01-23
feat: add ability to set custom timestamp for errors

watson:
image

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

PR #824, opened 2019-01-23
feat: add ability to set custom log message for errors

watson:
~~Depends on #823~~

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


watson:
@Qard just look at the last commit only

PR #826, opened 2019-01-24
refactor: simplify API for resuming traces

watson:
Simplify the API for marking a span or transaction as a child of another span or transaction, either local to the process or across processes via the `traceparent` header.

A parent can now be provided either as a Transaction, Span, or TraceContext object, or as a `traceparent` string. And no matter which format the parent is provided in, the parameter name is `childOf`.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/instrumentation/trace-context.js
+++ b/lib/instrumentation/trace-context.js
@@ -2,12 +2,6 @@
 
 const { randomFillSync } = require('random-poly-fill') // TODO: Remove when Node.js 6 is no longer supported
 
-// Avoid circular require dependency
-let GenericSpan = {}
-process.nextTick(function () {
-  GenericSpan = require('./generic-span')
-})
-
 const SIZES = {
   version: 1,
   traceId: 16,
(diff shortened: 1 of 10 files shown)


watson:
Expanded the scope of this PR quite a bit

watson:
This isn't a breaking change as the commit which added support for the `traceparent` config option hasn't been released yet

lib/instrumentation/trace-context.js
```diff
@@ -126,9 +132,12 @@ class TraceContext {
     })
   }
 
-  static startOrResume (traceparent, conf) {
-    return isValidHeader(traceparent)
-      ? resume(traceparent)
+  static startOrResume (childOf, conf) {
+    if (childOf instanceof TraceContext) return childOf.child()
+    if (childOf instanceof GenericSpan) return childOf._context.child()
```
Qard: Could also just check it's an object and has a `_context` property that is `instanceof TraceContext`. That way you don't need the `GenericSpan` circular dependency an the complicated logic surrounding that.

watson: Fixed

PR #829, opened 2019-01-30
feat: extract filters

Qard:
### Checklist

- [x] Implement code
- [x] Add tests


kuisathaverat:
The build was terminated because it spends more than an hour to finish, we have a timeout set of 1 hour. Is that enough? Shall we bump the limit?

kuisathaverat:
jenkins run the tests please

watson:
With the current setup of the TAV tests they often take more than an hour to finish. We had a similar problem on Travis where we changed the cutoff time to 4 hours if I recall correctly (but they never take more than 1:20h max)

kuisathaverat:
I have increased the priority of that task, let's see if I can pick it this week, it should be easy

PR #833, opened 2019-01-31
refactor: extract trace-context

Qard:
This extracts `trace-context` to a separate module.


watson:
@Qard According to npm, the URL for this repo should be https://github.com/elastic/apm-nodejs-trace-context, but I get a 404. Could it be that you've marked it private and only have access your self by any chance?

Qard:
Yep. It should be public now.

PR #842, opened 2019-02-07
fix(express): ensure correct transaction names

### LINKED ISSUE #443: Aggregated transactions on APM dashboard when using express.Router
Similar routes will be aggregated into same group of transactions on the APM dashboard view when using express.Router aka "mini-app"s only when the application is minified.

For example:
catRoutes and dogRoutes have a router.get('/', ...);

app.use('/cats', catRoutes);
app.use('/dogs', dogRoutes);

After requesting /cats/ and /dogs/, the APM dashboard will display "GET /" which includes both transactions with the proper urls displayed in the transaction sample.

I believe the expected behavior should display "GET /cats/" and "GET /dogs" separately.

- OS: Linux
- Node.js version: 8.11.3
- APM Server version: 6.3.1
- Agent version: 1.8.2
- I am calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)

Qard:
Fixes elastic/apm-agent-nodejs#443

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -2,24 +2,52 @@
 
 var symbols = require('../symbols')
 
+function ensureSlash (value) {
+  return value[0] === '/' ? value : '/' + value
+}
+
+function excludeRoot (value) {
+  return value !== '/'
+}
+
+function join (parts) {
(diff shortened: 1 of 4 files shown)


roncohen:
nice @Qard !

PR #850, opened 2019-02-13
Improve koa transaction name

posquit0:
In `koa` application, the transaction name is `${method} unknown route` if the request respond before reaching `koa-router` middleware.

With this change, the transaction name will be set properly. This change doesn't affect to `koa-router`'s instrumentation because it happens before `koa-router`'s instrumentation. The transaction name will be overwrited by `koa-router` if the request reached to `koa-router` middleware.

This change needs `koa` version `>=0.1.0` because of `createContext` method. 

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/koa.js
+++ b/lib/instrumentation/modules/koa.js
@@ -1,10 +1,43 @@
 'use strict'
 
+var semver = require('semver')
+
+var shimmer = require('../shimmer')
+
 module.exports = function (koa, agent, version, enabled) {
   if (!enabled) return koa
 
   if (!agent._conf.frameworkName) agent._conf.frameworkName = 'koa'
   if (!agent._conf.frameworkVersion) agent._conf.frameworkVersion = version
 
+  if (!semver.satisfies(version, '>=0.1.0')) {
+    agent.logger.debug('koa version %s not supported - aborting...', version)
+    return koa
+  }
+
+  agent.logger.debug('shimming koa prototype.createContext function')
+  shimmer.wrap(koa.prototype, 'createContext', function (orig) {
+    return function (req, res) {
+      var context = orig.apply(this, arguments)
+
+      var method = context.method
+      var path = context.path
+
+      if (typeof method !== 'string') {
+        agent.logger.debug('unexpected method type in koa prototype.createContext: %s', typeof method)
+        return context
+      }
+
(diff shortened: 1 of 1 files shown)


watson:
Ah interesting. I didn't know it was common to respond to requests before they were handled by koa-router.

The reason why we don't just always default to `req.path` (no matter if you use koa or something else), is because this could lead to an explosion of different transaction groups as the transaction name is used in the key for making the group unique. Elasticsearch and in turn the UI doesn't deal well with a very high amount of unique transactions. So this could give users a bad experience if they have a lot of unique URL's - e.g. if you have a product or user ID as part of the URL.

With this in mind, I'm very hesitant to making this the default behavior. However, I think it would be a great feature if it's behind a config option that users can enable if they need it. The Java agent already have a config option for this:

https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-use-path-as-transaction-name

This config option would take effect no matter which framework you were using. What do you think about this approach?

posquit0:
Oh, I agree that this change could bother users when the URL path includes some params. I didn't care about the case you said.

It would be great if `apm-agent-nodejs` also support the flag `use_path_as_transaction_name` like JAVA agent. Some middlewares respond early before reaching the router middleware. As you said, this feature could be used in other frameworks as well!

Could you take a time to support this feature?

watson:
I don't have the bandwidth to look into this right now. But if you are up for it, you could modify your PR to include a new `usePathAsTransactionName` config option and change it to apply for all http requests. Otherwise, we'll be happy to add it to our roadmap and look into it when we get the time :)

ICYMI: You can always call `agent.setTransactionName(req.method + ' ' + req.path)` your self in your middleware if you respond early. But I agree it would be nice to have this as a feature of the agent so you don't have to do it "manually".

posquit0:
Okay! I'll try to dive into this feature when I have time. :)
Thanks!

watson:
FYI: I've created an issue to add this before mentioned config option: #904

kuisathaverat:
jenkins run the tests please

watson:
If I'm not mistaken this will be solved by the new config option implemented in #907, so I'll close this now

watson:
The solution for this added in #907 have now been published to npm as part of the new version of the agent: v2.8.0

PR #861, opened 2019-02-20
[APM-CI] some exceptions are not convert to String when you call error step error step

kuisathaverat:
* Convert the exception always to string before to call error

### Checklist

- [x] Implement code
- [NA] Add tests
- [NA] Update documentation


Qard:
Some formatting errors in the commit message:

- It's too long
- It should be be structured like `type(subsystem): message`

PR #866, opened 2019-02-21
Add nodemailer instrumentation

kschingiz:
This PR adds instrumentation for nodemailer so it let's you track sending emails in transactions and track exceptions during sending a email. 
 
### Checklist

- [x] Add tests
- [x] Implement code
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/nodemailer.js
+++ b/lib/instrumentation/modules/nodemailer.js
@@ -1,0 +1,69 @@
+var shimmer = require('../shimmer')
+
+module.exports = function (nodemailer, agent, version, enabled) {
+  if (!enabled) return nodemailer
+  agent.logger.debug('shimming nodemailer.createTransport')
+  shimmer.wrap(nodemailer, 'createTransport', function (orig) {
+    return function () {
+      var transport = orig.apply(this, arguments)
+
+      shimTransport(transport)
+      return transport
+    }
+  })
+
+  function shimTransport (transport) {
+    agent.logger.debug('shimming transport.sendMail')
+
+    shimmer.wrap(transport, 'sendMail', function (orig) {
+      return function () {
+        const transaction = agent.currentTransaction ||
+                                        agent.startTransaction('Nodemailer', 'Nodemailer')
+
+        const span = transaction.startSpan('Send Email', 'nodemailer.sendMail')
+        if (!span) {
+          return orig.apply(this, arguments)
+        }
+
+        function endTransaction (transaction) {
+          if (transaction.type === 'Nodemailer') {
+            transaction.end()
+          }
+        }
+
+        // callback is provided
+        if (typeof arguments[1] === 'function') {
+          const cb = arguments[1]
+
+          return orig.call(this, arguments[0], function (err, result) {
+            if (err) {
+              agent.captureError(err)
+            }
+            span.end()
+            endTransaction(transaction)
+
+            cb(err, result)
+          })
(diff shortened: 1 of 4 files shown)


lib/instrumentation/modules/nodemailer.js
```diff
@@ -0,0 +1,69 @@
...
+
+  function shimTransport (transport) {
+    agent.logger.debug('shimming transport.sendMail')
+
+    shimmer.wrap(transport, 'sendMail', function (orig) {
+      return function () {
+        const transaction = agent.currentTransaction ||
+                                        agent.startTransaction('Nodemailer', 'Nodemailer')
+
+        const span = transaction.startSpan('Send Email', 'nodemailer.sendMail')
```
Qard: I don't think `agent.startTransaction(...)` should be used here. Transactions are meant to _contain_ a bunch of logically related things, like the async activity within HTTP requests or background jobs.

Rather than calling `agent.startTransaction(...)` and then `transaction.startSpan(...)`, it would be better to just use `agent.startSpan(...)`.

lib/instrumentation/modules/nodemailer.js
```diff
@@ -0,0 +1,69 @@
...
+            resolve(result)
+          }, (err) => {
+            agent.captureError(err)
+
+            span.end()
+            endTransaction(transaction)
+
+            reject(err)
+          })
+        })
```
Qard: Rather than wrapping this in a new promise and having all that code duplication between success and failure handlers and the callback form, I'd just do something like this:

```js
function after (err) {
  if (err) agent.captureError(err)
  span.end()
}

// Handle cleanup for callback-form
if (typeof arguments[1] === 'function') {
  shimmer.wrap(arguments, 1, cb => {
    return function (err) {
      after(err)
      return cb.apply(this, arguments)
    }
  })
}

var result = orig.apply(this, arguments)

// Handle cleanup for promise-form
if (result && typeof result.then === 'function') {
  result.then(
    () => after(),
    after
  )
}

return result
```

lib/instrumentation/modules/nodemailer.js
kschingiz: I just wanted to catch every email sent to clients, if it's happening not under built-in transactions, then I auto create new one. But I think you are right, will fix it today.

lib/instrumentation/modules/nodemailer.js
kschingiz: Thanks! I will update today.

> I'm currently working on an interface to allow custom instrumentation patches to be sideloaded easily

That's great! I am developing also elastic apm for meteor and experienced issue when I was not able to easily integrate my agent with yours, it forced me to hack your agent as well:
https://github.com/kschingiz/meteor-elastic-apm/blob/master/meteor-elastic-apm.js#L18

When are you going to merge PR #803? I can block my PR and wait for new release.

Qard: It should be released very soon, I think. The PR is just waiting for review at the moment. If it all looks good, we might be able to land it early next week. :)

kschingiz: @Qard Great! can we then close my PR? If you are not planning to add emailers, it should be as a separate module, I think

kuisathaverat:
jenkins run the tests please

Qard:
The patch registry has landed and will be released soon. Closing this now.

PR #882, opened 2019-02-28
test: be more relaxed with minimum duration

watson:


Qard:
Weird. Looks like this is passing on Travis, but on Github it still appears as in-progress. 🤔

watson:
I guess the hook just failed to fire. I'll merge it.

PR #886, opened 2019-02-28
chore(package): bump restify dev dependency to ^8.0.0

watson:
Changelog: https://github.com/restify/node-restify/blob/master/CHANGELOG.md#800-2019-02-20

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "pg": "^7.5.0",
     "redis": "^2.8.0",
     "request": "^2.88.0",
-    "restify": "^7.2.1",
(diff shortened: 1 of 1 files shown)


watson:
I just realized that this new version drops support for Node.js 6, so maybe we shouldn't have this as the default installed version as this might make v6 tests fail in the future without warning. So I'm thinking that we instead update `.tav.yml` to only test this version on Node.js 8+ but still keep the dev dependency for the old version in package.json.

I'll update the PR

watson:
I'll close this PR and create another

PR #894, opened 2019-03-04
docs: add missing ToC links to README.md

watson:


watson:
The Jenkins failures can be ignored... I think it's just a bug with Jenkins

PR #895, opened 2019-03-04
chore(package): bump elastic-apm-http-client to ^7.1.1

### LINKED ISSUE #888: Ensure request.headers are always strings (or array of strings)
We're considering making this a JSON Schema requirement. So just in case, we need to make sure that headers are always strings or array of strings.

Even if this doesn't make it into the JSON Schema, it's probably a good thing to ensure nonetheless.

watson:
This version includes a fix that ensures that all reported http headers
are always strings or array of strings.

Fixes #888

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "console-log-level": "^1.4.0",
     "cookie": "^0.3.1",
     "core-util-is": "^1.0.2",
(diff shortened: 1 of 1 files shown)


watson:
@Qard Since you approved the PR in the http client I assume you're ok with me just merging this. Afterwards I'll release a new version.

Issue #878, opened 2019-02-27
Consider custom router ?


cuiweiqiang:
**Is your feature request related to a problem? Please describe.**

I'm using egg.js to develop project. In this project, I want to use apm to monitor performance which from fronted to backend service.

In this project apm can not naming transaction with pathname, because of apm use http module, but egg.js use a module forked from `koa-router`

**Describe the solution you'd like**

Egg.js fork `koa-router` and rename [egg-router](https://github.com/eggjs/egg-router), they add some additional features, but not change the property. So apm can use `koa-router` module to parse request in egg.js in theory. I notice that, it do judge with package name in `Instrumentation`. So can add a parameter like `customRouterMap`, for exmaple:
```js
  customRouterMap: {
    "@eggjs/router": {
        name: "koa-router",
        version: "7.4.0"
    }
  }
```
put it in start func, then apm can use `koa-router` module to naming transaction.

There has a simple implementation in [my fork](https://github.com/cuiweiqiang/apm-agent-nodejs/tree/customRouterMap).

Please checkout whether it is feasible .

watson:
Interesting use-case. Thanks for spending time thinking about this coming up with a possible solution. This seems like a general solution for "when module A is being required, patch it with the code for module B" - i.e. not limited to routers only.

While this does seem like a very specific edge case, I do struggle to find an alternative. So maybe we should add support for this 🤔 

It of course comes with no guarantees as `egg-router` might change down the line in a way that means the code for patching `koa-router` no longer works on it. And without any tests in place to detect this, you're flying blind.

Let me just give it a think...

(for my own later reference, here's a link that shows the diff for just the patch: https://github.com/elastic/apm-agent-nodejs/compare/fd507f6838391467fa6076335c4c2379914d1419...cuiweiqiang:customRouterMap)

Qard:
It could change, but even _more_ likely is the version numbers not matching, so our use of version numbers to decide whether or not to apply a given patch would no longer work. I'm not sure there is a way to do this with any reasonable degree of stability. With the patch registry incoming, I think it's better to just have users add their own patches for more obscure things. In this case, they could just do this:

```js
agent.registerPatch('@eggjs/router', require('elastic-apm-node/lib/instrumentation/modules/koa-router'))
```

watson:
Good point about the version number @Qard - I didn't think about that.

[...]

PR #903, opened 2019-03-05
fix: add parseurl library to support node frameworks other than Express to get the right pathname

yuxizhe:
Add parseurl library to support node frameworks other than Express to get the right pathname.

 If using koa without koa-router, for example in koa React server side rending project. We can get right path now, instead of "unknown route".

elastic/apm-agent-nodejs#878 elastic/apm-agent-nodejs#850

### DIFF
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -1,5 +1,6 @@
 'use strict'
 
+var parseUrl = require('parseurl')
 var symbols = require('../symbols')
 
 function ensureSlash (value) {
@@ -15,8 +16,12 @@ function join (parts) {
 }
 
 // This works for both express AND restify
-function routePath (route) {
-  if (!route) return ''
+function routePath (req) {
+  var route = req.route
+  // support other node framework
+  if (!route) {
+    return parseUrl(req).pathname
+  }
   return route.path || (route.regexp && route.regexp.source) || ''
 }
 
@@ -35,7 +40,7 @@ function getPathFromRequest (req, useBase) {
   }
 
   var path = getStackPath(req)
-  var route = routePath(req.route)
+  var route = routePath(req)
 
   if (route) {
     return path ? join([ path, route ]) : route
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "measured-reporting": "^1.39.1",
     "object-filter-sequence": "^1.0.0",
     "original-url": "^1.2.2",
+    "parseurl": "^1.3.2",
     "read-pkg-up": "^4.0.0",
     "redact-secrets": "^1.0.0",
     "relative-microtime": "^2.0.0",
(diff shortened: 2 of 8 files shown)


watson:
Will `req._parsedUrl.pathname` always be the exact path requested? E.g. if someone makes a GET request to `http://example.com/users/42?foo=bar`, then its value would be `/users/42`?

yuxizhe:
> Will `req._parsedUrl.pathname` always be the exact path requested? E.g. if someone makes a GET request to `http://example.com/users/42?foo=bar`, then its value would be `/users/42`?

yes, in library [parseurl](https://github.com/pillarjs/parseurl) 

https://github.com/pillarjs/parseurl/blob/6b72e98683e1030cd21045e31dc379f953e1e618/test/test.js#L24

and req._parsedUrl is created by the parseurl library which is used by Express' Router when handling an incoming request. [stackoverflow](https://stackoverflow.com/questions/37285119/what-javascript-library-sets-the-parsedurl-property-on-a-request-object)

So I add parseurl to get the right path name for node framworks other than Express.

yuxizhe:
#878 
#850

watson:
This would unfortunately interfere with how 404 routes would be reported as those would no longer be grouped together under "unknown route" but instead each be reported with their unique path. 

Even without 404 requests, this could unfortunately easily lead to an explosion of unique transaction groups in case some part of the path was dynamic - e.g. if it contained a user ID for instance. Normally we group all transactions to the same route together even if they contain a dynamic element. With this chance, those routes would each get their own unique group.

But I do see the value of this in cases where the URL's doesn't contain any dynamic content and where you might not have that many random 404 requests.

Instead I feel this is something that should be done on a more generic level, with an optional config option called `userPathAsTransactionName` like the one we have [in the Java agent](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-use-path-as-transaction-name). By setting this to `true` it would use the path instead of the router information.

That being said, I also just wanted to mention that you can always manually set the name that you want the transaction to have without having to change the behavior of this agent. All you need to do is to call [`apm.setTransactionName()`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-set-transaction-name) when the request is received by your server. You can do this already now.

watson:
I've created an issue to add this before mentioned config option: #904

yuxizhe:
@watson Thanks for your replay!  May I have a chance to finish or participate in this issue? #904

watson:
Absolutely - all help is appreciated ❤️ And I'll be happy to discuss the best approach before you get started and guide you if needed. We should also check with @posquit0 as he also mentioned that he might get some time to look into it, so I'm not sure if he already made some progress.

posquit0:
I don't have any progress for this problem yet. I'll be happy if this problem can be solved by @yuxizhe ! :)

PR #905, opened 2019-03-06
fix no active transaction found for module tedious

w0dm4n:
Error when using the module tedious for MSSQL driver 

> no active transaction found - cannot build new span

### DIFF
--- a/lib/instrumentation/modules/tedious.js
+++ b/lib/instrumentation/modules/tedious.js
@@ -38,7 +38,9 @@ module.exports = function (tedious, agent, version, enabled) {
       }
 
       makeRequest (request) {
-        const span = agent.startSpan(null, 'db.mssql.query')
+        var trans = agent.startTransaction("mssql.driver", "function")
+        this.currentTransaction = trans
+	const span = agent.startSpan(null, 'db.mssql.query')
         if (!span) {
           return super.makeRequest(...arguments)
         }


watson:
Thanks for submitting the PR. I'm afraid however that this will not work as intended.

A transaction will be automatically created when ever the agent discovers an incoming HTTP request, so each of the database instrumentations shouldn't create transactions them selfs. If the MSSQL query is being performed during the scope of this HTTP request, you should not get the "no active transaction found" error message.

That being said, there are several scenarios when you might get this error message when a MSSQL query is being performed:

- The query is being performed outside the scope of an HTTP request. Maybe your app doesn't expose an HTTP server, or maybe it happens as part of a background job. These are perfect valid use-cases, but the way you should normally deal with this is to manually start a new transaction in your own codebase every time a new logical piece of work is being started. For more information about this see the [Custom Transactions](https://www.elastic.co/guide/en/apm/agent/nodejs/current/custom-transactions.html) documentation.
- In some cases, even though there should be an active transaction, your application or one of your dependencies might interfere with the agents ability to correctly identify the active transaction. This is just an unfortunate nature of how Node.js handles async execution and is something we're constantly trying to circumvent. But it's a bit of a cat and mouse game as new modules are released all the time that might cause such interference.

In your case, is the MSSQL query you're trying to instrument performed in the context of an incoming HTTP request to your application?

w0dm4n:
Hello, i don't have any HTTP server, i just tried to use the Tedious module by starting the agent from the entry point of my code, it seems like it was able to hook correctly the Tedious modules (passing by 
 packages mssql2) but then any span couldn't just do any transaction, that's why i did this passthrough for creating new transaction.

watson:
Would it be possible in your use-case to [start and end transactions manually](https://www.elastic.co/guide/en/apm/agent/nodejs/current/custom-transactions.html) in your code base? If you do so, the tedious spans will automatically be created without you having to do anything else then just starting the transaction

kuisathaverat:
jenkins run the tests please

watson:
Closing this PR as I don't think it's the right approach. If you disagree, feel free to leave a comment :)

I'd be happy to help you create [custom transactions](https://www.elastic.co/guide/en/apm/agent/nodejs/current/custom-transactions.html). You can always open a topic in [our Discuss forum](https://discuss.elastic.co/c/apm) and we can take it from there :)

PR #907, opened 2019-03-06
feat(config): add usePathAsTransactionName config option

### LINKED ISSUE #904: Add userPathAsTransactionName config option
This config option should behave the same as the [`use_path_as_transaction_name`](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-use-path-as-transaction-name) config option of the Java agent.

It should be off by default, but when set to `true` it should use the URL path as the transaction name instead of the router information, e.g. `GET /foo/42`.

This will solve the issues brought up in elastic/apm-agent-nodejs#850 and elastic/apm-agent-nodejs#903

yuxizhe:
Closes elastic/apm-agent-nodejs#904 
Add usePathAsTransactionName config option.
elastic/apm-agent-nodejs#903 
elastic/apm-agent-nodejs#850 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -1,5 +1,11 @@
 'use strict'
 
+var parseUrl
+try {
+  parseUrl = require('parseurl')
+} catch (e) {
+  parseUrl = require('url').parse
+}
 var symbols = require('../symbols')
 
 function ensureSlash (value) {
@@ -28,7 +34,7 @@ function getStackPath (req) {
 
 // This function is also able to extract the path from a Restify request as
 // it's storing the route name on req.route.path as well
-function getPathFromRequest (req, useBase) {
+function getPathFromRequest (req, useBase, usePathAsTransactionName) {
   // Static serving route
   if (req[symbols.staticFile]) {
     return 'static file'
@@ -43,6 +49,12 @@ function getPathFromRequest (req, useBase) {
 
   if (useBase) {
     return path
+  }
+
+  // Enable usePathAsTransactionName config
+  if (usePathAsTransactionName) {
+    const parsed = parseUrl(req)
+    return parsed && parsed.pathname
   }
 }
 
--- a/test/config.js
+++ b/test/config.js
@@ -57,7 +57,8 @@ var optionFixtures = [
   ['kubernetesNodeName', 'KUBERNETES_NODE_NAME'],
   ['kubernetesNamespace', 'KUBERNETES_NAMESPACE'],
   ['kubernetesPodName', 'KUBERNETES_POD_NAME'],
-  ['kubernetesPodUID', 'KUBERNETES_POD_UID']
+  ['kubernetesPodUID', 'KUBERNETES_POD_UID'],
+  ['usePathAsTransactionName', 'USE_PATH_AS_TRANSACTION_NAME', false]
 ]
 
 var falsyValues = [false, 'false']
@@ -477,6 +478,43 @@ captureBodyTests.forEach(function (captureBodyTest) {
     req.body = 'test'
 
     agent.captureError(new Error('wat'), { request: req })
+
+    var trans = agent.startTransaction()
+    trans.req = req
+    trans.end()
+  })
+})
+
+var usePathAsTransactionNameTests = [
+  { value: true, url: '/foo/bar?baz=2', transactionName: 'GET /foo/bar' },
+  { value: false, url: '/foo/bar?baz=2', transactionName: 'GET unknown route' }
+]
+
+usePathAsTransactionNameTests.forEach(function (usePathAsTransactionNameTest) {
+  test('usePathAsTransactionName => ' + usePathAsTransactionNameTest.value, function (t) {
+    t.plan(2)
+
+    var agent = Agent()
+    agent.start({
+      serviceName: 'test',
+      captureExceptions: false,
+      usePathAsTransactionName: usePathAsTransactionNameTest.value
+    })
+
+    var sendTransaction = agent._transport.sendTransaction
+    agent._transport.sendTransaction = function (trans, cb) {
+      t.ok(trans)
+      t.equal(trans.name, usePathAsTransactionNameTest.transactionName)
+      if (cb) process.nextTick(cb)
+    }
+    t.on('end', function () {
+      agent._transport.sendTransaction = sendTransaction
+    })
+
+    var req = new IncomingMessage()
+    req.socket = { remoteAddress: '127.0.0.1' }
+    req.url = usePathAsTransactionNameTest.url
+    req.method = 'GET'
 
     var trans = agent.startTransaction()
     trans.req = req
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -382,6 +382,15 @@ A capture location stack trace is never generated for uncaught exceptions.
 
 Set this option to `false` to disable capture of stack traces for measured spans during instrumentation.
 
+[[use-path-as-transaction-name]]
+==== `usePathAsTransactionName`
+
+* *Type:* Boolean
+* *Default:* `false`
+* *Env:* `ELASTIC_APM_USE_PATH_AS_TRANSACTION_NAME`
+
+Set this option to `true` to use the URL path as the transaction name if no other route could be determined. If the agent do not support your router, you can set this option to `true` to use specific URL path as the transaction name instead of `GET unknown route`.
+
 [[source-context-error]]
 ==== `sourceLinesErrorAppFrames` + `sourceLinesErrorLibraryFrames`
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -65,6 +65,7 @@ var DEFAULTS = {
   kubernetesPodUID: undefined,
   captureHeaders: true,
   metricsInterval: '30s',
+  usePathAsTransactionName: false,
   addPatch: undefined
 }
 
@@ -108,6 +109,7 @@ var ENV_TABLE = {
   kubernetesPodUID: 'ELASTIC_APM_KUBERNETES_POD_UID',
   captureHeaders: 'ELASTIC_APM_CAPTURE_HEADERS',
   metricsInterval: 'ELASTIC_APM_METRICS_INTERVAL',
+  usePathAsTransactionName: 'ELASTIC_APM_USE_PATH_AS_TRANSACTION_NAME',
   addPatch: 'ELASTIC_APM_ADD_PATCH'
 }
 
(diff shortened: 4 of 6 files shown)


lib/instrumentation/express-utils.js
```diff
@@ -28,12 +29,17 @@ function getStackPath (req) {
...
-function getPathFromRequest (req, useBase) {
+function getPathFromRequest (req, useBase, usePathAsTransactionName) {
   // Static serving route
   if (req[symbols.staticFile]) {
     return 'static file'
   }
 
+  // Enable usePathAsTransactionName config
+  if (usePathAsTransactionName) {
+    return req._parsedUrl ? req._parsedUrl.pathname : parseUrl(req).pathname
```
watson: If I understand the idea behind the `parseurl` module correctly, the idea is that you can just call it again with the same request and would get the same parsed object back. So you shouldn't need to access `req._parsedUrl` your self, as that's what `parseUrl(req)` does for you.

Also, `parseUrl(req)` can potentially return `undefined`, so we would need to guard against that.

This should take care of both those things:

```suggestion
    const parsed = parseUrl(req)
    return parsed && parsed.pathname
```

watson: Come to think of it, I'm not really sure if it's worth it using `parseurl` here.

Cons:

- It's an extra dependency
- It mutates the request object
- It only gives the user a benefit if they them selfs (or their framework) already uses the parseurl module. If not, this module doesn't give us any speed improvements

Pros:

- If the user or their framework uses the parseurl module, this will result in one allocation less than normal

@Qard any thoughts on this?

lib/instrumentation/express-utils.js
```diff
@@ -28,12 +29,17 @@ function getStackPath (req) {
...
 // it's storing the route name on req.route.path as well
-function getPathFromRequest (req, useBase) {
+function getPathFromRequest (req, useBase, usePathAsTransactionName) {
   // Static serving route
   if (req[symbols.staticFile]) {
     return 'static file'
   }
 
+  // Enable usePathAsTransactionName config
+  if (usePathAsTransactionName) {
```
watson: To do exactly what the docs describe, this should probably be moved up above the first if-sentence, so that even static routes will have their name set by the pathname

kuisathaverat:
jenkins run the tests please

lib/instrumentation/express-utils.js
Qard: We could just do a try/catch require of it. If the user has it in their code, we'll find it and we can check existence of the function to decide if we want to apply the logic. 🤔

watson: Good point @Qard. Yes I think that's the way forward. I'll make a new review

lib/instrumentation/express-utils.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+var parseUrl = require('parseurl')
```
watson: As @Qard suggested in a previous commit, let's drop the dependency to `parseurl` and instead only use it if the user already is using it in their application - and fall back to the regular `url.parse` function in core if not:

```suggestion
var parseUrl
try {
  parseUrl = require('parseurl')
} catch (e) {
  parseUrl = require('url').parse
}
```

This way we get the best of both worlds: If the user isn't already using `parseurl`, we'll not pollute the request object. If they are, they'll probably expect it to happen anyway.

yuxizhe: so cool 👍

watson: I think that our [dependency-check](https://github.com/maxogden/dependency-check) linter will complain if it sees the `parseurl` module being required and not being listed in our package.json dependencies. So we need to ignore that module.

You do that by updating this line:
https://github.com/elastic/apm-agent-nodejs/blob/5a8e06ded78953755b492be5b6f64369be85fe4b/package.json#L16

I think you just need to add `-i parseurl` to the first command, like so:

```
dependency-check . -i async_hooks -i parseurl
```

watson: Yep, it didn't like it: https://travis-ci.org/elastic/apm-agent-nodejs/jobs/506371767#L527

yuxizhe: fixed :) , I learn a lot from this code review

lib/instrumentation/express-utils.js
```diff
@@ -28,7 +34,13 @@ function getStackPath (req) {
 
 // This function is also able to extract the path from a Restify request as
 // it's storing the route name on req.route.path as well
-function getPathFromRequest (req, useBase) {
+function getPathFromRequest (req, useBase, usePathAsTransactionName) {
+  // Enable usePathAsTransactionName config
+  if (usePathAsTransactionName) {
```
Qard: You will also need to check that `parseUrl` exists here.

watson: No, it will always exist as we fall back to the core module above

Qard: Oh, my bad. Misread. 😅

yuxizhe:
@watson  got it, I will add tests soon

test/config.js
```diff
@@ -482,6 +483,40 @@ captureBodyTests.forEach(function (captureBodyTest) {
...
+    })
+
+    var req = new IncomingMessage()
+    req.url = usePathAsTransactionNameTest.url
+    req.method = 'GET'
+
+    var trans = agent.startTransaction()
+    trans.req = req
+    trans.end()
+})
```
watson: Seems like you're missing an ending `}` here:

```suggestion
  })
}
```

yuxizhe: oh, my bad 
fixed :)

test/config.js
```diff
@@ -482,6 +483,41 @@ captureBodyTests.forEach(function (captureBodyTest) {
...
+    })
+
+    var req = new IncomingMessage()
+    req.url = usePathAsTransactionNameTest.url
+    req.method = 'GET'
+
+    var trans = agent.startTransaction()
+    trans.req = req
+    trans.end()
+ })
```
watson: I think our linter will complain about the indentation here. I should be two spaces :)

yuxizhe: they all stuck here, maybe something wrong in my code?
image

watson: Interesting 🤔 I'll take a look

watson: It was because `IncomingMessage` expected to have a `socket` property. When it didn't have that it threw an exception. But that exception was swallowed because `captureExceptions` wasn't set to `false`. After I did that I could see what the problem was:

```
/Users/watson/code/node_modules/elastic-apm-node/lib/parsers.js:99
      remote_address: req.socket.remoteAddress,
                                 ^

TypeError: Cannot read property 'remoteAddress' of undefined
    at Object.exports.getContextFromRequest (/Users/watson/code/node_modules/elastic-apm-node/lib/parsers.js:99:34)
    at Transaction.toJSON (/Users/watson/code/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:137:41)
    at Transaction._encode (/Users/watson/code/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:153:15)
    at Instrumentation.addEndedTransaction (/Users/watson/code/node_modules/elastic-apm-node/lib/instrumentation/index.js:112:65)
    at Instrumentation.wrappedAddEndedTransaction [as addEndedTransaction] (/Users/watson/code/node_modules/elastic-apm-node/lib/instrumentation/async-hooks.js:53:34)
    at Transaction.end (/Users/watson/code/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:217:32)
    at Test.<anonymous> (/Users/watson/code/node_modules/elastic-apm-node/test/config.js:518:11)
    at Test.bound [as _cb] (/Users/watson/code/node_modules/elastic-apm-node/node_modules/tape/lib/test.js:76:32)
    at Test.run (/Users/watson/code/node_modules/elastic-apm-node/node_modules/tape/lib/test.js:95:10)
    at Test.bound [as run] (/Users/watson/code/node_modules/elastic-apm-node/node_modules/tape/lib/test.js:76:32)
    at Immediate.next (/Users/watson/code/node_modules/elastic-apm-node/node_modules/tape/lib/results.js:71:15)
    at processImmediate (timers.js:637:19)
```

I took the liberty to push a fix to your branch: https://github.com/elastic/apm-agent-nodejs/pull/907/commits/4d3e23c8f01e118b310af6dfe54224e0daa4aecb

yuxizhe: I see, thanks for your help 👍

lib/instrumentation/express-utils.js
```diff
@@ -28,7 +34,13 @@ function getStackPath (req) {
...
+    const parsed = parseUrl(req)
+    return parsed && parsed.pathname
+  }
```
Qard: Do we want this at the start rather than at the end as a fallback when no other viable route name is determined?

watson: That was my idea yes. I did consider what you are suggesting, but the config option is called `usePathAsTransactionName`, so I think users would expect that if they set it to `true` that it would overrule everything.

watson: 🤔 on the other hand, now that I re-read [the docs](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-use-path-as-transaction-name) for the same config option in Java, it does seem like they use it as a fallback. Is that correctly understood @felixbarny?

yuxizhe: docs said `If your URLs contain path parameters like /user/$userId, you should be very careful when enabling this flag, as it can lead to an explosion of transaction groups. `  
But if its a fallback when no other viable route name is determined, all ` /user/$userId` will be in one group, its not what the docs say

felixbarny: In the Java agent, this option is only relevant if no other route could be determined. The default transaction name would then be just `GET` if this option is disabled. If it's enabled, it would then use the URL path, for example, `/users/1`, `/users/2` etc. That's why there is also [`url_groups`](https://www.elastic.co/guide/en/apm/agent/java/current/config-http.html#config-url-groups). If you define the pattern `/users/*`, the URLs will be grouped so that there won't be an explosion of transaction groups.

watson: @yuxizhe no worries. You implemented it as we had agreed. And it's not wrong per se. But we just realized that the Java agent actually does it slightly different. So now we're just discussing if we should update this code (and docs) to do what the Java agent does.

yuxizhe: I put it at the end as a fallback and update docs. I think we need url_groups config too.

watson: Thanks for updating. I talked to @felixbarny today and we discussed that we at some point should probably replace this config option for one that can be set to both works as a fallback and as an overwrite - both use-cases are valid. But for now, your latest change is fine. At least this means that we align with the Java agent 😃 

At some point, we'll probably implement `urlGroups` as well, but I don't see that being a showstopper for this PR 👍

watson:
jenkins run the tests please

watson:
This feature has now been published to npm as version 2.8.0

PR #912, opened 2019-03-07
fix(metrics): use noop logger in metrics reporter

### LINKED ISSUE #899: Metrics code can be too chatty
Sometimes the metrics code will log the following to STDOUT (or maybe it's STDERR??):

```
{"name":"Reporter","hostname":"watson.localdomain","pid":33099,"level":50,"msg":"Failed to send metrics to signal fx Error: value is not defined\n    at Object.of (/Users/watson/code/node_modules/elastic-apm-node/node_modules/optional-js/index.js:9:19)\n    at MetricsReporter._reportMetricsWithInterval (/Users/watson/code/node_modules/elastic-apm-node/node_modules/measured-reporting/lib/reporters/Reporter.js:177:16)\n    at Immediate.setImmediate (/Users/watson/code/node_modules/elastic-apm-node/node_modules/measured-reporting/lib/reporters/Reporter.js:143:14)\n    at Immediate.elasticAPMCallbackWrapper [as _onImmediate] (/Users/watson/code/node_modules/elastic-apm-node/lib/instrumentation/index.js:206:27)\n    at processImmediate (timers.js:637:19)","time":"2019-03-05T09:48:49.129Z","v":0}
```

This can easily be reproduced by running `node test/config.js`.

vigneshshanmugam:
+ fix #899 

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/metrics/index.js
+++ b/lib/metrics/index.js
@@ -5,6 +5,15 @@ const MetricsRegistry = require('./registry')
 const registrySymbol = Symbol('metrics-registry')
 const agentSymbol = Symbol('metrics-agent')
 
+class NoopLogger {
+  debug () { }
+  error () { }
+  fatal () { }
+  info () { }
+  trace () { }
+  warn () { }
+}
+
 class Metrics {
   constructor (agent) {
     this[agentSymbol] = agent
@@ -17,7 +26,8 @@ class Metrics {
       reporterOptions: {
         defaultReportingIntervalInSeconds: metricsInterval,
         enabled: metricsInterval !== 0,
-        unrefTimers: true
+        unrefTimers: true,
+        logger: new NoopLogger()
       }
     })
   }
--- a/test/config.js
+++ b/test/config.js
@@ -19,6 +19,8 @@ var Agent = require('./_agent')
 var config = require('../lib/config')
 var Instrumentation = require('../lib/instrumentation')
 
+process.env.ELASTIC_APM_METRICS_INTERVAL = '0'
+
 var optionFixtures = [
   ['serviceName', 'SERVICE_NAME', 'elastic-apm-node'],
   ['secretToken', 'SECRET_TOKEN'],


Qard:
Just a note: this doesn't stop bunyan from being installed, so we still need https://github.com/yaorg/node-measured/pull/61 to land before the agent will install properly on Windows.

watson:
I assume this is somewhat related to #899?

Qard:
Yeah, by default measured logs with bunyan. We should probably be using our own logger, though we don’t have control over what level things are logged at in measured and it may not match what we want...

watson:
It's normally an anti-pattern for modules like this to log to stdout/stderr IMO. I'd prefer if it didn't at all. If we can't get this changed, I'm considering if we should fork it

Qard:
I have commit access to it, so we can make changes as-needed. I think the built-in logging makes sense--not with bunyan though...

We can just change the logger to something that matches our expectations. It's easy to give it a logger that logs nothing. I agree that it should _probably_ be silent by default though.

watson:
My bad. I didn't know that the default usage of measured-reporting was to log to stdout. In that case depending on a logger like pino is obviously not an anti-pattern. But it still would be nice if we didn't have to add it to the dependency tree 🤷‍♂️

Qard:
Yeah. I can suggest pulling the log reporter out to another module and having the core module just export an abstract interface to create a concrete implementation on top of. 🤔

vigneshshanmugam:
I do agree that we cannot control the external deps (bunyan in this case). Can we do this instead to prevent it from getting installed? 
```json
"bunyan": "file:../empty-file.js"
``` 

We include this in our `package.json` and let npm link do the magic. This is not a perfect solution, But we can avoid it from getting installed. We used to follow some similar approach by using yarn resolutions https://yarnpkg.com/lang/en/docs/selective-version-resolutions/.

watson:
@vigneshshanmugam I'd prefer if we didn't have to resort to hacks like that.

I'm leaning more and more towards dropping this module as a dependency all together as it sounds like we're actually using it wrong. Instead we should probably just fork it - unless of course we can change the scope of the module to not be focused on logging to STDOUT if the author is ok with it

Qard:
I'm a collaborator on measured, so we shouldn't have any trouble making whatever changes are needed. Just need to decide what they are. I have a PR open at the moment, switching it to use pino here: https://github.com/yaorg/node-measured/pull/61

vigneshshanmugam:
@StephanBenning But I think even switching to pino wont help us with logging it to stdout right? can we pass a custom stream to the logger in the tests and avoid using the `result.stdout` for validation?

watson:
@vigneshshanmugam wrong Stephen ;)

But you're right, switching to pino will not change that it's logging. But it will fix the compliation issue, which is the most pressing one. But you're right that we don't want it to log either. A quick fix could be to pass in a no-op logger as an argument. But the right fix would be to refactor the module to not log to stdout by default at all, and hence not have a dependency either pino or bunyan. @Qard and I just had a discussion about it and that's our end goal

vigneshshanmugam:
Oops, fixed my comment. Sure sounds good to me. Do you want me to make changes in this PR as well for passing a noop logger in tests? Or we wait for the proper fix by not logging to stdout in the other module?

watson:
I'll leave that up to @Qard

Qard:
I would just do the no-op logger thing here, that way this can be fixed parallel to the bunyan install thing rather than blocking on it.

vigneshshanmugam:
Sure will do the changes.

Qard:
Looks like it's still using the agent logger rather than a null logger. Were you going to change that?

vigneshshanmugam:
@Qard Do we need it still? I was thinking disabling metrics would as well disable the logging.

Qard:
I think it would still potentially log when users use it, so we should still get a null logger in there.

watson:
@Qard I think we should just change this PR to use a null-logger instead :)

vigneshshanmugam:
@Qard @watson replaced the logger with noop logger, Please check now.

test/_agent.js
```diff
@@ -8,13 +8,35 @@ var Filters = require('object-filter-sequence')
...
+  info () { return noopFn }
+  fatal () { return noopFn }
+  error () { return noopFn }
+}
+
+class AgentWithNoopLogger extends Agent {
+  constructor () {
+    super()
+    this.defaultLogger = new NoopLogger()
+    Object.defineProperty(this, 'logger', {
```
vigneshshanmugam: the reason for doing this is because, even if we do `agent._config({ logger: new NoopLogger()} ), It later gets changed in individual tests with original logger which does not  seem to be an ideal behaviour.

Qard:
Oh, actually, we need the MetricsReporter to be given the null logger. This just stops the log output when running the test, but we don't want it for users of the agent either.

watson:
This was just released in v2.7.0

PR #914, opened 2019-03-08
feat(template): add template shared to support jade and pug

yuxizhe:
add  Jade( a node template engine) support,
Display Jade render time in APM

image

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


watson:
Thanks for taking the time to look into this and make a PR.

We're pretty close to landing #803, which will allow the agent to load 3rd party instrumentations from standalone modules. This means that we don't necessarily have to add all instrumentations to the core agent. I'll just talk to the rest of the team to discuss if this is something that makes sense to have in core or if it makes more sense as a separate module.

Btw, while jade is technically deprecated and replaced by pug, it still has more downloads than pug. However, pug is probably the future. Have you looked into how different pug is from jade and if the same instrumentation could be used for pug?

yuxizhe:
@watson 
Glad to hear this news, #803 would make it convenient to add customized modules. I agree that core agent do not need to contain all instrumentations. But node template engines are widely used, and Pug/Jade  both use `compile()` function  as well as `handlebars`. All of them can share the same instrumentation. I think It is necessary to add some instrumentations that can handle series of similar 
 modules other than repetitive 3rd party instrumentations.

kuisathaverat:
jenkins run the tests please

Qard:
Have you seen the `http-shared.js` in the `instrumentation` directory? We could do a similar thing for template modules where it just applies a generic instrumentation to whichever modules conform to the interface.

yuxizhe:
@Qard Oh I see, thanks for your suggestion. I will have a try.

Qard:
@yuxizhe Do you want to update this PR to use a `template-shared.js` interface, as suggested? Otherwise, I can take over. We'll also want tests and an update to the supported technologies documentation.

yuxizhe:
@Qard yes, I would like to finish it soon, but I'm a little busy these days. I think I can update this PR this week :)

Qard:
Okay, no worries. 😄 

Feel free to ping me here if you have any questions.

kuisathaverat:
jenkins run the tests please

yuxizhe:
hi @Qard , template-utils may support several template modules such as `jade`,`pug`,`handbars`,`react-dom/server`,`vue-server-renderer`.  Should I add all these files in `instrumentation/modules` folder and each file require the shared template-utils, or just modify the `instrumentation/index.js` file like this
```js

  var templateModuleList = ['jade','pug','handbars','react-dom/server','vue-server-renderer']
  if (templateModuleList.indexOf(name) > -1) {
    return require('./modules/template-shared')(exports, this._agent, version, enabled, name)
  }

  return require('./modules/' + name)(exports, this._agent, version, enabled)
}
```

Which way do you prefer?

Qard:
`instrumentation/modules/jade.js` should require the template utils module. We don't want to create instrumentation for those other libraries right now, we just want the capability to do that easily in the future, if demand for it exists. Adding a new module to our instrumentation list means we are committing to supporting it, so we don't want to do that unless there is specific demand for it and we have decided it is worth the cost of maintaining that ourselves, otherwise we would recommend using the new patch registry feature.

yuxizhe:
@Qard  PR updated with tests & documentation

Qard:
By the way, there appears to be an issue with pug 2.0.0. I think it might be a bug, so we can probably just exclude that from our support range and in the tav config file.

package.json
```diff
@@ -127,6 +127,7 @@
     "inquirer": "^0.12.0",
     "ioredis": "^4.3.0",
     "is-my-json-valid": "^2.19.0",
+    "jade": "~1.11.0",
```
watson: Please use the `^` syntax 🙂

```suggestion
    "jade": "^1.11.0",
```

.tav.yml
```diff
@@ -189,6 +189,12 @@ elasticsearch:
 handlebars:
   versions: '*'
   commands: node test/instrumentation/modules/handlebars.js
+jade:
+  versions: '*'
+  commands: node test/instrumentation/modules/jade.js
+pug:
+  versions: '*'
```
watson: As @Qard mentioned, you should just be able to exclude version 2.0.0 of pug from this list by doing so:

```suggestion
  versions: '<2.0.0 >2.0.0'
```

watson: If you want to run this locally first to see which versions are supported, you can run this command:

```sh
TAV=pug ./node_modules/.bin/tav --compat
```

This will install all versions in the allowed range and print out if they are failing or not and allow you to see if there's more versions than just 2.0.0 that are failing. It will not print out the error though. To view that, just replace `--compat` with `--quiet`.

.tav.yml
```diff
@@ -189,6 +189,12 @@ elasticsearch:
 handlebars:
   versions: '*'
   commands: node test/instrumentation/modules/handlebars.js
+jade:
+  versions: '*'
```
watson: It looks like the API for Jade was significantly different in version 0.5.5 and below which is why the TAV tests are failing. Since this version is 8 years old, I vote for just not supporting that. In that case, just change this:

```suggestion
  versions: '>0.5.5'
```

docs/supported-technologies.asciidoc
```diff
@@ -80,6 +80,8 @@ The Node.js agent will automatically instrument the following modules to give yo
 |https://www.npmjs.com/package/elasticsearch-js[elasticsearch-js] |>=8.0.0 |Will instrument all queries
 |https://www.npmjs.com/package/graphql[graphql] |>=0.7.0 <15.0.0 |Will instrument all queries
 |https://www.npmjs.com/package/handlebars[handlebars] |* |Will instrument compile and render calls
+|https://www.npmjs.com/package/jade[jade] |* |Will instrument compile and render calls
```
watson: Because of the lack of support for older Jade versions, we should also update this:

```suggestion
|https://www.npmjs.com/package/jade[jade] |>=0.5.6 |Will instrument compile and render calls
```

lib/instrumentation/template-shared.js
```diff
@@ -0,0 +1,40 @@
...
+
+      agent.logger.debug('intercepted call to %s compile %o', moduleName, {
+        id: id,
+        input: input
+      })
+
+      var ret = original.apply(this, arguments)
+      if (span) span.end()
+
+      return wrapTemplate(ret)
```
watson: To make sure this doesn't throw an exception in case `ret` isn't a function, we should add a type-test, eg:

```suggestion
      return typeof ret === 'function' ? wrapTemplate(ret) : ret
```

.tav.yml
yuxizhe: I found it took a lot of time to test from 0.0.0 to 2.0.0., what about we just set jade >1.0.0 and pug >2.0.0 ？

watson: We, unfortunately, have the same issue with all our other dependency tests. So far we've just accepted the time it took and limited our test-coverage to where we actually support it. So I vote to keep testing the entire supported range.

.tav.yml
yuxizhe: I want to test it locally but got error `zsh: no such file or directory: ./node_modules/.bin/tav` .

yuxizhe: oh, I see. here it is https://github.com/watson/test-all-versions

.tav.yml
```diff
@@ -189,6 +189,12 @@ elasticsearch:
...
+  commands: node test/instrumentation/modules/jade.js
+pug:
+  versions: '<2.0.0 >2.0.0'
```
yuxizhe:
```suggestion
  versions: '0.1.0 || >2.0.0'
```
image

docs/supported-technologies.asciidoc
```diff
@@ -80,6 +80,8 @@ The Node.js agent will automatically instrument the following modules to give yo
...
+|https://www.npmjs.com/package/pug[pug] |* |Will instrument compile and render calls
```
yuxizhe:
```suggestion
|https://www.npmjs.com/package/pug[pug] |'0.1.0 || >=2.0.1' |Will instrument compile and render calls
```

.tav.yml
watson: the tav binary is a devDependency and will be installed automatically when you run `npm install`. If you installed using `npm install --production` you'll not have it

docs/supported-technologies.asciidoc
watson: We don't exclude broken versions from the "supported-technologies" page. So here it's fine to list just `>=0.1.0` (in this case I'd guess v0.0.0 was just broken as well, so feel free to just list `*` if you like)

.tav.yml
watson: 😅 ha yeah, good catch

watson:
Thanks for taking this on. I know it's not easy with all the old broken versions of the modules 😅

kuisathaverat:
jenkins run the tests please

docs/supported-technologies.asciidoc
```diff
@@ -80,6 +80,8 @@ The Node.js agent will automatically instrument the following modules to give yo
 |https://www.npmjs.com/package/elasticsearch-js[elasticsearch-js] |>=8.0.0 |Will instrument all queries
 |https://www.npmjs.com/package/graphql[graphql] |>=0.7.0 <15.0.0 |Will instrument all queries
 |https://www.npmjs.com/package/handlebars[handlebars] |* |Will instrument compile and render calls
+|https://www.npmjs.com/package/jade[jade] |>=0.5.6 |Will instrument compile and render calls
+|https://www.npmjs.com/package/pug[pug] |'0.1.0 || >=2.0.1' |Will instrument compile and render calls
```
watson: We don't exclude buggy module versions from our supported version range, that could quickly become a very long list and doesn't really add any informant information. If users were to install these they would not work anyway - with or without this agent... Some modules have many buggy versions 😬 We just exclude them from our test-matrix as they obviously can't be tested. This makes this list a lot easier to read.

```suggestion
|https://www.npmjs.com/package/pug[pug] |'>=0.1.0' |Will instrument compile and render calls
```

yuxizhe: OK, got it! I missed the review before. :)

watson:
jenkins run the tests please

watson:
One Travis test keeps timing out because it's taking too long. This is unrelated to this PR. I'll go ahead and merge

watson:
Thanks for all your help creating these two instrumentations ❤️ 
It will be part of our upcoming release 😃

watson:
Aaaand now released in v2.9.0 🎉

PR #919, opened 2019-03-12
test: fix intake API integration test

watson:
The span transaction id is no longer a required property


watson:
I'll merge this now as the Jenkins + AppVeyor tests have completed. Travis is backed up with other things and getting this landed before Greenkeeper starts freaking out would be really great.

PR #926, opened 2019-03-16
chore: add TypeScript typings file

watson:
Closes elastic/apm-agent-nodejs#586

### Checklist

- [x] Implement code
- [x] Fix TODO's
- [x] Add tests
- [x] Add TypeScript usage docs


watson:
I'm obviously doing something wrong with the shared function definitions. They have different icons in VS Code compared to the ones defined inline (e.g. `addTags` vs. `addFilter`):

image

index.d.ts
```diff
@@ -0,0 +1,204 @@
...
+  username?: string | null | undefined;
+  email?: string | null | undefined;
+}
+
+interface ParameterizedMessageObject {
+  message: string;
+  params: Array<any>;
+}
+
+interface Logger {
```
delvedor: Pay attention that often a logger has also other methods, such as `child` for example.

start.d.ts
```diff
@@ -0,0 +1,2 @@
+import agent from './';
```
delvedor: Why you need this file?

watson: Because there's two ways to require the agent:

```js
const agent = require('elastic-apm-node')
```

And:

```js
const agent = require('elastic-apm-node/start')
```

So I think that's the only way to do it right?

index.d.ts
watson: Ah ok, is there a way to say that it has these methods + allow any other method?

start.d.ts
delvedor: Ok, I didn't know that!

> So I think that's the only way to do it right?

The only way to be sure, I guess, is to try it with VSCode or the tsc compiler.

index.d.ts
delvedor: I would try with something like following.
```ts
type anyFn = (...args: any[]) => any;
export interface Logger {
  fatal (msg: string, ...args: any[]): void;
  fatal (obj: {}, msg?: string, ...args: any[]): void;
  ...
  [key: string]: anyFn;
}
```
But I always had some issue with `[key: string]` usage. Mainly because I don't really understand what ts expects from me.

watson:
I asked on SO about how to fix the issue with the shared class methods: https://stackoverflow.com/questions/55208211/how-to-share-a-typescript-function-definition-between-two-classes

start.d.ts
watson: I tried to require the both versions in VS Code and it seemed to work fine. So I think it's ok. I just wasn't sure if there were a better way to do it or if my way was ok

index.d.ts
watson: Ah that's simple. Thanks. It's just saying that you can have any property in the `Logger` interface of the type defined - in this case `anyFn`.

I also just found this in the TypeScript docs:

image

watson:
🤔 I just realized that this "feature" makes `.d.ts` files a bit annoying for users: https://github.com/Microsoft/vscode/issues/68782

index.d.ts
```diff
@@ -168,13 +185,13 @@ type UncaughtExceptionFn = (err: Error) => void;
...
-type FilterFn = (payload: object) => object | Falsy;
+type FilterFn = (payload: object) => object | Falsy | void;
 
 type LambdaFn =
   | ((handler: LambdaHandlerFn) => LambdaHandlerFn)
   | ((type: string, handler: LambdaHandlerFn) => LambdaHandlerFn);
-type LambdaHandlerFn = (event: object, context: object, callback: LambdaHandlerCallbackFn) => any;
-type LambdaHandlerCallbackFn = (err?: Error | null | undefined, result?: any) => void;
+type LambdaHandlerFn = (event: object, context: object, callback: LambdaHandlerCallbackFn) => any; // TODO: Is the `any` return type correct?
+type LambdaHandlerCallbackFn = (err?: Error | null | undefined, result?: any) => void; // TODO: Can `result` really be `any`?
```
Qard: [aws-lambda](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/index.d.ts#L1019-L1043) uses the `any` type, so... 🤷‍♂️

index.d.ts
```diff
@@ -93,7 +110,7 @@ interface AgentConfigOptions {
   logLevel?: LogLevel;
   logger?: Logger;
   metricsInterval?: string | number; // TODO: Do we officially want to support numbers?
-  payloadLogFile?: string; // TODO: Do we want to advertise this?
+  payloadLogFile?: string; // TODO: Should we advertise this API?
```
Qard: Any reason _not_ to? It's a helpful debugging feature. 🤔

index.d.ts
```diff
@@ -2,40 +2,57 @@
...
+  // Transport
+  addFilter (fn: FilterFn): void;
+  addErrorFilter (fn: FilterFn): void;
+  addSpanFilter (fn: FilterFn): void;
+  addTransactionFilter (fn: FilterFn): void;
   flush (callback: Function): void;
   destroy (): void;
+
+  // Utils
+  logger: Logger; // TODO: Should we advertise this API?
```
Qard: It's been there for a long time, and I don't expect we'll move it any time soon, so why not? 🤷‍♂️

index.d.ts
watson: Good point. I was just thinking that it wasn't in our official docs, but now that I just checked I can see that it's in the README.md file.

index.d.ts
watson: I just wasn't sure what the user-case from a users point of view to use this was and it's not documented. It's more of an internal thing. But you're right, that if we ever wanted to remove it we probably should make it a breaking change anyway, so in that case we might as well add it to the typings file

index.d.ts
watson: Hmm I wonder if I somehow could depend on those types inside of this file. That way I don't have to replicate them

Qard: Yes, but I think you might need to add aws-lambda to the dependencies, even though we don't use it for anything code-related. 🤔 

I think we can probably just change the `LambdaHandlerFn` type to return `void | Promise<any>`. Also, I _think_ the `event` needs to be `any`--it takes any arbitrary input, so it could be _anything_.

watson: I fixed it by importing the `aws-lambda` types directly

index.d.ts
```diff
@@ -0,0 +1,205 @@
...
+  email?: string;
+}
+
+interface ParameterizedMessageObject {
+  message: string;
+  params: Array<any>;
+}
+
+interface Logger {
+  fatal (msg: string, ...args: any[]): void;
```
vigneshshanmugam: can create an interface for the logger functions
```ts
    interface LoggerFn {
        (msg: string, ...args: any[]): void;
        (obj: object, msg?: string, ...args: any[]): void;
    }
 

   {
     warn: LoggerFn,
     debug: LoggerFn
   }

```

start.d.ts
vigneshshanmugam: you can avoid creating this file by doing 
```ts
declare module 'elastic-apm-node/start' {
   export * from './'
}
```

index.d.ts
watson: Neat

start.d.ts
watson: I'm getting this error in VS Code if I do:

image

index.d.ts
watson: I've reverted this optimization again because the auto-complete feature in VS Code wasn't as nice when the methods were defined as properties.

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+
+declare class Span extends GenericSpan {
+  // TODO: The following should not be documented right? constructor(), customStackTrace(), setDbContext()
+  transaction: Transaction;
+  name: string | null; // TODO: Should we allow null?
+
+  end (endTime?: number): void;
+}
+
+export interface AgentConfigOptions {
```
watson: @sqren @delvedor after advice from @Qard, I've exported all the interfaces and types that might make sense for a TypeScript user to use for typing inside their app.

This allows the user to write something like this:

```js
import agent, { AgentConfigOptions } from 'elastic-apm-node'

function starter (options: AgentConfigOptions) {
  agent.start(options)
}

starter({...})
```

What do you think about that?

sqren: THANK YOU! This is often something that is not exported but it makes it so much easier to wrap code, when the types for the arguments is available!

index.d.ts
vigneshshanmugam: I don't see any difference with both changes in auto completion.

index.d.ts
```diff
@@ -0,0 +1,208 @@
+/// <reference types="node" />
+/// <reference types="aws-lambda" />
+/// <reference types="connect" />
+
+import { IncomingMessage, ServerResponse } from 'http';
+import { Handler } from 'aws-lambda';
+import { ErrorHandleFunction } from 'connect';
+
+export default agent;
```
vigneshshanmugam: This would mean that users would need to make typing work in this way, 

Screenshot 2019-03-20 at 11 41 00

I think we should do this instead 
```ts
export = Agent
```
Screenshot 2019-03-20 at 11 42 01

index.d.ts
```diff
@@ -0,0 +1,208 @@
+/// <reference types="node" />
+/// <reference types="aws-lambda" />
+/// <reference types="connect" />
```
watson: I've been a little unsure about what this did and what the difference was compared to adding `@types/connect` as a dev-dependency. This seem a duplicate of the devDependencies.

According to the [documentation for this directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-), it simply instructs the compiler to inline that particular module in this file when compiling (at least that's how I read it). If so, the import statements below should not be necessary right? And if so, I think you'd either add the `@types/*` devDependencies + the `import` statements _OR_ you use the `<reference types="..." />` directives.

However, if I remove the `import` statements below, VS Code complains that it doesn't know about the different types previously imported. But maybe I need to prefix them with something? Or maybe I misunderstood this whole thing.

@delvedor @sqren any ideas?

index.d.ts
vigneshshanmugam: If we want to make other exports work, everything should be namespaced inside Agent and we can do 
```ts
export = Agent
export as namespace Agent

declare namespace Agent {

}
```

watson: I think you got the API of the module wrong. We're not exporting the `Agent` object, but an instance of that object

start.d.ts
vigneshshanmugam: sorry my bad, we can do this `export default Agent`

index.d.ts
watson: 🤔 though you are correct about something not working regarding the export... not really sure what's wrong though.

I thought `export default foo` was the same as `module.exports = foo`. But apparently not 100%

watson: How would you write the CommonJS style `module.exports = foo` as an ES Modules export?

start.d.ts
watson: That would export the `Agent` object though which we're not doing in real life. We're exporting an instance of the `Agent` object

index.d.ts
watson: If you write it as `key: value`, VS Code thinks that this is a property and not a method. Details: https://stackoverflow.com/questions/55208211/how-to-share-a-typescript-function-definition-between-two-classes

index.d.ts
vigneshshanmugam: > I thought export default foo was the same as module.exports = foo. But apparently not 100%

I am not sure though, its node's internal resolution. I guess this is the equivalent of `export default foo`
```js
module.exports = {
  __esModule: true,
  default: foo
}
```

vigneshshanmugam: The reason because import works and require doesn't is because in ES6, the default export is also a named export and `default` is a reserved keyword. Since we have typescript definition in ES6 export and my example was importing it as a common js, the default export has to be explicitly used. But if we used `import agent from 'elastic-apm-node'` it would work simply since the default export would be resolved properly.

vigneshshanmugam: As suggested before `export = Agent` and using namespaces we can guarantee it works both on commonjs and es6 modules. More details here http://www.typescriptlang.org/docs/handbook/declaration-files/deep-dive.html

test/script/run_tests.sh
```diff
@@ -53,6 +53,8 @@ run_test_suite () {
   else
     nyc node test/test.js
   fi
+
+  npm run test-types
```
watson: Not sure if it's good to run this after the main tests 🤔 On the other hand it would probably give weird errors if say there were a bug in the agent code somewhere and we triggered that bug by accident from inside the type tests if they ran before the regular agent tests. So I'd prefer if we could run them after the main tests.

@kuisathaverat do you think Jenkins would be ok with running this command after the `JUNIT` block above?

test/types/index.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function filter1 (payload: object) {
+  return payload
+}
+function filter2 (payload: object) {
+  return false
+}
+function filter3 (payload: object) {}
+agent.addFilter(filter1)
+agent.addFilter(filter2)
```
watson: @delvedor @sqren This test fails because it detects that `filter2` returns a boolean, but the `addFilter` function expects the return value to be `false`. Obviously in this case, that boolean value is `false`, but it's somehow not allowing it. Any idea how I can update the types to allow a falsy value here in a way that would allow the user to return false inside the `FilterFn`?

watson: Here's the VS Code error:

image

watson:
@Qard could use your input on the remaining TODO's 😃

test/script/run_tests.sh
kuisathaverat: if this command does not delete or overwrite the result files from tests and coverage, it'd be ok

index.d.ts
sqren: I don't understand those directives tbh. I'd always go for the devDependencies + static import option. Maybe with the exception of Node which cannot be imported as a module.

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+export interface SpanOptions {
+  childOf?: Transaction | Span | string // TODO: This technically accepts other values, but we might not want to document these?
+}
+
+export type CaptureBody = 'off' | 'errors' | 'transactions' | 'all';
+export type CaptureErrorLogStackTraces = 'never' | 'messages' | 'always';
+export type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'fatal';
+
+export type CaptureErrorCallback = (err: Error | null, id: string) => void;
+export type FilterFn = (payload: object) => object | Falsy | void;
```
delvedor: I might be wrong, but void should be included in the Falsy definition, isn't it?

watson: `void` isn't technically a falsy value in the pure JavaScript sense of the word. It's more of a lack of explicit `return` from a function which is a concept something TypeScript know about right?

test/script/run_tests.sh
watson: It shouldn't. So I guess that's ok then. Thanks 😃

index.d.ts
delvedor: By reading [this definition](https://www.typescriptlang.org/docs/handbook/basic-types.html#void) it appears that `void` can be either `null` or `undefined`, which are inside your `Falsy` definition.
I'm not sure how works TypeScript in this regard, but as you know, JavaScript has an implicit `return undefined` if nothing is specified.

index.d.ts
watson: So you think that I could just leave them all out? If I remove them I don't seem to get any errors. 

@delvedor why are you using this in fastify and elasticsearch-js?

delvedor: > @delvedor why are you using this in fastify and elasticsearch-js?

Good question, I don't know. We have always used this notation.

test/types/index.ts
sqren: Typescript is expanding the type of `false` to boolean. You could narrow it explicitly:

```ts
function filter2(payload: object): false {
  return false;
}
```

But that's pretty ugly. Another approach is something like this:
```ts
const filter4: Filter = (payload: object) => {
  return false;
};
```

A bit better but still not great.

index.d.ts
watson: I read the entire thing but only got more confused 😅

What did you say was the way to express `module.exports = new Agent()` in a `.d.ts` file?

index.d.ts
watson: Hmm you might be half right. it seems that `void` is almost like a synonym for `undefined`. Check out this example:

image

The only valid uses of return here is `return` by it self or `return undefined`. If I type `return null` it complains with the error you see.

However, if I don't declare `void` as a valid return type, I'm forced to add a `return`:

image

test/types/index.ts
watson: Hmm I see. Maybe I should change the API to be more TypeScript friendly then

watson: On the other hand that would be a breaking change, so I can't do that in this iteration 😞🤬

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+  ): void;
+
+  // Transactions
+  startTransaction (
+    name?: string,
+    type?: string,
+    options?: TransactionOptions
+  ): Transaction | null;
+  setTransactionName (name: string): void;
+  endTransaction (result?: string, endTime?: number): void; // TODO: Should we allow number as well for result?
```
Qard: I think it makes sense to align these types with whatever apm-server accepts. 🤔

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+  addTransactionFilter (fn: FilterFn): void;
+  flush (callback: Function): void;
+  destroy (): void;
+
+  // Utils
+  logger: Logger;
+}
+
+declare class GenericSpan implements Taggable {
+  // TODO: The following should not be documented right? constructor(), timestamp, ended, id, traceId, parentId, sampled, duration()
```
Qard: We don't currently have docs for timestamp, ended, id, traceId, parentId and sampled, so I think omitting them is fine. We could have a _separate_ discussion if it's a good idea to _make_ some of those public, but that could just be a follow-up PR.

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+  type: string | null; // TODO: Should we allow null?
```
Qard: I think so, yes. Otherwise it would have to be explicitly set to pass the interface test. 🤔

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  setTag (name: string, value: TagValue): boolean;
+  addTags (tags: Tags): boolean;
+}
+
+declare class Transaction extends GenericSpan implements StartSpanFn {
+  // TODO: The following should not be documented right? constructor(), setUserContext(), setCustomContext(), toJSON(), setDefaultName(), setDefaultNameFromRequest()
+  name: string | null; // TODO: Should we allow null?
+  result: string; // TODO: Should we also document number?
```
Qard: See comments above for these three TODOs.

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  startSpan (name?: string, type?: string, options?: SpanOptions): Span | null;
+  ensureParentId (): string;
+  end (result?: string, endTime?: number): void; // TODO: Should we allow number as well for result?
+}
+
+declare class Span extends GenericSpan {
+  // TODO: The following should not be documented right? constructor(), customStackTrace(), setDbContext()
+  transaction: Transaction;
+  name: string | null; // TODO: Should we allow null?
```
Qard: Same as GenericSpan type, I think it's valid for a Span to have a null name. We apply a fallback internally, but that's not exposed to the outer surface of the object to work for any runtime checks one might try to do to validate this.

index.d.ts
```diff
@@ -0,0 +1,208 @@
...
+  debug (msg: string, ...args: any[]): void;
+  debug (obj: {}, msg?: string, ...args: any[]): void;
+  trace (msg: string, ...args: any[]): void;
+  trace (obj: {}, msg?: string, ...args: any[]): void;
+  [propName: string]: any;
+}
+
+export interface TransactionOptions {
+  startTime?: number;
+  childOf?: Transaction | Span | string; // TODO: This technically accepts other values, but we might not want to document these?
```
Qard: If someone asks, we can always expand it later. 🤷‍♂️

Qard:
Could we do more to test the exported types are usable in user code?

index.d.ts
watson: The APM Server only supports string, but we're stringifying the result before sending it to the server, and I could see a use-case for using an exit or status code as a result. Allowing people to parse in a number (that we stringify for them internally) might be a normal use-case?

index.d.ts
watson: What interface test?

Qard: I mean for an object to be considered a valid GenericSpan by the type checker it would need to have the type set first, I think.

index.d.ts
Qard: Gotcha. Sounds reasonable to me.

index.d.ts
watson: The type default to the string `"custom"`. So actually I don't think it's normally ever `null` during normal operation 🤔

Qard: Ah, right...good point. For some reason I was thinking that was just an encoding default, but it actually sets that in the constructor. 👍

watson: So I guess I'll just remove the `null` type from here then. Then we can open up later if we find a use-case

index.d.ts
Qard: What you're already doing is the correct way to export an instance of a thing.

```ts
declare const agent: Agent;
export default agent;
```

watson: I thought so as well. The problem is just that it only works if the user is using ES Modules. If the user is using CommonJS, Intellisense wrongly thinks that they need to import the module like so for it to be valid:

```js
const agnet = require('elastic-apm-node').default
```

watson: And while we could make a `default` property available on the agent object that pointed to the agent object it self, that would just be a hack to make the typings happy, which I think is kind of silly

Qard: Oh, actually `export default agent` maps to the ESM `default` value, while `export = agent` would map directly to a commonjs module export.

watson: Ha, @sqren just told me the same thing. @vigneshshanmugam actually also wrote it above, but I got confused about the namespace stuff in the same comment so I never got it to work. But now it works locally. The only downside is that you can't export other things if you use the `export = agent` syntax. Which means that we can't export our types. But I can see from a technical perspective why that is and I guess there's no way around that. So I'll update it and remove all our other exports

test/types/index.ts
watson: Fixed

index.d.ts
watson: Fixed

index.d.ts
watson: Fixed

index.d.ts
```diff
@@ -0,0 +1,214 @@
+/// <reference types="node" />
+/// <reference types="aws-lambda" />
+/// <reference types="connect" />
+
+import { IncomingMessage, ServerResponse } from 'http';
+import { Handler } from 'aws-lambda';
+import { ErrorHandleFunction } from 'connect';
+
+export = agent;
```
watson: Changed to using this syntax instead of `export default agent`.

But unfortunately this means that TypeScript users will need import the module differently from Babel users and users of the experimental ES Module support in Node.js core:

```js
// Style 1: Works only with TypeScript
import * as agent from 'elastic-apm-node'

// Style 2: Works only with Babel and `node --experimental-modules`
import agent from 'elastic-apm-node'
```

If a TypeScript user uses the second style, it will break.

vigneshshanmugam: AFAIK, Typescript users can still use the second style by setting  allowSyntheticDefaultImports flag in compiler options.

watson: Oh you're right! I didn't know about that config option. Do you know how common that is to use and if it's well known?

vigneshshanmugam: Yeah its pretty common, We have used it extensively in lots of projects where the code must run both on Node.js as well as on the browser.

watson: Cool I guess I'll just add some docs then describing how to use it just in case. If anything because we've already documented to Babel stuff, so it would be weird if we didn't also document this.

watson: After a bit of testing, it turns out that `allowSyntheticDefaultImports` only "silences" the errors, but the JavaScript code it produces still thinks that there should be a `default` property  on the export. But if I instead use `esModuleInterop`, it works as expected

watson:
Updated ES Modules docs page to mention TypeScript as well:

image

index.d.ts
watson: If I install this branch as a module in a clean folder (i.e. no devDependencies), and try to run `tsc` on a file that does nothing else than imports `elastic-apm-node`, then I get these errors:

image

I have to manually install `@types/node`, `@types/aws-lambda`, and `@types/connect` for `tsc` to stop complaining.

It seems VS Code is smarter though and will just treat all non-installed types as `any`.

watson: I concluded that all TypeScript developers will have `@types/node` installed already, so this isn't really a big deal. Whereas most TypeScript users of our module will probably not use the lambda API or the connect API, and as such will not have any of those dependencies installed them selfs. Requiring users to have these `@types` installed is therefore very inconvenient for the majority of our TypeScript users. So I've now inlined those types instead into our `index.d.ts` file and removed the dependency.

watson:
I think this is ready now. I could spend time looking into using JS/TSDocs instead. But that might as well be another PR.

index.d.ts
```diff
@@ -0,0 +1,280 @@
...
+interface Taggable {
+  setTag (name: string, value: TagValue): boolean;
+  addTags (tags: Tags): boolean;
+}
+
+interface StartSpanFn {
+  startSpan (name?: string, type?: string, options?: SpanOptions): Span | null;
+}
+
+// Inlined from @types/aws-lambda - start
```
vigneshshanmugam: Just curious, any particular reason for inlining the external types vs not referencing them directly?

index.d.ts
```diff
@@ -0,0 +1,280 @@
+/// <reference types="node" />
```
Qard: The node reference doesn't cause issues with only being installed as a dev dependency? Is there something special about it?

watson: It's left there with the assumption that people already using TypeScript will already have `@types/node` installed them selfs. It will only be an issue if you run `tsc` and don't have `@types/node` installed. And you only do that if you're compiling TypeScript code.

watson: See this conversation of details: https://github.com/elastic/apm-agent-nodejs/pull/926#discussion_r267279580

index.d.ts
watson: It would add an unnecessary burden on all TypeScript users as it would force them to manually install `@types/connect` and `@types/aws-lambda`. See this conversation for details: https://github.com/elastic/apm-agent-nodejs/pull/926#discussion_r267279580

vigneshshanmugam: Cool that makes sense.

watson:
testing some hooks... please ignore 😎

nik9000:
I'm looking at the docs-build failure. I believe it is safe to ignore but I'll track it down!

nik9000:
> I'm looking at the docs-build failure. I believe it is safe to ignore but I'll track it down!

I've tracked down the issue and will submit a PR to fix it to the infra repo. Sorry for the trouble. Thanks for being my first victims!

nik9000:
@elasticmachcine, run docs build.

nik9000:
@elasticmachine, run docs build.

nik9000:
> > I'm looking at the docs-build failure. I believe it is safe to ignore but I'll track it down!
> 
> I've tracked down the issue and will submit a PR to fix it to the infra repo. Sorry for the trouble. Thanks for being my first victims!

Fixed!

watson:
jenkins run tests please

index.d.ts
watson: FYI, I just updated it so that the inlined styles are wrapped in a namespace

spalger:
Thank you!!! 🌮

smhmayboudi:
I think it is better to use the npm package definitions instead of writing them.
```
import AwsLambda from "aws-lambda";
import Connect from "connect";
```


https://github.com/elastic/apm-agent-nodejs/blob/c56f894138c81f0b07146f86c3ba7139bb68a59c/index.d.ts#L327

https://github.com/elastic/apm-agent-nodejs/blob/c56f894138c81f0b07146f86c3ba7139bb68a59c/index.d.ts#L389

smhmayboudi:
I copy and paste this file to my project. Is there another way to reach the interfaces and class definitions?

smhmayboudi:
I read the docs and find out `logUncaughtExceptions` is not defined in `AgentConfigOptions`. Would you please fix it.

PR #931, opened 2019-03-19
fix(lambda): do not wrap context

### LINKED ISSUE #909: AWS Lambda Context Issue
**Describe the bug**

AWS Lambda APM copies `context` properties onto a new object. When using `context.callbackWaitsForEmptyEventLoop`, which is a setter function, it doesn't do anything on this object because it was copied as a regular boolean property.

**To Reproduce**

1. Create a lambda function that keeps the event loop open (e.g. a database connection)
2. Wrap in `apm.lambda()`
3. Try to set `context.callbackWaitsForEmptyEventLoop = false` and see the lambda function timeout

**Expected behavior**

- OS: AWS Lambda
- Node.js version: 8.x
- Agent version: 2.6.0

This combined with https://github.com/elastic/apm-nodejs-http-client/issues/46 made for a long day trying to integrate Elastic APM. Let me know if there's anything I can do to make a speedy bugfix! Happy to submit effective PRs - I believe this one should just be mutating `context.{success,fail}` directly.

Qard:
Fixes #909

### DIFF
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -47,7 +47,7 @@ module.exports = function elasticApmAwsLambda (agent) {
     captureContext(trans, payload, context, result)
     trans.end()
     return () => {
-      agent.flush(callback)
+      agent.flush(() => callback())
     }
   }
 
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const shimmer = require('./instrumentation/shimmer')
+
 module.exports = function elasticApmAwsLambda (agent) {
   function sendTransactionFn (trans) {
     return () => {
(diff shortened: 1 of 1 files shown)


lib/lambda.js
```diff
@@ -50,36 +22,43 @@ module.exports = function elasticApmAwsLambda (agent) {
   }
 
   function wrapContext (trans, payload, context) {
-    const succeed = wrapLambdaCallback(trans, payload, context, (_, res) => context.succeed(res))
-    const done = wrapLambdaCallback(trans, payload, context, (err, res) => context.done(err, res))
-    const fail = wrapLambdaCallback(trans, payload, context, (err) => context.fail(err))
+    shimmer.wrap(context, 'succeed', (succeed) => {
+      return function wrappedSucceed (result) {
+        var bound = succeed.bind(this, result)
```
watson: We should probably just keep using `const` here (and elsewhere in the file)

Qard:
The context functions are the _old_ way of doing things. Originally it was just `(payload, context)`, then they added the callback awhile later, then they added promise support awhile after that. There's still lots of code in the wild that uses the context object functions, and the promise code actually uses it internally too.

lib/lambda.js
Qard: Good catch. 😅

Qard: Fixed.

lib/lambda.js
```diff
@@ -50,36 +22,43 @@ module.exports = function elasticApmAwsLambda (agent) {
...
+    shimmer.wrap(context, 'done', (done) => {
+      return wrapLambdaCallback(trans, payload, context, done)
     })
   }
 
+  function captureAndMakeCompleter (trans, payload, context, result, callback) {
+    captureContext(trans, payload, context, result)
+    trans.end()
+    return () => {
+      agent.flush(callback)
```
watson: The flush callback might be called with an Error object. Is the `callback` ok with that? I couldn't find the docs for the AWS API so I'm not entirely sure what happens. I think maybe we should just log it and call the callback without to not interfere with the normal execution of the function.

lib/lambda.js
```diff
@@ -100,19 +79,12 @@ module.exports = function elasticApmAwsLambda (agent) {
...
 
-      const result = fn.call(this, payload, context, callback)
-
-      return result && typeof result.then === 'function'
-        ? result.then(
-          passFn(trans, payload, context),
-          failFn(trans, payload, context)
-        )
-        : result
+      return fn.call(this, payload, context, callback)
```
watson: How will the transaction be ended in case there's no `callback`?

Qard: The context methods will either be called directly by the user, or internally via the promise execution.

lib/lambda.js
Qard: fixed

lib/lambda.js
```diff
@@ -50,36 +22,43 @@ module.exports = function elasticApmAwsLambda (agent) {
...
+    shimmer.wrap(context, 'done', (done) => {
+      return wrapLambdaCallback(trans, payload, context, done)
     })
   }
 
+  function captureAndMakeCompleter (trans, payload, context, result, callback) {
+    captureContext(trans, payload, context, result)
+    trans.end()
+    return () => {
+      agent.flush(() => callback())
```
watson: My initial thought was to log the error here. What's the reason for silencing it?

Qard: No reason. I can do that too.

Qard: fixed

Issue #935, opened 2019-03-21
Transaction name is incorrect when request processing has an error exception


jiangzm:
**Describe the bug**
In the use of the express framework, the transaction default name is inaccurate when processing requests with exceptions, so what is seen in the kibana system is part of the real path or route.

**To Reproduce**
1. Use this config '...'
2. Then call '....'
3. Then do '....'

**Expected behavior**

**Environment (please complete the following information)**
- OS: macOS 10.14.3
- Node.js version: 10.15.3
- APM Server version: 6.6.2
- Agent version: ^2.6.0

**How are you starting the agent? (please tick one of the boxes)**
- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**
example code.
```javascript
const apm = require('elastic-apm-node').start({
  serviceName: "apm-log",
  serviceVersion: "1.0",
  serverUrl: 'http://localhost:8200'
})

const express = require('express')
const route = module.exports = new express.Router()

route.get('/orders/:id', function (req, res) {
  throw new Error("foo")
})
route.get('/products/:id', function (req, res) {
  res.send('bar')
})

const app = express()
app.use('/api', route)
```

When processing a request normally, the currentTransaction name is `“GET /api/products/:id”`.
When processing a request exception, the currentTransaction name is `“GET /products/:id”`.

jiangzm:
[layer.js `#97`](https://github.com/expressjs/express/blob/master/lib/router/layer.js#L97) handler code in `express` fx:
```javascript
Layer.prototype.handle_request = function handle(req, res, next) {
  var fn = this.handle;

  if (fn.length > 3) {
    // not a standard request handler
    return next();
  }

  try {
    fn(req, res, next);
  } catch (err) {
    next(err);
  }
};
```
Executing this code  `'next(err)'` will pop the last executed route path.

[express.js`#50`](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/express.js#L50) warp code in `apm-agent-nodejs`：
```javascript
if (!layer.route && layerPath && typeof next === 'function') {
    safePush(req, symbols.expressMountStack, layerPath)
    arguments[2] = function () {
        req[symbols.expressMountStack].pop()
        return next.apply(this, arguments)
    }
}
```

watson:
[...]

PR #938, opened 2019-03-21
fix(express): transaction default name is incorrect

jiangzm:
Fixes elastic/apm-agent-nodejs#935 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -47,7 +47,9 @@ module.exports = function (express, agent, { version, enabled }) {
             if (!layer.route && layerPath && typeof next === 'function') {
               safePush(req, symbols.expressMountStack, layerPath)
               arguments[2] = function () {
-                req[symbols.expressMountStack].pop()
+                if (!(req.route && arguments.length && arguments[0] instanceof Error)) {
+                  req[symbols.expressMountStack].pop()
+                }
                 return next.apply(this, arguments)
               }
             }
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -47,7 +47,7 @@ module.exports = function (express, agent, { version, enabled }) {
             if (!layer.route && layerPath && typeof next === 'function') {
               safePush(req, symbols.expressMountStack, layerPath)
               arguments[2] = function () {
-                if (!(req.route && arguments.length && arguments[0] instanceof Error)) {
+                if (!(req.route && arguments[0] instanceof Error)) {
                   req[symbols.expressMountStack].pop()
                 }
                 return next.apply(this, arguments)
--- a/test/instrumentation/modules/express/basic.js
+++ b/test/instrumentation/modules/express/basic.js
@@ -214,6 +214,45 @@ test('sub-routers include base path', function (t) {
   })
 })
 
+test('sub-routers throw exception', function (t) {
+  t.plan(6)
+
+  resetAgent(function (data) {
+    t.equal(data.transactions.length, 1, 'has a transaction')
+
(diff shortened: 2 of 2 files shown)


lib/instrumentation/modules/express.js
```diff
@@ -47,7 +47,9 @@ module.exports = function (express, agent, version, enabled) {
             if (!layer.route && layerPath && typeof next === 'function') {
               safePush(req, symbols.expressMountStack, layerPath)
               arguments[2] = function () {
-                req[symbols.expressMountStack].pop()
+                if (!(req.route && arguments.length && arguments[0] instanceof Error)) {
```
watson: What's the reason for the `arguments.length` check?

jiangzm: Oh, the `arguments.length`  code is not required. I will remove it.

test/instrumentation/modules/express.js
```diff
@@ -214,6 +214,45 @@ test('sub-routers include base path', function (t) {
...
+
+    var trans = data.transactions[0]
+    t.equal(trans.name, 'GET /api/:name', 'transaction name is GET /api/:name')
+    t.equal(trans.type, 'request', 'transaction type is request')
+  })
+
+  var error = new Error('hello')
+  var captureError = agent.captureError
+  agent.captureError = function (err, data) {
+    t.equal(err, error, 'has the expected error')
```
jiangzm: For another question, I also reproduced locally in Express v4.4.5. My initial judgment is that the version did not throw the error exception parameter to the outer layer. 
But I still don't have a better solution to fix it except to comment this test code first. 

``` javascript
  t.plan(5)
  ......
  agent.captureError = function (err, data) {
    // t.equal(err, error, 'has the expected error')
    // t.ok(data, 'captured data with error')
  }
```
Can I fix this test error result like this?

watson: I just ran a test locally and it does seem like the error makes it to the root router. So I'm not sure why the agent doesn't pick it up... will just investigate a little more. But it might be an error unrelated to this.

watson: Ah, I found the issue. The agent expects the `finalhandler` module to be used by express so it can be notified of errors in case a custom error handler hasn't been manually added as middleware. But the `finalhandler` module wasn't added until Express 4.5.0, the version just after 4.4.5. That's why it's not calling `agent.captureError` in those early versions.

However, checking that the error is caught is not really something related to this test, so I suggest that you instead just change the custom `agent.captureError` function to be a no-op and always expect 4 assertions no matter which version of Express is being used.

jiangzm: Thanks for your suggestion, I am modifying the code to a fixed 4 assertions.

test/instrumentation/modules/express.js
```diff
@@ -10,7 +10,6 @@ var http = require('http')
 
 var express = require('express')
 var test = require('tape')
-
```
watson: Could you add this blank line back please? We have an unwritten "rule" in our code style where node modules from npm are separated from internal modules by a blank line 🙂

kuisathaverat:
jenkins run the tests please

test/instrumentation/modules/express.js
jiangzm: ok, It seems that another rule failed.
```javascript
agent.captureError = function (err, data) {}
```
// change to
```javascript
agent.captureError = function () {}
```
I have modified it

watson:
jenkins run the tests please

watson:
Just published a new version of the agent with this fix to npm (v2.8.0)

PR #943, opened 2019-03-25
fix: don't return anything from the module patch functions

watson:


watson:
jenkins run tests please

PR #952, opened 2019-03-27
fix: don't add traceparent header to S3 requests

watson:
This would make the request fail by invalidating the request signature and return this error:

    SignatureDoesNotMatch: The request signature we calculated does not match the signature you provided. Check your key and signing method.

Brought to my attention by this post: https://discuss.elastic.co/t/aws-s3-getobject-failure-signature-mismatch-while-updating-apm-to-2-x/173275/3

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -4,6 +4,8 @@ var url = require('url')
 
 var endOfStream = require('end-of-stream')
 
+const s3Host = /\.s3\.amazonaws\.com$/
+
 const transactionForResponse = new WeakMap()
 exports.transactionForResponse = transactionForResponse
 
@@ -123,7 +125,7 @@ exports.traceOutgoingRequest = function (agent, moduleName) {
       // to indicate requested services should not be sampled.
       // Use the transaction context as the parent, in this case.
       var parent = span || agent.currentTransaction
-      if (parent && parent._context) {
+      if (parent && parent._context && shouldPropagateTraceContext(options.host)) {
         options.headers['elastic-apm-traceparent'] = parent._context.toString()
       }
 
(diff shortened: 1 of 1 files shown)


watson:
I wanted to add tests, but it turned out to be really complicated to do without spending too much time on it. If you know of a quick and easy way to test this, please let me know, otherwise I think it's ok to just let it be.

watson:
Btw, I did test this locally using this script, so it should work:

```js
'use strict'

const agent = require('./').start({
  metricsInterval: 0,
  captureExceptions: false,
  logLevel: 'trace'
})

const AWS = require('aws-sdk')

AWS.config.update({
  accessKeyId: '...',
  secretAccessKey: '...'
})

var s3 = new AWS.S3()

var params = {
  Bucket: 'elastic-apm-test',
  Body: 'hello world ' + Date.now(),
  Key: 'test'
}

agent.startTransaction('s3.upload')
s3.upload(params, function (err, data) {
  agent.endTransaction(err ? 'error' : 'success')
  if (err) throw err
  console.log('Uploaded in:', data.Location)
})
```

ptariche:
@watson this issue is persisting with regional hosts like s3.us-west-2.amazonaws.com

watson:
@ptariche Good point. I didn't consider those. Either we just test for `*.amazonaws.com` or we add an extra test for `*.s3.*.amazonaws.com` to limit this to s3 only. But I actually suspect that this is an issue for other services than just s3. But I also fear that you might want to keep this header for some requests - e.g. to aws lambda 🤔 

What do you think?

ptariche:
Perhaps *.amazonaws.com

watson:
@ptariche I've created an issue for this: #1055

PR #954, opened 2019-03-27
fix: instrument http/https.get requests

### LINKED ISSUE #953: Outgoing http.get requests are not traced
It seems like only `http.request` requests are traced.

Related Discuss topic: https://discuss.elastic.co/t/apm-not-correlating-trace/172758/3

watson:
For `http.get` this was broken in Node.js 8 and up
For `https.get` this was broken in Node.js 9 and up

Fixes #953

### DIFF
--- a/test/instrumentation/modules/http/aborted-requests-enabled.js
+++ b/test/instrumentation/modules/http/aborted-requests-enabled.js
@@ -45,7 +45,7 @@ test('client-side abort below error threshold - call end', function (t) {
 
   server.listen(function () {
     var port = server.address().port
-    clientReq = http.get('http://localhost:' + port, function (res) {
+    clientReq = get('http://localhost:' + port, function (res) {
       t.fail('should not call http.get callback')
     })
     clientReq.on('error', function (err) {
@@ -88,7 +88,7 @@ test('client-side abort above error threshold - call end', function (t) {
 
   server.listen(function () {
     var port = server.address().port
-    clientReq = http.get('http://localhost:' + port, function (res) {
+    clientReq = get('http://localhost:' + port, function (res) {
       t.fail('should not call http.get callback')
     })
     clientReq.on('error', function (err) {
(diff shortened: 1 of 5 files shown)


watson:
Tip: The diff for this PR is best viewed while ignoring whitespace:

image

watson:
Hmm yeah. Issue seems to be that we're now capturing some `http.get` spans that we didn't catch before. So we're getting spans we didn't expect

watson:
@Qard It should work now. I'm not proud of the fix though. It would be nice to have a way to make requests that you somehow mark as "not to be traced". But that might be out of scope of this PR?

test/instrumentation/modules/http/aborted-requests-enabled.js
```diff
@@ -420,3 +420,8 @@ function resetAgent (cb) {
   agent._instrumentation.currentTransaction = null
   agent._transport = mockClient(1, cb)
 }
+
+function get () {
+  agent._instrumentation.currentTransaction = null
```
Qard: How is there a `currentTransaction` set in the first place, when these tests are resetting the agent before each test? 🤔

watson: It's somehow leaking from the previous test inside the `listen` callback - but only sometimes. I tried to figure out why but couldn't without extensive debugging.

watson: And funny enough, it's only an issue on Node.js 8.x 🤷‍♂️ So there must be something in regards to how `server.listen` works there.

Qard: I wonder if that's a bug in how we're using the `patch-async.js` stuff, because it patches the internal `net._listen2` method. 🤔

watson: I don't think it's related to `patch-async.js` as this also happens when using Async Hooks as far as I ca see

watson:
I agree. Maybe we should make an issue to look into this later

Issue #955, opened 2019-03-27
Logger is not initialized before usage


blakeembrey:
**Describe the bug**

When running in AWS Lambda with webpack, the dynamic requires are returning numbers. This, in turn, is causing an attempt to call `this._agent.logger`. Unfortunately, the logger is unavailable and this crashes the request.

**To Reproduce**

1. Webpack bundle a simple import, e.g. with https://github.com/zeit/ncc
2. Import APM agent
3. See crash

**Expected behavior**

Doesn't crash.

**Environment (please complete the following information)**
- OS: [e.g. Linux]
- Node.js version:
- APM Server version:
- Agent version:

**How are you starting the agent? (please tick one of the boxes)**
- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

blakeembrey:
![image](https://user-images.githubusercontent.com/1088987/55119774-6c1fcb80-50b0-11e9-8b9f-9679791a43db.png)

blakeembrey:
Some easy fixes:

1. Avoid `require.resolve` here since it returns a number in webpack, use `require` directly
2. Avoid erroring on a number
3. Initialize the logger before instrumentation

watson:
Oh, it's definitely an error on our part trying to use the logger before it's initialized - this should be fixed regardless of any Webpack issues. I hadn't considered how Webpack changes the `require` behavior and so we should probably make Webpack part of our test suite so this doesn't happen again.

We'll take a look at your PR's and see how we can best fix this - Thanks for taking the time to fix this 👍

watson:
I found a reference to the number being returned by `require.resolve` in Webpack here:
https://webpack.js.org/api/module-methods/#requireresolve

![image](https://user-images.githubusercontent.com/10602/55143150-6e306b80-513e-11e9-9dfe-ff37ada40765.png)

watson:
I did some early testing with the two PR's you made and it seems like while it stops complaining, it somehow doesn't instrument the application any more 🤔 It's most likely related to how we're patching the require function that just doesn't work with Webpack

blakeembrey:
@watson Oh yeah, FWIW it has never worked for me. I’m 99% sure it’s because you try to read the `package.json` which won’t exist. This is purely so I can use the APM client again, I’m tracing everything manually.

PR #956, opened 2019-03-28
Initialize config/logger before instrumentation

blakeembrey:
Fixes #955.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

PR #957, opened 2019-03-28
fix(instrumentation): explicitly use `require`

blakeembrey:
Closes the webpack issue causing https://github.com/elastic/apm-agent-nodejs/issues/955. Personally, I'd just avoid the API where people can define paths at all - it'll never work in bundlers and it just adds additional code. I'm not sure I understand the value except late requires, but that seems negligible.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -68,7 +68,7 @@ function Instrumentation (agent) {
   this._patches = new NamedArray()
 
   for (let mod of MODULES) {
-    this.addPatch(mod, require.resolve(`./modules/${mod}`))
+    this.addPatch(mod, require(`./modules/${mod}`))
   }
 }
 


kuisathaverat:
jenkins run the tests please

Qard:
This would mean _all_ instrumentation patch files are loaded, even if they aren't used. There could be overhead implications to that. 🤔

blakeembrey:
@Qard I definitely agree it's an issue, would love to see a proper fix as the issue blocks running APM master on AWS Lambda. Let me know what solution you'd prefer.

Qard:
Why does it block running on AWS Lambda? It seems to me like this is only webpack-related, not Lambda-related. 🤔

blakeembrey:
@Qard You are correct, it blocks anyone that does AWS Lambda via Webpack. Which is at least me and anyone using https://github.com/serverless-heaven/serverless-webpack or https://github.com/zeit/ncc.

In any case, that doesn't seem relevant. I'd like to implement the solution you prefer so I can unblock myself.

blakeembrey:
@Qard if you're concerned about late requires for disabled hooks, maybe we just can disable the string check for now and check non-function until the patch is called? Or include `number` in the check?

Qard:
Ah, gotcha. I've never seen webpack used with Lambda before. 😅 

Anyway, there's definitely an issue here with how webpack changes how requires work and therefore breaks our require patching. Can you explain what you mean about the number thing? Do you mean the check in `Instrumentation.prototype._patchModule`? If we can get something in there to support the format that webpack uses, I'm definitely onboard with that. Just want to be sure we are careful about how we're applying this because it's easy to get wrong and break apps, which we obviously don't want to do. 🙀

blakeembrey:
@Qard Yes, Webpack returns a number instead of a string (ref: https://github.com/elastic/apm-agent-nodejs/issues/955#issuecomment-477503012). Would it be safe to assume that we don't do the checks when you register, but when it's used (e.g. let `require` fail explicitly)? Or just check for string or number? 

The other thing is that none of these hooks can actually run under webpack anyway, is there a way we could eventually split out instrumentation into a generic package and have it register modules explicitly instead of on require? I think this code might be fine if we decide it should just log a bunch on bad inputs, and instead only https://github.com/elastic/apm-agent-nodejs/pull/956 is necessary.

Qard:
Yeah, splitting out instrumentations is actually one of my goals of the patch registry stuff, but that will definitely take awhile. 😅 

At the moment, the model is to give a string name for a module to load and that will not be used until attempting to `require` the matching module at some point in the future. At _that_ point, it will try to `require` the patch file and _only then_ will log a warning if it's unable to resolve to a valid patch function.

Another possibility is to make string form load immediately and special-case the internal patch loading to use function form with internal require calls to load the corresponding patch functions. Something like this:

```js
for (let mod of MODULES) {
  this.addPatch(mod, (exports, name, version, enabled) => {
    const patch = require(`./modules/${mod}`)
    return patch(exports, name, version, enabled)
  })
}
```

lib/instrumentation/index.js
```diff
@@ -68,7 +68,7 @@ function Instrumentation (agent) {
   this._patches = new NamedArray()
 
   for (let mod of MODULES) {
-    this.addPatch(mod, require.resolve(`./modules/${mod}`))
+    this.addPatch(mod, require(`./modules/${mod}`))
```
watson:
```suggestion
    this.addPatch(mod, (exports, name, version, enabled) => {
      // Lazy require so that we don't have to use `require.resolve` which
      // would fail in combination with Webpack. For more info see:
      // https://github.com/elastic/apm-agent-nodejs/pull/957
      const patch = require(`./modules/${mod}`)
      return patch(exports, name, version, enabled)
    })
```

Qard:
@blakeembrey Do you still plan on working on this PR? I can take over, if you like. :)

blakeembrey:
Sorry, please feel free to!

PR #958, opened 2019-03-28
fix(fastify): send framework name/version with metadata

watson:
Forgot this in the original PR


watson:
I'll make another PR doing this differently

PR #966, opened 2019-03-30
feat: add agent.setFramework() method

### LINKED ISSUE #688: Framework name and version not always correctly detected
For details see elastic/apm-integration-testing#241

watson:
Fixes #688

This is how the new docs will look:

image

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -73,10 +73,10 @@ otherwise returns `false`.
 
 [small]#Added in: v2.8.0#
 
-* `options` {uri-object}[`<Object>`]
-** `name` {uri-string}[`<string>`] Framework name.
-** `version` {uri-string}[`<string>`] Framework version.
-** `overwrite` {uri-boolean}[`<boolean>`] If set to `false`,
+* `options` +{type-object}+
+** `name` +{type-string}+ Framework name.
+** `version` +{type-string}+ Framework version.
+** `overwrite` +{type-boolean}+ If set to `false`,
   the <<framework-name,`frameworkName`>> and <<framework-version,`frameworkVersion`>> provided as <<configuration,config options>> will not be overwritten.
   **Default:** `true`.
 
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -68,6 +68,21 @@ Use `isStarted()` to check if the agent has already started.
 Returns `true` if the agent has started, 
 otherwise returns `false`.
 
+[[apm-set-framework]]
+==== `apm.setFramework(options)`
+
+[small]#Added in: v2.8.0#
+
+* `options` {uri-object}[`<Object>`]
+** `name` {uri-string}[`<string>`] Framework name.
+** `version` {uri-string}[`<string>`] Framework version.
+** `overwrite` {uri-boolean}[`<boolean>`] If set to `false`,
+  the <<framework-name,`frameworkName`>> and <<framework-version,`frameworkVersion`>> provided as <<configuration,config options>> will not be overwritten.
+  **Default:** `true`.
+
+Set or change the <<framework-name,`frameworkName`>> or <<framework-version,`frameworkVersion`>> after the agent has started.
+These config options can also be provided as part of the <<configuration,regular agent configuration>>.
+
 [[apm-add-filter]]
 ==== `apm.addFilter()`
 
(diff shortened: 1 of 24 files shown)


docs/index.asciidoc
```diff
@@ -2,6 +2,10 @@
 :server-branch: 6.5
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
+:uri-string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type
```
bmorelli25: What if we build the links into the attributes to make referencing them even easier? So, something like this:
```suggestion
:uri-string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type[<string>]
```

Then, in the API docs, we reference them like this: `+{uri-string}+`. Saves us from having to put the entire link in the docs over and over.

docs/agent-api.asciidoc
```diff
@@ -68,6 +68,21 @@ Use `isStarted()` to check if the agent has already started.
 Returns `true` if the agent has started, 
 otherwise returns `false`.
 
+[[apm-set-framework]]
+==== `apm.setFramework(options)`
+
+[small]#Added in: v2.8.0#
+
+* `options` {uri-object}[`<Object>`]
+** `name` {uri-string}[`<string>`] Framework name.
```
bmorelli25: See comment below for an explanation. If you go with that recommendation, this line becomes:

```suggestion
** `name` +{uri-string}+ Framework name.
```

lib/agent.js
```diff
@@ -171,6 +171,14 @@ Agent.prototype.start = function (opts) {
   return this
 }
 
+Agent.prototype.setFramework = function ({ name, version, overwrite = true }) {
```
Qard: Any particular reason you chose to use an object here, versus just the three things as parameters? Anticipating collecting more info in the future?

test/instrumentation/modules/express/set-framework.js
```diff
@@ -0,0 +1,21 @@
...
+  assert.strictEqual(name, 'express')
+  assert.strictEqual(version, require('express/package').version)
+  assert.strictEqual(overwrite, false)
+}
+
+const assert = require('assert')
+
+require('express')
+
+assert.strictEqual(asserts, 1)
```
Qard: Does it work to just mix assert-only tests like this with our tape-based stuff?

test/types/index.ts
```diff
@@ -8,6 +8,12 @@ agent.start({
 function started (bool: boolean) {}
 started(agent.isStarted())
 
+agent.setFramework({})
+agent.setFramework({ name: 'foo' })
+agent.setFramework({ name: 'foo', version: 'bar' })
+agent.setFramework({ version: 'bar' })
```
Qard: Should we consider it valid to include `version` yet omit `name`? Not sure how useful it is to have a version for something when we don't know what it is. 😅

lib/agent.js
watson: No, my main concern was just not to have a boolean as a standalone argument as that's not really that user-friendly. How about something like this?

```js
function (name, version, { overwrite = true }) {...}
```

test/instrumentation/modules/express/set-framework.js
watson: It should work ok as we're just spawning `node <test-file>` from within `test/test.js`. We use the `assert` module in a few other tests as well.

The reason why I went with `assert` here was to reduce the number of modules being required to ensure there were no unintended side-effects, which was also why I added it as a separate file (I tried to integrate it into the main test file for each framework first, but that had major drawbacks).

test/types/index.ts
watson: Maybe the name was set via the `frameworkName` config option? I agree its probably not that common, but I didn't see a reason to disallow it.

lib/agent.js
watson: The main reason why I went with this API was so that you could provide `version` without having to provide the `name`

docs/index.asciidoc
watson: I like! This is clever

test/types/index.ts
Qard: Ah, that's fair.

lib/agent.js
Qard: I think it's fine as-is. Just wondered if there was a particular reason for the design which I was missing.

PR #972, opened 2019-04-02
feat(debug): initiate debugging on trace

### LINKED ISSUE #961: Change DEBUG=true to logLevel: 'trace'
To output the agent config today, you have to both set the config `logLevel: 'debug'` (or `'trace'`) and the environment variable `DEBUG=true`. It would be easier if you just had to set `logLevel: 'trace'`.

The reason why we went with the `DEBUG` environment variable originally was that it was easy to detect, so that [the entire code block](https://github.com/elastic/apm-agent-nodejs/blob/e1026382d8c6987137c00ffda83eb4e1acf292e6/lib/agent.js#L138-L161) could be skipped if it wasn't needed. But it's just too complicated of an API. Instead the if-statement should just check if `logLevel === 'trace'`

curtisjohnson1:
fixes #961 

### Checklist

- [x] Implement code
- [ ] Add tests
- [x] Update documentation

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -93,6 +93,6 @@ Please create a new topic in the https://discuss.elastic.co/c/apm[Elastic APM di
 To enable agent debugging and capture enough information,
 perform these steps:
 
-1. Start your app with the environment variable `ELASTIC_APM_LOG_LEVEL=debug` or add `logLevel: 'debug'` to the config file or `start(options)` call (see <<log-level,`logLevel`>> for details)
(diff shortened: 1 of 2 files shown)


curtisjohnson1:
No problem. I can do that

curtisjohnson1:
I've fixed the logger issue you mentioned, although it's squashed that commit into my original one. Apologies about that

watson:
jenkins run the test please

watson:
jenkins run the tests please

watson:
Thanks for your help 🥇 

I've just published a new version of the agent to npm with your commit

PR #974, opened 2019-04-02
fix: don't add response listener to outgoing requests

watson:
The very act of adding a `response` listener to outgoing requests could trigger a memory leak in case the user didn't also attach a `response` listener.

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -102,6 +102,8 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
 
   return function (orig) {
     return function (...args) {
+      // TODO: See if we can delay the creation of span until the `response`
+      // event is fired, while still having it have the correct stack trace
       var span = agent.startSpan(null, spanType)
       var id = span && span.transaction.id
 
@@ -144,6 +146,9 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
 
       span.name = req.method + ' ' + req._headers.host + url.parse(req.path).pathname
 
+      // TODO: Research if it's possible to add this to the prototype instead.
+      // Or if it's somehow preferable to listen for when a `response` listener
+      // is added instead of when `response` is emitted.
       const emit = req.emit
       req.emit = function (type, res) {
         if (type === 'response') onresponse(res)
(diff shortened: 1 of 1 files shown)


watson:
I'd like to rewrite a big chunk of the logic for instrumenting outgoing requests, but I think this is the lowest hanging fruit that we should fix first.

lib/instrumentation/http-shared.js
```diff
@@ -143,7 +143,12 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       ins.bindEmitter(req)
 
       span.name = req.method + ' ' + req._headers.host + url.parse(req.path).pathname
-      req.on('response', onresponse)
+
+      const emit = req.emit
+      req.emit = function (type, res) {
+        if (type === 'response') onresponse(res)
+        return emit.apply(req, arguments)
+      }
```
Qard: Is there some reason you didn't use shimmer here?

watson: I was concerned that maybe the emit function had already been shimmed and therefore the 2nd call would be ignored. But maybe that's not the case 🤔

Qard: Hmm...okay. It'd be nice if there was a better way to prevent multiple patching that didn't prevent _intentional_ multiple patching? 🤔

watson: I'll update it to use shimmer

watson: Oh wait, I actually remember another bigger reason now why I didn't use shimmer: This happens for every outgoing http request and I wanted to reduce the overhead as much as possible and the `shimmer.wrap` function has a slightly larger overhead. But it's probably not that big of a difference, so maybe I should just use shimmer?

Qard: Can we find a way to apply this patch on the prototype instead? I'd prefer we avoid building a new closure for every call. 🤔

Qard: Actually, if we're planning a rewrite anyway, it's probably fine to just land this as-is to fix the memory leak issue and just take a note somewhere to look at moving the patch to the prototype in the rewrite.

watson: 👍 I'll add a TODO comment

watson:
jenkins, run tests please

PR #975, opened 2019-04-02
[docs] Standardize API reference

### LINKED ISSUE #925: Standardize API docs
Today the standard in which the different APIs are documented varies quite a bit throughout the documentation. For consistency, it would be best to align on one way to document each API method.

Here's a proposal given in the form of an example of a documented API method where the first argument is required and the two latter arguments are optional. The style is mainly inspired by the [official Node.js documentation](https://nodejs.org/api/).

```asciidoc
[[apm-my-function]]
==== `agent.myFunction(foo[, options][, callback])`

[small]#Added in: v1.2.3#

* `foo` {uri-number}[`<number>`] Some description of this argument.
* `options` {uri-object}[`<Object>`]
** `bar` {uri-string}[`<string>`] Some description of this argument. **Default:** 42.
** `baz` {uri-boolean}[`<boolean>`] | {uri-symbol}[`<symbol>`] Some description of this argument.
* `callback` {uri-function}[`<Function>`]
** `err` {uri-error}[`<Error>`]
** `exitCode` {uri-number}[`<integer>`]
* Returns: <<agent-api,`<Agent>`>>

Full description goes here...
```

Somewhere central we need to define these attributes:

```asciidoc
:uri-number: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Number_type
:uri-string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type
:uri-boolean: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type
:uri-symbol: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Symbol_type
:uri-object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object
:uri-function: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function
:uri-error: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
```

Here's the rendered result:

image

Obviously, this documentation could maybe be generated based on a [TypeScript Declaration File](https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html) or [JSDoc](http://usejsdoc.org/). So we should consider that as well. The TypeScript file would only have the signatures and types of course, but nothing of the actual documentation, so that might be a no-go.

@Qard @bmorelli25: What do you think?

bmorelli25:
Here's a WIP PR that fixes #925. Still lots to clean up, but wanted to get this up.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -535,7 +535,7 @@ NOTE: If there's no active transaction available,
 `null` will be returned.
 
 [[apm-handle-uncaught-exceptions]]
-==== `apm.handleUncaughtExceptions(callback)`
+==== `apm.handleUncaughtExceptions([callback])`
 
 // [small]#Added in: #
 
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -65,12 +65,12 @@ otherwise returns `false`.
 
 [small]#Added in: v2.8.0#
 
-* `options` +{type-object}+
+* `options` +{type-object}+ The following options are supported:
 ** `name` +{type-string}+ Framework name.
 ** `version` +{type-string}+ Framework version.
 ** `overwrite` +{type-boolean}+ If set to `false`,
   the <<framework-name,`frameworkName`>> and <<framework-version,`frameworkVersion`>> provided as <<configuration,config options>> will not be overwritten.
-  **Default:** `true`.
+  *Default:* `true`.
 
 Set or change the <<framework-name,`frameworkName`>> or <<framework-version,`frameworkVersion`>> after the agent has started.
 These config options can also be provided as part of the <<configuration,regular agent configuration>>.
@@ -220,7 +220,7 @@ Tags are key/value pairs that are indexed by Elasticsearch and therefore searcha
 
 [small]#Added in: v1.5.0#
 
-* `tags` +{type-object}+
+* `tags` +{type-object}+ Contains key/value pairs:
 ** `name` +{type-string}+
 Any periods (`.`), asterisks (`*`), or double quotation marks (`"`) will be replaced by underscores (`_`),
 as those characters have special meaning in Elasticsearch
@@ -239,7 +239,6 @@ Tags are key/value pairs that are indexed by Elasticsearch and therefore searcha
 ==== `apm.captureError(error[, options][, callback])`
 
 // [small]#Added in: #
-// this section needs work
 
 * `error` - Can be either an `Error` object,
 a <<message-strings,message string>>,
@@ -416,11 +415,11 @@ NOTE: `apm.middleware.connect` _must_ be added to the middleware stack _before_ 
 
 * `name` +{type-string}+ The name of the transaction.
 You can always set this later via <<transaction-name,`transaction.name`>> or <<apm-set-transaction-name,`apm.setTransactionName()`>>.
-Defaults to `unnamed`
+*Default:* `unnamed`
 
 * `type` +{type-string}+ The type of transaction.
 You can always set this later via <<transaction-type,`transaction.type`>>.
-Defaults to `custom`
+*Default:* `custom`
 
 * `options` +{type-object}+ The following options are supported:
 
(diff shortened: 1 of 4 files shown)


bmorelli25:
Sorry this took so long, I got sucked into release projects. I was able to add a lot of the _added_ dates, but others were complicated or too time consuming to find.

docs/index.asciidoc
```diff
@@ -2,9 +2,13 @@
 :server-branch: 6.5
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
-:type-string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type[`<string>`]
-:type-boolean: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type[`<boolean>`]
-:type-object: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object[`<Object>`]
+:type-string: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type[<string>]
```
watson: 🤔 how come this works? I would have thought that we needed to add the inline code-block backtick (`), but even without them it all seems to render corrctly with a monospace font when used.

docs/agent-api.asciidoc
```diff
@@ -236,71 +214,46 @@ it will also get tagged with the same tags.
...
-
-[source,js]
-----
-apm.captureError(new Error('boom!'))
-----
-
-Arguments:
+// [small]#Added in: #
 
 * `error` - Can be either an `Error` object,
```
watson: We should add types for this as well

docs/agent-api.asciidoc
```diff
@@ -236,71 +214,46 @@ it will also get tagged with the same tags.
...
 a <<message-strings,message string>>,
 or a <<parameterized-message-object,special parameterized message object>>
 
 * `options` - The following options are supported:
```
watson: We should add the type for this as well

docs/agent-api.asciidoc
```diff
@@ -73,23 +65,20 @@ otherwise returns `false`.
...
 
 [[apm-add-filter]]
-==== `apm.addFilter()`
+==== `apm.addFilter(callback)`
 
-[source,js]
-----
-apm.addFilter(callback)
-----
+// [small]#Added in: #
```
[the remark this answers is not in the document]
bmorelli25: I figured it might be handy for later, but I can remove if you'd like

docs/agent-api.asciidoc
```diff
@@ -131,55 +120,43 @@ Though you can also use filter functions to add new contextual information to th
...
 [[apm-set-user-context]]
-==== `apm.setUserContext()`
+==== `apm.setUserContext(context)`
 
-[source,js]
-----
-apm.setUserContext(context)
-----
+[small]#Added in: 0.1.0#
+// note that id, username, email properties are all optional
```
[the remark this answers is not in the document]
bmorelli25: Ahh good catch, I need to update this

docs/index.asciidoc
bmorelli25: magic

You can't describe text format within the attribute, you have to do it where the attribute is referenced. It works because of the plus signs we use to reference the attributes: `+{type-object}+`. The plus signs do the same as the backticks, but they allow attribute substitution.

docs/agent-api.asciidoc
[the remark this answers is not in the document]
watson: No problem, feel free to keep these 👍 Just wanted to hear your thoughts

docs/index.asciidoc
watson: Ohhhh TIL! 🤓

bmorelli25:
Thanks for catching a few of those missed substitutions!

RE: long headlines - Yeah, this was one of my concerns as well. I was actually pleasantly surprised to see it only happens three times (2x agent API, 1x transaction API).   Unfortunately, we really have no _perfect_ solution for API documentation right now. There are going to be compromises, and this is such a minor compromise for such a big improvement 🤷‍♂️I'm okay with it

watson:
We'll just ignore the commit message lint error. I'll fix it when I merge

PR #981, opened 2019-04-05
fix(agent): set package.json

### LINKED ISSUE #960: Support running DEBUG=true together with -r elastic-apm-node/start
If the agent is started using the `-r` flag, and you try to set the environment variable `DEBUG=true`, the agent will throw the following exception:

```
internal/validators.js:125
    throw new ERR_INVALID_ARG_TYPE(name, 'string', value);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type object
    at validateString (internal/validators.js:125:11)
    at Object.dirname (path.js:1260:5)
    at Agent.start (/Users/watson/code/node_modules/elastic-apm-node/lib/agent.js:141:24)
    at Object.<anonymous> (/Users/watson/code/node_modules/elastic-apm-node/start.js:3:32)
    at Module._compile (internal/modules/cjs/loader.js:738:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:749:10)
    at Module.load (internal/modules/cjs/loader.js:630:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:570:12)
    at Function.Module._load (internal/modules/cjs/loader.js:562:3)
    at Module.require (internal/modules/cjs/loader.js:667:17)
```

The issue is that the [`_ancestors`](https://github.com/elastic/apm-agent-nodejs/blob/e1026382d8c6987137c00ffda83eb4e1acf292e6/lib/agent.js#L140) array contains `null` as the last element:

```js
[ '/Users/watson/code/node_modules/elastic-apm-node/index.js',
  '/Users/watson/code/node_modules/elastic-apm-node/start.js',
  null ]
```

### LINKED ISSUE #973: Debug mode throws if agent is started using -r elastic-apm-node/start
The following line will throw an exception if the agent was started using `node -r elastic-apm-node/start app.js`:

https://github.com/elastic/apm-agent-nodejs/blob/b1e2929022ae369649f5cc2bcab2bf25aaa512da/lib/agent.js#L140

The issue is that `_ancestors` in this case will look something like this:

```js
[ '/Users/watson/code/node_modules/elastic-apm-node/index.js',
  '/Users/watson/code/node_modules/elastic-apm-node/start.js',
  null ]
```

Notice how the last element is `null`, which is what will cause the exception.

Besides the last element being `null`, there's also an issue in that the users own script (`app.js` in this example) isn't listed here, which means we can't get the correct base-path, which means we can't read their package.json. So in cases where the agent is started via the `-r` command line flag, we'll not be able to read the content of the package.json unless we find another way.

curtisjohnson1:
Fixes elastic/apm-agent-nodejs#960 

This fix sets the users `package.json` into the `logger.trace` object and also removes any falsy values from the array returned from the _ancestors while loop. 

I believe this also Fixes elastic/apm-agent-nodejs#973 

### Checklist

- [x] Implement code
- [ ] Add tests

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -136,10 +136,8 @@ Agent.prototype.start = function (opts) {
     this._conf.active = false
     return this
   } else if (this._conf.logLevel === 'trace') {
-    var _ancestors = ancestors(module).filter(Boolean)
-    var userModule = process.argv[1]
-    _ancestors.push(userModule)
-    var basedir = path.dirname(_ancestors[_ancestors.length - 1])
+    var _ancestors = ancestors(module)
+    var basedir = path.dirname(process.argv[1])
     var stackObj = {}
     Error.captureStackTrace(stackObj)
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -136,7 +136,9 @@ Agent.prototype.start = function (opts) {
     this._conf.active = false
     return this
   } else if (this._conf.logLevel === 'trace') {
-    var _ancestors = ancestors(module)
+    var _ancestors = ancestors(module).filter(Boolean)
+    var userModule = process.argv[1]
+    _ancestors.push(userModule)
     var basedir = path.dirname(_ancestors[_ancestors.length - 1])
     var stackObj = {}
     Error.captureStackTrace(stackObj)


lib/agent.js
```diff
@@ -136,7 +136,9 @@ Agent.prototype.start = function (opts) {
     this._conf.active = false
     return this
   } else if (this._conf.logLevel === 'trace') {
-    var _ancestors = ancestors(module)
+    var _ancestors = ancestors(module).filter(Boolean)
+    var userModule = process.argv[1]
+    _ancestors.push(userModule)
     var basedir = path.dirname(_ancestors[_ancestors.length - 1])
```
Qard: Seems to me like `_ancestors` should be left as-is and, instead, the `basedir` should just be set to `path.dirname(process.argv[1])` rather than pulling off the ancestors list. What do you think, @watson?

watson: Interesting approach. I hadn't considered that. Lemme just check some edge cases...

watson: That all seems fine. I don't know why I didn't think of that approach previously 😄

watson:
Not sure why I opened 2 issues for the same thing btw, thanks for discovering that @curtisjohnson1

lib/agent.js
curtisjohnson1: Great. I'll re-implement to include @Qard's suggestion and set `basedir` to be `path.dirname(process.argv[1])`  

There does seem to be an issue with how ancestors loops through the module. It always pulls `null` into the array, which is why I filtered out the value. But that's perhaps an issue with `ancestors` or the `module` itself?

watson:
jenkins, run tests please

watson:
The Jenkins failures are unrelated (it's because we fixed a bug in a downstream dependency so that it now works as expected, which in turn changes some of the output of our tests when they run inside Docker containers, which our Jenkins tests does)

watson:
Thanks for fixing this. Your patch will part of the next release 🎉

curtisjohnson1:
Happy to help!

watson:
Now released in v2.9.0

PR #1006, opened 2019-04-12
chore: add code comment about not changing _context

watson:


lib/instrumentation/generic-span.js
```diff
@@ -10,7 +10,7 @@ module.exports = GenericSpan
 
 function GenericSpan (agent, type, opts) {
   this._timer = new Timer(opts.timer, opts.startTime)
-  this._context = TraceParent.startOrResume(opts.childOf, agent._conf)
+  this._context = TraceParent.startOrResume(opts.childOf, agent._conf) // _context is used by the OT bridge, and should unfortunately therefore be considered public
```
Qard: Perhaps we should also make a major bump to the OT bridge _now_ to use the new API, then bump our dependency on that so there's some extra overlap time rather than a hard break at the next major of the agent?

watson: That would still leave users exposed who use an old version of the OT bridge. If they upgraded the agent to a newer version without upgrading the OT bridge it would blow up. The _only_ way to make a safe upgrade path is by bumping the major version of this agent.

Qard: Yes, but if they use a `^2.0.0` of our agent, and we bump the dependency within our agent, the would be able to use that early. I don't mean that we should _remove_ support for `_context` before a major bump in the agent, just that we should shift the community early rather than doing a hard-break that requires timed major releases of the OT bindings _and_ the agent together. By updating the OT bindings now, we don't have to worry about that extra work later when we will already be likely dealing with a bunch of other breaking stuff.

watson: No matter if we do it now or later, it will be the same amount of work as I see it. It requires a major bump no matter what. When we do that major bump shouldn't change anything for the end user

Qard: It's only a breaking change from one side though. Changing the agent requires the OT interface is also immediately changed, because it removes a feature being depended on. Changing the OT bindings, on the other hand, does _not_ require changing the agent, because the feature it depends on is already released as a minor. The OT bindings could just bump its own major and release immediately with a peer dependency range of `^2.9.0`, then the community can upgrade that any time between now when the agent bumps to `3.x` rather than being _required_ to bump both together later.

It doesn't likely benefit the users much right now, but it _does_ allow us to split up the work a bit to make it easier on ourselves later.

PR #1013, opened 2019-04-16
fix(express): improve names for routes added via app.use()

### LINKED ISSUE #1008: Express transaction names broken in v2.6.0+
**To Reproduce**
```js
require('elastic-apm-node').start();

const express = require('express');
const server = express();

// apm-agent-nodejs < 2.6.0: setting default transaction name: GET /some-path
// apm-agent-nodejs >= 2.6.0: setting default transaction name: GET unknown route
server.use('/some-path', (req, res) => {
    res.send('Hi');
});

server.listen(1212);
```

After a quick search, elastic/apm-agent-nodejs#842 looks like a possible reason.

watson:
Previously all `app.use()` routes would be treated as unknown routes. With this commit, only the following routes will be considered unknown:

- `app.use(fn)`
- `app.use('/', fn)`

All other routes, e.g. `app.use('/foo', fn)`, will have their transactions named accordingly. Note that the HTTP method will still be part of the transaction name, even if the route is defined using `app.use()`. I.e. `GET /foo` and `POST /foo` will be part of two different transaction groups even though they might be served by the same router function.

Fixes elastic/apm-agent-nodejs#1008

### DIFF
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -9,7 +9,7 @@ try {
 }
 var symbols = require('../symbols')
 
-function ensureSlash (value) {
+function normalizeSlash (value) {
   return value[0] === '/' ? value : '/' + value
 }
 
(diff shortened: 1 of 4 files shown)


MartinKolarik:
Any chance this can be merged and published as a bugfix soon?

Qard:
This is still in-progress and @watson is on vacation right now. He should get back to it when he returns.

watson:
@Qard Should be ready for review now. A lot in the diff is white-space related, so it reads better if you ignore that

Qard:
Seems to be failing tests.

watson:
Weird - not sure what changed after I rebased. Will take a look tomorrow

watson:
Ah, it's the new handle properties test that's failing now which came from `master` because of the rebase. Somehow that test doesn't like this new route naming logic

watson:
@Qard Fixed

PR #1019, opened 2019-04-16
chore: rename tags to labels

### LINKED ISSUE #962: Rename tags to labels
To align with [ECS](https://github.com/elastic/ecs), we're renaming "tags" to "labels". The old "tags" apis should still be available until the next major version.

Qard:
I've renamed tags to labels in the API, as suggested by [ECS](https://github.com/elastic/ecs). This includes renaming in the docs. I'm unsure if the old anchors should be left. Do we want to leave the whole sections in-tact with some sort of deprecation notice? I tried searching the docs for other deprecation warnings and didn't find anything to copy. 🤔 

Fixes #962

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -212,7 +212,7 @@ You can set multiple labels on the same transaction.
 If an error happens during the current transaction,
 it will also get tagged with the same label.
 
-Tags are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via <<apm-set-custom-context,`apm.setCustomContext()`>>).
+Labels are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via <<apm-set-custom-context,`apm.setCustomContext()`>>).
 
 
 [[apm-add-labels]]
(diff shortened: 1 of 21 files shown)


bmorelli25:
I'll look more in depth at this tomorrow, but I want to make a few comments while they're fresh in my mind:

* This change will impact at least one documentation link from Kibana to the agent. See my https://github.com/elastic/apm-server/issues/2111#issuecomment-483740288. Easy enough for me to change in Kib, just don't want to forget about it.

* There is a fun way to "deprecate" old anchors and we should do that here. We remove the anchors from their current location and add them to a `redirects` file. This will make sure any old links remain working. See this file: https://github.com/elastic/apm-agent-nodejs/blob/9e3696cb25b03b2c68a7c4b7ab9dedc54309bdd3/docs/redirects.asciidoc#L1-L9

* I wonder if we can add something like this for each change? Below or on top of the "Added in" tag.

```asciidoc
[small]#Renamed from `apm.setTag()` to `apm.setLabel()`: v1.5.0#
```

Let me know your thoughts. I'm happy to fill out the redirects file and add in the "renamed" tags to this PR tomorrow.

Qard:
I added a commit with those API change notes in the docs.

bmorelli25:
@Qard - I added two commits to this PR:

* docs: tags to labels f9af72c - fixes a few additional cases of tags
* docs: add redirects for old headers 8f867ab - adds redirects for the old headers, which I've confirmed are working

watson:
@bmorelli25 since we're still supporting the old API, shouldn't we keep the docs for those around but just add a deprecation notice in the docs to use the new API?

bmorelli25:
@watson Yes, you're right. Qard mentioned deprecation early on and I missed that. I had it in my head that the endpoints were being removed for some reason. 

@Qard - I pushed one more commit to this PR.

This commit leaves the original api endpoint documentation, and adds new sections for the new endpoints below it. It also adds a big **DEPRECATED** warning that links to the current function.

Here's what this looks like. Thoughts?
Screen Shot 2019-04-18 at 8 25 03 AM

Qard:
If no further changes are required, I intend to land this on Monday.

PR #1031, opened 2019-04-25
added support for default tags

derbl4ck:
This pull request adds support for "defaultTags" applied to all transactions and spans. This solves an Issue according to:
- Elastic Support Case 00327560
- Thomas Watson's (wa7son) comment ["It's currently not possible to set "global" tags etc that will automatically be applied to all transactions."](https://discuss.elastic.co/t/apm-nodejs-agent-how-to-use-setusercontext-settag-setcustomcontext/158800/6)
- and is might be related with [Issue 42](https://github.com/elastic/apm/issues/42)

### Checklist

- [x] Implement code
- [ ] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -237,6 +237,23 @@ it will also get tagged with the same label.
 
 Labels are key/value pairs that are indexed by Elasticsearch and therefore searchable (as opposed to data set via <<apm-set-custom-context,`apm.setCustomContext()`>>).
 
+[[apm-set-default-tags]]
+==== `apm.setDefaultTags({ [name]: value })`
+
+[small]#Added in: v2.10.0#
+
+* `tags` +{type-object}+ Contains key/value pairs:
+** `name` +{type-string}+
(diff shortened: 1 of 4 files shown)


elasticcla:
Hi @derbl4ck, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/apm-agent-nodejs/pull/1031.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

derbl4ck:
@elasticcla fixed :)

derbl4ck:
@Qard I think it's related with your [commit](https://github.com/elastic/apm-agent-nodejs/pull/1020/commits/30e1a7449dbc62df4b57463d11376f09cad2505b)?

derbl4ck:
Since tags should be renamed to labels in the future, there is no reason to implement the provided function. In the future, the environment variable `ELASTIC_APM_GLOBAL_LABELS` will realize the desired behavior.

Qard:
Yes, this is directly supported by my PR. We generally go for config options wherever possible, rather than functions like this. It makes controlling that more flexible. It could come from environment variables, a config file or code.

PR #1033, opened 2019-04-26
added doc example for globalLabels env var

derbl4ck:
This pr just adds an example of how to configure the new globalLabels environment variable.

### Checklist

- [x] Update documentation


elasticcla:
Hi @derbl4ck, we have found your signature in our records, but it seems like you have signed with a different e-mail than the one used in your Git [commit](https://github.com/elastic/apm-agent-nodejs/pull/1033.patch). Can you please add both of these e-mails into your Github profile (they can be hidden), so we can match your e-mails to your Github profile?

PR #1041, opened 2019-04-30
fix(config): do not use ELASTIC_APM_ prefix for k8s

### LINKED ISSUE #1028: Remove "ELASTIC_APM_" from  k8s data environment variables
Node.js implementation of this feature https://github.com/elastic/apm/issues/21#issuecomment-483229490 should be changed to align with the other agents.

Qard:
This makes kubernetes environment variables consistent with other agents while not breaking the prior behaviour. We should consider deprecating and removing the old environment variables in the next major.

Fixes #1028 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -641,7 +641,7 @@ it will be parsed out of the `/proc/self/cgroup` file.
 ==== `kubernetesNodeName`
 
 * *Type:* String
-* *Env:* `ELASTIC_APM_KUBERNETES_NODE_NAME`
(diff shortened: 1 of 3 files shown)


Qard:
@bmorelli25 What about docs redirects?

bmorelli25:
Good to go! None of the section IDs (e.g. `[[kubernetes-node-name]]`) need to change, so none of the urls will change.

Qard:
Ah, right. Thanks. :)

PR #1049, opened 2019-05-06
docs: document additional capture error options

### LINKED ISSUE #932: Document missing captureError options
The `options` that can be given to the `captureError` function are not all documented.

Currently only the following options are [documented](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-capture-error):

- `timestamp`
- `message`
- `user`
- `custom`

The following are not:

- `request`
- `response`
- `handled`
- `tags`

kevindoveton:
Fixes elastic/apm-agent-nodejs#932 

I've only just started working with this library so I may have misjudged what some of the options are doing.

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -324,13 +324,13 @@ it's possible to use this option to supply an additional message string that wil
 
 ** `custom` - See <<metadata,metadata section>> for details about this option
 
-** `request` +{type-incomingmessage}+ You can associate an error with information about the incoming request to gain additional context such as the request url, headers and cookies. 
-However in most cases the agent will detect if an error was in response to a http request and automatically add the request details for you. 
-See <<http-requests, http requests section>> for more details.
+** `request` +{type-incomingmessage}+ You can associate an error with information about the incoming request to gain additional context such as the request url, headers, and cookies.
+However, in most cases, the agent will detect if an error was in response to an http request and automatically add the request details for you.
+See <<http-requests,http requests section>> for more details.
 
-** `response` +{type-serverresponse}+ You can associate an error with information about the http response to get additional details such as status code and headers. 
-However in most cases the agent will detect if an error occured during a http request and automatically response details for you. 
-See <<http-responses, http responses section>> for more details.
+** `response` +{type-serverresponse}+ You can associate an error with information about the http response to get additional details such as status code and headers.
+However, in most cases, the agent will detect if an error occured during an http request and automatically add response details for you.
+See <<http-responses,http responses section>> for more details.
 
 ** `handled` +{type-boolean}+ Adds additional context to the exception to show whether the error is handled or uncaught.
 *Default:* `true`.
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -324,6 +324,14 @@ it's possible to use this option to supply an additional message string that wil
 
 ** `custom` - See <<metadata,metadata section>> for details about this option
 
+** `request` You can associate an error with an instance of `http.IncomingMessage` to gain additional context such as the request url, headers and cookies. However in most cases the agent will detect if an error was in response to a  http request and automatically add the request details for you. See <<http-requests, http requests section>> for more details.
+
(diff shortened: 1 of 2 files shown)


docs/agent-api.asciidoc
```diff
@@ -324,6 +324,14 @@ it's possible to use this option to supply an additional message string that wil
 
 ** `custom` - See <<metadata,metadata section>> for details about this option
 
+** `request` You can associate an error with an instance of `http.IncomingMessage` to gain additional context such as the request url, headers and cookies. However in most cases the agent will detect if an error was in response to a  http request and automatically add the request details for you. See <<http-requests, http requests section>> for more details.
```
watson: Besides adding a few line-breaks to adhere to our [One Sentence Per Line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) rule from the AsciiDoc Recommended Practices, I took the liberty to move the information about `http.IncomingMessage` to the beginning of the line so it's similar to the rest of the API docs ☺️

This, however, requires that you add the following line to the `docs/index.asciidoc` file:

```asciidoc
:type-incomingmessage: https://nodejs.org/api/http.html#http_class_http_incomingmessage[<http.IncomingMessage>]
```

Here's the complete change:

```suggestion
** `request` +{type-incomingmessage}+ You can associate an error with information about the incoming request to gain additional context such as the request url, headers, and cookies.
However, in most cases, the agent will detect if an error was in response to an http request and automatically add the request details for you.
See <<http-requests,http requests section>> for more details.
```

docs/agent-api.asciidoc
```diff
@@ -324,6 +324,14 @@ it's possible to use this option to supply an additional message string that wil
...
+
+** `response` You can associate an error with an instance of `http.ServerResponse` to get additional details such as status code and headers. However in most cases the agent will detect if an error occured during a http request and automatically response details for you. See <<http-responses, http responses section>> for more details.
```
watson: Similar to above, this change suggestion requires that you add the following line to the `docs/index.asciidoc` file:

```asciidoc
:type-serverresponse: https://nodejs.org/api/http.html#http_class_http_serverresponse[<http.ServerResponse>]
```

Here's the complete change:

```suggestion
** `response` +{type-serverresponse}+ You can associate an error with information about the http response to get additional details such as status code and headers.
However, in most cases, the agent will detect if an error occurs during an http request and automatically response details for you.
See <<http-responses,http responses section>> for more details.
```

docs/agent-api.asciidoc
```diff
@@ -324,6 +324,14 @@ it's possible to use this option to supply an additional message string that wil
...
+
+** `handled` +{type-boolean}+ true by default. Adds additional context to the exception to show whether the error is handled or uncaught.
```
watson: We try to put the default value at the end using this format (it's not used throughout all our docs yet, so it might not have been obvious, but slowly we're transitioning to this):

```suggestion
** `handled` +{type-boolean}+ Adds additional context to the exception to show whether the error is handled or uncaught.
*Default:* `true`.
```

docs/agent-api.asciidoc
```diff
@@ -324,6 +324,14 @@ it's possible to use this option to supply an additional message string that wil
...
+
+** `tags` +{type-object}+ Add aditional context with tags, these tags will be merged with the tags inside the current transaction.
```
watson: We [recently](https://github.com/elastic/apm-agent-nodejs/pull/1019) renamed tags to labels. While the `tags` option is still supported for backward compatibility, we could just decide to document the `labels` version to keep the docs simple:

```suggestion
** `labels` +{type-object}+ Add additional context with labels, these labels will be added to the error along with the labels from the current transaction.
See the <<apm-add-labels,`apm.addLabels()`>> method for details about the format.
```

kevindoveton:
Thanks for the suggestions, I've added those in :)

watson:
jenkins, run tests please

watson:
Thanks a lot!

The commit message linter will most likely fail when the tests are done running, but we'll just ignore that and fix it when we merge 🙂

watson:
Thanks again. You should see it live in our `master` docs in about an hour: https://www.elastic.co/guide/en/apm/agent/nodejs/master/agent-api.html#apm-capture-error

Then it will be part of our `current` docs when we do the next release.

PR #1050, opened 2019-05-06
[APM-CI][All] Make the trigger by comment case-insensitive

v1v:


watson:
The tests failed because of an invalid commit message. Since this change is trivial, I'll just change the commit message when I merge 🙂

PR #1058, opened 2019-05-08
chore(package): update ws to version 7.0.0

watson:


watson:
The failing test is unrelated and can be safely ignored

PR #1074, opened 2019-05-21
fix: url.parse expects req.url not req

Qard:
The `parseurl` module receives a request object while `url.parse` expects to receive the url string. This should fix that. I'm not sure how to test this though, as it loads the url module as a fallback when first running the module. An alternative would maybe be to just add `parseurl` to the dependencies and _always_ use it. 🤔 

### Checklist

- [x] Implement code
- [ ] Add tests

### DIFF
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -4,7 +4,8 @@ var parseUrl
 try {
   parseUrl = require('parseurl')
 } catch (e) {
-  parseUrl = require('url').parse
+  const url = require('url')
+  parseUrl = req => url.parse(req.url)]
 }
 var symbols = require('../symbols')
 
--- a/lib/instrumentation/express-utils.js
+++ b/lib/instrumentation/express-utils.js
@@ -5,7 +5,7 @@ try {
   parseUrl = require('parseurl')
 } catch (e) {
   const url = require('url')
-  parseUrl = req => url.parse(req.url)]
+  parseUrl = req => url.parse(req.url)
 }
 var symbols = require('../symbols')
 


lib/instrumentation/express-utils.js
```diff
@@ -4,7 +4,8 @@ var parseUrl
 try {
   parseUrl = require('parseurl')
 } catch (e) {
-  parseUrl = require('url').parse
+  const url = require('url')
+  parseUrl = req => url.parse(req.url)]
```
watson:
```suggestion
  parseUrl = req => url.parse(req.url)
```

watson:
Good catch! [Our tests](https://github.com/elastic/apm-agent-nodejs/blob/0aaa7e0ad792d5a4ce94e0b0f7421c66a7407641/test/config.js#L560-L590) should actually already have caught this as far as I can see. We should look into why they didn't as part of this PR 🤔 

I don't like adding `parseurl` as a dependency as it's messing with the `req` object in a way normal users might not like.

watson:
Found the reason why our tests didn't catch this:

image

Because we already have `connect`, `express`, `finalhandler`, and `koa` as devDependencies, we indirectly have `parseurl` installed when running our tests and therefore our tests will not throw an error and will instead use the fallback when requiring the `parseurl` module. So that code-path isn't tested at all.

Not sure how we can test this without literally running `rm -fr node_modules/parseurl` ... maybe that's what you meant and I just didn't get it 😅

watson:
The failing tests are just because Travis is being difficult this morning. Let's just ignore those

watson:
After giving it some thought, I think it's best to just merge this as is so we can get it released today. I'll create a separate issue to update the tests

Qard:
Yes, the indirect dev dependency is why I wasn't sure if there was a good way to test this and why I felt the existing tests didn't cover it.

PR #1080, opened 2019-05-21
test: verify internal http requests are ignored

### LINKED ISSUE #738: Add test to verify that APM Server requests are not traced

Qard:
Fix #738
Close #794

### DIFF
--- a/test/integration/skip-internal-http-spans.js
+++ b/test/integration/skip-internal-http-spans.js
@@ -1,0 +1,61 @@
+'use strict'
+
+const http = require('http')
+const zlib = require('zlib')
+
+const getPort = require('get-port')
+const ndjson = require('ndjson')
+const test = require('tape')
+
+const agent = require('../../')
+
+test('should not throw on socket close', t => {
+  getPort().then(port => {
+    agent.start({
+      serviceName: 'test',
+      serverUrl: 'http://localhost:' + port,
+      captureExceptions: false,
+      metricsInterval: 0
+    })
+
+    const seen = {
+      metadata: 0,
+      transaction: 0,
+      span: 0,
+      error: 0
+    }
+
+    const expected = {
+      metadata: 1,
+      transaction: 1,
+      span: 0,
+      error: 0
+    }
+
+    const server = http.createServer((req, res) => {
+      req
+        .pipe(zlib.createGunzip())
+        .pipe(ndjson.parse())
+        .on('data', data => {
+          const key = Object.keys(data)[0]
+          seen[key] = (seen[key] || 0) + 1
+        })
+        .on('end', () => res.end())
+    })
+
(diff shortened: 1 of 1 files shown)


watson:
I just tried to re-introduce the error that was fixed in #735 by applying this patch and that this test is supposed to valdiate:

```diff
diff --git a/lib/agent.js b/lib/agent.js
index 1c5d3035..5fdecb63 100644
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -109,7 +109,7 @@ Agent.prototype._config = function (opts) {
     ? parseUrl(this._conf.serverUrl)
     : { host: 'localhost:8200', port: '8200' }
 
-  this._conf.serverHost = url.host
+  this._conf.serverHost = url.hostname
   this._conf.serverPort = parseInt(url.port, 10)
 }
 
```

But this test doesn't fail as I would expect it to 🤔

Qard:
As far as I can tell, the issue doesn't actually _exist_ in v2, even without that "fix" in-place, as our http client pulls the http request out-of-band from the request with a stream that isn't patched with `bindEmitter`.

watson:
I found the issues:

1. All the modules were required before the agent was started, so hence the agent did not register any outgoing http request to the APM Server because it was not instrumenting the `http` module
1. The flush would produce an outgoing http request, but since there was no second flush, the agent would not get a chance to record a span for that outgoing http request

I've pushed two commits to fix these things. But they fail in a way that leads me to believe that we're still recording spans for http requests to the APM Server.

watson:
Did some more digging and found the following:

1. The mock APM Server in this test instruments itself, which leads to weird unrelated errors. It would probably be better if the APM Server ran in another process somehow
1. Since the transaction is ended before the agent flushed, the span for the outgoing http request is never created

watson:
It should technically work now. I worked around the issue with the APM Server being instrumented by just ignoring all transactions and only looking at spans and metadata. This is not a beautiful fix and I don't like it, but it works.

watson:
If you try to apply the diff that I pasted above you'll see that the test fails - so the test seems to be working as intended

Qard:
Thanks for picking this up. I had just quickly thrown this together at the end of the day yesterday and wasn't able to dig to deep into it. It looks good to me now. Apparently I can't approve the changes though as it's my PR and the PR owner can't approve it. 🤷‍♂

watson:
Cool! The failing tests are unrelated. I'll go ahead and merge 😃

PR #1083, opened 2019-05-22
test(travis): split TAV builds into 5 groups per Node.js version

watson:
Travis was getting to a point where the builds would time out because they reached the two-hour max limit. By splitting the TAV tests into five groups per Node.js version instead of 4, we should hopefully get a little more breathing room.


Qard:
Even after splitting, it _still_ appears to be timing out... 😞

watson:
This is madness!

watson:
I finally got it to a level where all 5 builds take approximately the same amount of time to complete

Issue #1072, opened 2019-05-19
Update "npm run docs" to use the new AsciiDoctor builder


watson:
When running `npm run docs` from the command line, it still uses the old asciidocs builder. We should remove the old code and update it to use the new AsciiDoctor builder

watson:
@bmorelli25 We'd love your help with this when you have time 🙂

bmorelli25:
This looks like fun! I can definitely help

Issue #1071, opened 2019-05-19
Remove Travis docs build


watson:
With the new AsciiDoctor build that's now started automatically via Jenkins, we no longer need the build that runs inside Travis. This build also uses the old asciidoc builder, which is no longer used in production and it might therefore give false results.

PR #1085, opened 2019-05-22
[docs] update npm run docs

bmorelli25:
* Fixes https://github.com/elastic/apm-agent-nodejs/issues/1072 by using docker and asciidoctor to build the docs.
* Fixes https://github.com/elastic/apm-agent-nodejs/issues/1071 by removing the docs build from travis

I was bored last night so I played around with this. I think I need your help, not sure what the best way to have a "build" and an "open" script is. In this potential solution I added a `docs:open` script that opens the documentation and a `docs:build` script that just builds it. Not sure what a best practice would be though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
-    "docs": "./docs/scripts/build_docs.sh apm-agent-nodejs ./docs ./build",
+    "docs:open": "PREVIEW=1 ./docs/scripts/build_docs.sh apm-agent-nodejs ./docs ./build",
+    "docs:build": "./docs/scripts/build_docs.sh apm-agent-nodejs ./docs ./build",
     "lint": "standard",
     "lint:fix": "standard --fix",
     "lint:commit": "test/lint-commits.sh",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -5,7 +5,8 @@
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
-    "docs": "./docs/scripts/build_docs.sh apm-agent-nodejs ./docs ./build",
```
watson: We have a Travis test that currently expects this script to be there:

https://github.com/elastic/apm-agent-nodejs/blob/9abb4aa926fdca65e70c570735f8f91985905952/.travis.yml#L63-L67

So we should either update that or remove the build from Travis, which would also fix #1071

bmorelli25: Ah, thanks, forgot we talked about this as well!

package.json
```diff
@@ -5,7 +5,8 @@
...
+    "docs:open": "PREVIEW=1 ./docs/scripts/build_docs.sh apm-agent-nodejs ./docs ./build",
```
Qard: I believe you could also do `PREVIEW=1 npm run docs:build` to reduce duplication.

bmorelli25: oooo I like that! Let me try

bmorelli25: Nice, it works!

bmorelli25:
I'll let one of you look at the failing travis build and decide if this is ok to merge or not

watson:
It's a bug that's fixed in master, but unrelated to this PR, so I'll go ahead and merge 😊

PR #1089, opened 2019-05-24
fix: don't add traceparent header to signed AWS requests

### LINKED ISSUE #1055: Don't add traceparent header to some/all AWS endpoints
We recently stopped adding the `elastic-apm-traceparent` header to outgoing HTTP requests to `*.s3.amazonaws.com` hosts (see #952).

But if the user is using regional hosts like `s3.us-west-2.amazonaws.com`, this will still break. At a minimum, we should therefore also apply the same logic to hosts with the pattern `*.s3.*.amazonaws.com`.

But should strongly consider expanding the query to `*.amazonaws.com`. But before doing so we should consider if this will affect our ability to capture distributed traces with AWS Lambda functions somewhere in the trace (and potentially also other AWS services).

watson:
If the outgoing HTTP request is signed with an AWS version 4 signature, this traceparent header will not be added by the agent.

Fixes #1055

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -4,8 +4,6 @@ var url = require('url')
 
 var endOfStream = require('end-of-stream')
 
-const s3Host = /\.s3\.amazonaws\.com$/
-
 const transactionForResponse = new WeakMap()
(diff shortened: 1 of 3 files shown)


watson:
Added tests

PR #1092, opened 2019-05-27
fix(metrics): report correct CPU usage on Linux

### LINKED ISSUE #1068: CPU metrics mismatch
**Describe the bug**
APM agent often pushes 0% process and system CPU usage, it should have at least some value greater than zero as the process is still running. Even comparing the stats on the container itself does not match the CPU usage provided by APM. 

**To Reproduce**
1. Deploy APM enabled node application to a container (Linux). 
2. Compare container stats to APM data
3. Compare process metrics within the container to APM data
4. Screenshot of this comparison attached.

**Expected behavior**
APM agent should push correct CPU metrics instead of 0%

- OS: Linux
- Node.js version: 8.10.0
- APM Server version: 6.7.0
- Agent version: 2.6.0, 2.11.1

- [X] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)

**Additional context**

- Screenshot

metrics-comapre

watson:
CPU usage metrics are reported to the APM Server at a set interval. Instead of looking at the time spend on CPU since the last reporting, the CPU usage metrics on Linux only looked at the last 100th fraction of that time. E.g. if reporting metrics every 10 seconds (the default), the CPU usage numbers in each reporting would only cover the last 100ms - leaving a big gap of 9900ms unreported.

In reality, the entire 10 seconds was calculated, but in chunks of 100ms. And only the calculation for the last 100ms interval was kept. So not only was the data from the 9900ms lost, it also incurred an
unnecessary calculation overhead. This overhead is now also avoided.

Fixes #1068 

(snuck in some minor unrelated cleanup as will - hope you don't mind)

### DIFF
--- a/lib/metrics/platforms/linux/stats.js
+++ b/lib/metrics/platforms/linux/stats.js
@@ -173,13 +173,13 @@ class Stats {
     this.inProgress = false
   }
 
-  start (interval, cb) {
+  start (reportingIntervalInSeconds, cb) {
     if (this.timer) this.stop()
 
     const timer = setInterval(() => {
       if (cb) cb(this.toJSON())
       this.reload()
-    }, interval)
+    }, reportingIntervalInSeconds * 1000)
 
     this.timer = timer
     timer.unref()
--- a/lib/metrics/platforms/linux/index.js
+++ b/lib/metrics/platforms/linux/index.js
@@ -2,9 +2,9 @@
 
 const Stats = require('./stats')
 
-module.exports = function createSystemMetrics (registry, interval) {
+module.exports = function createSystemMetrics (registry, reportingIntervalInSeconds) {
   const stats = new Stats()
-  stats.start(interval)
+  stats.start(reportingIntervalInSeconds)
 
   const metrics = [
     'system.cpu.total.norm.pct',
--- a/lib/metrics/registry.js
+++ b/lib/metrics/registry.js
@@ -12,7 +12,7 @@ const createSystemMetrics = process.platform === 'linux'
 const defaultReporterOptions = {
   defaultDimensions: {
     hostname: os.hostname(),
-    env: process.env.NODE_ENV || 'development'
+    env: process.env.NODE_ENV || 'development' // TODO: Use environment config option once it lands
   }
 }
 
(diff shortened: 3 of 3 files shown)


watson:
Now that this PR fixes our custom interval, I think we suddenly have a new issue:

Since the Linux metrics have its own `setInterval` separate from the metrics reporter library, I guess we run the risk of a race condition: I.e. just before the first batch of metrics are reported to the APM Server, our custom internal fires and calculates the metrics and stores them in memory, then a few milliseconds later the other interval fires and the in-memory metrics are reported. We then wait for the `metricsInterval` to pass, but due to random fluctuations in the timings of the intervals, the metrics reporter interval callback kicks off first and reports the same metrics as the last run, and then a few milliseconds later, our custom interval fires and updates the in-memory metrics - too late for the metrics reporter to pick it up.

The reverse could also happen: That the in-memory metrics are updated twice within the same reporting interval.

I think however that it's best to land this PR first as the issue this PR is fixing is much more important. And then we can make another issue for fixing this new one.

lib/metrics/platforms/linux/stats.js
```diff
@@ -173,13 +173,13 @@ class Stats {
...
 
-  start (interval, cb) {
+  start (reportingIntervalInSeconds, cb) {
     if (this.timer) this.stop()
 
     const timer = setInterval(() => {
       if (cb) cb(this.toJSON())
       this.reload()
-    }, interval)
+    }, reportingIntervalInSeconds * 1000)
```
Qard: Ah, right...forgot about this. I actually fixed it in the runtime metrics PR in a different place. https://github.com/elastic/apm-agent-nodejs/pull/1021/files#diff-a419a7bc99109d34ec9019e4409917fdR27

watson: Yeah, I was going back and forth about if I should do it when I called the `createSystemMetrics` function or inside where it used in the `setInterval` function. I settled on the latter because it felt a tiny bit better to me keeping it the same unit until it was specifically needed in another unit.

Qard: Sounds good. I'll make sure the runtime metrics PR aligns now. :)

PR #1098, opened 2019-05-29
Add support for renamed hapi => @hapi/hapi module

### LINKED ISSUE #1097: HapiJS update to new package paths
HapiJS moved their packages to '@hapi/hapi' instead of just 'hapi' and new versions are published there. The agent needs updated to support newer hapijs installs.

Update the agent to support both old and new versions and paths

I'm not sure there are alternatives...

**Additional context**
Creating the issue so I can get a fix and PR started :)

bgardner87:
Added support for HapiJS modules updated location at @hapi/hapi

I copied the existing hapi implementations and trimmed out most of the semver checks as they don't apply to the new module paths. Copied and trimmed the tests as well. This is working in my tests so far but hoping to get code review and suggestion process started.

Closes #1097

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


watson:
Thank for letting us know about this and creating the PR 💯 

I'd prefer if we could find a way to avoid duplicating the files as that will be hard to keep in sync when fixing bugs, though that will require quite a bit of refactoring. But I'll be happy to help you land this PR if you have the time 😃 

To achieve this, one approach would be to allow a module to have zero or more "aliases", described in an array of strings where the first element in the array is also the file path for the patched module on disk:

```diff
diff --git a/lib/instrumentation/index.js b/lib/instrumentation/index.js
index acbe95aa..f0748654 100644
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -23,7 +23,7 @@ var MODULES = [
   'generic-pool',
   'graphql',
   'handlebars',
-  'hapi',
+  ['hapi', '@hapi/hapi'],
   'http',
   'https',
   'http2',
@@ -70,12 +70,17 @@ function Instrumentation (agent) {
   this._patches = new NamedArray()
 
   for (let mod of MODULES) {
-    this.addPatch(mod, (exports, name, version, enabled) => {
-      // Lazy require so that we don't have to use `require.resolve` which
-      // would fail in combination with Webpack. For more info see:
-      // https://github.com/elastic/apm-agent-nodejs/pull/957
-      const patch = require(`./modules/${mod}`)
-      return patch(exports, name, version, enabled)
+    if (!Array.isArray(mod)) mod = [mod]
+    const pathName = mod[0]
+
+    mod.forEach(mod => {
+      this.addPatch(mod, (exports, name, version, enabled) => {
+        // Lazy require so that we don't have to use `require.resolve` which
+        // would fail in combination with Webpack. For more info see:
+        // https://github.com/elastic/apm-agent-nodejs/pull/957
+        const patch = require(`./modules/${pathName}`)
+        return patch(exports, name, version, enabled)
+      })
     })
   }
 }
```

To avoid duplicating the tests, we could just wrap the `basic.js` test file in a loop that essentially runs the tests twice once for each module name.

Duplicating the `set-framework.js` test file is the easiest I think as it needs a clean state of the process when it runs. And it doesn't contain that many lines anyway, so I'm ok with that.

You'd also need to duplicate the `hapi` section in the `.tav.yml` file and add `@hapi/hapi` to the TAV tests in both `.travis.yml` and `test/.jenkins_tav.yml`.

watson:
Glad I can help. Forgot to mention that you probably want to add `@hapi/hapi` to `docs/supported-technologies.asciidoc` as well.

bgardner87:
OK, I think it's closer now. I'm not sure exactly what to do with .tav.yml, it doesn't seem like an update would be necessary?

Let me know what you think when you have a moment.

I'm currently having difficulty executing the test suite locally, the cassandra container won't stay up for me yet. I can run the hapi tests directly though and they are all passing, so moving the right direction.

watson:
What I was thinking regarding `.tav.yml` was something like this:

```diff
diff --git a/.tav.yml b/.tav.yml
index 5c131e8e..1c8a0fc7 100644
--- a/.tav.yml
+++ b/.tav.yml
@@ -211,6 +211,10 @@ hapi-async-await:
   node: '>=8.2'
   versions: '>=17.0.0'
   commands: node test/instrumentation/modules/hapi/basic.js
+'@hapi/hapi':
+  node: '>=8.2'
+  versions: '>=17.0.0'
+  commands: node test/instrumentation/modules/hapi/basic.js
 
 tedious:
   name: tedious
```

However, I just realized I hadn't thought it all the way through when it comes to having `test/instrumentation/modules/hapi/basic.js` run the tests for both `hapi` and `@hapi/hapi` as this means that the tests are not separated when running these TAV tests.

Either we just ignore it and accept that some tests are run multiple times, or we could improve the `basic.js` script slightly to take an argument, e.g:

```diff
diff --git a/.tav.yml b/.tav.yml
index 5c131e8e..a587a0f6 100644
--- a/.tav.yml
+++ b/.tav.yml
@@ -205,12 +205,16 @@ pug:
 hapi-no-async-await:
   name: hapi
   versions: '>=9.0.1 <17.0.0'
-  commands: node test/instrumentation/modules/hapi/basic.js
+  commands: node test/instrumentation/modules/hapi/basic.js hapi
 hapi-async-await:
   name: hapi
   node: '>=8.2'
   versions: '>=17.0.0'
-  commands: node test/instrumentation/modules/hapi/basic.js
+  commands: node test/instrumentation/modules/hapi/basic.js hapi
+'@hapi/hapi':
+  node: '>=8.2'
+  versions: '>=17.0.0'
+  commands: node test/instrumentation/modules/hapi/basic.js @hapi/hapi
 
 tedious:
   name: tedious
```

Then inside of `test/instrumentation/modules/hapi/basic.js` you could check if `process.argv[2]` was set. If it was you could use that for the content of `moduleNames`, if not, you'd default to both modules.

What do you think of that approach? Is it too "magic"?

Qard:
Could maybe split the basic test into a shared file that just exports everything as a function which accepts the module name and have two separate test files that use it.

```js
// shared file
module.exports = (name) => {
   // ...
}

// hapi/basic.js
require('./shared-file')('hapi')

// @hapi/hapi/basic.js
require('./shared-file')('@hapi/hapi')
```

bgardner87:
Finally getting back to this, sorry for the delay!

I updated the .tav.yml and refactored the basic.js file based on the above suggestions, splitting it in 2 and using a shared file.

lib/instrumentation/index.js
```diff
@@ -70,12 +70,17 @@ function Instrumentation (agent) {
...
-      // Lazy require so that we don't have to use `require.resolve` which
-      // would fail in combination with Webpack. For more info see:
-      // https://github.com/elastic/apm-agent-nodejs/pull/957
-      const patch = require(`./modules/${mod}`)
-      return patch(exports, name, version, enabled)
+    if (!Array.isArray(mod)) mod = [mod]
+    const pathName = mod[0]
+
+    mod.forEach(mod => {
+      this.addPatch(mod, (exports, name, version, enabled) => {
```
Qard: Rather than this nested loop here, maybe we should just make `addPatch(...)` also accept arrays of module names and put the loop there? 🤔

watson: I'm good either way. I guess it doesn't hurt to make the API a little more flexible to support stuff like this better.

@bgardner87 to support this, just add some logic to the `addPatch` function so it also supports being called with an array of strings. I think it would make sense to also add this support to the `removePatch` function as well. After updating the functions, you also need to update [the API docs](https://github.com/elastic/apm-agent-nodejs/blob/master/docs/agent-api.asciidoc#apmaddpatchname-handler) to explain that the first argument to the function can be an array of strings.

.tav.yml
```diff
@@ -210,16 +210,19 @@ hapi-no-async-await:
...
   versions: '>=17.0.0'
   commands:
-    - node test/instrumentation/modules/hapi/basic.js
+    - node test/instrumentation/modules/hapi/basic-legacy-path.js
     - node test/instrumentation/modules/hapi/set-framework.js
-
+'@hapi/hapi':
+  node: '>=8.2'
+  versions: '>=17.0.0'
+  commands: node test/instrumentation/modules/hapi/basic.js
```
watson: You need to add `set-framework-2.js` to this list of commands as well (just as above)

bgardner87:
I believe I have it all, let me know what you think!

bgardner87:
Looking at API docs now

lib/instrumentation/index.js
```diff
@@ -70,30 +70,41 @@ function Instrumentation (agent) {
...
-  this._startHook()
+  modules.forEach(mod => {
+    const type = typeof handler
+    if (type !== 'function' && type !== 'string') {
+      this._agent.logger.error('Invalid patch handler type:', type)
+      return
+    }
+
+    this._patches.add(mod, handler)
+    this._startHook()
```
Qard: The call to `this._startHook()` could be moved out of the loop as it doesn't need to be called between patches being added.

lib/instrumentation/index.js
```diff
@@ -70,30 +70,41 @@ function Instrumentation (agent) {
...
-  if (type !== 'function' && type !== 'string') {
-    this._agent.logger.error('Invalid patch handler type:', type)
-    return
-  }
+Instrumentation.prototype.addPatch = function (modules, handler) {
+  if (!Array.isArray(modules)) modules = [modules]
 
-  this._patches.add(name, handler)
-  this._startHook()
+  modules.forEach(mod => {
```
Qard: This could be `for (let mod of modules) {`.

lib/instrumentation/index.js
```diff
@@ -70,30 +70,41 @@ function Instrumentation (agent) {
   this._patches = new NamedArray()
 
   for (let mod of MODULES) {
+    if (!Array.isArray(mod)) mod = [mod]
+    const pathName = mod[0]
```
Qard: Maybe `const pathName = Array.isArray(mod) ? mod[0] : mod`, so you don't need to build an array you don't actually need.

lib/instrumentation/index.js
```diff
@@ -70,30 +70,41 @@ function Instrumentation (agent) {
...
 
-Instrumentation.prototype.removePatch = function (name, handler) {
-  this._patches.delete(name, handler)
-  this._startHook()
+Instrumentation.prototype.removePatch = function (modules, handler) {
+  if (!Array.isArray(modules)) modules = [modules]
+
+  modules.forEach(mod => {
+    this._patches.delete(mod, handler)
+    this._startHook()
```
Qard: Again, the call to `this._startHook()` doesn't need to be made between each change, you could just put it after the loop. Also, for/let loop too, please. :)

PR #1106, opened 2019-05-31
feat(config): add environment option

### LINKED ISSUE #88: Add app.environment to payload sent to APM Server

Qard:
~~Depends on elastic/apm-nodejs-http-client#58~~
Fixes #88

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


test/config.js
```diff
@@ -59,7 +59,8 @@ var optionFixtures = [
   ['kubernetesNamespace', 'KUBERNETES_NAMESPACE'],
   ['kubernetesPodName', 'KUBERNETES_POD_NAME'],
   ['kubernetesPodUID', 'KUBERNETES_POD_UID'],
-  ['usePathAsTransactionName', 'USE_PATH_AS_TRANSACTION_NAME', false]
+  ['usePathAsTransactionName', 'USE_PATH_AS_TRANSACTION_NAME', false],
+  ['environment', 'ENVIRONMENT']
```
watson: I think we can add a test to test the default value like so:

```suggestion
  ['environment', 'ENVIRONMENT', 'development']
```

Qard:
Is there a reason _not_ to just land this? Seems like from it being already implemented in the other agents that at least the agent side of things is basically already decided? 🤔

watson:
It was considered a buggy feature as the UI would show invalid data if used. We therefore wanted to pull it from the other agents, but that would have been a breaking change. So instead of opted for a warning in the documentation not to use it for the agents who already released it. Until the UI is released that supports it, it will technically still be problematic if used.

docs/configuration.asciidoc
```diff
@@ -91,6 +91,17 @@ var options = {
...
+===== `environment`
+
+* *Type:* String
+* *Default:* `process.env.NODE_ENV || 'development'`
+* *Env:* `ELASTIC_APM_ENVIRONMENT`
+
+The environment name is automatically logged along with all errors and transactions.
+If you want to overwrite this,
+use this option.
+
```
bmorelli25: @Qard Can you add the following as a new paragraph after line 103:
```
Environments allow you to easily filter data on a global level in the APM UI.
It's important to be consistent when naming environments across agents.
See {kibana-ref}/filters.html#environment-selector[environment selector] in the Kibana UI for more information.

NOTE: This feature is fully supported in the APM UI in Kibana versions >= 7.2.
You must use the query bar to filter for a specific environment in versions prior to 7.2.
```

Qard: fixed

watson:
@bmorelli25 The docs-build is still failing. Is that because the new docs have not yet published? If so, how can we merge this before 7.2?

bmorelli25:
@watson unfortunately, the link checker in CI still isn't perfect. I'm not sure why it's failing, but this link does work. I think this PR is ok to merge as is.

PR #1114, opened 2019-06-05
feat: allow manual instrumentation when instrument is set to false

### LINKED ISSUE #858: instrument: false should disable automatic instrumentation
Today setting `instrument: false` will disable both automatic and manual instrumentation. This doesn't really make any sense. Instead I think this option should only disable automatic instrumentation.

This should probably be coordinated with the other agents. Today we seem to all have implemented `instrument` a little bit differently. E.g. in Python setting it to `false`, will only disable automatic creation of spans, in Java it will allow manual instrumentation, in Ruby and Node.js it will not etc.

Qard:
Fixes elastic/apm-agent-nodejs#858

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -130,9 +130,7 @@ Instrumentation.prototype.clearPatches = function (modules) {
 Instrumentation.modules = Object.freeze(MODULES)
 
 Instrumentation.prototype.start = function () {
-  if (!this._agent._conf.instrument) return
   if (this._started) return
-
   this._started = true
 
   if (this._agent._conf.asyncHooks && semver.gte(process.version, '8.2.0')) {
@@ -164,7 +162,7 @@ Instrumentation.prototype._startHook = function () {
   this._agent.logger.debug('adding hook to Node.js module loader')
 
   this._hook = hook(this._patches.keys, function (exports, name, basedir) {
-    var enabled = !disabled.has(name)
+    var enabled = self._agent._conf.instrument && !disabled.has(name)
     var pkg, version
 
     if (basedir) {
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -119,7 +119,7 @@ You must use the query bar to filter for a specific environment in versions prio
 * *Default:* `true`
 * *Env:* `ELASTIC_APM_INSTRUMENT`
 
-A boolean specifying if the agent should collect performance metrics for the app.
+A boolean specifying if the agent should automatically apply instrumentation to supported modules when they are loaded.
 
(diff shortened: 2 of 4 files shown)


lib/instrumentation/index.js
```diff
@@ -138,7 +136,7 @@ Instrumentation.prototype._startHook = function () {
   this._agent.logger.debug('adding hook to Node.js module loader')
 
   this._hook = hook(this._patches.keys, function (exports, name, basedir) {
-    var enabled = !disabled.has(name)
+    var enabled = !disabled.has(name) && self._agent._conf.instrument
```
Qard: Flip config order because checking boolean is faster than set

watson:
We should probably put this in v3.0.0 as it could be considered a breaking change

Qard:
True. Marking as breaking change.

watson:
Jenkins run the tests please

watson:
I'm able to reproduce the Travis failure locally. The issue is that the CPU just starts spinning forever during the `test/config.js` test. It's happening on the latest Node.js v10, but not v12 when running `nyc node test/config.js`, but doesn't happen when running `node test/config.js`. So something that nyc is doing is causing something

watson:
It's now Monday, and I upgraded to Safari 13 over the weekend, which required me to agree on a new XCode License Agreement before git would work in the terminal again. And I'm no longer able to reproduce the error locally

image

watson:
Ok, I can reproduce it again - but not every time. Turns out the error doesn't always happen, so sometimes you need to run the test command a few times before it occurs. So I usually run it like this:

```
while true; do ./node_modules/.bin/nyc node test/config.js || break; done
```

I made a core dump while the script was stuck in the infinite loop and discovered that it was related to async hooks. When disabling async hooks in the `instrument: false allows manual instrumentation` test, the error disappeared. This obviously isn't a solution but might help in determining the cause.

watson:
I think I narrowed it down to multiple async hooks being enabled during the `test/config.js` tests. If I put the return value of `asyncHooks.createHook()` into a global variable on which I call `.disable()` if it's already present when trying to create a new hook, then the problem goes away.

Working on a fix...

watson:
I've fixed the CI issue in #1373. Once that's merged, `master` needs to be merged into this PR and it should hopefully be ready to merge

watson:
I think you accidentally force pushed a version that removed my last commit. I assume that was just because you rebased with `master` without pulling first?

I'll force push a new version with the extra commits. Hope you don't mind 🙂

watson:
jenkins run the tests please

Qard:
Ah, sorry about that. I missed your commits. 😅

Qard:
jenkins run the tests please

PR #1118, opened 2019-06-11
ci(travis): add Node.js v12 nightly (no async hooks) build

watson:
This will not trigger until in master, so there's no need to wait for passing tests


watson:
Pressed "update branch" by a mistake instead of merge. Will just merge now and not wait for the tests to run again

PR #1127, opened 2019-06-12
test: fix broken intake API schema tests

watson:
These broke because 4 new files were added in:
https://github.com/elastic/apm-server/pull/2265


watson:
Travis is currently broken. I've reported it them and will merge without Travis tests for now.

PR #1129, opened 2019-06-12
Force Trusty distribution instead of relying on default

cotsog:
See https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@
 # different machines:
 # - With `sudo: false`: https://travis-ci.org/opbeat/opbeat-node/builds/252301998
 # - With `sudo: required`: https://travis-ci.org/opbeat/opbeat-node/builds/252302576
+sudo: required
 dist: trusty
 
 language: node_js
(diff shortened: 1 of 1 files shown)


watson:
Wow, thanks for the quick PR ❤️

watson:
Possibly `sudo: required` doesn't do any difference anymore, but if removing it, we should also remove the above comment. So, for now, I'll just leave it in. Will look into if it can be removed in another PR

watson:
All looks good now 👍 I'll fix the commit message linting error during "Squash and merge".

Thanks again for the quick help 💯

Qard:
We should probably also test if xenial works and maybe just switch to that. Trusty is _ancient_ at this point. 😅

watson:
Travis unfortunately doesn't support Cassandra on Xenial yet, and hence our tests fails when they run on Xenial - which is what just happened as we were rolled over automatically from Trusty to Xenial a few hours ago

PR #1130, opened 2019-06-12
chore(config): sort configs alphabetically

Qard:
This should make it easier to catch when the Typescript typings get out-of-sync, and also adds some configs that were missing in the typings.


watson:
Good catch 💯

PR #1139, opened 2019-06-18
chore(package): bump all devDependencies

watson:


watson:
Jenkins is unstable, so the failing tests are unrelated

Issue #1131, opened 2019-06-13
feat(instrumentation): add memcached


mikefaut:
**Is your feature request related to a problem? Please describe.**
There currently isn't support for memcached.

**Describe the solution you'd like**
A memcached handler could be added to /lib/instrumentation/modules

**Describe alternatives you've considered**
Currently I overwrite/wrap the memcached requests with a span.

**Additional context**
The code I use is below, which I use to wrap the set and get functions. 
```
function wrapFunctionWithAPMSpan(func, name) {
    return (...args) => {
        let span = apm.startSpan(name)
        return func(...args).then((result) => {
            span.end()
            return result
        }).catch((err) => {
            span.end()
            throw (err)
        })
    }
}
```

watson:
Hi @mikefaut, thanks for opening this issue 😃

I would love to see automatic instrumentation of memcached in the agent as well 👍 Our roadmap is currently pretty full, but hopefully, we can get to it soon.

mikefaut:
Are there any caveats for implementing this? I might look at doing it when/if I get a chance. I'd probably use the redis instrumentation as a template.

watson:
I don't think so, but it depends a little on the internals of the memcached module. Some modules have obvious hooks where it's easy to get notified when for instance a new query starts and when it ends, and what type of query it is, while others make it a lot harder.

watson:
The API to add automatic instrumentation for new modules is public btw, so you can easily try it out without having to modify the module source code. You just need to call the [`agent.addPatch`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-add-patch) function with a patch function.

We'll happily help you through the PR process as well.

mikefaut:
I wrote a patch for this. I'll still need to do a bit of testing but I could use some feedback. What is the process for PRs? Are there some docs? Also do you have ideas on how I can do version testing?

watson:
That's great to hear. We have documented our contribution guidelines in the [`CONTRIBUTING.md`](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md) file. I suggest you take a look at that. In there's also a specific section for adding support for new modules:
https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#adding-support-for-new-modules

Point number 6 in the list is a bit outdated as the `test/.jenkins_tav.yml` file have been moved to a different folder, so you can skip that for now.

[...]

PR #1144, opened 2019-06-19
feat(instrumentation): add memcached

mikefaut:
This instruments memcached so that key functions have spans. 

Resolves: elastic/apm-agent-nodejs#1131 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,46 @@
...
+  return memcached
+
+  // Wrap the generic command that is used to build touch, get, gets etc
+  function wrapCommand (original) {
+    return function wrappedCommand () {
+      if (typeof arguments[0] === 'function') {
+        var query = arguments[0]()
+        // If the callback is not a function the user doesn't care about result
+        if (typeof query.callback === 'function') {
+          var span = agent.startSpan(`memcached.${query.type}`)
```
mikefaut: I'm wondering if this should include the key as well. 
It makes it look like:
image
Open to suggestions if its better to add it as a label etc...

Qard: Probably better as a label. @watson Do you agree?

test/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,52 @@
...
+          })
+        })
+      })
+    })
+  })
+})
+
+function resetAgent (cb) {
+  agent._instrumentation.currentTransaction = null
+  agent._transport = mockClient(11, cb)
```
watson: The callback to `resetAgent` is supposed to be called when the mock client has sent the number of events indicated here. In the above test, you only create 7 spans + 1 transaction - so 8 events in total - but it's waiting for 11. So the callback is never called. If you change it to 8, it will call the callback and you can do your final tests inside that.

```suggestion
  agent._transport = mockClient(8, cb)
```

test/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,52 @@
...
+          cache.touch('foo', 300, (err) => {
+            t.error(err)
+            cache.del('foo', (err) => {
+              t.error(err)
+              cache.get('foo', (err, data) => {
+                t.error(err)
+                t.equal(data, undefined)
+                agent.endTransaction()
+                agent.flush()
+                t.end()
```
watson: Just remove this line as the test should not end until the `resetAgent` callback is called (within which you need to check that the `data` is as expected).

test/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,52 @@
...
+  serviceName: 'test',
+  secretToken: 'test',
+  captureExceptions: false,
+  metricsInterval: 0
+})
+
+var test = require('tape')
+var mockClient = require('../../_mock_http_client')
+test(function (t) {
+  resetAgent(function (data) {
```
watson: Within this callback, you need to check that the events sent by the agent are as expected.

All the events will be contained in the `data` argument, which will have the following format:

```js
{
  length: 8,
  spans: [...],
  transactions: [...],
  errors: [...]
}
```

You can see an example of how we validate this in the handlebar tests:

https://github.com/elastic/apm-agent-nodejs/blob/a907f573864da7e391491622012642ad811cac5b/test/instrumentation/modules/handlebars.js#L19-L39

lib/instrumentation/modules/memcached.js
watson: @Qard I had the page open in GitHub while doing the review, so didn't see your comment until now. A label is fine as well, but you can't see it at a glance. So if this is relevant information on the waterfall screen, then it's probably better to have it in the name. Is there any downside to showing it in the name?

watson:
I noticed that you only show support for the latest version of memcached. I assume it's just because you haven't checked if it works with versions prior to that?

If you want to check which versions are supported, you can use the `tav` command to run the test file against each released version.

For a simple compatibility check run the following command, where `*` is any semver range you want to check, `*` being all versions (this requires memcached to be running locally):

```
./node_modules/.bin/tav memcached '*' 'node test/instrumentation/modules/memcached.js' --compat
```

The get details about failures, run it without the `--compat` argument, possibly with `-q` instead to be less chatty

watson:
Forgot to mention that it would also be great if you could add the ability to start/stop memcached not in Docker. You can do that by applying this diff:

```diff
diff --git a/test/script/local-deps-start.sh b/test/script/local-deps-start.sh
index 6c13b585..216a1c66 100755
--- a/test/script/local-deps-start.sh
+++ b/test/script/local-deps-start.sh
@@ -4,5 +4,6 @@ pg_ctl -D /usr/local/var/postgres start
 mongod --fork --config /usr/local/etc/mongod.conf --pidfilepath /tmp/mongod.pid >/tmp/mongod.log 2>&1
 elasticsearch -p /tmp/elasticsearch.pid --daemonize && echo 'waiting for elasticsearch to start...' && ./node_modules/.bin/wait-on tcp:9200
 cassandra -p /tmp/cassandra.pid &> /tmp/cassandra.log
+memcached -d -P /tmp/memcached.pid
 redis-server /usr/local/etc/redis.conf --daemonize yes
 mysql.server start
diff --git a/test/script/local-deps-stop.sh b/test/script/local-deps-stop.sh
index e983f428..cf5cfb14 100755
--- a/test/script/local-deps-stop.sh
+++ b/test/script/local-deps-stop.sh
@@ -4,5 +4,6 @@ pg_ctl -D /usr/local/var/postgres stop
 kill `cat /tmp/mongod.pid`
 kill `cat /tmp/elasticsearch.pid`
 kill `cat /tmp/cassandra.pid`
+kill `cat /tmp/memcached.pid`
 redis-cli shutdown
 mysql.server stop
```

lib/instrumentation/modules/memcached.js
Qard: Depends. With SQL we have the summary which removes unique data and just retains the query structure, allowing similar queries to overlap, here we don't have that which could result in an explosion of unique names. Would that matter much for fingerprinting, or do you think just going by the type is enough? 🤔

mikefaut:
Thanks for the feedback. I should get a chance on Thursday to fix this.

lib/instrumentation/modules/memcached.js
watson: I was thinking it was kind of like including the table name as we do in SQL - but I guess the analogy doesn't quite hold. Let's leave it out of the name and just make sure to include it using `span.setDbContext()` instead.

I was thinking we should include the query type in there as well so that it will be `get foo` instead of just `foo`. Or `getMulti foo,bar` instead of just `foo,bar`:

```js
span.setDbContext({ statement: `${query.type} ${query.key}`, type: 'memcached' })
```

mikefaut:
I've tried to run the tav command to test versions and versions lower than 2.2.0 fail to install because of node-gyp issues. I dunno when I'll get some more time to look at it.

I can't seem to get the env var for the memcached test to be set. I tried looking at how it is done for redis but seem to be missing something.

test/docker-compose.yml
```diff
@@ -102,6 +102,17 @@ services:
       timeout: 10s
       retries: 5
 
+  memcached:
+    image: memcached:alpine
+    user: 11211
```
Qard: I assume it's a mistake, setting the user to the port number? 🤔

test/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,58 @@
...
+    t.equal(data.transactions.length, 1)
+    t.equal(data.spans.length, 7)
+    for (var span of data.spans) {
+      t.ok(/^memcached\.(get|set|replace|touch|delete)$/.test(span.name))
+      t.ok(/^db\.memcached\.(get|set|replace|touch|delete)$/.test(span.type))
+      t.ok(/^(get|set|replace|touch|delete) foo$/.test(span.context.db.statement))
```
Qard: Rather than this overly permissive loop, can you check for each span in order, so we know we're actually getting all different values?

test/docker-compose.yml
mikefaut: The docker image for memcached has a user with 11211. There is a version of the alpine image that errors when it is run as root.

test/instrumentation/modules/memcached.js
mikefaut: Will do.

test/docker-compose.yml
Qard: Weird choice by whoever made that image. 🤔 

Anyway, ignore this thread. :)

Qard:
Looks like there's a failure on Node.js 6.x due to `@hapi/hoek` using not-yet-supported syntax. Is hoek in the dependency tree of memcached somewhere, or that something else depending on that version? We _should_ have `hapi` locked down already, but maybe they released something new that broke things without a major bump? 🤔

mikefaut:
@hapi/hoek isn't in the dependency tree for memcached. 
https://github.com/hapijs/hoek/commit/af4868e206424891eeea919741a2cab6dfe4d411#diff-9b79c950bc87af2a515ddd86ee317d30L17
It seems they have patched the issue we're seeing. Maybe we can explicitly use a version of hapi/hoek.
I've updated the tests to be more strict about checking for values.
The ci still seems to have an issue with the MEMCACHED_HOST env var, not sure if I'm missing something or if memcached is failing to run.

test/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,74 @@
...
+              })
+            })
+          })
+        })
+      })
+    })
+  })
+})
+
+function resetAgent(cb) {
```
Qard: To appease the linter gods. 😅 

```suggestion
function resetAgent(cb) {
function resetAgent (cb) {
```

Qard:
The test results seem to vary between Node.js versions. https://travis-ci.org/elastic/apm-agent-nodejs/builds/563899835

Are you able to identify if there are some functional differences to the library behaviour depending on the Node.js version? We may need some version checks or to make the test code less order-dependent.

mikefaut:
There don't seem to be any differences between node versions. It's strange though, it looks like the first two spans are swapping. It is recording the get span before the set span. Also when ELASTIC_APM_ASYNC_HOOKS=false, all the tests pass for all the versions.

Qard:
Ah, sounds like a data race. Can you change the test to make it not dependent on the _order_ the spans are received but rather only on the _presence_ of certain spans? The `findObjInArray` function in `test/_utils.js` might help you here, it's used in a few places in the test suite to find spans that match a given name without expecting them to be at a specific _position_.

test/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,74 @@
...
+          t.equal(data, 'fizz')
+          cache.touch('foo', 300, (err) => {
+            t.error(err)
+            cache.del('foo', (err) => {
+              t.error(err)
+              cache.get('foo', (err, data) => {
+                t.error(err)
+                t.equal(data, undefined)
+                agent.endTransaction()
+                agent.flush()
```
watson: The test is hanging after it completes because the socket to memcached is still open. To make the test exist right when it finishes, just call `cache.end()`:

```suggestion
                agent.flush()
                cache.end()
```

.tav.yml
```diff
@@ -254,3 +254,7 @@ fastify:
 finalhandler:
   versions: '*'
   commands: node test/instrumentation/modules/finalhandler.js
+
+memcached:
+  versions: '>2.2.0'
```
watson: I just did a quick test and it looks to me like we support all the way down to v2.2.0, so this should be changed to:

```suggestion
  versions: '>=2.2.0'
```

(which is also what the if-statement in the patch file says)

docs/supported-technologies.asciidoc
```diff
@@ -83,6 +83,7 @@ The Node.js agent will automatically instrument the following modules to give yo
 |https://www.npmjs.com/package/jade[jade] |>=0.5.6 |Will instrument compile and render calls
 |https://www.npmjs.com/package/pug[pug] |'>=0.1.0' |Will instrument compile and render calls
 |https://www.npmjs.com/package/ioredis[ioredis] |>=2.0.0 <5.0.0 |Will instrument all queries
+|https://www.npmjs.com/package/memcached[memcached] |>=2.2.2  |Will instrument all commands.
```
watson: To align with the if-statement in the patch file:

```suggestion
|https://www.npmjs.com/package/memcached[memcached] |>=2.2.0  |Will instrument all commands.
```

mikefaut:
I can't seem to access: continuous-integration/jenkins/pr-merge to see what is wrong there.
It looks like memcached isn't running/picking up the env var again. Do I need to make appveyor install scripts like there are for cassandra etc?

watson:
Thanks for the update @mikefaut - I'll take a look now.

Regarding `continuous-integration/jenkins/pr-merge` your issue accessing, then Jenkins was updated over the weekend with a security patch witch somehow makes you see a 404 page the first time you click it. **If you just refresh the page when you get the 404, you should get in.** But obviously this is not a good experience. I already reported this bug internally yesterday, and we're working on a fix.

watson:
The test failure is just a linting issue. Travis is probably easier for you to access and get an overview with. Here's the error: https://travis-ci.org/elastic/apm-agent-nodejs/jobs/568777798#L626

test/docker-compose.yml
watson: Docker complains about this not being a string. If I wrap it in quotes, it seems to work

.ci/docker/docker-compose.yml
```diff
@@ -101,6 +101,17 @@ services:
       interval: 30s
       timeout: 10s
       retries: 5
+  
+  memcached:
+    image: memcached:alpine
+    user: 11211
```
watson: Docker complains about this not being a string. If I wrap it in quotes, it seems to work

lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,47 @@
...
+          arguments[0] = function queryCompiler () {
+            return query
+          }
+        }
+      }
+      return original.apply(this, arguments)
+
+      function wrapCallback (cb) {
+        return function wrappedCallback () {
+          if (span && !span.ended) span.end()
```
watson: Just a tiny nit when you're fixing the docker-compose.yml issue anyway: There's no need to check if the span is ended as it can only end in one way in this implementation.

```suggestion
          if (span) span.end()
```

lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,47 @@
...
+
+  // Wrap the generic command that is used to build touch, get, gets etc
+  function wrapCommand (original) {
+    return function wrappedCommand () {
+      if (typeof arguments[0] === 'function') {
+        var query = arguments[0]()
+        // If the callback is not a function the user doesn't care about result
+        if (query && typeof query.callback === 'function') {
+          var span = agent.startSpan(`memcached.${query.type}`, `db.memcached.${query.type}`)
+          span.setDbContext({ statement: `${query.type} ${query.key}`, type: 'memcached' })
```
watson: Missed this one the last time around:

The return value of `startSpan()` can be `null` depending on how the agent is operating, so we need to add a guard on this line to be sure we don't get an error.

But since we already have one on the line after, I suggest we just wrap both this and the next line in an if-block, eg:

```js
if (span) {
  span.setDbContext({ statement: `${query.type} ${query.key}`, type: 'memcached' })
  var id = span.id
}
```

test/docker-compose.yml
watson:
```suggestion
    user: "11211"
```

.ci/docker/docker-compose.yml
watson:
```suggestion
    user: "11211"
```

watson:
jenkins run the tests please

watson:
Just ignore the Jenkins test errors. it's failing because it needs me to first approve you as a contributor. It's a new system that we've set up, and I haven't yet figured out how it works 😅

watson:
Ah, forgot about AppVeyor. I'll just look into how to setup memcached on Windows.. hang on

watson:
Got memcached installed on AppVeyor now, but the latest one I could find was version 1.4.4. Unfortunately the `TOUCH` command wasn't added until v1.4.8. So one of the tests are still failing. I'll see if I can find a newer version...

watson:
Turns out memcached is actually not officially supported on Windows (the v1.4.4 build I found was made by a 3rd party). So there's no reason to try to get the tests to run. I've pushed a new commit which just bails if we detect windows

watson:
jenkins run the tests please

watson:
@v1v I just noticed that Jenkins didn't run the memcached TAV tests, so I just added you as a reviewer as it might be related to the changes to the `.ci` files in this PR. Could you verify if all looks ok?

v1v:
Let me see what's going on

v1v:
For some reason, I'm not able to reproduce the same context locally, and neither the log provides further insights on what's the result of git region match method. I'll add some debug to be able to know what it does under the hood.

watson:
@v1v could it be because those memcached test/lib files were not in `master`, but are added in this commit?

v1v:
> @v1v could it be because those memcached test/lib files were not in master, but are added in this commit?

I know what you meant, but I don't think so, as the git diff happens between the origin/master...$commit, more precisely the last commit is: https://github.com/elastic/apm-agent-nodejs/pull/1144/commits/9a2e5572934cecc7b4e4e490edf949b625a44e8b

```
git diff --name-only origin/master...9a2e5572934cecc7b4e4e490edf949b625a44e8b
```

So if I run the same command locally, as shown below, it does list all the files which have been changed so far:

```
git diff --name-only origin/master...9a2e5572934cecc7b4e4e490edf949b625a44e8b
.ci/.jenkins_tav.yml
.ci/docker/docker-compose-all.yml
.ci/docker/docker-compose-memcached.yml
.ci/docker/docker-compose-node-test.yml
.ci/docker/docker-compose.yml
.ci/scripts/test.sh
.tav.yml
.travis.yml
docs/agent-api.asciidoc
docs/index.asciidoc
docs/supported-technologies.asciidoc
lib/instrumentation/index.js
lib/instrumentation/modules/memcached.js
package.json
test/config.js
test/docker-compose.ci.yml
test/docker-compose.yml
test/instrumentation/modules/hapi/basic-legacy-path.js
test/instrumentation/modules/hapi/basic.js
test/instrumentation/modules/hapi/set-framework-2.js
test/instrumentation/modules/hapi/shared.js
test/instrumentation/modules/memcached.js
test/script/local-deps-start.sh
test/script/local-deps-stop.sh
```

The above command is also shown in the console output, please refers to:
- [log](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-1144/42/pipeline/23#step-55-log-1)

And in fact the following command, in charge to grep for the regex, does match with the file:
- [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-1144/42/pipeline/23#step-56-log-2)

And the function in charge to trigger the TAV smartly didn't care about the above output at all as it went through the `else` condition:
- https://github.com/elastic/apm-agent-nodejs/blob/master/Jenkinsfile#L332

Maybe the smartTAV algorithm got some corner case that It was not considered when It was implemented. 

So as soon as I get some idea, I'll leave a comment in this PR for keeping track of this.

watson:
@v1v Got'ya... Could it be the updates to the `.ci/*` directory in this PR that somehow is missing something for memcached to be registered as valid TAV test case?

v1v:
> @v1v Got'ya... Could it be the updates to the .ci/* directory in this PR that somehow is missing something for memcached to be registered as valid TAV test case?

I don't think it's about the .ci itself as the `TAV Test — TAV Test disabled passed` GitHub check matches with the preconditions before even running any scripts:
- https://github.com/elastic/apm-agent-nodejs/blob/master/Jenkinsfile#L333

To ensure memcachde TAV works as expected, I'll then say jenkins run the module tests for memcached

v1v:
jenkins run the module tests for memcached

v1v:
For the record, the comment https://github.com/elastic/apm-agent-nodejs/pull/1144#issuecomment-519501781 did trigger the TAV subset stage:
- See [here](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-1144/43/pipeline/1185)

nik9000:

watson:
This should be ready to land now. I'm just waiting for a review from @Qard.

When merging, we should rename the commit to:

```
feat(memcached): add support for v2.2.0 and higher
```

lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,49 @@
...
+  function wrapCommand (original) {
+    return function wrappedCommand () {
+      if (typeof arguments[0] === 'function') {
+        var query = arguments[0]()
+        // If the callback is not a function the user doesn't care about result
+        if (query && typeof query.callback === 'function') {
+          var span = agent.startSpan(`memcached.${query.type}`, `db.memcached.${query.type}`)
+          if (span) {
+            span.setDbContext({ statement: `${query.type} ${query.key}`, type: 'memcached' })
+            var id = span.id
```
Qard: Not a fan of using a nested `var` that escapes its scope like this. Could we maybe update this file to use `const`/`let` instead and avoid this weird pattern?

lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,49 @@
...
+          }
+          agent.logger.debug('intercepted call to memcached.prototype.command %o', { id: id, type: query.type })
+          query.callback = wrapCallback(query.callback)
+          // Rewrite the query compiler with the wrapped callback
+          arguments[0] = function queryCompiler () {
```
Qard: Does this need to be outside of the `if (query && typeof query.callback === 'function')`? If you're trying to avoid side effects from calling the function multiple times, it won't work if that check fails. I'm not really sure what the purpose of this wrap here is though. 🤔

lib/instrumentation/modules/memcached.js
mikefaut: I'm happy to use let and move the section that uses span.id inside the scope so its not escaped. The other modules I've looked at briefly for inspiration are using var though so I thought it was a standard.

watson: This code is technically fine but could be easier to read if we moved the variable declaration outside of the if-statement. I didn't comment on the use of `var` as I know we have some very old code from the Node.js 0.8 days still around in our code-base. And I knew you probably got inspired from some of that 😅 

Today I would define `id` outside of the if-statement using `let` and set it just like today inside the if-statement.

lib/instrumentation/modules/memcached.js
mikefaut: It does not need to be outside. We only need to change arguments[0] when we wrap the callback. I'm not worried about it being called multiple times because it doesn't mutate state. Their argument[0] returns a static object with a variable callback(set by user)

watson: I think it makes sense to not call the query compiler more than once, so I like the pattern. But yeah, it should probably be outside of the if block - good catch! I'd probably put it right after the line where the `query` variable is set so that the two are next to each other. That makes it easier to understand IMO.

watson: @mikefaut Ups, sorry. Didn't see your comment when I wrote mine. GitHub sometimes doesn't update very well 😅

watson: Now reading your comment, I agree that it's fine to leave inside for the current implementation of `memcached`. And though they probably will never change this behavior, wouldn't it be more future proof / safe to put it outside of the if-block? I'm fine either way though.

lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,48 @@
...
+  function wrapCommand (original) {
+    return function wrappedCommand () {
+      if (typeof arguments[0] === 'function') {
+        var query = arguments[0]()
+        // If the callback is not a function the user doesn't care about result
+        if (query && typeof query.callback === 'function') {
+          var span = agent.startSpan(`memcached.${query.type}`, `db.memcached.${query.type}`)
+          if (span) {
+            span.setDbContext({ statement: `${query.type} ${query.key}`, type: 'memcached' })
+            agent.logger.debug('intercepted call to memcached.prototype.command %o', { id: span.id, type: query.type })
```
watson: I know this probably wasn't that clear, but it was actually on purpose that this debug log statement was logged even if no span was created. This better allows us to debug the flow of the problem even if spans are not generated. That's why the `span.id` was put into a variable that would then either be `undefined` or contain the spna id.

But I do like that we don't have to wrap the callback if no span is created. That's a clear improvement.

How about we just move this line outside of the `if (span)` block (just before) and change it accordingly:

```diff
-            agent.logger.debug('intercepted call to memcached.prototype.command %o', { id: span.id, type: query.type })
+            agent.logger.debug('intercepted call to memcached.prototype.command %o', { id: span && span.id, type: query.type })
```

lib/instrumentation/modules/memcached.js
```diff
@@ -0,0 +1,48 @@
...
+            arguments[0] = function queryCompiler () {
+              return query
+            }
+          }
+        }
+      }
+      return original.apply(this, arguments)
+
+      function wrapCallback (cb, span) {
+        return function wrappedCallback () {
```
watson: As we don't use this more than once in this patch, if you wanted to simplify the code you could just move this function up inline, and just set it directly on the `query.callback` line. But now we're just into code cleanup territory.

mikefaut: It starts to become a little hard to read then because we'd need to write the original function to a placeholder and then call that placeholder as the callback.

watson:
jenkins run the module tests for memcached

PR #1146, opened 2019-06-20
fix: only log deprecation warning once per function

watson:
The previous implementation would spam the logs if one of these functions were called one or more times per request for instance.

While you're sure to get peoples attention by this, it's also super annoying for the user, which wasn't the intention. So this PR dials it down a bit.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -201,7 +201,10 @@ Agent.prototype.setCustomContext = function (context) {
 }
 
 Agent.prototype.setTag = function (key, value) {
-  this.logger.warn('Called deprecated method: apm.setTag(...)')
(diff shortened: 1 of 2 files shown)


watson:
The test failures are just an internal Jenkins error, so just ignore it.

Qard:
Maybe rather than duplicating this logic for each function, we should have a deprecate function like Node.js core:

```js
function deprecate (message, fn) {
  var called = false
  return function deprecated () {
    if (!called) {
      called = true
      this.logger.warn(message)
    }
    return fn.apply(this, arguments)
  }
}

Agent.prototype.setTag = deprecate('Called deprecated method: apm.setTag(...)', function (key, value) {
  return this.setLabel(key, value)
})
```

PR #1171, opened 2019-06-27
ci: notify on cleanup stage

kuisathaverat:
notify on cleanup stage, this ensures all the build info is reported


watson:
The test failures seem to be related to Jenkins itself rather than our tests. Gonna merge regardless

PR #1172, opened 2019-06-27
ci(jenkins): add lint commit messages stage

### LINKED ISSUE #1157: Lint commit messages as separate build on Jenkins
Today Jenkins will lint the commits in the PR for each stage of the build before running the actual tests. Instead, this should be done just once as a separate build stage just as it's done on Travis. This way it doesn't hinder the actual tests from running even if the commit message is invalid.

cc @elastic/observablt-robots

kuisathaverat:
* add lint commit messages stage
* use node:12 Docker image

Closes #1157

### DIFF
--- a/.ci/scripts/lint-commits.sh
+++ b/.ci/scripts/lint-commits.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -xeo pipefail
 
-export PATH=${PATH}:/$(pwd)/node_modules:/$(pwd)/node_modules/.bin
+export PATH=$(npm bin):${PATH}
 export HOME=$(pwd)
 
 npm install
--- a/.ci/scripts/lint-commits.sh
+++ b/.ci/scripts/lint-commits.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
+set -xeo pipefail
 
-set -xe
+export PATH=${PATH}:/$(pwd)/node_modules:/$(pwd)/node_modules/.bin
+export HOME=$(pwd)
+
+npm install
 
 if [[ -z "${CHANGE_ID}" ]]; then
   # If on master, just test the latest commit
--- a/.ci/scripts/test_basic.sh
+++ b/.ci/scripts/test_basic.sh
@@ -12,7 +12,6 @@ npm --version
 
 standard
 npm run test:deps
-.ci/scripts/lint-commits.sh
 npm run test:types
 npm run test:babel
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,6 +46,27 @@ pipeline {
       }
     }
     /**
+      Lint commit messages
+    */
+    stage('Lint commit messages') {
+      agent { label 'docker && immutable' }
+      options { skipDefaultCheckout() }
+      environment {
+        HOME = "${env.WORKSPACE}"
+      }
+      steps {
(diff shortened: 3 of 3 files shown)


.ci/scripts/lint-commits.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
+set -xeo pipefail
 
-set -xe
+export PATH=${PATH}:/$(pwd)/node_modules:/$(pwd)/node_modules/.bin
+export HOME=$(pwd)
+
+npm install @commitlint/cli
```
watson: This should already be installed as a dev-dependency when running `npm install` - or is this because we're not installing the normal modules as part of this stage?

watson:
(added "Closes #1172" so it will close the issue automatically when merged)

.ci/scripts/lint-commits.sh
kuisathaverat: I was trying to make it quicker, I mean not have to download all dependencies

watson: Good idea. Just wasn't sure if you also ran `npm install`. But if you don't then this is better 👍

.ci/scripts/lint-commits.sh
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
+set -xeo pipefail
 
-set -xe
+export PATH=${PATH}:/$(pwd)/node_modules:/$(pwd)/node_modules/.bin
```
watson: This technically shouldn't be a problem, but why add `./node_modules` to the `PATH`? If there's no need for it it's less confusing if it's removed so we only have `./node_modules/.bin`. Though normally I'm in favor of just running the command with the full path instead of modifying `PATH`. But in this case, it's pretty self-contained, so I guess it doesn't have any weird side-effects 😅

.ci/scripts/lint-commits.sh
```diff
@@ -1,6 +1,10 @@
...
+export HOME=$(pwd)
+
+npm install
```
watson: I didn't mind you just installing the module you needed. I just didn't know if you _also_ installed all the other modules, in which case it didn't make any sense. So feel free to swap it back.

kuisathaverat: there is not too much difference a few seconds so it is fine to make npm install

.ci/scripts/lint-commits.sh
kuisathaverat: it is only for CI use

watson: What's the reason you have `/$(pwd)/node_modules` in there as well besides `/$(pwd)/node_modules/.bin`?

kuisathaverat: those are the folders where  Node.js put binaries, that's it.

Qard: There's binaries in `/$(pwd)/node_modules/.bin`, but there shouldn't be anything directly in `/$(pwd)/node_modules`, so we can remove that.

.ci/scripts/lint-commits.sh
Qard: Doing an `npm install` every time we run the linter seems a bit weird to me. 🤔

kuisathaverat: this was my first proposal, I've changed, now, Should I change it again to the first approach?

PR #1180, opened 2019-07-01
ci(travis): re-add docs build to Travis

watson:
Currently the shared 'elasticsearch-ci/docs-build' CI system is broken and is always failing (we're working on removing it so GitHub builds doesn't fail all the time).

Until we get it back, we need another way to build docs as part of CI. This approach will not do link checking, but it's better than nothing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,6 @@ jobs:
     -
       script: npm run docs:build
       name: "Build Docs"
-      perl: '5.26'
 
     # Commit Messages
     -
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,12 @@ jobs:
       node_js: '8'
       env: ELASTIC_APM_ASYNC_HOOKS=false
 
+    # Docs
+    -
+      script: npm run docs:build
+      name: "Build Docs"
+      perl: '5.26'
+
     # Commit Messages
     -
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -61,6 +61,12 @@ jobs:
       node_js: '8'
       env: ELASTIC_APM_ASYNC_HOOKS=false
 
+    # Docs
+    -
+      script: npm run docs:build
+      name: "Build Docs"
+      perl: '5.26'
```
watson: @bmorelli25 I'm not sure if perl is still needed?

bmorelli25: Good question. This is now using the new script, so I would guess not. I'm not 100% sure though

watson: 👍 I'll try to remove it and see what happens to the tests

watson: It worked 💯

watson:
@bmorelli25 how come the `elasticsearch-ci/docs-build` didn't fail here? 🤔 I thought it was currently in a broken state and would always fail...

watson:
Closing 😅 Turns out I misunderstood and that the errors we saw on other PR's in relation to `elasticsearch-ci/docs-build` was only for old PR's. New branch/PR builds will not be affected (unless it happens again). But we'll take our chances for now

PR #1181, opened 2019-07-02
ci(jenkins): move schedule to a scheduler job

kuisathaverat:
Right now, we have the cron task configured on the Jenkins file, this means that every branch and PR uses these settings, so we launch ever branch and PR every day. We want only to launch on daily period only the master branch so we would schedule the task on a special job for that. From here we can schedule daily or weekly whatever we want on a controlled way.

* Daily scheduled job
* Weekly scheduled job
* remove cron from the Jenkinsfile


kuisathaverat:
it is the purpose of run this daily, see elastic/apm-agent-nodejs/issues/378

Issue #1152, opened 2019-06-24
No need to run TAV tests all the time in Jenkins


watson:
Today the TAV tests run all the time on Jenkins which ties up a lot of resources. This is not needed as we have some degree of knowledge of when we don't need to run them.

The TAV tests only need to run under the following circumstances:
- Only run on PR builds and not on commits to regular branches like `master` etc (unless it's a "cron" job as described in #1135, in which case we want to run it on `master`)
- The PR branch being build doesn't start with `greenkeeper/`

watson:
cc @elastic/observablt-robots

kuisathaverat:
it'd check changes on modules to trigger only the TAV test needed for PRs, on scheduled builds it would launch 'em all

watson:
The approach you're describing sounds like the one proposed in #378.

Checking for changes to modules requires a few other things to be in place first, so I was thinking this was too big of a feature to add if we wanted it quickly. So it would be easier, for now, to simply implement my proposed solution. If you want to implement the solution described in #378, we need to first have a chat about how exactly we need this to work before proceeding.

> [...] on scheduled builds it would launch 'em all

Does that mean that TAV tests currently will not run on scheduled builds but only on PRs and when we push to `master`? We don't want them to run on new branch commits, but we do need them to run on scheduled builds (basically as described in the description).

kuisathaverat:
>Does that mean that TAV tests currently will not run on scheduled builds but only on PRs and when we push to master? 

currently a build it is scheduled for every branch, this build executes the current pipeline that launches the TAV test, 'em all.
https://github.com/elastic/apm-agent-nodejs/blob/master/Jenkinsfile#L24

>We don't want them to run on new branch commits, but we do need them to run on scheduled builds (basically as described in the description).

this is what you are requesting in the issue, isn't it?

watson:
Maybe I misunderstood what "scheduled builds" mean. We need to run a build automatically on `master` once a day as described in #1135 where the TAV tests are included. If that's what a scheduled build is, then we need to run them during those. But not on PR branches - only on `master`.

> this is what you are requesting in the issue, isn't it?

Yes

v1v:
This particular issue has been addressed with https://github.com/elastic/apm-agent-nodejs/pull/1181

PR #1182, opened 2019-07-02
ci(jenkins): ignore the branches created from the bots

v1v:
Some slight improvements to trigger builds based on different events. 

## Highlights
- `periodic-folder-trigger` property has been changed to `7d` to avoid the overhead based on the number of branches/prs for this particular repo.
- `greenkeeper` branches are excluded from the main pipeline.


.ci/jobs/apm-agent-nodejs-mbp.yml
```diff
@@ -26,6 +26,9 @@
         - regular-branches: true
         - change-request:
             ignore-target-only-changes: false
+        - named-branches:
+            - wildcards-name:
+                excludes: 'greenkeeper*'
```
kuisathaverat: this will build only the PR from greenkeeper, right now is building the branch and the PR.

v1v:
It seems the local commit linting and the one in travis differ though:

```
 npm run lint:commit

> elastic-apm-node@2.11.6 lint:commit /Users/vmartinez/work/src/github.com/elastic/apm-agent-nodejs
> test/lint-commits.sh
```

So the current setup locally is about 72 chars 

```
 header must not be longer than 72 characters, current length is 96 [header-max-length]
```

Not sure where that particular configuration is defined though

watson:
Yeah, the one on Travis knows the ID of the pull request (1182 in this case) and takes into account that when you "Squash and merge" it will add it to the end of the commit message ` (#1182)`. That's why it can't be longer than 64 chars.

watson:
I assume this fixes #1152?

If so, please update the PR description with `Fixes #1152` so the two are tired correctly together.

v1v:
It's not really specific for that particular issue but it's a kind of related though. In other words, this was to reduce the overkilling of the polling in Jenkins with less frequency.
As it seems related though I'll add that particular information in the description.

watson:
Ah ok. Thanks for the explanation. When you say "kind of related", it doesn't sound like you think it actually fixes it though. But since you wrote `Fixes #1152` in the body, I'm a little confused if it actually fixes it or not?

v1v:
Let me rephrase what I just said, this particular PR was raised after discussing @kuisathaverat and I the status of the nodejs pipeline as there are a few dozens of branches/PRs and find out we could potentially reduce the overhead in the Jenkins instance by reducing the polling time from once a day to once a week. Let's say it's more about to optimize the Jenkins internals.

On the other hand, the greenkeeper branches seem to be related to the bots and therefore this particular PR does ignore the greenkeeper branches from the Multibranch Project Pipeline. In other words, it will keep building any PRs from any forked/branches but will ignore any branches which are named `greenkeeper*`. This will reduce the number of builds per branches which are already validated by PRs. The current Multibranch Project Pipeline in Jenkins does care of any tags, branches and PRs as shown here [mbp(https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/)

Then, I didn't know about the #1152 which does tackle a more specific use case, so that's the reason I said it's a kind of related. 

Please let me know If I need to explain further the reason of this PR. tx

watson:
Ok, it sounds like it doesn't cover every aspect of the issue then. I've removed the "Fixes" keyword for now.

But it seems to me that if we make a PR that fixes the issue, then this PR would not actually be needed anymore. If I'm not mistaken, you want to reduce the overhead of running builds on branches created by Greenkeeper, mainly because running the entire TAV test suite takes up too many resources. But as described in the issue, then TAV should not be running for any type of greenkeeper branch or PR, so if we implemented that, then I think it wouldn't be a problem right?

Even further, we don't need _any_ form of scheduled builds to run on Greenkeeper branches or PR's at all - not daily, not weekly. We just want the initial build that kicks off because the branch or the PR is created. Any build after that will not be used for anything and would be wasted as Greenkeeper only ever looks at the first build.

How about we update this PR to include all of this to kill two birds with one stone?

kuisathaverat:
this PR is needed even do we implemented #1152, the way that greenkeeper makes the updates by creating a local branch and a PR should be achieved, we do not want to build the Branch and the PR is a waste of CI time because is the same thing.

>Even further, we don't need any form of scheduled builds to run on Greenkeeper branches or PR's at all 

What happens if you take a week to merge de PR and you have changes on the master that are untested because the PR was not re-build with those changes? 

>How about we update this PR to include all of this to kill two birds with one stone?

IMHO small changes are best than 1000 changes in a PR, we do not want a PR open for months

watson:
> What happens if you take a week to merge de PR and you have changes on the master that are untested because the PR was not re-build with those changes?

Yeah keeping the scheduled PR builds would make sense. But the scheduled branch builds are not needed as the issue created because of those branches are created just after the first build. So any build after that doesn't change anything.

> IMHO small changes are best than 1000 changes in a PR, we do not want a PR open for months

Agreed. Was just trying to understand what this PR was doing.

v1v:
@watson, I guess @kuisathaverat clarified further and much better the context of this particular PR, therefore I guess we are now aligned about the aim of this PR.  #1152 will be tackled in another PR and then we get the benefits of small changes that solve specific scenarios.

Does it sound correct? If so, I just need your approval :)

PR #1185, opened 2019-07-02
chore: add 'npm run docker:dev' to open cwd in Docker

watson:
This command will run a temporary Docker container with access to the root of your local agent directory and place you in a bash shell inside it.

This allows you to test things in a Linux environment when you're not running Linux yourself.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,6 +14,7 @@ TESTING.md
 # Folders to ignore
 .ci
 .github
+dev-utils
 docs
 test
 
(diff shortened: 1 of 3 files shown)


Qard:
Might be nice to also have a way to start it together with `./test/docker-compose.yml` to have all the databases up too. Might make it easier to work on specific instrumentations, if a Linux issue comes up. 🤔

watson:
I would like to have the option to boot it up without any databases as you many times don't need those. And starting those in Docker on my computer means it grinds to a hold, so I usually always run them natively. But I guess we could make different configs so we can do both.

However, I don't have the knowledge of how to set that up in Docker, so it would take me too long to do for this PR unfortunately. So maybe it could be done in a separate PR?

PR #1184, opened 2019-07-02
test(tav): include knex@^0.18.0 in TAV test-matrix

watson:

PR #1189, opened 2019-07-03
fix(knex): abort early on unsupported version of knex

watson:
Knex v0.18.0 was just released, and as has come apparent in #1184, we don't yet support this version.

The issue is in this line:

https://github.com/elastic/apm-agent-nodejs/blob/fe8ff8ac6c71a6a21ba7e0dd787e570e6655f498/lib/instrumentation/modules/knex.js#L36

The `runner` object no longer have a `query` function in knex v0.18.0 and above.


watson:
I've created a separate issue to add support for knex 0.18.0: #1190

PR #1197, opened 2019-07-07
feat: add support for APM Agent Configuration via Kibana

### LINKED ISSUE #1125: Add ability to change agent sampleRate from within Kibana
For details see elastic/apm#76

watson:
Depends on elastic/apm-nodejs-http-client#66
Closes elastic/apm-agent-nodejs#1125

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [ ] Update dependency to released version

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -126,7 +126,7 @@ Note that both `active` and `instrument` needs to be `true` for instrumentation 
 
 * *Type:* Boolean
 * *Default:* `true`
-* *Env:* `ELASTIC_APM_REMOTE_CONFIG`
+* *Env:* `ELASTIC_APM_CENTRAL_CONFIG`
 
 Activate APM Agent Configuration via Kibana.
 If set to `true`, the client will poll the APM Server regularly for new agent configuration.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -131,7 +131,7 @@ Note that both `active` and `instrument` needs to be `true` for instrumentation 
 Activate APM Agent Configuration via Kibana.
 If set to `true`, the client will poll the APM Server regularly for new agent configuration.
 
-NOTE: This feature requires APM Server v7.3 or later.
+NOTE: This feature requires APM Server v7.3 or later and that the APM Server is configured with `kibana.enabled: true`.
 
 [[async-hooks]]
 ==== `asyncHooks`
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -121,8 +121,8 @@ A boolean specifying if the agent should collect performance metrics for the app
 
 Note that both `active` and `instrument` needs to be `true` for instrumentation to be running.
 
-[[remote-config]]
-==== `remoteConfig`
+[[central-config]]
+==== `centralConfig`
 
 * *Type:* Boolean
 * *Default:* `false`
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -131,6 +131,8 @@ Note that both `active` and `instrument` needs to be `true` for instrumentation 
 Activate APM Agent Configuration via Kibana.
 If set to `true`, the client will poll the APM Server regularly for new agent configuration.
 
+NOTE: This feature requires APM Server v7.3 or later.
+
 [[async-hooks]]
 ==== `asyncHooks`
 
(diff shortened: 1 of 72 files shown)


watson:
The diff might look a bit daunting, but this is mainly due to a refactoring I had to do to ensure the `normalize` function could be called multiple times (first when the initial config is loaded, and later when we receive new config from the APM Server). **If you ignore whitespace it should look a lot simpler.**

The issue was that the properties `ignoreUrlStr`, `ignoreUrlRegExp` etc all were set on the new config options object every time it was given to `normalize`. Instead, these are now properties on a new real config object and set in the constructor the first time the config is loaded. And `normalize` now just expects them to be there.

There's a lot of related refactoring that can be done in the `config.js` file, but I tried to keep it minimal to not make the PR to confusing.

test/remote-config-disabled.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const http = require('http')
+
+const test = require('tape')
+
+test('remote config disabled', function (t) {
+  const server = http.createServer((req, res) => {
+    t.fail('should not poll APM Server for config')
```
Qard: I would make this check the request URL in case we add some other route in the future which _should_ be made before the timeout.

watson: Fixed

docs/configuration.asciidoc
```diff
@@ -120,6 +120,18 @@ A boolean specifying if the agent should collect performance metrics for the app
...
+==== `remoteConfig`
+
+* *Type:* Boolean
+* *Default:* `false`
+* *Env:* `ELASTIC_APM_REMOTE_CONFIG`
+
+Activate APM Agent Configuration via Kibana.
+If set to `true`, the client will poll the APM Server regularly for new agent configuration.
+
+NOTE: This feature requires APM Server v7.3 or later.
```
watson: @bmorelli25 we probably want to align on this wording across all agents. And we probably want to also add that this feature is disabled by default in the APM Server (as far as I know). So a user needs v7.3 + they need to enable it in `apm-server.yml` before it works. I don't know if it's going to be enabled by default in Elastic Cloud -@roncohen do you know?

watson: I updated the line to include information about the APM Server config option (see updated version in PR)

watson: Forgot to mention: According to @simitt, the feature will be off by default in the APM Server for 7.3

test/config.js
```diff
@@ -21,46 +21,49 @@ var config = require('../lib/config')
...
-  ['stackTraceLimit', 'STACK_TRACE_LIMIT', 50],
-  ['captureExceptions', 'CAPTURE_EXCEPTIONS', true],
-  ['filterHttpHeaders', 'FILTER_HTTP_HEADERS', true],
+  ['asyncHooks', 'ASYNC_HOOKS', true],
+  ['captureBody', 'CAPTURE_BODY', 'off'],
   ['captureErrorLogStackTraces', 'CAPTURE_ERROR_LOG_STACK_TRACES', config.CAPTURE_ERROR_LOG_STACK_TRACES_MESSAGES],
+  ['captureExceptions', 'CAPTURE_EXCEPTIONS', true],
   ['captureSpanStackTraces', 'CAPTURE_SPAN_STACK_TRACES', true],
-  ['captureBody', 'CAPTURE_BODY', 'off'],
+  ['centralConfig', 'CENTRAL_CONFIG', true],
```
watson: This is one of two lines that I added (the diff is a little hard to read as I reordered the array to be alphabetically sorted)

test/config.js
```diff
@@ -21,46 +21,49 @@ var config = require('../lib/config')
...
+  ['frameworkVersion', 'FRAMEWORK_VERSION'],
+  ['hostname', 'HOSTNAME'],
   ['instrument', 'INSTRUMENT', true],
-  ['asyncHooks', 'ASYNC_HOOKS', true],
+  ['kubernetesNamespace', 'KUBERNETES_NAMESPACE'],
+  ['kubernetesNodeName', 'KUBERNETES_NODE_NAME'],
+  ['kubernetesPodName', 'KUBERNETES_POD_NAME'],
+  ['kubernetesPodUID', 'KUBERNETES_POD_UID'],
+  ['logLevel', 'LOG_LEVEL', 'info'],
+  ['metricsInterval', 'METRICS_INTERVAL', 30],
```
watson: This is one of two lines that I added (the diff is a little hard to read as I reordered the array to be alphabetically sorted)

PR #1202, opened 2019-07-09
fix: use npm@6.9.0 fixed version for Node.js < 8

kuisathaverat:
Node.js 6 is failing to install `http-signature`, locally with docker failed in the same package, outside Docker fails on an optional dependency and can continue, it is a little bit erratic. So I have forced to install `npm@6.9.0` (latest is 6.10.0) and it works so there is something that breaks the `npm install` on `Node.js 6` and `npm@6.10.0`

```
node_tests_1     | 801 http fetch GET 304 https://registry.npmjs.org/http-signature/-/http-signature-1.2.0.tgz 84ms (from cache)
node_tests_1     | 802 silly fetchPackageMetaData error for http-signature@^1.2.0 zlib: zlib binding closed
```

this start happens 6 days ago that it is the same date that npm@6.10.0 was release.

### DIFF
--- a/.ci/docker/node-container/Dockerfile
+++ b/.ci/docker/node-container/Dockerfile
@@ -5,7 +5,7 @@ WORKDIR /npm
 RUN ( [ "${NODE_VERSION%%.*}" -le 8  ] \
   && echo "Node.js ${NODE_VERSION} - Manual install npm" \
   && mkdir -p /npm/node_modules \
-  && npm install npm \
+  && npm install npm@6.9.0 \
   && rm /usr/local/bin/npm \
   && ln -s /npm/node_modules/npm/bin/npm-cli.js /usr/local/bin/npm ) || exit 0
 
--- a/.ci/scripts/docker-test.sh
+++ b/.ci/scripts/docker-test.sh
@@ -5,11 +5,7 @@ export
 id
 node --version
 npm --version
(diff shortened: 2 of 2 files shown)


.ci/docker/node-container/Dockerfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /npm
 RUN ( [ "${NODE_VERSION%%.*}" -le 8  ] \
```
watson: `-le 8` also includes the latest version of Node.js 8 right? I think we intended it only to include Node.js 8.1.0 or below. While not technically related to this PR I just wanted to point it out as this fix will then also affect the latest version of Node.js in that case. But in that case, feel free to change it in another PR if that's easier 😃

kuisathaverat: we try to keep it simpler, to check <8.1 make it a little more complex and IMHO it not worth

watson: Taking a second look at this, isn't `NODE_VERSION` hardcoded to be `8` in this Dockerfile btw?

kuisathaverat: no, is not, it is the default value if you do not pass a `--build-arg NODE_VERSION=42` to the `docker build` command

PR #1219, opened 2019-07-15
feat(metrics): breakdown graphs

### LINKED ISSUE #1064: Agent breakdown metrics
Implementation of agent breakdown metrics (https://github.com/elastic/apm/issues/78)

Checklist/TL;DR version of the design:

- [x] [Calculate self-time of transactions and spans](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.ondan294nbpt)
  - [x] Prerequisite: Spans have a reference to their parent
  - [x] Add a child duration reentrant timer to transactions and spans
- [x] Add [feature flag](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.12wulrccoi9k)
- Metrics registry enhancements
  - [x] Add support for [timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.qk36t41ikda3)
  - [ ] [Limit the max amount of timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.tzyh7loxf2is)
  - [x] Add support for counters
  - [ ] Optional: [guard against concurrent reads/writes](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.aler9ixddp0e)
- [x] [Track span self-times on the transaction](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.g5pdtefj3ag)
  - [x] Prerequisite: Spans have a reference to their transaction
  - On span end:
    - [x] track the span self-time on the corresponding transaction
  - On transaction end
    - [x] track all self-times on the metric registry
    - [x] track the transaction's self time with the `span.type` `app`
    - [x] track the transaction's duration on the metrics registry
- [Reporting metrics](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.gkxkryn9w02s)
  - [x] Add support for top-level labels for `transaction.name`, `transaction.type`, `span.type` and `span.subtype`
  - [x] Reset counters for Timers and the `transaction.breakdown.count` counter every time the metrics are reported
- [x] [Acceptance tests](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.7m6qydtkxod4)
- [ ] [Evaluate the performance impact](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.68ixudlgfzac)
- [x] [Document metrics](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.68ixudlgfzac)

Qard:
This expands the timer and transaction classes to enable capturing breakdown graph metrics. I don't really like the flatten/unflatten thing, but I'm not sure there's a better way, since measured doesn't support the nested dimensions we need to separate transaction and span data. 🤔 

Closes #1064

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -16,15 +16,16 @@ module.exports = Span
 util.inherits(Span, GenericSpan)
 
 function Span (transaction, name, type, opts) {
+  var childOf = transaction._agent._instrumentation.activeSpan || transaction
   if (!opts) {
     opts = {
-      timer: transaction._timer,
-      childOf: transaction._agent._instrumentation.activeSpan || transaction
+      timer: childOf._timer,
+      childOf
     }
   } else {
-    opts.timer = transaction._timer
+    opts.timer = childOf._timer
     if (!opts.childOf) {
-      opts.childOf = transaction._agent._instrumentation.activeSpan || transaction
+      opts.childOf = childOf
     }
   }
 
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -62,6 +62,7 @@ Span.prototype.end = function (endTime) {
   this.ended = true
   this._agent.logger.debug('ended span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type })
   this._agent._instrumentation.addEndedSpan(this)
+  this.transaction._captureBreakdown(this)
 }
 
 Span.prototype.setDbContext = function (context) {
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -2,6 +2,9 @@
 
 var util = require('util')
 
+var ObjectIdentityMap = require('object-identity-map')
+var entries = require('object.entries')
+
 var getPathFromRequest = require('./express-utils').getPathFromRequest
 var GenericSpan = require('./generic-span')
 var parsers = require('../parsers')
@@ -30,6 +33,7 @@ function Transaction (agent, name, type, opts = {}) {
   this._droppedSpans = 0
   this._contextLost = false // TODO: Send this up to the server some how
   this._abortTime = 0
+  this._breakdownTimings = new ObjectIdentityMap()
 }
 
 Object.defineProperty(Transaction.prototype, 'name', {
@@ -199,6 +203,7 @@ Transaction.prototype.end = function (result, endTime) {
 
   this._timer.end(endTime)
   this.ended = true
+  this._captureBreakdown(this)
 
   var trans = this._agent._instrumentation.currentTransaction
 
@@ -217,4 +222,73 @@ Transaction.prototype.end = function (result, endTime) {
 
   this._agent._instrumentation.addEndedTransaction(this)
   this._agent.logger.debug('ended transaction %o', { trans: this.id, parent: this.parentId, trace: this.traceId, type: this.type, result: this.result, name: this.name })
+}
+
+Transaction.prototype._captureBreakdown = function (span) {
+  const metrics = this._agent._metrics
+
+  // Record span data
+  captureBreakdown(this, {
+    transaction: transactionBreakdownDetails(this),
+    span: spanBreakdownDetails(span)
+  }, span._timer.selfTime)
+
+  // Record transaction data
+  if (span instanceof Transaction) {
+    const { duration } = this._timer
+    const labels = flattenBreakdown({
+      transaction: transactionBreakdownDetails(this)
+    })
+
+    metrics.getOrCreateCounter('transaction.duration.count', labels).inc()
+    metrics.getOrCreateCounter('transaction.duration.sum.us', labels).inc(duration)
+    metrics.getOrCreateCounter('transaction.breakdown.count', labels).inc(this.sampled ? 1 : 0)
+
+    this._breakdownTimings.forEach(({ labels, time, count }) => {
+      labels = flattenBreakdown(labels)
+      metrics.getOrCreateCounter('span.self_time.count', labels).inc(count)
+      metrics.getOrCreateCounter('span.self_time.sum.us', labels).inc(time)
+    })
+  }
+}
+
+function transactionBreakdownDetails ({ name, type } = {}) {
+  return {
+    name,
+    type
+  }
+}
+
+function spanBreakdownDetails (span) {
+  if (span instanceof Transaction) {
+    return {
+      type: 'app'
+    }
+  }
+
+  const [ type, subtype ] = span.type.split('.')
+  return {
+    type,
+    subtype
+  }
+}
+
+function captureBreakdown (transaction, labels, time) {
+  const build = () => ({ labels, count: 0, time: 0 })
+  const counter = transaction._breakdownTimings.ensure(labels, build)
+  counter.time += time
+  counter.count++
+}
+
+function flattenBreakdown (source, target = {}, prefix = '') {
+  for (const [ key, value ] of entries(source)) {
+    if (typeof value === 'undefined') continue
+    if (typeof value === 'object') {
+      flattenBreakdown(value, target, `${prefix}${key}::`)
+    } else {
+      target[`${prefix}${key}`] = value
+    }
+  }
+
+  return target
 }
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -225,13 +225,17 @@ Transaction.prototype.end = function (result, endTime) {
 }
 
 Transaction.prototype._captureBreakdown = function (span) {
-  const metrics = this._agent._metrics
+  const agent = this._agent
+  const metrics = agent._metrics
+  const conf = agent._conf
 
   // Record span data
-  captureBreakdown(this, {
-    transaction: transactionBreakdownDetails(this),
-    span: spanBreakdownDetails(span)
-  }, span._timer.selfTime)
+  if (this.sampled && conf.breakdownMetrics) {
+    captureBreakdown(this, {
+      transaction: transactionBreakdownDetails(this),
+      span: spanBreakdownDetails(span)
+    }, span._timer.selfTime)
+  }
 
   // Record transaction data
   if (span instanceof Transaction) {
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -202,8 +202,8 @@ Transaction.prototype.end = function (result, endTime) {
   if (!this._defaultName && this.req) this.setDefaultNameFromRequest()
 
   this._timer.end(endTime)
-  this.ended = true
   this._captureBreakdown(this)
+  this.ended = true
 
   var trans = this._agent._instrumentation.currentTransaction
 
@@ -225,6 +225,7 @@ Transaction.prototype.end = function (result, endTime) {
 }
 
 Transaction.prototype._captureBreakdown = function (span) {
+  if (this.ended) return
   const agent = this._agent
   const metrics = agent._metrics
   const conf = agent._conf
--- a/test/metrics/breakdown.js
+++ b/test/metrics/breakdown.js
@@ -35,9 +35,13 @@ if (process.platform === 'linux') {
 }
 
 test('includes breakdown when sampling', t => {
-  resetAgent(7, (data) => {
+  const conf = {
+    metricsInterval: 1
+  }
+
+  resetAgent(7, conf, (data) => {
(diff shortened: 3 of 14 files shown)


lib/instrumentation/transaction.js
```diff
@@ -218,3 +222,71 @@ Transaction.prototype.end = function (result, endTime) {
...
+  return {
+    type,
+    subtype
+  }
+}
+
+function captureBreakdown (transaction, labels, time) {
+  const build = () => ({ labels, value: 0 })
+  const counter = transaction._breakdownTimings.ensure(labels, build)
+  counter.value += time
```
felixbarny: Seems like you are not capturing the count per span type. For example when there are two `db.mysql` spans, each taking 1µs, the metricset should have `span.self_time.count=2` and `span.self_time.sum.us`.

But maybe I'm just not understanding how the ObjectIdentityMap works. If that contains two different counters in that example for the same (equal) key (with different object instances), it should work.

Qard: Ah, no. It's only counting the time, not the number occurrences, within the request. I'll fix that. 😅

Qard: Fixed

kuisathaverat:
jenkins run the test

lib/instrumentation/transaction.js
```diff
@@ -218,3 +223,72 @@ Transaction.prototype.end = function (result, endTime) {
...
+  counter.count++
+}
+
+function flattenBreakdown (source, target = {}, prefix = '') {
+  for (const [ key, value ] of entries(source)) {
+    if (typeof value === 'undefined') continue
+    if (typeof value === 'object') {
+      flattenBreakdown(source[key], target, `${prefix}${key}::`)
+    } else {
+      target[`${prefix}${key}`] = source[key]
```
axw: should that be `= value`?

lib/instrumentation/transaction.js
```diff
@@ -218,3 +223,72 @@ Transaction.prototype.end = function (result, endTime) {
...
+  const counter = transaction._breakdownTimings.ensure(labels, build)
+  counter.time += time
+  counter.count++
+}
+
+function flattenBreakdown (source, target = {}, prefix = '') {
+  for (const [ key, value ] of entries(source)) {
+    if (typeof value === 'undefined') continue
+    if (typeof value === 'object') {
+      flattenBreakdown(source[key], target, `${prefix}${key}::`)
```
axw: is there a technical reason to use `::` over `.`?
(sorry if it's a dumb question, I'm asking for my own education)

test/metrics/breakdown.js
```diff
@@ -0,0 +1,191 @@
...
+  'nodejs.eventloop.delay.ns',
+  'nodejs.memory.heap.allocated.bytes',
+  'nodejs.memory.heap.used.bytes',
+  'nodejs.eventloop.delay.avg.ms'
+]
+
+if (process.platform === 'linux') {
+  basicMetrics.push('system.process.memory.size')
+}
+
```
axw: Are you intending to add the acceptance tests from the spec to check the values too? I think it would be a good idea to do that. IIANM, you should be able to check exact values by specifying all the start and end times.

lib/agent.js
```diff
@@ -257,7 +257,7 @@ Agent.prototype.addSpanFilter = function (fn) {
 
 Agent.prototype.captureError = function (err, opts, cb) {
   if (typeof opts === 'function') return this.captureError(err, null, opts)
-
+  console.log(err)
```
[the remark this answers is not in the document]
Qard: Not intentional. 😅

lib/instrumentation/transaction.js
Qard: Less likely to be in an object key on purpose. 🤷‍♂

axw: Thought that might be it, but I also thought we wouldn't ever be including objects in the map. Anyway, I was just curious - thanks!

lib/agent.js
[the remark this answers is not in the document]

lib/instrumentation/transaction.js
Qard: Either works. Changed it anyway, just to be clearer. 🤷‍♂

test/metrics/breakdown.js
Qard: Working on it. Should hopefully have them up later today. 👍

lib/instrumentation/span.js
```diff
@@ -42,6 +43,8 @@ function Span (transaction, name, type, opts) {
     this._recordStackTrace()
   }
 
+  this.transaction._captureBreakdown(this)
```
axw: What's the rationale behind moving this here? IIANM, this will cause spans to be included in the breakdown if they're started before but end after the transaction ends, and they would have a "self_time.sum.us" of zero. Spans should not be reported if they start or end after the transaction.

axw: The last 3 acceptance tests in the breakdown doc should cover this.

Qard: It needs to be captured at the start to be able to properly capture the self-time it should subtract from the parent.

axw: This is what I just did locally to test the second last acceptance test. Note that I had to move the _captureBreakdown call back to span end, otherwise it fails.

```diff
diff --git a/lib/instrumentation/span.js b/lib/instrumentation/span.js
index cf9a8ce..6e532b4 100644
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -43,8 +43,6 @@ function Span (transaction, name, type, opts) {
     this._recordStackTrace()
   }
 
-  this.transaction._captureBreakdown(this)
-
   this._agent.logger.debug('start span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: name, type: type })
 }
 
@@ -62,6 +60,7 @@ Span.prototype.end = function (endTime) {
   this._timer.end(endTime)
   this._agent._instrumentation._recoverTransaction(this.transaction)
 
+  this.transaction._captureBreakdown(this)
   this.ended = true
   this._agent.logger.debug('ended span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type })
   this._agent._instrumentation.addEndedSpan(this)
diff --git a/test/metrics/breakdown.js b/test/metrics/breakdown.js
index 1e389a6..1e8e094 100644
--- a/test/metrics/breakdown.js
+++ b/test/metrics/breakdown.js
@@ -460,6 +460,36 @@ test('acceptance', t => {
     })
   })
 
+  t.test('with span ending after transaction', t => {
+    resetAgent(8, conf, ({ metricsets }) => {
+      const found = {
+        transaction: finders.transaction(metricsets),
+        transaction_span: finders['transaction span'](metricsets),
+      }
+
+      t.ok(found.transaction, 'found a transaction metricset')
+      t.deepEqual(found.transaction.samples, {
+        'transaction.duration.count': { value: 1 },
+        'transaction.duration.sum.us': { value: 20 },
+        'transaction.breakdown.count': { value: 1 }
+      }, 'sample values match')
+
+      t.ok(found.transaction_span, 'found a transaction span metricset')
+      t.deepEqual(found.transaction_span.samples, {
+        'span.self_time.count': { value: 1 },
+        'span.self_time.sum.us': { value: 10 }
+      }, 'sample values match')
+
+      agent._metrics.stop()
+      t.end()
+    })
+
+    var transaction = agent.startTransaction('foo', 'bar', { startTime: 0 })
+    var span = agent.startSpan('s name', 's type', { startTime: 10 })
+    transaction.end(null, 20)
+    if (span) span.end(30)
+  })
+
```

Qard: All the acceptance tests are now implemented, with some improvements to the existing ones to more closely conform to what is in the design doc.

test/metrics/breakdown.js
```diff
@@ -0,0 +1,596 @@
...
+    agent._instrumentation.activeSpan = span0
+
+    transaction.end(null, 20)
+    var span1 = agent.startSpan('SELECT *', 'db.mysql', { startTime: 20, childOf: span0 })
+    if (span0) span0.end(30)
+    if (span1) span1.end(30)
+  })
+
+  t.test('with other sub-span extending beyond end', t => {
+    resetAgent(4, conf, ({ metricsets }) => {
```
axw: should be 5 I think: 3 metricsets + transaction + span

test/metrics/breakdown.js
```diff
@@ -0,0 +1,596 @@
...
+    })
+
+    var transaction = agent.startTransaction('foo', 'bar', { startTime: 0 })
+    var span = agent.startSpan('SELECT *', 'db.mysql', { startTime: 10 })
+    transaction.end(null, 20)
+    if (span) span.end(30)
+  })
+
+  t.test('with other sub-span starting after end', t => {
+    resetAgent(5, conf, ({ metricsets }) => {
```
axw: should be 4 I think: 3 metricsets + transaction
(no span because it starts after transaction)

lib/instrumentation/transaction.js
```diff
@@ -218,3 +223,76 @@ Transaction.prototype.end = function (result, endTime) {
   this._agent._instrumentation.addEndedTransaction(this)
   this._agent.logger.debug('ended transaction %o', { trans: this.id, parent: this.parentId, trace: this.traceId, type: this.type, result: this.result, name: this.name })
 }
+
+Transaction.prototype._captureBreakdown = function (span) {
+  const agent = this._agent
```
axw: Not sure how worthwhile it is, but you could make a minor optimisation here by returning if `this.ended`, moving `this.ended = true` directly after the call to `this._captureBreakdown(this)` in `Transaction.end`. That way spans that end after the transaction won't accumulate anything in the transaction.

lib/instrumentation/span.js
```diff
@@ -16,15 +16,16 @@ module.exports = Span
 util.inherits(Span, GenericSpan)
 
 function Span (transaction, name, type, opts) {
+  var childOf = transaction._agent._instrumentation.activeSpan || transaction
```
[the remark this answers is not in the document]
Qard: Those are two very different things: `childOf` is just expressing the parent -> child relationship of spans while `activeSpan` maintains the async call graph to know the causality of things so it can link things back to the correct transaction, since the JS part of Node.js is a single-thread event loop.

test/metrics/breakdown.js
Qard: fixed

test/metrics/breakdown.js
Qard: fixed

lib/instrumentation/transaction.js
Qard: Good call, I'll make a change for that later. Just on my way to give a talk now.

lib/instrumentation/span.js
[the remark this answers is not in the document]
axw: Hmmm yes, I suppose technically the parent isn't of the same transaction.

test/sourcemaps/index.js
```diff
@@ -88,6 +88,10 @@ test('fails', function (t) {
 
 function onError (t, assert) {
   agent._transport = {
+    flush () {},
+    sendTransaction () {},
+    sendSpan () {},
+    sendMetricSet () {},
```
watson: @Qard as far as I can see this shouldn't be affected by the change in this PR. If I comment it out and run the test, it parses fine. Do you remember it was added it?

Qard: If you comment out `sendMetricSet` it may fail intermittently because the test does not set `metricsInterval: 0`. We should _probably_ actually be setting that too, but I wanted to have those functions in there just for _completeness_. Something could change in the future which results in those functions getting _called_ in there. At the least, `flush(...)` has a high probably of being called before it completes.

Issue #378, opened 2018-05-28
[APM-CI][Node.js] Smarter TAV tests


Qard:
Currently the entire TAV test suite runs on every PR, which often isn't necessary. I'm thinking it would be a good idea to write a script to detect which modules changed, if any, and run only those parts of the TAV test suite on PRs. We'd then leave running the full suite up to a regular cron job. This should help to significantly cut down CI feedback time on PRs.

alvarolobato:
This will be improved with the new build system we are working on.

kuisathaverat:
duplicated https://github.com/elastic/apm-agent-nodejs/issues/1152

watson:
These are two different issues. We already have the feature described in #1152 running on Travis, just not on Jenkins. This issue is for making that feature even smarter.

kuisathaverat:
for us, it is mostly the same, the trigger for the test stage is a path match in the changes, for example, if you have changes on the path /module/my_module it will trigger the TAV test for that module.

watson:
The other issue is of higher importance and also covers what happens when you merge into master and how scheduled builds behave. This issue is only about PR builds. But if you want to cover both issues in a single fix and it doesn't take much longer to do, that would be awesome 😃 In that case, there are a few things we need for it in order to work with our workflow (these all concern them selfs with PR builds - scheduled builds and branch builds are not covered by this issue):

- We need the TAV tests for a given module to run automatically only if the corresponding module file has been updated in `lib/instrumentation/modules/*` or in `test/instrumentation/modules/*`.
- Even if no changes have been made to a given module file or module test file(s), we need to be able to manually trigger running its TAV tests.
- We also need a way to manually trigger running _all_ TAV tests.

v1v:
> - We need the TAV tests for a given module to run automatically only if the corresponding module file has been updated in lib/instrumentation/modules/* or in test/instrumentation/modules/*.

This is something feasible.

> - Even if no changes have been made to a given module file or module test file(s), we need to be able to manually trigger running its TAV tests.

That's not really easy to solve though. I'd rather say let's keep this particular requirement outside of this scope for now.

> - We also need a way to manually trigger running all TAV tests.

This is something doable. 

I'll start working on this particular issue now.

kuisathaverat:
>We also need a way to manually trigger running all TAV tests.

It could be a job that you can trigger manually or with a comment from GitHub, basically a pipeline with the comment trigger.

v1v:
[...]

PR #1228, opened 2019-07-18
ci: smarter tav tests for pull requests

v1v:
fixes https://github.com/elastic/apm-agent-nodejs/issues/378

## Highlights
- If the change list contains any related files to the `instrumentation` folders then it will run the TAVs which match them.
- If the GitHub comment contains `jenkins run the module tests for ([module,]+|ALL)` then it will trigger the TAVs for the modules in the comment or all the TAVs if the `ALL` is in the comment.
- Trigger by comment will run all the stages in the Pipeline (aka, build, test, TAV, Docs and so on...)
- `tests_ci` boolean parameter to enable/disable the tests stage in the pipeline when running the Pipeline from the UI. Default: `true`, aka it will run the tests always.
- `TAV *` will be called:
  - either `TAV Test` if it runs all the TAVs 
  - or `TAV Test Subset` if it runs a subset of TAVs and will also include a description `TAV Test comment-triggered` to highlight in GitHub why it was a subset.

## Test Cases
- https://github.com/elastic/apm-agent-nodejs/pull/1228/commits/6da4911ef26efe4d5312cb952bf901a237ee5a96 caused to run the TAV for `handlebars` and `mysql`.

image

image

- Forcing the `params.Run_As_Master_Branch` caused to run all the TAVs. See [build](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp%2FPR-1228/detail/PR-1228/7/pipeline)

- https://github.com/elastic/apm-agent-nodejs/pull/1228#issuecomment-513311236 caused to run the mysql modules only (to clarify the test stage is disabled in this case only for testing purposes):

image

- https://github.com/elastic/apm-agent-nodejs/pull/1228#issuecomment-514148741 caused to run the TAV Subset

image

- `Run_As_Master_Branch` parameter does trigger all the TAVs even though when the above test scenario was still running, that's expected!

image


v1v:
jenkins run the tests for the module mysql

v1v:
jenkins run the tests for the module mysql

v1v:
jenkins run the tests for the module mysql

v1v:
jenkins run the tests for the module mysql

v1v:
jenkins run the tests for the module mysql

v1v:
jenkins run the tests for the module mysql

v1v:
jenkins run the tests for the module http

Qard:
Is this still WIP? I see passing tests--and it looks great having the separate status checks, by the way!

v1v:
@Qard , as soon as we agree on https://github.com/elastic/apm-agent-nodejs/issues/378 I'll update this PR to be reviewed, does it sound like a good plan?

v1v:
jenkins run the module tests for mysql

Jenkinsfile
```diff
@@ -24,12 +24,13 @@ pipeline {
     quietPeriod(10)
   }
   triggers {
-    issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?tests(?:\\W+please)?.*')
+    issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?(?:module\\W+)?tests(?:\\W+please)?.*')
```
kuisathaverat: even do the regexp is complimentary, I'd split it in two if it possible, to know that there are two types of comment trigger

Jenkinsfile
```diff
@@ -115,18 +129,42 @@ pipeline {
...
-          unstash 'source'
-          dir("${BASE_DIR}"){
-            script {
-              def node = readYaml(file: '.ci/.jenkins_tav_nodejs.yml')
-              def tav = readYaml(file: '.ci/.jenkins_tav.yml')
+        deleteDir()
+        unstash 'source'
+        dir("${BASE_DIR}"){
+          script {
+            def ghContextName = 'TAV Test'
```
kuisathaverat: I'd extract this to a function to give context about what it does.

Jenkinsfile
v1v: I'm afraid the `issueCommentTrigger` can be declared only once. I could potentially create two groups but the regex would be too long.
The proposed one supports the current use cases:
- `jennkins run the tests` and `jenkins run tests`
- `jenkins run the module tests for` and `jenkins run module tests for`

In any case, the filtering of running TAV is handled later on in the TAV stage.

what do you think if I leave it as it is? Or do you have another use case in mind?

Jenkinsfile
v1v: See https://github.com/elastic/apm-agent-nodejs/pull/1228/commits/adb4b08e14059269931f4f9c6806a34ef7f7597d

PR #1230, opened 2019-07-18
fix(graphql): use custom transaction type to graphql, fix #765

### LINKED ISSUE #765: Custom transaction type for incoming GraphQL requests
If we detect an incoming GraphQL request, we'll change the name of the transaction from say `POST /graphql` to `<query> (/graphql)` where `<query>` is the name of the GraphQL query. Example: `hello (/graphql)`.

We should consider changing the transaction type to `graphql` as well instead if the default `request`.

Update: We decided we want to do this

sibelius:
This will create a new type for GraphQL requests, so each GraphQL query could be compared to others

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


Qard:
Thanks for the contribution! Would you mind adding a test for this too, please?

Qard:
It looks like the tests at `test\instrumentation\modules\apollo-server-express.js` and  `test\instrumentation\modules\express-graphql.js` will need to be updated to match this change.

sibelius:
is there an easy way to run this failing tests?

Qard:
If you mean locally, you can just do `node $test_path`, so `node test\instrumentation\modules\apollo-server-express.js` and `node test\instrumentation\modules\express-graphql.js`. To trigger them on the CI, you can force push or ask me and I can trigger another run.

test/instrumentation/modules/graphql.js
```diff
@@ -62,7 +62,7 @@ test('graphql.graphql - transaction ended', function (t) {
     var trans = data.transactions[0]
 
     t.equal(trans.name, 'foo')
-    t.equal(trans.type, 'custom')
+    t.equal(trans.type, 'graphql')
```
Qard: Looks like you made these test changes in the wrong file. It's `express-graphql`, not `graphql`.

Qard:
Looks like that last commit _removed_ the changes from the wrong test file but didn't add the required changes to the correct file. Did you forget to add the file to the commit? 🤔

Issue #1158, opened 2019-06-25
Add ELASTIC_APM_ASYNC_HOOKS=false dimension to regular tests on Jenkins


watson:
Currently, the regular tests on Jenkins (not TAV), run with default agent settings. We need an extra dimension where we also run them all with the environment variable `ELASTIC_APM_ASYNC_HOOKS=false`

cc @elastic/observablt-robots

PR #1231, opened 2019-07-18
ci(jenkins): add tests with elastic_apm_async_hooks=false

v1v:
Fixes https://github.com/elastic/apm-agent-nodejs/issues/1158

## Highlights
- Run in two batches with 
  - `ELASTIC_APM_ASYNC_HOOKS=true` for the first batch as used to be.
  - `ELASTIC_APM_ASYNC_HOOKS=false` for the last batch as asked in the issue.

## Pipeline UI

image

## How to test it locally?

```
export ELASTIC_APM_ASYNC_HOOKS=false
.ci/scripts/test.sh 12 ''
+ DOCKER_FOLDER=.ci/docker
+ NODE_VERSION=12
+ TAV_VERSIONS=
+ case ${TAV_VERSIONS} in
+ DOCKER_COMPOSE_FILE=docker-compose-all.yml
+ ELASTIC_APM_ASYNC_HOOKS=false
+ NODE_VERSION=12
...
Attaching to docker_node_tests_1
node_tests_1     | + export
node_tests_1     | + id
node_tests_1     | declare -x CASSANDRA_HOST="cassandra"
node_tests_1     | declare -x ELASTIC_APM_ASYNC_HOOKS="false"
node_tests_1     | declare -x ES_HOST="elasticsearch"
...
```


Qard:
Test failure appears unrelated. Flaky, maybe?

v1v:
IIRC, `Node.js-6.0` already got that flakiness and that's indeed where the test failed. I have not really changed anything in the tests but adding the support for the async hooks. I guess this is unrelated to this PR.

Qard:
I think we can just ignore it. Node.js 6.x has been end-of-life since April, and I don't have the time to dive into fixing it right now anyway. 😞

v1v:
If no issues, I'll then avoid the fail-fast _ONLY_ for the nodejs-6 version and use a warning, then the rest of the stages in the pipeline will be triggered. what do you think?

Qard:
Good idea. 👍 

Looks like mongodb-core is failing the TAV tests currently too. Just ignore that for now too and I'll investigate it later. 😅

Qard:
Trying to reproduce the `mongodb-core` failure in tav and no luck on any Node.js version so far. My theory currently is that, because the docker-compose doesn't specify a mongodb version, it pulled down a newer version than we've been testing against until now which is incompatible with the 1.x series of the `mongodb-core` module. Not yet verified... 🤔

Qard:
Would you mind rebasing this? I suspect the mongo failures were intermittent as I have not seen them on any other PRs, and there doesn't appear to be any changes in this PR which should impact that, as far as I can tell. 🤔

v1v:
I already rebased a few times, and it seems the previous last one (https://github.com/elastic/apm-agent-nodejs/pull/1231/commits/2db66b1c0c0d301a01d04fb9decd0455f54efb6d) passed. Please let me know if you need anything else

Jenkinsfile
```diff
@@ -285,7 +292,13 @@ def generateStep(Map params = [:]){
         dir("${BASE_DIR}"){
           retry(2){
             sleep randomNumber(min:10, max: 30)
-            sh(label: "Run Tests", script: """.ci/scripts/test.sh "${version}" "${tav}" "${edge}" """)
+            if (version?.startsWith('6')) {
+              catchError {
```
watson: @v1v Why differ the execution for Node.js 6 here?

v1v: Fair point, we have discussed this in the following comment:
- https://github.com/elastic/apm-agent-nodejs/pull/1231#issuecomment-514514260

The idea was as node.js 6 seems to be flaky and decided that rather than failing the whole pipeline to set the build as unstable if anything happens with nodejs6.

Jenkins pipelines provide three major different status:
- Stable
- Unstable
- Failure

With this implementation, the build status for nodejs6 in case it fails, would be unstable rather than failure.

Does it make sense?

watson:
@v1v Sorry for being late to the part here, but I just noticed another thing. The "async hooks false" tests should only run in the latest version of Node.js 8 and newer. Node 8.1 and below doesn't have support for async hooks, so running these tests on those versions will not have any effect and only tie up resources while the tests run.

PR #1237, opened 2019-07-19
feat(config): default serviceVersion to package version

### LINKED ISSUE #1234: Use package.json version as default serviceVersion
We currently use the `name` field in the `package.json` file as the default for the `serviceName` option. We should probably do the same with mapping the `version` field to `serviceVersion`.

yuxizhe:
fix #1234
### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -63,6 +63,7 @@ Disabling validation is normally required if using self-signed certificates.
 ==== `serviceVersion`
 
 * *Type:* String
+* *Default:* `version` field of `package.json`
 * *Env:* `ELASTIC_APM_SERVICE_VERSION`
 
 The version of the app currently running.
--- a/lib/config.js
+++ b/lib/config.js
@@ -29,9 +29,11 @@ if (fs.existsSync(confPath)) {
   }
 }
 
-let serviceName
+let serviceName, serviceVersion
 try {
-  serviceName = readPkgUp.sync().pkg.name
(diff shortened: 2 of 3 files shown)


Qard:
jenkins test this please

Qard:
Oh, looks like you will also need to update [this line](https://github.com/elastic/apm-agent-nodejs/blob/master/test/config.js#L30) to get the `version` field from the module `package.json` and put it in the third array position. Current `SERVICE_NAME` hardcodes this to match the package name, so maybe make that dynamically loaded from the `package.json` too while you're in there. 🤔

Qard:
jenkins run the tests

yuxizhe:
> Oh, looks like you will also need to update [this line](https://github.com/elastic/apm-agent-nodejs/blob/master/test/config.js#L30) to get the `version` field from the module `package.json` and put it in the third array position. Currently `SERVICE_NAME` hardcodes this to match the package name, so maybe make that dynamically loaded from the `package.json` too while you're in there. 🤔

oh I see, fixed :）

Qard:
jenkins run the tests please

Qard:
run tests

yuxizhe:
SERVICE_NAME also loaded from package.json now.

Qard:
jenkins run the tests

Qard:
jenkins run the tests please

Qard:
Test failures are unrelated. I've verified this works locally, so I will land as-is and investigate the failures separately.

PR #1246, opened 2019-07-29
chore: make hapi instrumentation support @hapi/hapi rename

Qard:
Closes #1098


Qard:
Not sure why `@hapi/hoek` is causing `wait-on` to fail. I feel like there must be missing content in that log. 🤔

Qard:
Ah...apparently `wait-on` changed the dependency on `joi` to the newer `@hapi/joi` in 3.3.0, which drops Node.js 6.x compatibility in a minor...

watson:
I think this is a duplicate of #1098?

Qard:
It's a squash + rebase of #1098 with a commit message rewrite so it'd actually pass the commit lint and trigger the tav tests on Travis.

docs/agent-api.asciidoc
```diff
@@ -690,12 +690,12 @@ Optionally, a type may also be provided to group lambdas together. By default,
 Read more lambda support in the <<lambda,Lambda>> article.
 
 [[apm-add-patch]]
-==== `apm.addPatch(name, handler)`
+==== `apm.addPatch(modules, handler)`
 
 [small]#Added in: v2.7.0#
 
-* `name` +{type-string}+
```
watson: The type(s) within the curly braces needs to match a type defined in [`docs/index.asciidoc`](https://raw.githubusercontent.com/elastic/apm-agent-nodejs/master/docs/index.asciidoc) so that the link works.

We should probably do something along the lines of what we did with the `handler` argument below where there are two possible types listed, e.g:

```asciidoc
* `modules` +{type-string}+ | +{type-string-array}+
```

We don't yet have a `type-string-array` in `docs/index.asciidoc`, so we need to add that as well, e.g:

```asciidoc
:type-string-array: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#String_type[<string[]>]
```

package.json
```diff
@@ -186,6 +187,6 @@
   },
   "coordinates": [
     55.778264,
-    12.593202
```
watson: This change is probably just due to a merge conflict. Should just be reverted.

test/config.js
```diff
@@ -671,7 +675,13 @@ test('disableInstrumentations', function (t) {
       }
 
       for (const mod of modules) {
-        require(mod)
+        if (Array.isArray(mod)) {
```
watson: Because the modules array was flattened above, this shouldn't be necessary right?

docs/agent-api.asciidoc
Qard: Fixed

package.json
Qard: Fixed

test/config.js
Qard: Fixed

PR #1251, opened 2019-07-31
chore: pin wait-on to 3.2.0

watson:
v3.3.0 doesn't support Node.js 6 anymore.


watson:
Changing it to a `chore` as this is a devDependency... for some reason Greenkeeper created it as a `fix`.

PR #1252, opened 2019-07-31
ci(travis): disable npm cache

watson:
The npm cache used to be off by default on Travis, but since July 2019 they changed it to be on by default. This hides errors in our tests as updated to dependencies are no longer installed.

For details see:
https://docs.travis-ci.com/user/caching/#npm-cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ services:
   - cassandra
   
 cache:
+  npm: false
   directories:
     - deb-cache
   


Qard:
Looks like even 8.1 is broken by the `wait-on` dependency bump. 😞

watson:
The minimum required version in the next major will be v8.6.0, so there's no need to worry about 8.1 anymore.

Qard:
Should this be marked breaking-change then?

watson:
(We should do this for Jenkins as well)

watson:
> Should this be marked breaking-change then?

No, the test errors should preferably be fixed here first so we don't have a broken 2.x branch. Hopefully the wait-on fix is the only one.

watson:
@Qard The fix with wait-on helped

watson:
> (We should do this for Jenkins as well)

We're not caching `node_modules` on Jenkins, so all is good there

PR #1262, opened 2019-08-07
feat: allow agent.clearPatches to be called with array of names

watson:
Just a few things we missed in #1246 

I added it all in one PR, but named it after the main change. Hope that's ok.

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -107,9 +107,13 @@ Instrumentation.prototype.removePatch = function (modules, handler) {
   })
 }
 
-Instrumentation.prototype.clearPatches = function (name) {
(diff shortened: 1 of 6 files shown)


lib/instrumentation/index.js
```diff
@@ -94,21 +94,28 @@ Instrumentation.prototype.addPatch = function (modules, handler) {
     }
 
     this._patches.add(mod, handler)
-    this._startHook()
```
watson: I don't think there was any particular reason why this was done inside of the loop, so I moved it outside.

Qard: That's fine. It only needs to be triggered before returning to user code, so it can be just once at the end. 👍

Issue #968, opened 2019-04-01
Instrument the @elastic/elasticsearch module


watson:
The new Elasticsearch client has been moved to the `@elastic` npm namespace, so its name is now `@elastic/elasticsearch`.

This is a complete rewrite of the client and I therefore assume the old instrumentation code doesn't work with it. So this is essentially a new instrumentation.

gingerwizard:
Kibana has now been instrumented using the node APM agent. Currently, Kibana uses a mixture of the old and new ES client but is migrating to this new `@elastic/elasticsearch` module.

The ES performance team are looking to use Kibana's instrumentation data to capture the requests made to Elasticsearch - see https://github.com/elastic/rally-internal-tracks/issues/21. Without this auto-instrumentation, we will need to instrument Kibana ourselves as the _search and _async_search endpoints (the primary ones of interest) are already using this new client.

PR #1266, opened 2019-08-08
Added instrumentation for Elasticsearch Node.js client

delvedor:
This pr adds the support for the new Elasticsearch client for Node.js.

**TODO:**
- [ ] we should figure out how to serialize the errors to avoid the loss of important information
- [ ] can we signal when the http request has effectively started?
- [x] add test
- [x] add documentation

Closes: https://github.com/elastic/apm-agent-nodejs/issues/968 
Related: https://github.com/elastic/elasticsearch-js/pull/932

### DIFF
--- a/.ci/.jenkins_tav.yml
+++ b/.ci/.jenkins_tav.yml
@@ -1,29 +1,30 @@
 TAV:
+  - @elastic/elasticsearch
+  - apollo-server-express
+  - bluebird
+  - cassandra-driver
+  - elasticsearch
+  - express
+  - express-graphql
+  - express-queue
+  - fastify
+  - finalhandler
   - generic-pool
+  - got
+  - graphql
+  - handlebars
+  - hapi
+  - ioredis
+  - jade
+  - knex
+  - koa-router
+  - mimic-response
+  - mongodb-core
   - mysql
   - mysql2
-  - redis
-  - koa-router
-  - handlebars
-  - jade
-  - pug
-  - mongodb-core
-  - finalhandler
-  - ioredis
   - pg
-  - cassandra-driver
-  - tedious
+  - pug
+  - redis
   - restify
-  - fastify
-  - mimic-response
-  - got
-  - bluebird
-  - apollo-server-express
-  - knex
+  - tedious
   - ws
-  - graphql
-  - express-graphql
-  - elasticsearch
-  - hapi
-  - express
-  - express-queue
--- a/lib/instrumentation/modules/@elastic/elasticsearch.js
+++ b/lib/instrumentation/modules/@elastic/elasticsearch.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const { setDbContext } = require('../../elasticsearch-shared')
+
 module.exports = function (elasticsearch, agent, { version, enabled }) {
   if (!enabled) return elasticsearch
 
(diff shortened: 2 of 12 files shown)


watson:
@delvedor we use [conventional commits](https://www.conventionalcommits.org/). Could I get you to rename the commit to something along the lines of `feat(elasticsearch): add support @elastic/elasticsearch client`?

watson:
Nevermind... I got the opportunity to do it my self when I had to check it out locally to add tests... I force pushed, so you'll need to pull it down again

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,80 @@
...
+      this.on('prepare-request', (err, { meta, params }) => {
+        hasPrepareRequestEvent = true
+        if (err) {
+          return agent.captureError(err)
+        }
+
+        generateSpan(activeSpans, meta, params)
+      })
+
+      this.on('request', (err, { meta }) => {
```
delvedor: In the `request` event the `body` and `querystring` values will be already serialized, would it be possible to add the `setDbContext` call here?
In this way, we don't need to serialize the body/querystring twice ;)

.ci/.jenkins_tav.yml
```diff
@@ -1,29 +1,30 @@
 TAV:
+  - @elastic/elasticsearch
```
watson: Sorry for the diff... I only added this line, but sorted all entries alphabetically

lib/instrumentation/modules/@elastic/elasticsearch.js
watson: I'd love the make the query more readable in the case of an array of queries. What I do with the old client is format it as NDJSON so that each query is on its own line. If we use the JSON that you formatted, it will be one long line containing all the queries... not really sure how to best do this 🤔

delvedor: What you mean with "array of queries"?

watson: Example:

```js
  const body = [
    {},
    {
      query: {
        query_string: {
          query: 'pants'
        }
      }
    }
  ]

  client.msearch({ body }, function () {...})
```

test/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,196 @@
...
+
+  client.ping(function (err, result) {
+    t.error(err)
+    agent.endTransaction()
+    agent.flush()
+  })
+})
+
+test('client.search with callback', function userLandCode (t) {
+  resetAgent(done(t, 'GET', '/_search', 'q=pants'))
```
watson: In the old client we actually report this as `{"q":"pants"}` because that's what the user provides. The fact that the client chooses the add this to the querystring is not really relevant. It would be nicest if we reported the query as it was provided by the user.

The downside of that is that we'll have to serialize it twice, so the question is if it's worth it.

package.json
```diff
@@ -109,6 +109,7 @@
     "@commitlint/cli": "^8.0.0",
     "@commitlint/config-conventional": "^8.0.0",
     "@commitlint/travis-cli": "^8.0.0",
+    "@elastic/elasticsearch": "elastic/elasticsearch-js#update-events",
```
watson: If this is merged into master, it will break ones the branch is deleted. But for now it's the only way to test our compatibility with the yet unreleased `prepare-request` event which will ship in 7.4.

Maybe we should wait to merge this PR until that PR branch has been merged?

lib/instrumentation/modules/@elastic/elasticsearch.js
delvedor: In that case you will get a `ndjson` serialized body.

v1v:
jenkins run the tests please

trentm:
Closing in favour of https://github.com/elastic/apm-agent-nodejs/pull/1877

PR #1273, opened 2019-08-09
feat(metrics): add metricsLimit option

### LINKED ISSUE #1249: Limit the max amount of timer metrics
We missed a bullet point from elastic/apm-agent-nodejs#1064 in the implementation (https://github.com/elastic/apm-agent-nodejs/pull/1219):

- [ ] [Limit the max amount of timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.tzyh7loxf2is)

Qard:
Fixes #1249

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/test/metrics/index.js
+++ b/test/metrics/index.js
@@ -145,6 +145,26 @@ test('reports expected metrics', function (t) {
   metrics.start()
 })
 
+test('applies metrics limit', function (t) {
+  agent = mockAgent({
+    metricsInterval: 10,
+    metricsLimit: 2,
+    hostname: 'foo',
+    environment: 'bar'
+  }, (metricset = {}) => {
+    t.equal(Object.keys(metricset.samples).length, 2, 'has expected number of metrics')
+    t.end()
+  })
+
+  metrics = new Metrics(agent)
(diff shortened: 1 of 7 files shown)


test/metrics/index.js
```diff
@@ -145,6 +145,21 @@ test('reports expected metrics', function (t) {
...
 })
 
+test('applies metrics limit', function (t) {
+  agent = mockAgent({
+    metricsInterval: 10,
+    metricsLimit: 2,
+    hostname: 'foo',
+    environment: 'bar'
+  }, (metricset = {}) => {
+    t.ok(Object.keys(metricset.samples).length <= 2, 'has expected number of metrics')
```
watson: I assume this test works on the basis that we by default collect more than two metrics, and hence if we set the limit to 2, we should hit it just by calling `metrics.start()`. But in that case wouldn't we expect the length to be exactly 2?

Qard: Yes to both things. 😅

Issue #1261, opened 2019-08-06
NodeJS v10.16.1 crash while using elastic-apm-node


ronkorland:
**Describe the bug**
Since I have started using elastic-apm-node, it has started throwing `Could not process callsites` errors and node process to crash :

Errors:
```
could not capture stack trace for span { span: '627f2bf485604de9',
  parent: '5c26008e6557b10e',
  trace: '5d8bc7df6238f7bf83bccf7a95c8ce0f',
  name: 'GET',
  type: 'cache.redis',
  err: 'Could not process callsites' }

could not capture stack trace for span { span: 'e1e0625c65b656ca',
  parent: '6ed44490829eabca',
  trace: 'c4f78b304cf844d368b5830712c5c88f',
  name: 'SELECT FROM `XXXX`',
  type: 'db.mysql.query',
  err: 'Could not process callsites' }
```

Crash error:
```
13 verbose stack Exit status 1
13 verbose stack     at EventEmitter.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/index.js:301:16)
13 verbose stack     at EventEmitter.emit (events.js:198:13)
13 verbose stack     at ChildProcess.<anonymous> (/usr/local/lib/node_modules/npm/node_modules/npm-lifecycle/lib/spawn.js:55:14)
13 verbose stack     at ChildProcess.emit (events.js:198:13)
13 verbose stack     at maybeClose (internal/child_process.js:982:16)
13 verbose stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:259:5)
```

**To Reproduce**
Use this config 
node v10.16.1 + npm v6.9.0 (docker base image: node:10-slim)
elastic-apm-node v2.13.0
apm-server v6.5.4
mongoose v5.6.2
ioredis v4.3.0
mysql v2.16.0
mysql2 v1.6.4 

**Environment (please complete the following information)**
- Docker base image: node:10-slim
- OS: Debian
- Node.js version: 10.16.1
- APM Server version: 6.5.4
- Agent version: 2.13.0

**How are you starting the agent? (please tick one of the boxes)**
- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

watson:
[...]

PR #1274, opened 2019-08-09
fix: be less chatty if span stack traces cannot be parsed

watson:
The same error was accidentally logged twice (once on the `debug` level and once on the `warning` level). We decided in #1261 to only log on the `debug` level.


lib/instrumentation/span.js
```diff
@@ -87,12 +87,7 @@ Span.prototype._recordStackTrace = function (obj) {
 
   // TODO: This is expensive! Consider if there's a way to cache some of this
   stackman.callsites(obj, function (err, callsites) {
-    if (err || !callsites) {
```
watson: I checked the code and if there's no error, there's always a `callsites` object

PR #1275, opened 2019-08-09
perf: use for-of instead of forEach

watson:
Changed all uses for `forEach` inside of the `lib` folder to be for-of loops. I left out two instances as they looked better with `forEach`:
- https://github.com/elastic/apm-agent-nodejs/blob/245aa4520ced15e1932bd7d79507739d27090c2f/lib/filters/http-headers.js#L24
- https://github.com/elastic/apm-agent-nodejs/blob/245aa4520ced15e1932bd7d79507739d27090c2f/lib/filters/http-headers.js#L34

There's still quite a few uses of `forEach` in the tests but I didn't feel the need to change those. Let me know if we should do that (though many of those use the `index` argument of the `forEach` callback, so I don't think it's worth it).

I also changed all uses of `let` in existing for-of loops to `const` where possible.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -369,9 +369,9 @@ Agent.prototype.captureError = function (err, opts, cb) {
         })
 
         if (callsites) {
-          callsites.forEach(function (callsite) {
+          for (const callsite of callsites) {
             parsers.parseCallsite(callsite, true, agent, next())
-          })
+          }
         }
       })
     } else {
--- a/lib/config.js
+++ b/lib/config.js
@@ -319,10 +319,10 @@ class Config {
           if (haveAccepted || haveErrors) {
             if (haveAccepted) msg += `\nAPM Server accepted ${err.accepted} events in the last request`
             if (haveErrors) {
-              err.errors.forEach(error => {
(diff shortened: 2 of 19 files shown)


watson:
I discovered that you can't just covert `forEach` to for-of when dealing with maps as they don't give the same result. On maps you have to add `.values()` before for-of'ing it so that you don't get the array containing key+value

watson:
@Qard Yeah, I'm probably a bit on the fence about it... But I can't really make up my mind of which approach is best. You're right that the merge conflicts can easily become problematic and that it's easier to just update things as you're editing the code anyway. But on the other hand, there's a lot of code we just will never touch again - at least not anytime soon. And I see often that users making PR's get inspired by our existing code style. So if they see `var` and `forEach` etc all over the place, they will copy those patterns. There's also the joy of feeling like you're working on a modern codebase which shouldn't be understated 😕

Qard:
I think old code is one of the areas where modularity helps a lot. It's normal that different parts of a system will be more up-to-date than others, and modularity helps to more clearly define those barriers.

Issue #1257, opened 2019-08-02
Calling endTransaction() on custom transaction throws when active: false


MartinKolarik:
**Describe the bug**

Running the following code with an inactive agent (`active: false`)

```js
apmClient.startTransaction();
apmClient.endTransaction();
```

throws this error in the latest version:

```
TypeError: Cannot read property 'getOrCreateCounter' of null
    at Metrics.getOrCreateCounter (node_modules\elastic-apm-node\lib\metrics\index.js:43:33)
    at Transaction._captureBreakdown (node_modules\elastic-apm-node\lib\instrumentation\transaction.js:248:13)
    at Transaction.end (node_modules\elastic-apm-node\lib\instrumentation\transaction.js:205:8)
    at Instrumentation.endTransaction (node_modules\elastic-apm-node\lib\instrumentation\index.js:235:27)
    at Agent.endTransaction (node_modules\elastic-apm-node\lib\agent.js:94:47)
```

**Environment (please complete the following information)**
- OS: Windows 10
- Node.js version: 12
- APM Server version: N/A
- Agent version: 2.13.0

**How are you starting the agent? (please tick one of the boxes)**
- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

PR #1278, opened 2019-08-12
fix: avoid throwing when agent is in active: false mode

bsyk:
Add check for whether the agent is active before trying to increment counters.  This avoids the issue of throwing when a counter is non-existent due to the agent being inactive.

Fixes: elastic/apm-agent-nodejs#1257

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


bsyk:
@watson can I get some eyes on this small PR please?

Qard:
I think it'd probably be better that we just move the logic for _using_ the counter interface into `lib/metrics/index.js` and do the check there rather than here as this issue would exist _anywhere_ that there's a counter being used. 🤔

bsyk:
Currently, Counter is only use within Transaction as far as I can tell.
Are you proposing a better 'started' check within Metrics to protect against `this[registrySymbol]` being null during this call?

```
getOrCreateCounter (...args) {
  return this[registrySymbol].getOrCreateCounter(...args)
}
```

Qard:
I'm proposing _replacing_ that with something like `incrementCounter(name, amount)` with an early-return if not active. The issue currently is that the code is _expecting_ that to return an object that it can call `inc(...)` on, which won't be the case if not active.

Qard:
Also, the reason for putting the check _there_ versus in that breakdown metrics code is that we may _in the future_ decide we want to capture _more_ metrics in _other_ places and would then need to duplicate that check everywhere else too. It's better to have a central place where the check happens, that way we don't accidentally miss it again in the future.

bsyk:
Makes sense. I'll update with a change that reflects that tomorrow.

Qard:
jenkins run the tests please

bsyk:
I can't reproduce the jenkins build failure locally - has this test been an issue in the past?

Qard:
jenkins run the tests please

watson:
This has now been released in v2.16.1 - Thanks again for fixing this 💯

PR #1281, opened 2019-08-13
ci(jenkins): async hook false runs in parallel

v1v:
## Highlights
- Async hooks false branches run in parallel, aka less waste of time for getting feedback.
- Nodejs-6.x with async hook false is disabled.
- `generateStep` is now responsible to inject the `ELASTIC_APM_ASYNC_HOOKS` if `disableAsync` parameter is true. The default behavior as used to be, aka without the env variable.

## Test Cases
- Nodejs6.X for async configuration is not enabled

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -84,7 +84,7 @@ pipeline {
               node['NODEJS_VERSION'].each{ version ->
                 parallelTasks["Node.js-${version}"] = generateStep(version: version)
                 if (version != '6') {
-                  parallelTasks["Node.js-${version}-async-hooks-false"] = generateStep(version: version, async: false)
+                  parallelTasks["Node.js-${version}-async-hooks-false"] = generateStep(version: version, disableAsync: true)
                 }
               }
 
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -83,17 +83,15 @@ pipeline {
               def parallelTasksWithoutAsyncHooks = [:]
               node['NODEJS_VERSION'].each{ version ->
                 parallelTasks["Node.js-${version}"] = generateStep(version: version)
-                parallelTasksWithoutAsyncHooks["Node.js-${version}-async-hooks-false"] = generateStep(version: version)
+                if (!version.startsWith('6')) {
+                  parallelTasks["Node.js-${version}-async-hooks-false"] = generateStep(version: version, disableAsync: true)
```
watson: I think the name can be a bit confusing as there's also a feature in Node.js called `async`. So to make our lifes easier in the future, I suggest naming it `disableAsyncHooks`.

PR #1277, opened 2019-08-12
ci(jenkins): run tests without async hooks on Node.js nightly

watson:
The diff might be a little hard to read, but I essentially just duplicated the "Nightly Test" stage.

I've also split up the edge yaml file so that we in the future can test different Node.js versions for RC and Nightly (which we will have to the moment this PR lands - just felt it was cleaner to do in two PRs)

PR #1283, opened 2019-08-14
ci(jenkins): run edge tests in parallel

watson:


watson:
Following up from #1277 - Turned out that the diff was much simpler than I had originally thought 😅

watson:
@v1v 🤔 I guess it wasn't as simple as I thought... Jenkins is complaining

v1v:
I was actually on the way to work on this in another PR, you were much faster :). Let me review it and I'll push the required changes if I'm able to make it work

v1v:
Parallel stages before being running, aka waiting for a worker

image

Then in the BlueOcean view, you see something like 

image

Even though those branches have been already triggered:

image

And already assigned with some workers:

image

BlueOcean doesn't show those stages with parallel branches nicely though, let' see how it looks like when the build is done.

watson:
No worries. As long as it looks ok in the end 👍 

Thanks for fixing this 😃

Issue #1254, opened 2019-08-02
Change code to use url.URL instead of url.parse


watson:
`url.parse` is a deprecated core API. We should update our code to use `url.URL` instead.

yuxizhe:
@watson  I'm happy to take this issue.

watson:
@yuxizhe that would be really helpful. But it might be a bit tricky though... before I opened this issue, I checked if I could easily do it myself. But it's unfortunately not a 1-to-1 replacement due to the differences in how the two functions behave. But we can discuss how to best implement it if you still want to take it on

yuxizhe:
@watson Yes, I reread the docs https://nodejs.org/api/url.html#url_url_strings_and_url_objects.

I search in this repo find there are a few places where we use `url.parse`, most of them are just use `.pathname` propertie. Because URL and UrlObject both have `.pathname` propertie, so I think in these places they can be easily replaced. 

Except for the code in 
https://github.com/elastic/apm-agent-nodejs/blob/248c94531781a8a77f4e64e836797888c1674618/lib/instrumentation/http-shared.js#L83-L95
https://github.com/elastic/apm-agent-nodejs/blob/248c94531781a8a77f4e64e836797888c1674618/lib/instrumentation/http-shared.js#L116

Object.assign(options, URL) mix options with URL in a different format
```js
{ [Symbol(context)]:
   URLContext {
     flags: 912,
     scheme: 'https:',
     username: '',
     password: '',
     host: 'aaa.com',
     port: null,
     path: [ 'a', 'b' ],
     query: 'a=1',
     fragment: null },
  [Symbol(query)]: URLSearchParams { 'a' => '1' } }
``` 
So in this place I won't simply replace it with URL, 
what about the code below:
```js
function formatURL (url) {
  return {
    href: url.href,
    pathname: url.pathname,
    protocol: url.protocol,
    host: url.host,
    port: url.port,
    hostname: url.hostname,
    hash: url.hash,
    search: url.search,
  }
}
// NOTE: This will also stringify and parse URL instances
// to a format which can be mixed into the options object.
function ensureUrl (v) {
  if (typeof v === 'string') {
    if (url.URL) {
      const url = new URL(v)
      return formatURL(url)
    } else {
      return url.parse(v)
    }
  } else if (v instanceof url.Url) {
    return v
  } else if (url.URL && v instanceof url.URL) { // check for url.URL because it wasn't added until Node.js 6.13.0
    return formatURL(v)
[...]

PR #1284, opened 2019-08-14
chore: prepare code style for new version of linter

watson:
We can't upgrade to the new version of 'standard' currently because we still use the deprecated `url.parse` API in several locations (see issue no #1254). But to ensure fewer merge conflicts between the 2.x and the upcoming 3.x branch), we'll fix as many future linting issues as possible already now.


watson:
Jenkins is stuck, but Travis passed, so I'll just merge

PR #1291, opened 2019-08-15
feat: use external as span type

### LINKED ISSUE #1225: Use `span.type: external` instead of `span.type: ext`
We should be using `span.type: external` instead of `span.type: ext` to align with the other agents in accordance with https://github.com/elastic/apm/issues/117

Qard:
Rather than the shortened 'ext' use full 'external' in span types

Fixes #1225

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -109,7 +109,7 @@ function ensureUrl (v) {
 }
 
 exports.traceOutgoingRequest = function (agent, moduleName, method) {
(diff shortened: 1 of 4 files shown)


v1v:
For the record, `continuous-integration/jenkins/pr-merge` should point out to `https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-1291/1/`. The reason why it's not reflecting that particular link is related to the test local environments in my end.

PR #1292, opened 2019-08-16
feat: split types

### LINKED ISSUE #713: Split span type into 3 fields: Type, Subtype and Action
Implement support for this API change: https://github.com/elastic/apm-server/pull/1618

This involves changing this API:

```js
agent.startSpan([name][, type])
```

To this:

```js
agent.startSpan([name][, type][, subType][, action])
```

If the function is called with only 2 arguments, it's assumed that `type` is of the old dot-separated format `type.subtype.action`. In that case the agent should split on `.` and fill `type`, `subType` and `action` accordingly.

We should also add a new `span.setType(type[, subType][, action])` function. If the type is set via `span.type = '...'` it's assumed that it's in the old dot-separate format and should automatically be split like if `agent.startSpan()` is called only with the `name` and `type` argument.

Qard:
This splits span and transaction type field into type, subtype and action.

Fixes elastic/apm-agent-nodejs#713
~~Depends on elastic/apm-nodejs-http-client#77~~

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


Qard:
This removes support for `startSpan(name, type, traceparent)` and `startTransaction(name, type, traceparent)`. Only the options object form would work after this change. It _also_ means the options object can now be in _any_ position after name.

lib/agent.js
```diff
@@ -86,7 +86,7 @@ Agent.prototype.clearPatches = function (modules) {
   return this._instrumentation.clearPatches.apply(this._instrumentation, arguments)
 }
 
-Agent.prototype.startTransaction = function (name, type, { startTime, childOf } = {}) {
+Agent.prototype.startTransaction = function (name, type, subtype, action, { startTime, childOf } = {}) {
```
watson: Remember to update the TypeScript typings as well

lib/agent.js
```diff
@@ -98,7 +98,7 @@ Agent.prototype.setTransactionName = function (name) {
   return this._instrumentation.setTransactionName.apply(this._instrumentation, arguments)
 }
 
-Agent.prototype.startSpan = function (name, type, { childOf } = {}) {
+Agent.prototype.startSpan = function (name, type, subtype, action, { childOf } = {}) {
```
watson: Remember to update the TypeScript typings as well

lib/instrumentation/generic-span.js
```diff
@@ -8,18 +8,37 @@ const TraceParent = require('traceparent')
 
 module.exports = GenericSpan
 
-function GenericSpan (agent, type, opts) {
+function GenericSpan (agent, ...args) {
+  let opts = {}
```
watson: I'd probably rewrite this to a ternary (and also guard against the `null` edge-case):

```js
const opts = typeof args[args.length - 1] === 'object'
  ? (args.pop() || {})
  : {}
```

lib/instrumentation/generic-span.js
```diff
@@ -101,3 +120,16 @@ GenericSpan.prototype.addLabels = function (labels) {
   }
   return true
 }
+
+GenericSpan.prototype.setType = function (type, subtype, action) {
+  if (type && type.indexOf('.') >= 0) {
```
watson: If we're making this a breaking change, we don't have to support the old format anymore as far as I can see. If so, we don't need the custom `Object.defineProperty` setter either, which also simplifies things 😃

lib/instrumentation/generic-span.js
```diff
@@ -101,3 +120,16 @@ GenericSpan.prototype.addLabels = function (labels) {
...
+    const parts = type.split('.')
+    type = parts.shift()
+    subtype = parts.shift()
+    action = parts.shift()
+  }
+
+  this._type = type || 'custom'
```
watson: I think we can probably drop the string "custom" as a fallback in this refactor as well. If I recall correctly, the APM UI will see `null` and group transactions with `null` under its own group named "Custom" in the UI. And spans with type `null` will get special treatment as well. Right @elastic/apm-ui?

lib/instrumentation/span.js
```diff
@@ -15,21 +15,19 @@ module.exports = Span
...
-    opts = {
-      timer: childOf._timer,
-      childOf
-    }
-  } else {
-    opts.timer = childOf._timer
-    if (!opts.childOf) {
-      opts.childOf = childOf
-    }
+  let opts = {}
```
watson: Consider a ternary just as with `GenericSpan`

lib/instrumentation/span.js
```diff
@@ -43,7 +41,7 @@ function Span (transaction, name, type, opts) {
     this._recordStackTrace()
   }
 
-  this._agent.logger.debug('start span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: name, type: type })
+  this._agent.logger.debug('start span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type })
```
watson: Now that `type` doesn't contain all the same info as before, we're losing some insights with this change. Wouldn't it be worth it to include `this.subtype` and `this.action` as well? (this goes for all debug output like this across the code-base)

lib/instrumentation/transaction.js
```diff
@@ -82,7 +82,7 @@ Transaction.prototype.setCustomContext = function (context) {
   this._custom = Object.assign(this._custom || {}, context)
 }
 
-Transaction.prototype.startSpan = function (name, type, opts) {
+Transaction.prototype.startSpan = function (name, ...args) {
```
watson: Remember to update the TypeScript typings as well

lib/instrumentation/generic-span.js
Qard: Hard-dropping support for setting it with the old format could break existing custom instrumentation though. I think it's probably best to put it through a deprecation cycle. 🤔

lib/instrumentation/transaction.js
Qard: fixed

lib/instrumentation/span.js
Qard: fixed

lib/instrumentation/span.js
Qard: fixed

lib/agent.js
Qard: fixed

lib/instrumentation/generic-span.js
Qard: fixed

lib/agent.js
Qard: fixed

lib/instrumentation/generic-span.js
Qard: fixed

Qard:
Added a commit removing support for dot-form types and now the test issues are gone for some reason. 🤷‍♂

lib/instrumentation/generic-span.js
watson: Did you check that the UI behaves as I think? I'm not 100%

sqren: We currently don't group transactions with `name` property. We did talk about it but never got around to it. 
Is this also something that should happen for transactions without `type`? 

>  And spans with type null

Spans are only displayed in the timeline view. Afair we show spans regardless of their type - the type is only used to display an icon.

watson: My bad... spans and transactions are always required to have a type property

docs/span-api.asciidoc
```diff
@@ -47,15 +47,44 @@ This can also be set via <<apm-start-span,`apm.startSpan()`>>.
 The type of span.
 This can also be set via <<apm-start-span,`apm.startSpan()`>>.
 
-The type is a hierarchical string used to group similar spans together.
```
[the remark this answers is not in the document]
bmorelli25: I thought we agreed on adding another tag with the new information. For example, like we did here:
https://github.com/elastic/apm-agent-nodejs/blob/d525a8a8584186458800b54abe20e2e027ca11ed/docs/span-api.asciidoc#L86-L88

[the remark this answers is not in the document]
watson: Ah yes, thanks

docs/transaction-api.asciidoc
```diff
@@ -60,17 +80,22 @@ This is typically the HTTP status code,
 or e.g. "success" or "failure" for a background task.
 
 [[transaction-start-span]]
-==== `transaction.startSpan([name][, type][, options])`
+==== `transaction.startSpan([name][, type][, subtype][, action][, options])`
 
 [small]#Added in: v2.0.0#
```
watson:
```suggestion
[small]#Added in: v2.0.0#

[small]#Split `type` into `type`, `subtype` and `action` in: v3.0.0# 
```

index.d.ts
```diff
@@ -42,17 +42,55 @@ declare class Agent implements Taggable, StartSpanFn {
...
   currentTransaction: Transaction | null;
 
   // Spans
-  startSpan (name?: string, type?: string, options?: SpanOptions): Span | null;
+  startSpan(
+    name?: string,
+    options?: SpanOptions
+  ): Span | null;
+  startSpan(
+    name: string,
```
watson: Same as above, if we want to support this:

```js
startSpan(null, 'my-type')
```

Then we should change the definition to:

```suggestion
    name: string | null,
```

Then it does allow `null` values. But this is something that a test should confirm if you add a test with `name` being `null`.

(same goes for the other versions of `startSpan` below...)

index.d.ts
```diff
@@ -80,8 +118,11 @@ declare class GenericSpan implements Taggable {
   // timestamp, ended, id, traceId, parentId, sampled, duration()
 
   type: string;
+  subtype: string;
+  action: string;
   traceparent: string;
 
+  setType (type: string, subtype: string, action: string): void;
```
watson: I think we need to allow for the following ways of calling `setType`:

```js
span.setType('a', 'b', 'c')
span.setType('a', 'b', null)
span.setType('a', null, null)
span.setType('a', 'b')
span.setType('a', null)
span.setType('a')
```

The following TypeScript Typings should allow for that and nothing else:

```suggestion
  setType (type: string, subtype?: string | null): void;
  setType (type: string, subtype: string, action?: string | null): void;
  setType (type: string, subtype?: null, action?: null): void;
```

And we should add tests for this in `test/types/index.ts` as well to test that this all works as expected 😄

index.d.ts
```diff
@@ -80,8 +118,11 @@ declare class GenericSpan implements Taggable {
   // timestamp, ended, id, traceId, parentId, sampled, duration()
 
   type: string;
+  subtype: string;
```
watson: This can be `null`:

```suggestion
  subtype: string | null;
```

lib/instrumentation/generic-span.js
```diff
@@ -8,16 +8,24 @@ const TraceParent = require('traceparent')
...
   this._context = TraceParent.startOrResume(opts.childOf, agent._conf) // _context is used by the OT bridge, and should unfortunately therefore be considered public
   this._agent = agent
   this._labels = null
 
   this.timestamp = this._timer.start
-  this.type = type || 'custom'
   this.ended = false
   this.sync = true
+
+  this.type = null
```
watson: We should probably keep the initial value as the string `custom`:

```suggestion
  this.type = null || 'custom'
```

Either that, or we somehow need to make sure that it gets set to `custom` in another way if no value is provided.

index.d.ts
```diff
@@ -80,8 +118,11 @@ declare class GenericSpan implements Taggable {
...
+  action: string;
```
watson: This can be `null`:

```suggestion
  action: string | null;
```

lib/instrumentation/generic-span.js
```diff
@@ -101,3 +109,9 @@ GenericSpan.prototype.addLabels = function (labels) {
   }
   return true
 }
+
+GenericSpan.prototype.setType = function (type, subtype, action) {
+  this.type = type || 'custom'
```
watson: I think if we set the initial value to `custom` (see constructor), then we can just drop the fallback here right?

```suggestion
  this.type = type
```

lib/instrumentation/generic-span.js
```diff
@@ -101,3 +109,9 @@ GenericSpan.prototype.addLabels = function (labels) {
   }
   return true
 }
+
+GenericSpan.prototype.setType = function (type, subtype, action) {
```
watson: To make sure the default values are always `null`, I think we can do this:

```suggestion
GenericSpan.prototype.setType = function (type, subtype = null, action = null) {
```

index.d.ts
```diff
@@ -42,17 +42,55 @@ declare class Agent implements Taggable, StartSpanFn {
...
     options?: TransactionOptions
   ): Transaction | null;
   setTransactionName (name: string): void;
   endTransaction (result?: string | number, endTime?: number): void;
   currentTransaction: Transaction | null;
 
   // Spans
-  startSpan (name?: string, type?: string, options?: SpanOptions): Span | null;
+  startSpan(
+    name?: string,
```
watson: In pure JavaScript we currently allow the user to start a span without providing a name, e.g. `startSpan(null, { ... })`. To allow this in TypeScript Typings, we should change it to:

```suggestion
    name?: string | null,
```

index.d.ts
```diff
@@ -238,7 +300,13 @@ interface Taggable {
 }
 
 interface StartSpanFn {
-  startSpan (name?: string, type?: string, options?: SpanOptions): Span | null;
+  startSpan (
```
watson: I think this interface should be the same as the `startSpan` definitions defined in `Transaction` and `Agent`

lib/instrumentation/generic-span.js
watson: yeah, I did consider that but reluctantly concluded that this is already possible with the current implementation. So while it's not ideal, we're not introducing a "bug". I think the real solution is to move the `custom` fallback to the encoding step at which time there's no way for the user to later overwrite it with an invalid value. What do you think of that?

docs/transaction-api.asciidoc

index.d.ts
Qard: fixed

index.d.ts
Qard: I just made every field optional _and_ `string | null`, so it can accept _any_ combination of strings, nulls or undefined.

index.d.ts
Qard: fixed

lib/instrumentation/generic-span.js
Qard: I changed it to just use `'custom'` as the fallback in the span/transaction encoding functions.

index.d.ts
Qard: fixed

lib/instrumentation/generic-span.js
Qard: I moved it to the encoding functions.

lib/instrumentation/generic-span.js
Qard: fixed

index.d.ts
Qard: fixed

index.d.ts
Qard: fixed

docs/transaction-api.asciidoc
```diff
@@ -41,6 +41,24 @@ The type of the transaction.
 
 There's a special type called `request` which is used by the agent for the transactions automatically created when an incoming HTTP request is detected.
 
+[[transaction-subtype]]
+==== `transaction.subtype`
+
+[small]#Added in: v0.1.0#
```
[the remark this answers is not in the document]
watson: This somehow got reverted back to `v0.1.0`

docs/transaction-api.asciidoc
```diff
@@ -41,6 +41,24 @@ The type of the transaction.
...
+
+* +{type-string}+ *Default:* `custom`
+
+The subtype of the transaction.
+
+[[transaction-action]]
+==== `transaction.action`
+
+[small]#Added in: v0.1.0#
```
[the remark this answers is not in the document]
watson: This somehow got reverted back to `v0.1.0`

lib/instrumentation/generic-span.js
watson: perfect 👍

docs/transaction-api.asciidoc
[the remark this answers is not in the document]
Qard: fixed...again. 😅

docs/transaction-api.asciidoc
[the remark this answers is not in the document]
Qard: fixed...again. 😅

Qard:
The remaining failure is from mongodb-core tav tests, which seems to be unrelated.

Qard:
jenkins run the tests please

watson:

watson:
It's really strange that the [mongodb-core tests](https://github.com/elastic/apm-agent-nodejs/blob/master/test/instrumentation/modules/mongodb-core.js) are all failing on TAV for mongodb-core version 1.3.21. They all fail to call the [`resetAgent`](https://github.com/elastic/apm-agent-nodejs/blob/210abb5aaacbfcfc2dd7d31a32150cfc701413d4/test/instrumentation/modules/mongodb-core.js#L23) callback for some reason.

The only thing that I can think of that's different for the MongoDB tests is that their name includes a period. But we don't split on it, so I'm not sure how that should change anything.

watson:
I just discovered that the failing mongodb-core tests are also failing in `master`. Here's the latest daily CI build on `master`: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/master/153/pipeline

So this shouldn't be considered a blocker for this PR

watson:
I've created a separate issue for the TAV failues: #1337

watson:

Qard:

PR #1297, opened 2019-08-20
ci(jenkins): avoid builds in the master worker

v1v:
## Highlights
- `master` worker could be used but causes bottlenecks in some cases.
- This will allow to use of another worker and get rid of any potential bottlenecks when the build queue is massive.

### DIFF
--- a/.ci/downstreamTests.groovy
+++ b/.ci/downstreamTests.groovy
@@ -40,7 +40,7 @@ pipeline {
     Checkout the code and stash it, to use it on other stages.
     */
     stage('Checkout') {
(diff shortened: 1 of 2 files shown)


watson:
Has this anything to do with the `master` branch or is the `master` worker something else?

Qard:
Something else. The master node is responsible for dispatching triggered jobs to worker nodes in the Jenkins cluster. The master node _itself_ can be a worker node too, but it's not recommended to use it that way in a cluster that gets a lot of jobs--especially long-running ones.

PR #1298, opened 2019-08-21
feat(http): add monitorIncomingHTTPRequests config

Qard:
PR for splitting disabling of outgoing http requests from incoming http requests.

Closes elastic/apm-agent-nodejs#434

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -61,7 +61,7 @@ var DEFAULTS = {
   logLevel: 'info',
   metricsInterval: '30s',
   metricsLimit: 1000,
-  monitorIncomingHTTPRequests: true,
+  instrumentIncomingHTTPRequests: true,
   centralConfig: true,
   serverTimeout: '30s',
   sourceLinesErrorAppFrames: 5,
@@ -99,6 +99,7 @@ var ENV_TABLE = {
   globalLabels: 'ELASTIC_APM_GLOBAL_LABELS',
   hostname: 'ELASTIC_APM_HOSTNAME',
   instrument: 'ELASTIC_APM_INSTRUMENT',
+  instrumentIncomingHTTPRequests: 'ELASTIC_APM_INSTRUMENT_INCOMING_HTTP_REQUESTS',
   kubernetesNamespace: ['ELASTIC_APM_KUBERNETES_NAMESPACE', 'KUBERNETES_NAMESPACE'],
   kubernetesNodeName: ['ELASTIC_APM_KUBERNETES_NODE_NAME', 'KUBERNETES_NODE_NAME'],
   kubernetesPodName: ['ELASTIC_APM_KUBERNETES_POD_NAME', 'KUBERNETES_POD_NAME'],
@@ -106,7 +107,6 @@ var ENV_TABLE = {
   logLevel: 'ELASTIC_APM_LOG_LEVEL',
   metricsInterval: 'ELASTIC_APM_METRICS_INTERVAL',
   metricsLimit: 'ELASTIC_APM_METRICS_LIMIT',
-  monitorIncomingHTTPRequests: 'ELASTIC_APM_MONITOR_INCOMING_HTTP_REQUESTS',
   payloadLogFile: 'ELASTIC_APM_PAYLOAD_LOG_FILE',
   centralConfig: 'ELASTIC_APM_CENTRAL_CONFIG',
   secretToken: 'ELASTIC_APM_SECRET_TOKEN',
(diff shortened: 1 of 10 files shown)


watson:
Tip: You can use the new Draft PR feature of GitHub for WIP PR's 😃

test/instrumentation/_agent.js
```diff
@@ -12,27 +13,14 @@ var sharedInstrumentation
...
 module.exports = function mockAgent (expected, cb) {
   var agent = {
-    _conf: {
-      serviceName: 'service-name',
-      active: true,
-      instrument: true,
-      captureSpanStackTraces: true,
+    _conf: config({
+      abortedErrorThreshold: '250ms',
+      asyncHooks: false,
```
watson: Why did you add this?

test/instrumentation/_agent.js
```diff
@@ -12,27 +13,14 @@ var sharedInstrumentation
...
+      breakdownMetrics: false,
```
watson: Why did you add this?

test/instrumentation/_agent.js
Qard: There was a test issue...I forget exactly what it broke. Technically though, the _omission_ of `asyncHooks: false` previously made it _implicitly_ false and this just restores the previous behaviour.

test/instrumentation/_agent.js
Qard: Just to be _extra_ sure it's not producing any metrics stuff that'd mess with event count results? 🤷‍♂

lib/config.js
```diff
@@ -69,6 +69,7 @@ var DEFAULTS = {
   logLevel: 'info',
   metricsInterval: '30s',
   metricsLimit: 1000,
+  monitorIncomingHTTPRequests: true,
```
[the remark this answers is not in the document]
Qard: I was actually thinking the same thing that maybe it should use "instrument" instead.

test/instrumentation/_agent.js
watson: But this is just a mock agent, so it shouldn't start any metrics stuff right?

test/instrumentation/_agent.js
watson: As this is a mock agent I'm just wondering where this makes a difference... and since you didn't add any new tests, it's weird that this update should change anything. So I'm a bit worried if it's covering over something else 🤔

test/instrumentation/_agent.js
Qard: No, it starts the metrics stuff. Not sure _why_, if it's being disabled again by `metricsInterval: 0`, but it's been that way for awhile. 🤷‍♂

watson: I originally thought that since we don't call the `agent.start()` function, then we never start the metrics, but now I just realize that we manually start the metrics in this mock agent:

https://github.com/elastic/apm-agent-nodejs/blob/8aa4344cedfb9515bca21c7efa89da4f4c6e2bcf/test/instrumentation/_agent.js#L34-L35

Do you remember why we do that? I mean, if we in the config completely disable it, then why try to start it?

lib/config.js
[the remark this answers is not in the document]
Qard: fixed

Qard:
I've narrowed the 12.0 failure to `http.request(...)` losing context before emitting the `'response'` event, which calls the response handler function, if given. I'm not sure yet exactly _where_ that context loss occurs between the the call and the handler. Possibly some internals in `net`? 🤔

Qard:
The failure is due to the issue fixed by https://github.com/nodejs/node/pull/27477, which didn't land until 12.2.0. As a result, all 12.x releases _before_ 12.2.0 will actually have a context loss in the response handler of all outgoing requests. Apparently our existing tests just happened to not cover that particular case. 🤔

test/instrumentation/modules/http/disabling.js
```diff
@@ -0,0 +1,162 @@
...
+
+    function ping (url, fn = res => res.resume()) {
+      return new Promise((resolve, reject) => {
+        const req = http.get(url, res => {
+          res.on('error', reject)
+          res.on('end', resolve)
+          fn(res)
+        })
+        req.on('error', reject)
+        req.end()
```
watson: No need to call `end()` on a `http.get()` request as they are automatically ended

lib/instrumentation/http-shared.js
```diff
@@ -156,6 +156,7 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       return req
 
       function onresponse (res) {
+        ins._recoverTransaction(span.transaction)
```
watson: It would be nice also get this as a separate PR so we can get it into 2.x

watson:
The PR description still says WIP, but I assume it's ready now?

Qard:
Yep, just forgot to update the description. 👍

PR #1300, opened 2019-08-21
fix: change agent active status log message to debug level

watson:
Previously, the agent would log the following message on the info level if `active: false`:

    Elastic APM agent is inactive due to configuration

This proved problematic for CLI applications that ships with the agent turned off by default.

In cases where `active` is set to `false` on purpose, it shouldn't be necessary to log this message on the info level, instead we now log this more specific message on the debug level:

    Elastic APM agent disabled (`active` is false)

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -128,7 +128,7 @@ Agent.prototype.start = function (opts) {
   }
 
   if (!this._conf.active) {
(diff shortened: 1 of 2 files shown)


watson:
@Qard what do you think of this change? It came up during my work on #43548 because this message was logged when you for instance run Kibana from the terminal:

```
yarn start --help
```

PR #1304, opened 2019-08-22
ci(jenkins): avoid using the any to skip the master

v1v:
## Highlights
- When working with ephemeral workers the only agent which might be always up is the `master` one, besides the static workers if any. So far, we don't have any after moving to windows ephemeral workers.
- This particular change will help to scale up when the build queue is long but will increase a bit the wait time.
- In other words, there are three major concepts regarding the overall time:
  - Waste of time for resources during the build
  - Waste of time for resources in the queue
  - Real build time.
- This particular fix will reduce the time for the first and second one.
- GIT_SHA rather than GIT_BASE_COMMIT as the last one is caused when doing a merge within a PR.


v1v:
The travis GitHub check is not required for this particular PR, I'll merge this PR now

PR #1308, opened 2019-08-23
ci(travis): don't run TAV, Nightly, or RC builds on Travis

watson:
Our Jenkins setup is now at a level where we can rely on that for running these builds.

On Jenkins, the TAV tests don't run automatically for each PR as it has been the case with Travis. Instead, you have to manually write a comment on the PR instructing Jenkins which TAV tests to run, eg:

- `jenkins run the tests please`
- `jenkins run the module tests for memcached`
- `jenkins run the module tests for memcached,hapi`
- `jenkins run the module tests for ALL`

(This PR also removes an unused `slack` section left over from Opbeat)


watson:
When merged, we can remove the cron job fra Travis as well

PR #1314, opened 2019-08-27
Revert "ci(jenkins): avoid using the any to skip the master (#1304)"

v1v:
This reverts commit 8b585c6814796d87203c031ede6a4ed9d959a66c.

## Highlights
- As long as we use the share step `gitCheckout` we might need to use the agent top-level to share the env variables between stages.
- It's not ideal but even though the changes were quite straight, its behavior was not as expected. 
- Let's keep the pipeline stable for now and we will figure out what's going on.

### DIFF
--- a/.ci/downstreamTests.groovy
+++ b/.ci/downstreamTests.groovy
@@ -11,7 +11,7 @@ it is need as field to store the results of the tests.
 @Field def nodeTasksGen
 
 pipeline {
-  agent none
+  agent any
   environment {
     REPO = 'apm-agent-nodejs'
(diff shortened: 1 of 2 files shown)


watson:
Do you need this to be part of the `2.x` branch as well? We're probably not doing any more 2.x releases, so if this needs to be part of the `2.x` branch, we need to do a backport once we know that we're not doing any more 2.x releases.

v1v:
That's indeed a good point, as the previous commits is already in the `2.x` branch, I'll need to raise another PR to backport it. Thanks for the heads up.

watson:
I’ll take care of the backporting if it comes to that 👍

PR #1319, opened 2019-08-30
fix(lambda): handle traceparent case-insensitively

### LINKED ISSUE #1315: Lambda instrumentation should not expect lowercase headers
**Describe the bug**

The `apm.lambda` instrumentation expects to find the trace header lower-cased:

https://github.com/elastic/apm-agent-nodejs/blob/d866894a810e5dedee1a857781712efb7beeb251/lib/lambda.js#L78

At least when testing with the local API Gateway (using `sam local start-api`), the headers are not lower-cased. In fact the gateway seems to transform them, so "foo: bar" becomes "Foo: bar".

I haven't found a formal definition of what the headers are _supposed_ to look like, but I've come across these pages which suggest that the headers aren't expected to be lower-cased:
 - https://docs.aws.amazon.com/lambda/latest/dg/with-on-demand-https.html (see example event)
 - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-known-issues.html ("Header names and query parameters are processed in a case-sensitive way" -- seems to conflict with `sam local start-api`, but that could just be a bug in the local API Gateway code.)
 - https://github.com/serverless/serverless/issues/2765#issuecomment-331158257

**To Reproduce**
1. Install the SAM CLI: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install.html
2. Create a new SAM app: `sam init -r nodejs10.x -n foo`
3. Modify foo/hello-world/app.js so that the response body includes the event:

```js
response = {
    'statusCode': 200,
    'body': JSON.stringify({
        message: 'hello world',
        input: event
    })
}
```

4. Start the local API Gateway: `cd foo && sam local start-api`
5. Send an HTTP request: `curl -H "Elastic-Apm-Traceparent: foobar" http://localhost:3000/hello`

You should see in the response that the headers contain the key `Elastic-Apm-Traceparent`, rather than `elastic-apm-traceparent`.

(I originally instrumented the app with the agent, and found that it was not honouring incoming elastic-apm-traceparent headers from the RUM agent. The above is a simplification.)

**Expected behavior**
When instrumenting the app with the agent, transactions should honour the incoming header regardless of what case it was sent to the API Gateway.

- OS: Linux
- Node.js version: 10.16.0
- APM Server version: 7.3.0
- Agent version: 2.16.0

axw:
Headers coming from API Gateway to Lambda may not be
lowercase, so when extracting the elastic-apm-traceparent
header, compare case-insensitively.

Fixes #1315

### DIFF
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const shimmer = require('./instrumentation/shimmer')
+const entries = require('object.entries') // TODO: Remove when Node.js 6 is no longer supported
 
 module.exports = function elasticApmAwsLambda (agent) {
   function captureContext (trans, payload, context, result) {
@@ -75,7 +76,7 @@ module.exports = function elasticApmAwsLambda (agent) {
     return function wrappedLambda (payload, context, callback) {
       let parentId
       if (payload.headers !== undefined) {
-        for (let [key, value] of Object.entries(payload.headers)) {
+        for (const [key, value] of entries(payload.headers)) {
           if (key.toLowerCase() === 'elastic-apm-traceparent') {
             parentId = value
             break
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -75,7 +75,12 @@ module.exports = function elasticApmAwsLambda (agent) {
     return function wrappedLambda (payload, context, callback) {
       let parentId
       if (payload.headers !== undefined) {
-        parentId = payload.headers['elastic-apm-traceparent']
+        for (let [key, value] of Object.entries(payload.headers)) {
+          if (key.toLowerCase() === 'elastic-apm-traceparent') {
+            parentId = value
+            break
+          }
+        }
       }
       const trans = agent.startTransaction(context.functionName, type, {
         childOf: parentId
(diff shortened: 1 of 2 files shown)


lib/lambda.js
```diff
@@ -75,7 +75,12 @@ module.exports = function elasticApmAwsLambda (agent) {
     return function wrappedLambda (payload, context, callback) {
       let parentId
       if (payload.headers !== undefined) {
-        parentId = payload.headers['elastic-apm-traceparent']
+        for (let [key, value] of Object.entries(payload.headers)) {
```
watson: `Object.entries` is unfortunately not available in Node.js 6. We have a polyfill you can use instead by requring `object.entries` in the top of the file like so:

```js
const entries = require('object.entries') // TODO: Remove when Node.js 6 is no longer supported

// ...

for (const [key, value] of entries(payload.headers)) {
  // ...
}
```

(I also changed `let` to `const` in the `for` loop as this is going to be a requirement by our linter soon, so we might as well ensure it's fixed already)

axw: Thanks for the review! Updated.

watson:
Awesome! Thanks for the help ☺️

PR #1324, opened 2019-09-02
fix(package): bump elastic-apm-http-client to version ^8.2.1

watson:


watson:
Getting ready to release a new version with the Lambda fix now, and it would be great to get this change in it as well... going to merge without a review as it's trivial.

PR #1325, opened 2019-09-03
ci(jenkins): decouple linting from CI pipeline

v1v:
## Highlights
- The main pipeline has been split into two `pipelines`.
- The main pipeline will evaluate the `linting check` stage in the branches only.
- The new pipeline will run the `linting-check` and `basic tests` for the PRs only.
- The new pipeline will create a new GitHub check called `apm-ci/linting/pr-merge`
-  If PRs from contributors modify any of the existing scritps then it will fail.
- Add linting coverage per Node.js version for `npm test:types`, `npm test:babel` and `npm test:esm`
- Basic linting coverage for Node.js version `12` which will run `npm run lint` and `npm run test:deps`
- `.ci/scripts/lint-commits.sh` is not required anymore as it runs within the `npm run lint` context.
- GitHub comment for running this pipeline: `jenkins run the lint please`

## Test Cases

*NOTE*: the current test cases have been verified using a local jenkins instance, that's the reason those GH checks are using the `v1v` account.

- When a PR is created then, a new GH check has been added

image

- When a PR does not contain any linting issues then GH check is green

image

and pipeline view

image

- When a PR does contain a linting issue (see https://github.com/elastic/apm-agent-nodejs/pull/1325/commits/f39cf8b2b675d0b915fef1cad036762ee929e00b) then GH check failed

image

and pipeline view

image

- When a PR does change any of the blacklisted files, then the validation will confirm everything works as expected.

```
14:48:40  [INFO] githubPrCheckApproved: Title: ci(jenkins): decouple linting from CI pipeline - User: v1v - Author Association: MEMBER
```

- Parallel branches

image

## Tasks
- [x] Validate locally


watson:
Great to see this progressing 😃 Just a few questions:

- We want to run linting in parallel with the regular tests, so the regular tests don't have to wait for the lint tests to complete. Does this PR still allow that?
- What do you mean with "If PRs from contributors modify any of the existing scritps then it will fail."?
- As far as I can see this currently only concerns commit message linting. It would be great to get code linting into this as well (performed by the `standard` command)

v1v:
> We want to run linting in parallel with the regular tests, so the regular tests don't have to wait for the lint tests to complete. Does this PR still allow that?

IIUC, I guess it does. Let me explain:
The main pipeline will behave, as usual, in other words, it will run in parallel all the tests for all the PRs and also the linting for the branches/tags. I have not changed the behavior about parallel executions, but excluding the linting for the pull requests.
The `.ci/scripts/test_basic.sh` and `.ci/scripts/lint-commits.sh` scripts will run for each PR in the new pipeline.

> What do you mean with "If PRs from contributors modify any of the existing scripts then it will fail."?

It's a security gate to ensure any changes in `.ci/scripts/test_basic.sh` or `.ci/scripts/lint-commits.sh` should be done by someone from the elastic organisation. This will ensure the CI environment cannot be compromised from the security point of view. 

In other words, those files don't change often and if so, most likely will be changed by us. If not, we wanna ensure the PR is evaluated beforehand as already done.

> As far as I can see this currently only concerns commit message linting. It would be great to get code linting into this as well (performed by the standard command)

It's actually launched within the `Basic tests` step:
- https://github.com/elastic/apm-agent-nodejs/pull/1325/files#diff-e80a1567a1f2a242e4c25e451d0e1f3fR27

Is that correct?

watson:
Yes, currently code linting is performed here before every test:

https://github.com/elastic/apm-agent-nodejs/blob/c8d1b9e2c29645190dc12bc270132cfbb685617c/.ci/scripts/test_basic.sh#L13

That's not necessary as the linting only needs to run once, and not once per every combination of our test matrix. So it would be great if we could move that to our linting pipeline also 😃

.ci/linting.groovy
```diff
@@ -0,0 +1,54 @@
...
+    disableResume()
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+    rateLimitBuilds(throttle: [count: 60, durationName: 'hour', userBoost: true])
+    quietPeriod(10)
+  }
+  triggers {
+    issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?lint(?:\\W+please)?.*')
+  }
+  stages {
+    stage('Basic Test') {
```
mdelapenya: Is `Basic Test` a descriptive enough name? As this is a linting pipeline, I can see another name such as `Linting tests` or similar.

wdyt?

.ci/linting.groovy
```diff
@@ -0,0 +1,76 @@
...
+
+def verifyChangesAreApproved() {
+  def ret = 0
+  catchError(buildResult: 'SUCCESS', message: 'Trap any errors') {
+    // This is required to populate the env variables
+    githubEnv()
+    if (!githubPrCheckApproved()) {
+       ret = sh(label: 'Validate changes',
+                script: '''
+                  files=".ci/scripts/test_basic.sh .ci/scripts/lint-commits.sh"
```
v1v:
```suggestion
                  files=".ci/scripts/test_basic.sh .ci/scripts/test_types_babel_esm.sh"
```

PR #1335, opened 2019-09-09
feat: add toString methods for logging

Qard:
This enables logging spans and transactions in a string form
which can be used for log correlation.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -783,3 +783,37 @@ apm.clearPatches('timers')
 
 apm.clearPatches(['timers'])
 ----
+
+[[apm-current-trace-ids]]
+==== `apm.currentTraceIds`
+
+[small]#Added in: #2.17.0
+
+Produces an object containing `trace.id` and either `transaction.id` or `span.id` when a current transaction or span is available.
+When no transaction or span is available it will return an empty object.
+This enables log correlation to APM traces with structured loggers.
+
+[source,js]
+----
+{
+  "trace.id": "abc123",
+  "transaction.id": "abc123"
+}
+// or ...
+{
+  "trace.id": "abc123",
+  "span.id": "abc123"
+}
+----
+
+All current trace id objects,
+including the empty form,
+include a `toString()` implementation.
+This enables log correlation to APM traces with text-only loggers.
+
+[source,js]
+----
+"trace.id=abc123 transaction.id=abc123"
+// or ...
+"trace.id=abc123 span.id=abc123"
+----
--- a/docs/span-api.asciidoc
+++ b/docs/span-api.asciidoc
@@ -133,13 +133,34 @@ it's converted to a string before being sent to the APM Server
 Add several labels on the span.
 You can add labels multiple times.
 
+[[span-ids]]
+==== `span.ids`
+
+[small]#Added in: #2.17.0
+
+Produces an object containing `span.id` and `trace.id`.
+This enables log correlation to APM traces with structured loggers.
+
+[source,js]
+----
+{
+  "trace.id": "abc123",
+  "span.id": "abc123"
+}
+----
+
 [[span-to-string]]
 ==== `span.toString()`
 
 [small]#Added in: #2.17.0
 
 Produces a string representation of the span to inject in log messages.
-This enables log correlation to APM traces.
+This enables log correlation to APM traces with text-only loggers.
+
+[source,js]
+----
+"trace.id=abc123 span.id=abc123"
+----
 
 [[span-end]]
 ==== `span.end([endTime])`
--- a/docs/span-api.asciidoc
+++ b/docs/span-api.asciidoc
@@ -133,6 +133,14 @@ it's converted to a string before being sent to the APM Server
 Add several labels on the span.
 You can add labels multiple times.
 
+[[span-to-string]]
+==== `span.toString()`
+
+[small]#Added in: #2.17.0
+
+Produces a string representation of the span to inject in log messages.
+This enables log correlation to APM traces.
+
 [[span-end]]
 ==== `span.end([endTime])`
 
--- a/docs/transaction-api.asciidoc
+++ b/docs/transaction-api.asciidoc
@@ -206,6 +206,14 @@ elasticApm.init({
 
 See the {apm-rum-ref}[JavaScript RUM agent documentation] for more information.
 
+[[transaction-to-string]]
+==== `transaction.toString()`
+
+[small]#Added in: #2.17.0
+
+Produces a string representation of the transaction to inject in log messages.
+This enables log correlation to APM traces.
+
 [[transaction-end]]
 ==== `transaction.end([result][, endTime])`
 
--- a/docs/transaction-api.asciidoc
+++ b/docs/transaction-api.asciidoc
@@ -206,13 +206,34 @@ elasticApm.init({
 
 See the {apm-rum-ref}[JavaScript RUM agent documentation] for more information.
 
+[[transaction-ids]]
+==== `transaction.ids`
+
+[small]#Added in: #2.17.0
+
+Produces an object containing `transaction.id` and `trace.id`.
+This enables log correlation to APM traces with structured loggers.
+
+[source,js]
+----
+{
+  "trace.id": "abc123",
+  "transaction.id": "abc123"
+}
+----
+
 [[transaction-to-string]]
 ==== `transaction.toString()`
 
 [small]#Added in: #2.17.0
 
 Produces a string representation of the transaction to inject in log messages.
-This enables log correlation to APM traces.
+This enables log correlation to APM traces with text-only loggers.
+
+[source,js]
+----
+"trace.id=abc123 transaction.id=abc123"
+----
 
 [[transaction-end]]
 ==== `transaction.end([result][, endTime])`
--- a/lib/instrumentation/ids.js
+++ b/lib/instrumentation/ids.js
@@ -1,0 +1,33 @@
+'use strict'
+
+const { stringify } = require('querystring')
+
+class Ids {
+  toString () {
+    return stringify(this, ' ', '=')
+  }
+}
+
+class SpanIds extends Ids {
+  constructor (span) {
+    super()
+    this['trace.id'] = span.traceId
+    this['span.id'] = span.id
+    Object.freeze(this)
+  }
+}
+
+class TransactionIds extends Ids {
+  constructor (transaction) {
+    super()
+    this['trace.id'] = transaction.traceId
+    this['transaction.id'] = transaction.id
+    Object.freeze(this)
+  }
+}
+
+module.exports = {
+  Ids,
+  SpanIds,
+  TransactionIds
+}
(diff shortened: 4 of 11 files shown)


lib/instrumentation/span.js
```diff
@@ -46,6 +47,13 @@ function Span (transaction, name, type, opts) {
   this._agent.logger.debug('start span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: name, type: type })
 }
 
+Span.prototype.toString = function () {
```
axw: This is enforcing a particular format which may be out of place with the user's logging. That's the main reason why I wasn't too keen on adding integration with unstructured loggers in the first place.

What about this as an alternative: add an `Agent.currentTraceIds` method, similar to `Agent.currentTraceparent`, which will return an object containing the IDs of the current span if there is one, or otherwise the current transaction if there is one.

Then users could use it like `querystring.stringify(apm.currentTraceIds, ',', '=')`, and it would also be used for structured logging, e.g. labels in winston.

WDYT?

Qard: We already have getters for the various ids, so that doesn't really seem all that useful to me. If the format generated by toString() doesn't produce what you want, you can always use those directly when writing your own log integration. 🤷‍♂

axw: You have to have the relevant object (span or transaction) before you can get its IDs, and there's some complexity here:

 - there may or may not be a transaction
 - there may or may not be a span
 - the agent currently doesn't produce no-op spans for non-sampled transactions, but that may not be the case forever; if we start creating no-op spans, then it wouldn't make sense to use the span ID

So without something like `currentTraceIds`, how would you get the labels for for the current span or transaction for a structured logger? I imagine something like the following, assuming no-op spans in the future:

```js
const span = apm.currentSpan
if (span && span.sampled) {
    return {'trace.id': span.traceId, 'span.id': span.id}
} else {
    const tx = apm.currentTransaction
    return tx ? {'trace.id': tx.traceId, 'transaction.id': tx.id} : {}
}
```

This is relevant for unstructured loggers too. They also need to call `toString` on the current-sampled-span-or-transaction, so the conditional parts of of that code still apply. Then if people want to format it differently, everyone needs to repeat that.

Maybe I'm missing a simpler way?

Qard: Fair point about having to deal with possible non-existence of spans/transactions. It puts extra work on the user for serialization though. Could maybe go for a hybrid approach, something like:

```js
const { stringify } = require('querystring')

class Ids {
  toString () {
    return stringify(this)
  }
}

class SpanIds extends Ids {
  constructor(span) {
    super()
    this['span.id'] = span.id
    this['trace.id'] = span.traceId
  }
}

class TransactionIds extends Ids {
  constructor(transaction) {
    super()
    this['transaction.id'] = transaction.id
    this['trace.id'] = transaction.traceId
  }
}

function currentTraceIds () {
  const span = apm.currentSpan
  if (span && span.sampled) {
      return new SpanIds(span)
  } else {
      const tx = apm.currentTransaction
      return tx ? new TransactionIds(tx) : new Ids()
  }
}
```

axw: That seems like a good approach to me. Then people who want to override the string format could either change `Ids.prototype.toString` or pass the object into a function depending on their use case.

BTW is it typical to use a querystring format in Node.js app logs? I would have thought space-separated would be more common, but I don't know.

Qard: I just pushed an update with a new API based on your feedback. Let me know what you think.

axw: The new API looks great to me, thanks. Good idea adding `.ids` to Transaction and Span too.

lib/instrumentation/ids.js
```diff
@@ -0,0 +1,31 @@
...
+  toString () {
+    return stringify(this, ',', '=')
+  }
+}
+
+class SpanIds extends Ids {
+  constructor (span) {
+    super()
+    this['trace.id'] = span.traceId
+    this['span.id'] = span.id
```
watson: Wouldn't it be nicer if the user could access this without having to use quotes? E.g. `span.spanId` instead of `span['span.id']`

lib/instrumentation/ids.js
```diff
@@ -0,0 +1,31 @@
...
+    Object.freeze(this)
```
watson: Just out of curiosity: Why the freeze?

lib/instrumentation/index.js
```diff
@@ -83,6 +84,17 @@ function Instrumentation (agent) {
   }
 }
 
+Object.defineProperty(Instrumentation.prototype, 'ids', {
+  get () {
+    var current = this.currentSpan || this.currentTransaction
```
watson:
```suggestion
    const current = this.currentSpan || this.currentTransaction
```

lib/instrumentation/index.js
```diff
@@ -83,6 +84,17 @@ function Instrumentation (agent) {
...
+    return current ? current.ids : new Ids()
+  }
+})
+
+Instrumentation.prototype.toString = function () {
```
watson: Having a `toString()` on the `Instrumentation` object seems a little odd to me... what's the use-case for this?

test/agent.js
```diff
@@ -224,6 +224,41 @@ test('#currentTraceparent', function (t) {
   })
 })
 
+test('#currentTraceIds', function (t) {
+  t.test('no active transaction or span', function (t) {
```
watson: It would probably be a good idea to also check the return value of `agent.currentTraceIds.toString()` in this and the two other tests below

lib/instrumentation/ids.js
Qard: I went with this form to align with the property names decided in the main design issue.

lib/instrumentation/ids.js
Qard: No particular reason. There's no reason it _should_ be mutable, so I figured why not just make it _immutable_. 🤷‍♂

lib/instrumentation/index.js
Qard: Just to align with the other places, but could probably just remove that. 🤷‍♂

test/agent.js
Qard: fixed

lib/instrumentation/index.js

lib/instrumentation/index.js

docs/agent-api.asciidoc
```diff
@@ -783,3 +783,17 @@ apm.clearPatches('timers')
 
 apm.clearPatches(['timers'])
 ----
+
+[[apm-current-trace-ids]]
+==== `apm.currentTraceIds`
+
+[small]#Added in: #2.17.0
+
+Produces an object containing `trace.id` and either `transaction.id` or `span.id` when a current transaction or span is available.
```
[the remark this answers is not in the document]

docs/span-api.asciidoc
```diff
@@ -133,6 +133,22 @@ it's converted to a string before being sent to the APM Server
 Add several labels on the span.
 You can add labels multiple times.
 
+[[span-ids]]
+==== `span.ids`
+
+[small]#Added in: #2.17.0
+
+Produces an object containing `span.id` and `trace.id`.
```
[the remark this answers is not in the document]
Qard: fixed

docs/span-api.asciidoc
```diff
@@ -133,6 +133,22 @@ it's converted to a string before being sent to the APM Server
...
+This enables log correlation to APM traces with structured loggers.
+
+[[span-to-string]]
+==== `span.toString()`
+
+[small]#Added in: #2.17.0
+
+Produces a string representation of the span to inject in log messages.
```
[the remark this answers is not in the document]
Qard: fixed

docs/transaction-api.asciidoc
```diff
@@ -206,6 +206,22 @@ elasticApm.init({
 
 See the {apm-rum-ref}[JavaScript RUM agent documentation] for more information.
 
+[[transaction-ids]]
+==== `transaction.ids`
+
+[small]#Added in: #2.17.0
+
+Produces an object containing `transaction.id` and `trace.id`.
```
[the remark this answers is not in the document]

docs/transaction-api.asciidoc
```diff
@@ -206,6 +206,22 @@ elasticApm.init({
...
+This enables log correlation to APM traces with structured loggers.
+
+[[transaction-to-string]]
+==== `transaction.toString()`
+
+[small]#Added in: #2.17.0
+
+Produces a string representation of the transaction to inject in log messages.
```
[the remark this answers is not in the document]

lib/instrumentation/ids.js
```diff
@@ -0,0 +1,31 @@
+const { stringify } = require('querystring')
```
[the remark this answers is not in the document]
Qard: fixed

lib/instrumentation/ids.js
watson: That's for the output format when using the `toString()` version right? I was more concerned with those people who needed to access the values directly by their property name... or maybe I've misunderstood how they are supposed to be used/

Qard: I understood it as _all_ forms of the logging object should look like that. @axw What do you think?

axw: @watson did you have a use case in mind? `apm.currentTraceIds.spanId` looks a bit nicer, but I'm not sure when you would want to do that. For log correlation purposes, I think it'll be typical to just log "the trace IDs", whatever they may be - span, transaction, or nothing, depending on the context and sampling.

lib/instrumentation/ids.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const { stringify } = require('querystring')
+
+class Ids {
+  toString () {
+    return stringify(this, ',', '=')
```
axw: Can we change this to space-separated a la logfmt? While there may not be a convention in Node.js, I think this is generally more standard across languages.

lib/instrumentation/ids.js
watson: I was thinking that the user wanted to specify the keys in the structured logs them selfs. If so they needed to access these properties. But if they just want to use the keys we provide, then you're right, they don't need to access them explicitly.

watson: Talked about this in the weekly meeting. Let's go with the current implementation 👍

lib/instrumentation/ids.js
Qard: done. 👍

docs/agent-api.asciidoc
```diff
@@ -783,3 +783,37 @@ apm.clearPatches('timers')
...
+----
+
+All current trace id objects,
+including the empty form,
+include a `toString()` implementation.
+This enables log correlation to APM traces with text-only loggers.
+
+[source,js]
+----
+"trace.id=abc123,transaction.id=abc123"
```
axw: this and a few more instances in docs below need updating now

Qard: Oh, good catch. Should be fixed now. 👍

PR #1340, opened 2019-09-11
feat: change code to use url.URL instead of url.parse

### LINKED ISSUE #1254: Change code to use url.URL instead of url.parse

yuxizhe:
close #1254

### Checklist

- [x] Implement code

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -84,6 +84,7 @@ function formatURL (item) {
   return {
     href: item.href,
     pathname: item.pathname,
+    path: item.pathname + item.search,
     protocol: item.protocol,
     host: item.host,
     port: item.port,
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -86,7 +86,7 @@ function formatURL (item) {
   return {
     href: item.href,
     pathname: item.pathname,
-    path: item.pathname + item.search,
+    path: item.pathname + (item.search || ''),
     protocol: item.protocol,
     host: item.host,
     port: item.port,
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -4,6 +4,8 @@ var url = require('url')
 
 var endOfStream = require('end-of-stream')
 
+var parsers = require('../parsers')
+
 const transactionForResponse = new WeakMap()
 exports.transactionForResponse = transactionForResponse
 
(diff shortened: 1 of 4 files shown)


Qard:
jenkins run the tests please

Qard:
Marked as breaking change as the `URL` class does not exist yet in Node.js 6 so we have to finish dropping v6 first.

lib/instrumentation/http-shared.js
```diff
@@ -80,15 +80,27 @@ function isRequestBlacklisted (agent, req) {
...
+    hash: item.hash,
+    search: item.search
+  }
+}
 // NOTE: This will also stringify and parse URL instances
 // to a format which can be mixed into the options object.
 function ensureUrl (v) {
-  if (typeof v === 'string') {
+  if (url.URL && (v instanceof url.URL || typeof v === 'string')) {
+    const urlObj = new url.URL(v.toString(), 'relative:///')
```
watson: If `v` is already an instance of `url.URL`, there really isn't a reason to first stringify it and then make it back into a `url.URL` instance. So it would probably be more performant to split this if-block into two - eg:

```js
if (typeof v === 'string') {
  retrun formatURL(url.URL ? new url.URL(v, 'relative:///') : url.parse(v))
} else if (url.URL && v instanceof url.URL) {
  return formatURL(v)
} else {
  return v
}
```

yuxizhe:
@watson All fixed, Thanks for your advice 👍

Qard:
jenkins run the tests please

Qard:
jenkins run the tests please

Qard:
So with this using _both_ `url.Url` _and_ `url.parse(...)`, are we targeting landing this in 2.x as a _not_ a breaking change? If so, it appears to be failing in Node.js 6.x and we should figure out why. If we want to just do this as a _breaking_ change in 3.x, it should be fine as-is.

watson:
I think it's easiest to target 2.x now and then we can just remove the `url.parse` fallback once we're in 3.x. That way we get the best of both worlds with almost no effort

watson:
I'll just look into the failing Node.js 6 test...

yuxizhe:
> I'll just look into the failing Node.js 6 test...

Very strange. It passed travis-ci node 6 tests, but failed in apm-ci node 6.  I can't reproduce that error on my computer with node 6.x.

watson:
I've seen this specific test failure on Jenkins a few times in other PRs lately. So I think the Node.js 6 tests are just flaky when running on Jenkins. I'll go ahead and run the tests one more time to see if it goes away.

watson:
jenkins run the tests please

watson:
This time around the

watson:
Ups, I pressed the wrong button 🤦‍♂ 

Was about to write that this time around the Node.js 6 tests passed just fine 😅

watson:
I've created another issue with the Node.js 6 failures on Jenkins: #1350

watson:
Thanks a lot for helping us fix this one @yuxizhe 💯 It'll be part of the next release 😃

yuxizhe:
> Thanks a lot for helping us fix this one @yuxizhe 💯 It'll be part of the next release 😃

My pleasure : )

PR #1346, opened 2019-09-13
feat: add support for @koa/router

### LINKED ISSUE #1343: Add/update instrumentation for @koa/router

The `koa-router` module seems to have become unmaintained as of late, resulting in a fork being made and placed under the "official" `@koa` namespace. See [this ticket](https://github.com/ZijianHe/koa-router/issues/513) for details.

Due to how this module shims middleware, moving to `@koa/router` would cause breakage.

Please consider including `@koa/router` in the instrumentation `MODULES` set.
Given how modules are required in the loader, there is sadly no straightforward way to link multiple entries to the same source file; symlinking might be dangerous, not sure how that'd work out cross-platform?

Made a quick attempt at using `agent.addPatch(...)` as suggested in #878. Had to tweak the validation logic in `lib/instrumentation/modules/koa-router.js` as well, as the `(!semver.satisfies(version, '>=5.2.0 <8'))` rejects `@koa/router`, which is at major version 8.

Would not consider this a permanent solution, regardless.

**Edit:** More details about the workaround, fixed typo.

watson:
The `koa-router` module is abandoned and a fork is being maintained under `@koa/router` instead.

This fork is published as a major bump (v8.0.0) of the original, but our patch still works fine on that version.

Closes #1343

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


watson:
jenkins run the module tests for koa-router,@koa/router

watson:
@elastic/observablt-robots it looks like Jenkins isn't running the TAV tests for `@koa/router`. Am I doing something wrong in the comment I posted above or do you know what might cause this?

v1v:
It seems the readYAML parse failed

```
while scanning for the next token
found character '@' that cannot start any token. (Do not use @ for indentation)
 in 'reader', line 3, column 5:
      - @koa/router
        ^
```

https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-1346/3/pipeline#step-1028-log-4

Likely it's required to be single-quoted when the TAV is created on the fly:
- https://github.com/elastic/apm-agent-nodejs/blob/7998c1ffa87a7c7cd89a028d378de381db53e0d1/Jenkinsfile#L395

watson:
@v1v I would expect so as well, but it is single-quoted in the file, so it should just work right?

https://github.com/elastic/apm-agent-nodejs/pull/1346/files#diff-a4e3129a0bc14e3977da8ff722024fe4R2

v1v:
When triggering the TAV subset with a github comment the file is created on the fly rather than using the  .ci/.jenkins_tav.yml. So the entry seems fine in that particular file.

I'll create a patch to create the entries which contain the `@` with single quotes. Meanwhile, you might need to run all the TAV rather than using the TAV subset feature.

watson:
Ah of cause! Maybe if I write the comment like `koa,'@koa/router'` it will generate the file with quotes as well. I'll just try and see. If that doesn't work I'll just do all of them as you suggest 😃

watson:
jenkins run the module tests for koa-router,'@koa/router'

watson:
@v1v Yeah it worked by adding the quotes to the GitHub comment. But it would obviously be better if we didn't have to do this. A simpler fix could simply be to just put them all in quotes no matter if they contained a `@` or not.

v1v:
I'll raise that PR with the suggested approach to quote those entries created on the fly. Thanks for finding the way to bypass it. :)

PR #1352, opened 2019-09-16
ci(jenkins): don't run TAV tests for Node.js core modules

watson:
Previously if there was a change to any of our patches for a Node.js core modules (e.g. `http`), Jenkins would try to run TAV tests for the `http` module. This is just a waste of resources as there are no TAV tests for core modules.

This commit will filter out any core module from the list of modules being considered for TAV tests.

### DIFF
--- a/.ci/scripts/get_tav.sh
+++ b/.ci/scripts/get_tav.sh
@@ -9,6 +9,44 @@ OUTPUT=$1
 GIT_DIFF=git-diff.txt
 CHANGES=changes.txt
 
+## Node.js core modules are not on npm and doesn't need to be tested in the
+## same way as regular modules does
+NODE_CORE_MODULES=(
+  assert
+  async_hooks
+  child_process
+  cluster
+  crypto
+  dns
+  domain
+  events
+  fs
+  http
+  http2
+  https
+  inspector
+  net
+  os
+  path
+  perf_hooks
+  punycode
+  querystring
+  readline
+  repl
+  stream
+  string_decoder
+  tls
+  trace_events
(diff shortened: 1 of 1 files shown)


watson:
@v1v Does this somehow conflict with the change in #1351 because I'm also adding quotes here?

v1v:
They are independent:
- #1351 refers only when there is a GitHub comment and matches with the TAV comment format. https://github.com/elastic/apm-agent-nodejs/blob/14abfe7421a9f0b21def388f142f71c1b9323a6b/Jenkinsfile#L383
- While this PR is regarding what source code changes have been pushed onto the PR. https://github.com/elastic/apm-agent-nodejs/blob/14abfe7421a9f0b21def388f142f71c1b9323a6b/Jenkinsfile#L401

v1v:
So if someone might change https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/http.js then this particular script should skip running the TAV for the module `http`, right?

watson:
@v1v Correct. I discovered this in #1340 where it ran the `http2` tests: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-1340/12/pipeline

Qard:
I think this PR may have missed dealing with the `TAV_UPDATED` flag in the Jenkinsfile. It's resulting in failed attempts to load a `.ci/.jenkins_generated_tav.yml` file.

watson:
@Qard fixed in #1363

PR #1354, opened 2019-09-16
modified semver check to include v5 express

nw:
Simple change to `express` instrumentation patch to not exit for versions greater then `^4.0.0`

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -11,7 +11,7 @@ module.exports = function (express, agent, { version, enabled }) {
 
   agent.setFramework({ name: 'express', version, overwrite: false })
 
-  if (!semver.satisfies(version, '^4.0.0')) {
+  if (!semver.gte(version, '4.0.0')) {
     agent.logger.debug('express version %s not supported - aborting...', version)
     return express
   }


watson:
Version 5 of Express is not yet released, so I'm not sure if it's safe to include it like this... Do you use the alpha version?

nw:
Ah makes sense. I’ve played around with it trivially but not in prod. It does work.

I guess I got confused by intent with the following condition explicitly targeting v5.

watson:
If we were to include Express 5 in what's supported, we would need to update [the TAV tests](https://github.com/elastic/apm-agent-nodejs/blob/6f84e0ced1c7334fd27fa4d69228c7cbad378aae/.tav.yml#L80-L86) to also test that version (since they are pre-release versions, they are not included in the normal `>=4` range), so that we don't get any surprises as breaking changes are still possible since the current v5 is only a pre-release alpha version.

But I'm not sure if it's worth it as I personally don't think v5 is about to land any time soon - if at all. The latest v5 pre-release is over a year old, so it seems like progress has stopped.

Qard:
Closing as we've decided not to support v5 for now. If you feel you need it, let us know and we will reconsider. We just don't want to add extra maintenance burden for something users are unlikely to be using, but if there _is_ a user that changes things. :)

nw:
Agreed

PR #1355, opened 2019-09-17
chore(package): bump 'standard' linter to ^14.3.1

watson:
Also fix the linting errors related to the new linting rules in this new version.


Qard:
This appears to result in syntax errors on old versions of Node.js.

watson:
This is a bug in standard. I'll fix it

watson:
Hmm seems like standard v14.3.1 hasn't actually been released yet... investigating...

watson:
14.3.1 has been released now

PR #1365, opened 2019-09-19
test: allow mock APM Server to be initiated without args

watson:
It would throw an error when trying to access `requests[0]` later in the function if no `mockOpts` object was provided. This made it hard to debug certain things.


watson:
Ups, accidentally pressed the "update branch" button instead of the merge button. I went ahead and merged right after though

PR #1367, opened 2019-09-20
test: fix http disabling test

watson:
The PR #1298 introduced a bug in the tests as it expected the span.type for outgoing http requests to start with `ext.*`. This was changed to `external.*` in the PR #1291, but #1298 was just never updated to be up to date with master before it was merged.


watson:
Merging without a review to get `master` back in working order

Qard:
Good catch. 👍

PR #1373, opened 2019-09-23
test: ensure proper cleanup between config tests

watson:
This was all the code that came out of trying to solve a weird CI related issue in #1114.

I thought it was best to submit it as a separate PR in case we want to backport it.

### DIFF
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -3,9 +3,18 @@
 const asyncHooks = require('async_hooks')
 const shimmer = require('./shimmer')
 
+// FOR INTERNAL TESTING PURPOSES ONLY!
+const resettable = process.env._ELASTIC_APM_ASYNC_HOOKS_RESETTABLE === 'true'
+let _asyncHook
+
 module.exports = function (ins) {
   const asyncHook = asyncHooks.createHook({ init, before, destroy })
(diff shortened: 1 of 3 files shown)


lib/instrumentation/async-hooks.js
```diff
@@ -3,10 +3,19 @@
...
+let _asyncHook
+
 module.exports = function (ins) {
   const asyncHook = asyncHooks.createHook({ init, before, destroy })
   const contexts = new WeakMap()
 
+  if (resettable) {
+    if (_asyncHook) _asyncHook.disable()
+    _asyncHook = asyncHook
+  }
```
Qard: Is there a reason not to just _always_ do this? 🤔

watson: Good question... I did so in the first iteration but I wasn't 100% sure if there were any implications of keeping a reference to this object so it couldn't be garbage collected. This way it only kept it around during testing - but it's probably irellevant 🤷‍♂

watson:
jenkins run the tests please

lib/instrumentation/async-hooks.js
Qard: Fair. Approved anyway just to get the fix in. I can clean it up later. 🤔

watson:
jenkins run the tests please

PR #1374, opened 2019-09-23
ci(jenkins): Mark Node.js stage as UNSTABLE and build as SUCCESS on failures

kuisathaverat:
Avoid hiding error on the Node.js 6 stage, this change will mark the build as SUCCESS and the stage as UNSTABLE.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -342,7 +342,7 @@ def generateStep(Map params = [:]){
           retry(2){
             sleep randomNumber(min:10, max: 30)
             if (version?.startsWith('6')) {
-              catchError {
+              catchError(buildResult: 'SUCCESS', stageResult: 'UNSTABLE') {
                 sh(label: 'Run Tests', script: """.ci/scripts/test.sh "${version}" "${tav}" "${edge}" """)
               }
             } else {


v1v:
Aha!! For some reason, I thought it will catch the error and won't fail the build. It does the first step but not the second one.

image

Thanks for the fix. ❤️

watson:
When you say that the build will be marked as `SUCCESS`, does that mean that we'll not see it on GitHub?

kuisathaverat:
>When you say that the build will be marked as SUCCESS, does that mean that we'll not see it on GitHub?

On Jenkins/GitHub the build will be SUCCESS (green).

watson:
Eh so if Node.js 6 fails, it will pass anyway? Why was it that we're giving Node.js 6 special treatment like this in the first place? Shouldn't it just break the build?

kuisathaverat:
>Eh so if Node.js 6 fails, it will pass anyway? Why was it that we're giving Node.js 6 special treatment like this in the first place? Shouldn't it just break the build?

at some point was requested to ignore Node.js 6 failures, we can mark the build to whatever result we want for me it is a FAILURE build and the catcherror block is something we do not need

watson:
I don't remember us wanting to ignore Node.js 6 failures. If the tests are flaky we should probably just look into why and fix it. This wasn't a problem on Travis, so I assume anything that causing the tests to fail on Jenkins is just a race condition and is therefore fixable as it's not related to Node.js itself.

Qard:
I think the ignoring was a misunderstanding. There was ci-related PR a few weeks ago that was failing on Node.js 6 and I said to just ignore the failure for now as we'd be dropping 6 soon anyway. This was before we added a bunch of other breaking changes to the list for the major bump. At the time it seemed like the PR could just land in 3.x and not worry about getting it working for 2.x.

v1v:
That particular `catchError` was added a couple of months ago:
- https://github.com/elastic/apm-agent-nodejs/pull/1231#issuecomment-514514260

Anyway, I guess this PR is superseded with https://github.com/elastic/apm-agent-nodejs/pull/1383

watson:
We will still need to fix this issue in the `2.x` branch

watson:
Closing in favor of #1402

PR #1383, opened 2019-09-24
chore: drop support for older Node.js versions

watson:
Node.js versions older than 8.6.0 + the 11.x branch are no longer supported.

~~Depends on #1387~~
Closes #1014

PR #1384, opened 2019-09-24
fix(package): bump dependencies

watson:
~~Depends on #1383~~

Upgrade dependencies that couldn't be upgraded previously because they required Node.js 8+.


Qard:
If this passes after fixing #1383 and rebasing, it should be fine to merge. 👍

PR #1391, opened 2019-09-25
ci(jenkins): 2.x branch on a daily basis

v1v:
## Highlights
- Enable the pipeline for the `2.x` branch on a daily basis.
- To clarify the benchmark stage only runs when a new commit is merged into the `master` branch.


watson:
Does this need to be backported into the `2.x` branch as well?

PR #1400, opened 2019-09-26
ci(jenkins): time out problematic test if running too long

watson:
This test has a tendency to get stuck on Jenkins and never complete. Instead of waiting 3 hours until Jenkins times out, we'll just time out the test after 10 seconds.

Obviously this doesn't fix the issue but merely stops us from consuming Jenkins resources.


watson:
jenkins run the tests please

PR #1402, opened 2019-09-26
ci(jenkins): don't suppress Node.js 6 errors

watson:
This is a replacement for #1374 to not suppress any errors on Node.js 6 in the `2.x` branch.


watson:
@v1v it's fine that it's not running the next pipelines... that will just put extra pressure on us to fix the test that is broken. It's better than accidentally ignoring a test that should not be ignored

PR #1403, opened 2019-09-26
[2.x] ci(jenkins): time out problematic test if running too long (#1400)

watson:
 - ci(jenkins): time out problematic test if running too long (#1400)


watson:
@Qard Weird that Jenkins didn't pick up on the long commit message (we need to look into that)... though this type of linting will get annoying quickly. We should probably discuss how to handle this

Qard:
The linting CI job only runs the code lint, not the commit lint. We should probably have another job for that.

watson:
Weird, we added it as a separate stage in #1172, but it must have been lost after that

watson:
Ah, found it! It was removed in #1325 - I'll make an issue to get it readded

watson:
For now I'll just modify the commit message to get this merged

PR #1409, opened 2019-09-27
ci(jenkins): end test properly even if it times out

watson:
This was missed in #1400


Qard:
jenkins run the tests please

PR #1410, opened 2019-09-27
[2.x] ci(jenkins): end test properly even if it times out (#1409)

watson:
 - ci(jenkins): end test properly even if it times out (#1409)


Qard:
AppVeyor failure is unrelated, merging anyway.

PR #1415, opened 2019-09-29
docs: update Node.js release schedule graphic

watson:
The original Node.js release schedule image from nodejs.org is an SVG. Previously we had to convert it to a PNG because of our docs page on elastic.co couldn't serve SVG files. This has now been fixed in https://github.com/elastic/docs/pull/1200, so this PR updates to use an SVG and updates the graphics as well to show newer versions of Node.js.

The new graphics looks like this:

image


Qard:
Now that svg is supported, can we just link to the file directly from the Node.js repo now? As I recall, the only reason we had our own png copy in the first place was just because our doc building didn't work with svgs. 🤔

watson:
Fine with me. Can you think of any issues with that approach @bmorelli25?

bmorelli25:
The downside is that we wouldn't know if the URL to the image changes. This could result in a broken image in the docs until we discovered it. If that's a risk you're willing to take, that approach is fine with me.

PR #1425, opened 2019-10-02
ci(jenkins): reuse top level agent

v1v:
### Highlights
- Let's reuse the top-level agent to reduce the number of fo requested agents and speed up just a bit the builds.
- Reset labels to be `linux && immutable` to help with what kind of OS is required to be used.
- Time improvements per Stage
  - From the very beginning to the tests stage used to take ~3 minutes and now 1.30 minutes.
  - TAV stage doesn't require a new worker, so roughly another 1minute approx of improvement.

### Test Cases
- TAV has been launched explicitly with the comment: https://github.com/elastic/apm-agent-nodejs/pull/1425#issuecomment-538056979

### DIFF
--- a/.ci/downstreamTests.groovy
+++ b/.ci/downstreamTests.groovy
@@ -11,7 +11,7 @@ it is need as field to store the results of the tests.
 @Field def nodeTasksGen
 
 pipeline {
-  agent { label 'immutable' }
+  agent { label 'linux && immutable' }
   environment {
     REPO = 'apm-agent-nodejs'
     BASE_DIR="src/github.com/elastic/${REPO}"
--- a/.ci/downstreamTests.groovy
+++ b/.ci/downstreamTests.groovy
@@ -11,7 +11,7 @@ it is need as field to store the results of the tests.
 @Field def nodeTasksGen
 
 pipeline {
-  agent any
+  agent { label 'immutable' }
   environment {
     REPO = 'apm-agent-nodejs'
     BASE_DIR="src/github.com/elastic/${REPO}"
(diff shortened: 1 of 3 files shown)


Qard:
Not really clear on how the agents thing works, but is the `docker` part not needed anymore? 🤔

v1v:
That's a good point. For simplicity, we do use labels when requesting a Worker, some of those labels are not required anymore, such as docker. The reason is that the current Linux workers got the below labels:
- docker
- immutable
- linux
- swarm
- ubuntu
- ubuntu-16.04
- virtualx86_64

And Windows workers got:
- swarm
- windows-2019-immutable

So docker is installed by default, just using `immutable` is sufficient to run the build in a linux worker with docker and some other tools installed by default.

v1v:
Regarding how the workers are provisioned. Our current CI infra supports ephemeral agents/workers. Therefore if a pipeline contains a bunch of stages, the worker can be defined at very top level and being reused between stages, or explicitly per stage.
Each worker can be used once, so minimizing the number of workers and reusing as much as possible will imply a few fewer minutes for the build to wait for resources. 
I have not analysed yet, how faster the builds would be, but for another project, with a few less number of stages, the improvement was about 5 minutes.
Happy to provide further details if it helps :)

Qard:
Sounds good to me. Is there still more work to do on this? It's still labelled with `in progress`.

v1v:
I'd like to monitor the improvement and also I'd include the label `linux && immutable` to add more semantic about the expected worker, it might help to understand what kind of worker is used by.

I'll change the status of this PR as soon as I solve the above two topics.

Thanks

v1v:
jenkins run the module tests for http

Issue #1427, opened 2019-10-02
[meta][docs] Documentation restructure


bmorelli25:
A meta issue for tracking documentation improvements outlined in https://github.com/elastic/apm-dev/issues/425.

**https://github.com/elastic/apm-agent-nodejs/pull/1435**
- [x] Add release notes to the documentation: https://github.com/elastic/apm/issues/132

**https://github.com/elastic/apm-agent-nodejs/pull/1428**
- [x] Top level heading and doc sync for log correlation: https://github.com/elastic/apm/issues/149
- [x] Use shared version attributes: https://github.com/elastic/apm/issues/150
- [x] Add a section on how the Agent does instrumentation: https://github.com/elastic/apm/issues/146
- [x] Rename the APM UI to "APM app": https://github.com/elastic/apm-dev/issues/567
- [x] Updates to the new documentation structure outlined [here](https://docs.google.com/spreadsheets/d/1P53LbNY7jFSLLJhOKw-nmDenQbrIFlY4EEFJGqT1kmY/edit#gid=0)
- [x] Closes https://github.com/elastic/apm-agent-nodejs/issues/1023.
- [x] Sync Central config docs: https://github.com/elastic/apm-agent-dotnet/pull/497#discussion_r326303325

bmorelli25:
Thanks @Qard for your help and patience. One Agent done!

PR #1428, opened 2019-10-02
[docs] Documentation restructure

### LINKED ISSUE #1023: [docs] Fix duplication on getting started pages and api calls
There's a fair amount of duplication in the documentation that I'd like to explore removing. Sections like this are duplicated across all getting started pages:

https://github.com/elastic/apm-agent-nodejs/blob/9e3696cb25b03b2c68a7c4b7ab9dedc54309bdd3/docs/express.asciidoc#L171-L184

Sections like this are duplicated across similar api calls. Although, I'm not sure if this particular example makes much sense to change:
https://github.com/elastic/apm-agent-nodejs/blob/9e3696cb25b03b2c68a7c4b7ab9dedc54309bdd3/docs/transaction-api.asciidoc#L100-L105

There are ways we can maintain this identical information in a single location and include it where it is necessary. This will prevent having to update the same information in multiple places as witnessed in https://github.com/elastic/apm-agent-nodejs/pull/1019

bmorelli25:
WIP. For https://github.com/elastic/apm-agent-nodejs/issues/1427.

This PR accomplishes the following:
* Use shared version attributes
* Rename the APM UI to "APM app"
* Updates to the new documentation structure
* Adds a top level heading for log correlation
* Adds a section on how the Agent does instrumentation
* Closes #1023 (duplication in getting started pages).

New layout (only one link changes)
Screen Shot 2019-10-02 at 1 40 36 PM

### DIFF
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -7,8 +7,6 @@ endif::[]
 
 == Introduction
 
-Welcome to the APM Node.js Agent documentation.
-
 The Elastic APM Node.js Agent sends performance metrics and errors to the APM Server.
 It has built-in support for the most popular frameworks and routers,
 as well as a simple API which allows you to instrument any application.
@@ -21,25 +19,3 @@ APM Agents work in conjunction with the {apm-server-ref-v}/index.html[APM Server
 {ref}/index.html[Elasticsearch],
 and {kibana-ref}/index.html[Kibana].
 Please view the {apm-overview-ref-v}/index.html[APM Overview] for details on how these components work together. 
-
-[float]
-[[get-started]]
-=== Get started
-
-To get you off the ground, we've prepared guides for the most popular web frameworks:
-
-* <<express,Get started with Express>>
-* <<hapi,Get started with hapi>>
-* <<koa,Get started with Koa>>
-* <<restify,Get started with Restify>>
-* <<fastify,Get started with Fastify>>
-
-Alternatively, you should check out our guide for <<custom-stack,getting started with a custom Node.js stack>>.
-To get an overview of which components of your application we instrument automatically,
-use the <<supported-technologies,Supported Technologies Overview>> page.
-
-Other useful documentation includes:
-
-- <<advanced-setup,Advanced Setup and Configuration>>
-- <<api,API Reference>>
-- <<troubleshooting,Troubleshooting>>
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -19,7 +19,7 @@ The agent auto-instruments <<compatibility-frameworks,supported frameworks>> and
 like HTTP requests and database queries. To do this, it ??
 This means that for the supported technologies, there are no code changes required.
 
-The Agent automatically ... to measure their duration and metadata (like the DB statement),
+The Agent automatically links module function calls to callback calls to measure their duration and metadata (like the DB statement),
 as well as HTTP related information (like the URL, parameters, and headers).
 
 These events, called Transactions and Spans, are sent to the APM Server.
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -16,7 +16,7 @@ as well as a simple API which allows you to instrument any application.
 === How does the Agent work?
 
 The agent auto-instruments <<compatibility-frameworks,supported frameworks>> and records interesting events,
-like HTTP requests and database queries. To do this, it ??
+like HTTP requests and database queries. To do this, it patches modules as they are loaded to capture when module functions and callbacks are called. Additionally, there are some cases where a module will be patched to allow tracing context to be propagated through the asynchronous continuation.
 This means that for the supported technologies, there are no code changes required.
 
 The Agent automatically links module function calls to callback calls to measure their duration and metadata (like the DB statement),
--- a/docs/intro.asciidoc
+++ b/docs/intro.asciidoc
@@ -12,6 +12,21 @@ It has built-in support for the most popular frameworks and routers,
 as well as a simple API which allows you to instrument any application.
 
 [float]
+[[how-it-works]]
+=== How does the Agent work?
+
+The agent auto-instruments <<compatibility-frameworks,supported frameworks>> and records interesting events,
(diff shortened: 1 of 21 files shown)


docs/intro.asciidoc
```diff
@@ -7,12 +7,25 @@ endif::[]
...
+The agent auto-instruments <<compatibility-frameworks,supported frameworks>> and records interesting events,
+like HTTP requests and database queries. To do this, it ??
+This means that for the supported technologies, there are no code changes required.
+
+The Agent automatically ... to measure their duration and metadata (like the DB statement),
+as well as HTTP related information (like the URL, parameters, and headers).
+
+These events, called Transactions and Spans, are sent to the APM Server.
+The APM Server converts them to a format suitable for Elasticsearch, and sends them to an Elasticsearch cluster.
+You can then use the APM app in Kibana to gain insight into latency issues and error culprits within your application.
```
bmorelli25: @Qard - Going to need your help filling in the blanks here. Doesn't need to be super implementation heavy, just a general idea of how the Agent works.

docs/intro.asciidoc
```diff
@@ -7,12 +7,25 @@ endif::[]
...
 The Elastic APM Node.js Agent sends performance metrics and errors to the APM Server.
 It has built-in support for the most popular frameworks and routers,
 as well as a simple API which allows you to instrument any application.
 
+[float]
+[[how-it-works]]
+=== How does the Agent work?
+
+The agent auto-instruments <<compatibility-frameworks,supported frameworks>> and records interesting events,
+like HTTP requests and database queries. To do this, it ??
```
Qard:
```suggestion
like HTTP requests and database queries. To do this, it patches modules as they are loaded to capture when module functions and callbacks are called. Additionally, there are some cases where a module will be patched to allow tracing context to be propagated through the asynchronous continuation.
```

docs/log-correlation.asciidoc
```diff
@@ -0,0 +1,22 @@
...
+:leveloffset: -1
+include::agent-api.asciidoc[tag=currentTraceIds]
+:leveloffset: +1
+
+Using your favorite logging framework, you'd then need to inject this information into your logs.
+
+When your logs contain the appropriate identifiers, the final step is to ingest them into the same
+Elasticsearch instance that contains your APM data. See
+{apm-overview-ref-v}/observability-integrations.html#_ingest_your_logs_into_elasticsearch[Ingest your logs into Elasticsearch]
+for more information.
```
bmorelli25: This page uses *tagged regions* to pull the `apm.currentTraceIds` information in from the `agent-api` file. Since you can't see how it looks here, here's how it renders. I'd like some feedback on this:
Screen Shot 2019-10-02 at 2 56 05 PM

docs/intro.asciidoc
```diff
@@ -7,12 +7,25 @@ endif::[]
...
+This means that for the supported technologies, there are no code changes required.
+
+The Agent automatically ... to measure their duration and metadata (like the DB statement),
```
Qard:
```suggestion
The Agent automatically links module function calls to callback calls to measure their duration and metadata (like the DB statement),
```

bmorelli25:
Fantastic, thank you! Still some work to do, so I probably won't open this up for a full review until tomorrow. I'll let you know!

docs/shared-set-up.asciidoc
```diff
@@ -0,0 +1,148 @@
+// Content in this file is shared across the Node.js documentation.
```
bmorelli25: @Qard -- This is for https://github.com/elastic/apm-agent-nodejs/issues/1023. All six of the getting started pages share a ton of content. The idea here is to move the shared content into one file and then include it in each of the getting started files. This adds a bit of complexity to the docs, but also makes it easier to update them if things change.

Curious to hear your thoughts on this approach. If you think it makes it too complicated, we can chat about alternative solutions.

Qard: Seems straightforward enough. Anything that makes it _easier_ to maintain docs while keeping a high _quality_ standard is a good thing. 😸

bmorelli25: Great!

bmorelli25:
It looks like it doesn't like the commit naming from the GitHub suggestions: `Update docs/intro.asciidoc`. Do I need to rename those?

Qard:
Yes, all commits need to follow the `type(subsystem): message` or `type: message` forms. I'd just squash the commits, but we can also do that when landing, assuming everything else passes.

bmorelli25:
Finally got this green @Qard

docs/supported-technologies.asciidoc
```diff
@@ -37,15 +37,16 @@ These are the frameworks that we officially support:
...
+|Framework |Version
+|<<express,Express>> |^4.0.0
+|<<hapi,hapi>> |>=9.0.0 <19.0.0
+|<<hapi,@hapi/hapi>> |>=17.9.0 <19.0.0
+|<<koa,Koa>> via koa-router or @koa/router |>=5.2.0 <9.0.0
+2+|Koa doesn't have a built in router, so we can't support Koa directly since we rely on
+router information for full support. We currently support the most popular Koa router called
+https://github.com/alexmingoia/koa-router[koa-router].
+|<<restify,Restify>> |>=5.2.0
+|<<fastify,Fastify>> |>=0.27.0
```
Qard: What does this change look like? It seems to be switching from three columns to two but koa remains with three columns. Does this render properly?

bmorelli25: Check it out here: http://apm-agent-nodejs_1428.docs-preview.app.elstc.co/guide/en/apm/agent/nodejs/master/supported-technologies.html#compatibility-frameworks

I think it makes the table easier to read. Thoughts?

bmorelli25:
I agree with backporting these changes. I expect a few collisions, but don't see any major problems getting this backported.

docs/supported-technologies.asciidoc
Qard: Cool, seems reasonable to me. 👍

PR #1429, opened 2019-10-03
fix: write stack trace of uncaught exceptions to STDERR

### LINKED ISSUE #1141: Log stack trace for uncaught exceptions to STDERR
Today the stack trace for an uncaught exception isn't logged to STDERR when it's being sent to the APM Server. This makes it impossible to see what's wrong during development without setting `captureExceptions: false` temporarily.

watson:
The stack trace of uncaught exceptions is normally written to STDERR. The `uncaughtException` listener inhibits this behavior, and it's therefore necessary to manually do this to not break expectations.

I ran into this issue in Kibana again today and it became too annoying not seeing the stack trace. The wonders of dogfooding 😉 

Fixes #1141

### Checklist

- [x] Implement code
- [x] Add tests


watson:
I've added a config option: `logUncaughtExceptions`

Issue #1392, opened 2019-09-25
Support disabling logger or default to noop logger


jsumners:
My organization has its own logging format (mostly similar to Pino's output). But I cannot boot our logger before this library due to issues. We also have log shippers and destinations that expect that specific format to be adhered to. So to prevent this module from writing incompatible logs I have to:

```js
const apm = require('elastic-apm-node').start({
  logger: Object.create(require('abstract-logging'))
});
```

At a very minimum, I'd like to see `logger: false` as a supported option to disable the logger. A more complete solution would be to use `abstract-logging` yourself internally and you'll automatically gain compatibility with the list of loggers in https://github.com/elastic/apm-agent-nodejs/issues/1187#issuecomment-514103468. Note: this is what we do in Fastify -- https://github.com/fastify/fastify/blob/0e39b83c3bde8a9cb6531ad2dbe960be75199e1d/lib/logger.js#L84-L88

Qard:
The equivalent of `logger: false` would be `logLevel: 'silent'`. The `logger` option should only be used when you _want_ to receive the logs.

As for supporting `abstract-logging`, that could be an interesting possibility, but it sounds like it might be unnecessary for your particular needs right now. 🤔

jsumners:
“Silent” is not documented as a possibility — https://github.com/elastic/apm-agent-nodejs/blob/master/docs/configuration.asciidoc#loglevel

Using `abstract-logging` internally to the module would solve this for me because I could ignore customizing the logger and let the noop logger do its thing.

Qard:
Ah, right..."silent" was just a test thing. Try "fatal", there shouldn't be anything on that log level so it should be effectively silent.

jsumners:
Yes, that does seem to be a way to get it solved for now. But it isn't a foolproof solution. There is no guarantee that `logger.fatal` will not be used in the future. It would really be nice to have the ability to disable the logger completely by `logger: false`.

Qard:
We don't _intend_ on it ever logging anything by default, so that should not change.

Qard:
It basically _is_ a guarantee. _We_ control all the code where the logger is used, and we _do not_ intend to ever log _anything_ by default as there are many users that _require_ process output to have a specific format with no extra output mixed in.

[...]

PR #1430, opened 2019-10-03
refactor: use pino as default logger

### LINKED ISSUE #335: Update debug log output to not add line breaks
When logging objects with multiple properties, the current logger will split the objects over multiple lines. This is often the case when the log level is set to `debug`. The old debug mode didn't do this.

Example of line split:

image

watson:
This fixes a long-standing issue where log output would contain line-breaks. This made automated parsing harder. Instead, all log output is now JSON formatted, so each log message is only one line.

To format the log output for human consumption in the terminal, use `pino-pretty`: https://github.com/pinojs/pino-pretty

This also adds a new `logLevel` config option value called `silent`.

As opposed to the old default logger, pino will throw if given an invalid log level.

Fixes elastic/apm-agent-nodejs#335


watson:
Good point.

Do you think it makes sense to include the `pid` and `hostname` like in the example above? (it's added by default by pino). It's possible to tell pino to change what's added to the structured logs by default.

A gotcha related to the default fields that we need to be aware of: Pino will for performance reasons not run the output through a JSON serializer, but will instead manually write it to STDOUT. This means that the keys are not validated. So it's possible to have two keys by the same name if one of the keys is contained in the default fields - e.g. `name`. So we need to make sure that we don't also add a key when calling the logger with a log message with the same name.

I like having the name of the package that generated the log line somewhere on the log line (which is what the `name` property is normally used for in pino. But we do use `name` in a few places as well. So we either have the change the default `name` to something else (e.g. `logger`) or we need to change the use of `name` everywhere else.

Qard:
If there's a way to control it, I'd vote for being explicit about it. Pino could change the set of defaults in the future to include more stuff we don't want/need.

I don't think stuff like pid is important given that each process has its own output stream and likely would be piped to its own unique file or wherever else--all messages on the stream would have the same pid, which is just a waste of space. 🤷‍♂ 

I like having the name of the package responsible for the message. Might be cool for each instrumentation file to create it's own child logger too, so it can report the name of the module it's coming from also. Could be very helpful for debugging agent issues if we can see the specific file a message came from in the log output.

watson:
> Pino could change the set of defaults in the future to include more stuff we don't want/need.

That would be a breaking change IMO and I'm pretty sure @mcollina would agree, so I'm not too worried about that.

> I don't think stuff like pid is important given that each process has its own output stream and likely would be piped to its own unique file or wherever else

I agree - I'll remove it 👍 I assume the same goes for the hostname

> Might be cool for each instrumentation file to create it's own child logger too, so it can report the name of the module it's coming from also.

I'll look into it

jsumners:
Please see issue #1392. It would solve this problem by providing the apm module with the standard log4j style logger interface and allow users of the module to fill in their own logger. I do suggest promoting Pino as the primary candidate for that user logger instance, though.

trentm:
This is obsoleted by #2001

PR #1439, opened 2019-10-08
fix: always end transaction when socket is closed prematurely

### LINKED ISSUE #1411: Transactions not always ended if client closes socket prematurely
[The first test](https://github.com/elastic/apm-agent-nodejs/blob/3e360ce82563ea35fb3f691d32c1d25e11674236/test/instrumentation/modules/http/aborted-requests-enabled.js#L15-L58) in `test/instrumentation/modules/http/aborted-requests-enabled.js` randomly hangs and never completes on Jenkins. It happens every 5 times or so we run a test on Jenkins and it happens on multiple different versions of Node.js.

This might be related to elastic/apm-agent-nodejs#1350

watson:
This fixes a race condition that in some cases meant that a request transaction was never ended if the underlying TCP socket was closed by the client and the server didn't discover this in time.

Fixes #1411

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -35,9 +35,10 @@ exports.instrumentRequest = function (agent, moduleName) {
           ins.bindEmitter(res)
 
           endOfStream(res, function (err) {
+            if (trans.ended) return
             if (!err) return trans.end()
 
-            if (agent._conf.errorOnAbortedRequests && !trans.ended) {
+            if (agent._conf.errorOnAbortedRequests) {
               var duration = trans._timer.elapsed()
               if (duration > (agent._conf.abortedErrorThreshold * 1000)) {
                 agent.captureError('Socket closed with active HTTP request (>' + agent._conf.abortedErrorThreshold + ' sec)', {
@@ -49,9 +50,12 @@ exports.instrumentRequest = function (agent, moduleName) {
 
             // Handle case where res.end is called after an error occurred on the
             // stream (e.g. if the underlying socket was prematurely closed)
-            res.on('prefinish', function () {
+            const end = res.end
+            res.end = function () {
+              const result = end.apply(this, arguments)
               trans.end()
-            })
+              return result
+            }
           })
         }
       }
--- a/test/instrumentation/modules/http/aborted-requests-enabled.js
+++ b/test/instrumentation/modules/http/aborted-requests-enabled.js
@@ -37,12 +37,13 @@ test('client-side abort below error threshold - call end', { timeout: 10000 }, f
   var server = http.createServer(function (req, res) {
     setTimeout(function () {
       clientReq.abort()
-      setTimeout(function () {
-        res.write('Hello') // server emits clientError if written in same tick as abort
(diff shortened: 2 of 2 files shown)


watson:
This PR doesn't attempt to mark the transaction with a different result in case the underlying socket is closed prematurely. I still think it would be a good feature to somehow "mark" the transaction in some way - whether by setting a different `result` or by other means. But that doesn't have to be part of this PR. If not, we just need to create another issue to track it.

lib/instrumentation/http-shared.js
```diff
@@ -49,9 +50,12 @@ exports.instrumentRequest = function (agent, moduleName) {
 
             // Handle case where res.end is called after an error occurred on the
             // stream (e.g. if the underlying socket was prematurely closed)
-            res.on('prefinish', function () {
+            const end = res.end
+            res.end = function () {
+              const result = end.apply(this, arguments)
               trans.end()
```
axw: @watson doesn't this assume that `prefinish` will be emitted synchronously? Did I misunderstand your comment https://github.com/elastic/apm-agent-nodejs/issues/1411#issuecomment-538927796?

> I guess we could instead listen for when someone calls response.end() and then end the transaction then. That would get around the stream related issues, but that might then wrongly end a transaction before the actual HTTP response is ended as any buffering happening internally in the response object after calling response.end() would be ignored (which is why we made our current implementation the way we did).

Won't this change end the transaction after buffering the data, but before the data is actually sent? That's why I thought we would keep listening for `prefinish`, and only end the transaction in `res.end` if `req.aborted === true`.

watson: No, this PR only uses the call to `response.end()` as a signal to end the transaction in case the underlying socket was closed prematurely - so as you said, any buffered data would be lost anyway. Under normal circumstances, we wait until the stream closes and then end the transaction. This is determined based on whether or not [the callback](https://github.com/elastic/apm-agent-nodejs/blob/37524b1a25d17c8cccc85a77c955cd29350206a4/lib/instrumentation/http-shared.js#L37) to `endOfStream` is called with an error or not.

Previously we only listened for the `prefinish` event in case we detected that the underlying socket closed prematurely. But the problem was that the `prefinish` event wasn't always emitted (if the race condition was triggered).

This PR instead relies on [`response.end()`](https://github.com/nodejs/node/blob/064e111515185529b6f943dc66929440557fd609/lib/_http_outgoing.js#L698-L766) being called in the cases where the stream encountered an error (like the socket being closed prematurely). Normally the `prefinish` event is emitted [right before](https://github.com/nodejs/node/blob/064e111515185529b6f943dc66929440557fd609/lib/_http_outgoing.js#L762) the `end` function runs to completion, so this shouldn't really change that behavior, except that we now always end the transaction, instead of only if the `prefinish` event was emitted.

axw: > This is determined based on whether the callback to endOfStream is called with an error or not.

Got it - I missed this fact. So if and only if the stream ends due to an error, then we end the transaction on `res.end`. Thanks for the explanation!

PR #1445, opened 2019-10-09
[2.x] fix: always end transaction when socket is closed prematurely (#1439)

watson:
 - fix: always end transaction when socket is closed prematurely (#1439)

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -33,9 +33,10 @@ exports.instrumentRequest = function (agent, moduleName) {
           ins.bindEmitter(res)
 
           endOfStream(res, function (err) {
+            if (trans.ended) return
             if (!err) return trans.end()
 
-            if (agent._conf.errorOnAbortedRequests && !trans.ended) {
+            if (agent._conf.errorOnAbortedRequests) {
               var duration = trans._timer.elapsed()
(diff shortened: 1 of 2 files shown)


Qard:
Failing on commit message length, apparently. Not sure why the original PR didn't catch that. 🤔

Qard:
Oh. because of the extra PR reference at the end. 😅

Qard:
How should we deal with that? We already leave space for the PR reference at the end of the original commit, but it doesn't take into account the doubling from backporting. 🤔

watson:
Yeah, it's a bit unfortunate. I think it would be best, in the long run, to move the PR id to the body just like we do in Node core. In the past, I've just resorted to manually editing the commit merge when squash-merging

Qard:
Maybe we could add something to the `backport` tool for filtering commit messages somehow? The commit reference is a standard format on GitHub, so it makes sense to have a config for `backport` to detect and remove it. The reference to the origin PR already exists in the backport PR anyway. 🤔

watson:
Could be (ping @sqren). But for now, I think it's ok to just manually edit the commit merge message 👍

sqren:
There are options to set the patterns for PR title and PR description but it's pretty simplistic. Currently no way to format the commit message automatically.

https://github.com/sqren/backport/blob/master/docs/configuration.md#prtitle

PR #1451, opened 2019-10-11
ci(jenkins): fix wrong label

v1v:
## Highlights
- Fix leftover from https://github.com/elastic/apm-agent-nodejs/pull/1425


watson:
Shouldn't this be backported to the `2.x` branch as well?

kuisathaverat:
2.x does not have `Test RC` stage so it does not have the bug.

PR #1463, opened 2019-10-23
feat(http,http2): add http context info

### LINKED ISSUE #1462: More info in HTTP spans
It would be nice to have more information in HTTP spans for outgoing requests, such as:
- parsed URL
- HTTP method
- response code

https://github.com/elastic/apm-server/pull/1438

Possibly related to https://github.com/elastic/apm-agent-nodejs/issues/1060

Qard:
Fixes #1462

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -3,6 +3,7 @@
 var url = require('url')
 
 var endOfStream = require('end-of-stream')
+var httpRequestToUrl = require('http-request-to-url')
 
 var { parseUrl } = require('../parsers')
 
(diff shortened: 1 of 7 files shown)


lib/instrumentation/http-shared.js
```diff
@@ -189,7 +190,16 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
 
         function onEnd () {
           agent.logger.debug('intercepted http.IncomingMessage end event %o', { id: id })
-          span.end()
+
+          httpRequestToUrl(req).then(url => {
```
watson: Isn't there a risk that this promise will never resolve in case `req.socket` is never populated for some reason?

Qard: Yep, there was also some test issues due to the timing, so I changed it. The url is fetched out-of-band now, so it'd just be `undefined` if the socket is never populated.

Qard:
I wasn't intending on `span.setHttpContext(...)` being a public API, so it shouldn't need docs/typings. The existing `span.setDbContext(...)` is like that already--it's not documented or included in the typings.

PR #1464, opened 2019-10-23
ci: add new Node.js versions to CI

watson:
Node.js 13 was just released yesterday which means that it will now be part of the regular releases and Node.js 14 will now be built on a nightly basis.

The latest LTS version also just changed from 10 to 12.


watson:
Odd that Jenkins is failing on not being able to locate `node:13`. As far as I can see on https://hub.docker.com/_/node/ it should be there

watson:

watson:
Hmm the Docker errors on Jenkins seem to happen at random and affects other Node.js versions as well, e.g:

```
Error response from daemon: Get https://registry-1.docker.io/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)
```

@v1v is this as known issue?

watson:
Jenkins, run the tests please

watson:
@kuisathaverat Just realized that Victor is on holiday. Adding you as a reviewer instead - I hope you don't mind

watson:

test/instrumentation/modules/tedious.js
```diff
@@ -33,10 +33,7 @@ if (semver.gte(version, '4.0.0')) {
       },
       options: {
         database: 'master',
-        encrypt: false,
-        cryptoCredentialsDetails: {
-          ciphers: 'RC4-MD5'
-        }
+        encrypt: false
```
Qard: Is this not needed anymore? 🤔

watson: `RC4-MD5` fails on Node.js 12 because it's not supported anymore. I couldn't figure out what else to use, so I tried to remove it and it worked just fine. That's one of the reasons why I wanted to run the TAV tests so we could see if this would also work in older versions of tedious.

watson: Looks like all TAV tests besides tedious pass.. but the tedious failures seem odd. I'll just try to re-run them in case they were one-offs.

watson:
jenkins run the module tests for tedious

test/instrumentation/modules/tedious.js
watson: 🤦‍♂ obviously the tedious TAV tests are unrelated to this change as the change only affects AppVeyor. Don't know what I was thinking there... so the TAV tests must be unrelated

watson:
jenkins run the tests please

watson:

kuisathaverat:
it, is a timeout to connect to Docker Hub, so `docker pull node:8.6` fails because of those timeouts we use to add the Docker images to the packer cache, I will add node:8.6 too

Qard:
jenkins run the tests please

Qard:

PR #1471, opened 2019-10-25
ci(jenkins): move Jenkinsfile inside .ci folder

kuisathaverat:
this PR would move the Jenkinsfile to the folder .ci, to have all related CI files inside that folder


watson:
Shouldn't this be backported to the `2.x` branch as well?

kuisathaverat:
yep, we should if we continue building 2.x from the CI

PR #1473, opened 2019-10-25
fix(tedious): support 6.5+

### LINKED ISSUE #1472: Tedious test timeout
The latest version of tedious is causing timeouts in the test runs.

Qard:
In 6.5, tedious began using a version of TypeScript which locks property descriptors of exports. This prevents us from being able to re-assign functions with shimmer.

Fixes #1472

### Checklist

- [x] Implement code

### DIFF
--- a/lib/instrumentation/modules/tedious.js
+++ b/lib/instrumentation/modules/tedious.js
@@ -3,8 +3,6 @@
 var semver = require('semver')
 var sqlSummary = require('sql-summary')
 
-var shimmer = require('../shimmer')
-
 module.exports = function (tedious, agent, { version, enabled }) {
   if (!enabled) return tedious
   if (!semver.satisfies(version, '>=0.0.5')) {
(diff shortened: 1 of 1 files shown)


watson:
We had the same problem with `graphql`. To ensure that the API we exported looked identical to the user we implemented the patch in a bit of special way - maybe you can use the same approach here: https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/graphql.js

Qard:
Nope, tried that, it didn't work. The fields now have `configurable: false`, so they can't be redefined.

watson:
It's the same problem with `graphql`. That's why we return a `wrapped` object:

https://github.com/elastic/apm-agent-nodejs/blob/bce3adbf063b1e08fb73b63c093c926eb97fa447/lib/instrumentation/modules/graphql.js#L19

We also use the `__lookupGetter__` and setter equivalent to not trigger the getters and setters but just check if they are there:

https://github.com/elastic/apm-agent-nodejs/blob/bce3adbf063b1e08fb73b63c093c926eb97fa447/lib/instrumentation/modules/graphql.js#L26-L27

That way we get to export the same object as they export even if that object changes. This makes the wrapped object more resilient to change down the line in case they add more properties or change the existing ones.

watson:
Btw, shouldn't this be backported to the `2.x` branch as well?

watson:
I've created a PR the fix I was thinking of: #1477

PR #1476, opened 2019-10-28
refactor: use smarter shimming of graphql module

watson:
Instead of relying on that the `__esModule` is the only property that not enumerable, use the smarter `shallow-clone-shim` module to ensure the clone is identical while still shimming the getters for `graphql` and `execute`.

### DIFF
--- a/lib/instrumentation/modules/graphql.js
+++ b/lib/instrumentation/modules/graphql.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var semver = require('semver')
+var clone = require('shallow-clone-shim')
 
 var getPathFromRequest = require('../express-utils').getPathFromRequest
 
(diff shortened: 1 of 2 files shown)


watson:
The `shallow-clone-shim` module can be used for all modules that have been generated using Babel, TypeScript etc and where it's therefore not possible to overwrite its properties directly.

I'd therefore also recommend using this to patch the `tedious` module.

Qard:
Looks like this can't be backported because `Object.getOwnPropertyDescriptors(...)` doesn't exist in Node.js 6.x.

Qard:
I've made a pull request to shallow-clone-shim to make it work properly in 6.x so this can get backported. https://github.com/watson/shallow-clone-shim/pull/1

watson:
Support for Node.js 6 has been added in `shallow-clone-shim@1.1.0`. If you upgrade to that version it should work 👍 

I've also released `shallow-clone-shim@2.0.0` which doesn't support Node.js 6 to keep the footprint of the module smaller. In `master` you can use that version and keep using the `1.x` version in the `2.x` branch

PR #1477, opened 2019-10-28
fix(tedious): ensure shimmed module exposes same API

watson:
Depends on #1476 (so that `shallow-clone-shim` is available)

Ensure that the tedious module is copied faithfully so that the interface of the shimmed module is as identical to the original module as possible.

The old implementation would convert getters to their returned values and would not copy non-enumerable values. The new implementation copies both enumerable and non-enumerable values and preserves getters and setters.

### DIFF
--- a/lib/instrumentation/modules/tedious.js
+++ b/lib/instrumentation/modules/tedious.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var semver = require('semver')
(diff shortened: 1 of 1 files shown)


watson:
Once/if #1476 is merged, this PR should just be rebased with `master` so that the diff is legible.

PR #1485, opened 2019-10-30
chore(knex): bump knex dependency and fix tests

Qard:
Fixes #1480

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -63,9 +63,15 @@ bluebird:
   commands:
     - node test/instrumentation/modules/bluebird/bluebird.js
     - node test/instrumentation/modules/bluebird/cancel.js
-knex:
+knex-old:
+  name: knex
   # v0.16.4 accidentally dropped support for Node.js 6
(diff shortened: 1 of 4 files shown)


watson:
@Qard If I understand this PR correctly, it removes the assertion for if the stack trace contains a frame from the user-land code if running knex v0.18 or newer. As described [in the docs](https://www.elastic.co/guide/en/apm/agent/nodejs/current/supported-technologies.html#compatibility-better-stack-traces), the reason why the agent patches knex in the first place was to ensure this user-land frame was in the stack trace. So if this is removed from the tests, we might as well not patch those versions of next at all. Or preferably, find an alternative way to show where in the user's code the db call originated...

Qard:
Hmm...well if we _don't_ update then we continue to have a big scary banner saying we have a vulnerable dependency, because everything before 0.19 has a SQL injection vulnerability in it. 😬 

I'll see if I can figure out a way to get the stack trace in the new versions. 🤔

watson:
I think you can just tell GitHub to ignore this particular vulnerability until you find a way to properly support v0.19+

If you click the banner there's a button to ignore it I think

PR #1497, opened 2019-10-31
fix(knex): make stack traces work in 0.18+

Qard:
I figured out how to get stack traces working properly in knex. It was a result of knex switching from Bluebird to native promises in 0.18. In the process, they also switched to using async functions, which our version of shimmer apparently was unable to patch correctly because it used a `toString()` check to identify functions rather than a simple `typeof`, and async functions have a different name than regular functions.

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/shimmer.js
+++ b/lib/instrumentation/shimmer.js
@@ -32,7 +32,7 @@ function logger () {
 }
 
 function isFunction (funktion) {
-  return funktion && {}.toString.call(funktion) === '[object Function]'
+  return typeof funktion === 'function'
 }
 
 function wrap (nodule, name, wrapper) {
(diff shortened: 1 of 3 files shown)


lib/instrumentation/shimmer.js
```diff
@@ -32,7 +32,7 @@ function logger () {
 }
 
 function isFunction (funktion) {
-  return funktion && {}.toString.call(funktion) === '[object Function]'
+  return typeof funktion === 'function'
```
watson: @Qard Wow, nice find!

This was also updated in the upstream `shimmer` module last year: https://github.com/othiym23/shimmer/commit/ec15ba2bf2cffc25622fca0d277ea68f0ae2d8ee

I wonder why `typeof` wasn't used originally.

Why was this making knex fail btw? I would love to know what knex did to fail this check 🤔 Did they use async or generator functions?

Qard: They started using async functions.

PR #1499, opened 2019-11-01
feat(parsers): collect error causes

### LINKED ISSUE #1196: Look into supporting chained exceptions
We're adding support for chained exceptions in Elastic APM (see elastic/apm#40).

However, we don't have a way to chain in exceptions in Node.js, but I was thinking we might provide an API that allows the user to record such things.

E.g. it's not uncommon in Node.js to say get a low-level error back from the database driver and then throw a more high-level error. Having access to both the low and the high-level error would be useful.

Qard:
Fixes elastic/apm-agent-nodejs#1196

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -5,6 +5,7 @@ var url = require('url')
 
 var afterAll = require('after-all-results')
 var basicAuth = require('basic-auth')
+var Causes = require('causes')
 var getUrlFromRequest = require('original-url')
 var httpHeaders = require('http-headers')
 var stringify = require('fast-safe-stringify')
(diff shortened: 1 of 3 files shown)


AnandChowdhary:
Hi @trentm and @graphaelli,

I contacted @Qard about requesting the npm package name `causes` (for a "developers can support causes like BLM in their projects"-kind of project), and they told me that since they made this package as part of this PR, I should ask Elastic whether they'd be up to letting me have it.

If you don't have any use for it for the time being, I'd love to have you consider this! 

http://npmjs.com/package/causes - I'm on npm as @`anandchowdhary`

trentm:
@AnandChowdhary (/cc @Qard), I am fine with passing the "causes" package name on.  

I'll close this PR -- once https://github.com/elastic/apm-agent-nodejs/pull/2094 goes in, this PR would need to be re-written or deal with the merge conflict.

Qard:
I've handed over the module. If you choose to revisit this and would like to make use of the code I had previous written, it's currently available at https://github.com/qard/causes. Feel free to use it, or not, as you choose. Happy to hand over the code rights or whatever. :)

trentm:
@Qard Thanks!

PR #1523, opened 2019-11-15
fix(ioredis): prevent unhandled promise rejection

### LINKED ISSUE #1518: Uncaught promise
**Describe the bug**

Uncaught promise!

There is a rule in Node.JS that you must surround all `new Promises` with a `catch`. This is typically done using `try ... catch` inbetween `async` `await` syntax, or using the chainable `Promise` `.catch()` interface. 

The fact that a promise was created without this structure indicates that the apm node programmer may not have understood a fundamental concept, and therefore you might also want to review all other Promise structures throughout the code, to be thorough about how these are handled.

The side-effect is that while APM is loaded, some redis errors are not catchable, therefore the application cannot recover from failure.

A secondary future effect is that the process may be forced to exit.

```
{
	"ts": "2019-11-13T21:46:58.039Z",
	"src": "internal/process/warning.js:27:3",
	"error": "(node:19157) UnhandledPromiseRejectionWarning: ReplyError: WRONGTYPE Operation against a key holding the wrong kind of value\n    at parseError (/workspace/projectX/node_modules/redis-parser/lib/parser.js:179:12)\n    at parseType (/workspace/projectX/node_modules/redis-parser/lib/parser.js:302:14)",
	"data": {}
}
{
	"ts": "2019-11-13T21:46:58.040Z",
	"src": "internal/process/warning.js:27:3",
	"error": "(node:19157) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)",
	"data": {}
}
{
	"ts": "2019-11-13T21:46:58.040Z",
	"src": "internal/process/warning.js:27:3",
	"error": "(node:19157) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.",
	"data": {}
}
```

**To Reproduce**

Steps to reproduce the behavior: 
1. Create a Hash by means necessary.
    `HSET a b c`
2. Then try to read it with the wrong command, using `ioredis` library with `elastic-apm-node` module loaded.
    `GET a`
3. Observe error thrown. While it is correct that Redis error should be thrown here, it is INCORRECT that it should be made UNCATCHABLE due to interference by the APM library, resulting in process exit.

**Expected behavior**

I expect when such an error is thrown, that I can catch and handle it. Like so:

```js
(async () => {
try {
  await redis.hget('a');
}
catch(e) {
  debugger; // should break here, but doesn't (only while APM is loaded)
}
})();
```

- OS: Linux
- Node.js version: v12.12.0
- APM Server version: N/A
- Agent version: 3.1.0

**Additional context**

APM is nice. Plz fix. Also we're a paying elastic enterprise customer (Ancestry)

package.json
```
    "elastic-apm-node": "^3.1.0",
    "ioredis": "^4.11.2",
```

axw:
Fixes #1518

The use of `finally` without updating `command.promise` causes a bifurcation. The onFinally handler will rethrow, and nothing can catch it, leading to an unhandled promise rejection.

Let's simplify the code a bit, and just use `.then` regardless, which avoids this issue.

### Checklist

- [x] Implement code
- [x] Add tests
~- [ ] Update TypeScript typings~
~- [ ] Update documentation~

### DIFF
--- a/lib/instrumentation/modules/ioredis.js
+++ b/lib/instrumentation/modules/ioredis.js
@@ -67,11 +67,7 @@ module.exports = function (ioredis, agent, { version, enabled }) {
           const endSpan = function () {
             if (!span.ended) span.end()
           }
-
-          if (typeof command.promise.finally === 'function') {
-            // Bluebird and Node.js 10+
-            command.promise.finally(endSpan)
-          } else if (typeof command.promise.then === 'function') {
+          if (typeof command.promise.then === 'function') {
             command.promise.then(endSpan).catch(endSpan)
           }
         }
--- a/test/instrumentation/modules/ioredis.js
+++ b/test/instrumentation/modules/ioredis.js
@@ -97,6 +97,37 @@ test('nested', function (t) {
   })
 })
 
+test('rejections_handled', function (t) {
+  // Make sure there are no unhandled promise rejections
+  // introduced by our promise handling. See #1518.
+  let unhandledRejection = false
+  function onUnhandledRejection (e) {
+    unhandledRejection = true
+  }
+  process.once('unhandledRejection', onUnhandledRejection)
+  t.on('end', () => {
+    process.removeListener('unhandledRejection', onUnhandledRejection)
+  })
+  agent._instrumentation.currentTransaction = null
+  agent._transport = mockClient(3, function () {
+    setTimeout(function () {
+      t.notOk(unhandledRejection)
+      t.end()
+    }, 0)
+  })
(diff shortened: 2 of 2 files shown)


watson:
I don't think we need to backport anything to 1.x. No one should be using that version anymore and I don't think the tests work on the branch

PR #1528, opened 2019-11-19
ci(jenkins): update argument (#1522)

v1v:
Backport #1522


watson:
@v1v Please use the `npm run backport` command when backporting as it makes it easier to later see which PR's are backport-PR's and removes most user-errors.

Tip: If you're not backporting your own PR, use `npm run backport -- --all` to see all commits by other authors

PR #1530, opened 2019-11-20
ci(jenkins): Update CI badges

mdelapenya:
## What is this PR doing?
It updates Jenkins badge for the master branch, using the unprotected one, so that anybody wit ViewStatus access can read it.

**protected** exposes the badge to users having at least Read permission on the job
**unprotected** exposes the badge to users having at least ViewStatus permission on the job

## Why is it important?
We want to keep al projects in sync in terms of CI, so that we can be even more transparent about the status of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@ If you have any feedback or questions,
 please post them on the [Discuss forum](https://discuss.elastic.co/c/apm).
 
 [![npm](https://img.shields.io/npm/v/elastic-apm-node.svg)](https://www.npmjs.com/package/elastic-apm-node)
-[![Build status](https://travis-ci.org/elastic/apm-agent-nodejs.svg?branch=2.x)](https://travis-ci.org/elastic/apm-agent-nodejs)
+[![Build Status](https://apm-ci.elastic.co/buildStatus/icon?job=apm-agent-nodejs%2Fapm-agent-nodejs-mbp%2F3.x)](https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/3.x/)
+[![Build status](https://travis-ci.org/elastic/apm-agent-nodejs.svg?branch=3.x)](https://travis-ci.org/elastic/apm-agent-nodejs)
 [![Standard - JavaScript Style Guide](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat)](https://github.com/standard/standard)
 
 ## Installation


mdelapenya:
Hey @watson, so then how do you update the Jenkins badge in this case? Each branch has its own CI build and the readme should point to the proper build.

What is your recommendation here?

axw:
@mdelapenya perhaps we shouldn't run tests for the current major, separately from master? i.e. just don't have a job for 3.x, and instead have ones for 1.x., 2.x, and master.

watson:
Don't think of the `3.x` branch as a "real" branch, but more like a "pointer" to the latest release on the `master` branch.

I would think of the badges as showing the build status of the latest released version, so in `master`, I'd make all the badges point to `3.x` branch. The badges are mostly there for the users (not for us) and they are shown on https://www.npmjs.com/package/elastic-apm-node as well, so it would look a little bad if say `master` was broken and then showed a red badge even though the latest release was green.

mdelapenya:
@axw @watson to exclude 3.x branch, it would be a matter of simply updating the JJBB descriptor (https://github.com/elastic/apm-agent-nodejs/blob/master/.ci/jobs/apm-agent-nodejs-mbp.yml#L15), the same way we do with Greenkeeper branches. The dark side of this is having to maintain that change with every upgrade in the branch name (i.e. from 3.x to 4.x).

Related to the use cases, we do consume internally those badges, as we want to know in a quick shot the status of all our repositories in the form of Jenkins jobs.

BTW, did you take a look #1527, which updated the badge for the master branch? Is it OK? We can revert it if needed.

To understand the branching model, 3.x is stable, right? and master is not. Reading the CONTRIBUTING.md document, we noticed the release process is manual. We can assist here to create automated releases based on git tags, in a similar manner to what we do for [the Ruby agent](https://github.com/elastic/apm-agent-ruby/blob/master/.ci/Jenkinsfile#L181-L215).

watson:
I personally wouldn't exclude the `3.x` branch - at least not if this is the one we want to show the build status off in badges. As I mentioned in my above comment I'd just updated the badges in `master` to point to `3.x` and leave it at that.

> BTW, did you take a look #1527, which updated the badge for the master branch? Is it OK? We can revert it if needed.

I can see that the badge was actually pointing to the `2.x` branch. This was an oversight on my part from when we bumped the major from `2` to `3`. No need to revert this one - instead we should make a PR to change it to `3.x` instead if y'all agree on my approach here.

watson:
Regarding the release process: There's currently no way that I know of to use 2FA for automated releases which is why we haven't enabled it so far.

mdelapenya:
> Regarding the release process: There's currently no way that I know of to use 2FA for automated releases which is why we haven't enabled it so far.

We discussed about that topic with the RUM team during last EAH, and we are going to work on finding a solution (if possible). We'll keep you in the loop

mdelapenya:
> I personally wouldn't exclude the `3.x` branch - at least not if this is the one we want to show the build status off in badges. As I mentioned in my above comment I'd just updated the badges in `master` to point to `3.x` and leave it at that.
> 
> > BTW, did you take a look #1527, which updated the badge for the master branch? Is it OK? We can revert it if needed.
> 
> I can see that the badge was actually pointing to the `2.x` branch. This was an oversight on my part from when we bumped the major from `2` to `3`. No need to revert this one - instead we should make a PR to change it to `3.x` instead if y'all agree on my approach here.

We are going to keep 3.x branch (we see it convenient even though we also run builds per git tags), and we'll send a PR updating Travis badge on master pointing to the latest pointer: 3.x

Thanks!

PR #1536, opened 2019-11-20
fix(package): bump shallow-clone-shim to ^2.0.0

watson:
Bumping non-dev-dependencies in separate PR's to make a future `git bisect` easier in case any of these has unforeseen side-effects


watson:
Jenkins integration test failures are unrelated (see elastic/apm-integration-testing#672)

PR #1539, opened 2019-11-20
ci(jenkins): Use 3.x branch for CI badges

mdelapenya:
As discussed in #1530, Travis' master branch badge must point to latest release: 3.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ If you have any feedback or questions,
 please post them on the [Discuss forum](https://discuss.elastic.co/c/apm).
 
 [![npm](https://img.shields.io/npm/v/elastic-apm-node.svg)](https://www.npmjs.com/package/elastic-apm-node)
-[![Build Status](https://apm-ci.elastic.co/buildStatus/icon?job=apm-agent-nodejs%2Fapm-agent-nodejs-mbp%2Fmaster)](https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/master/)
-[![Build status](https://travis-ci.org/elastic/apm-agent-nodejs.svg?branch=master)](https://travis-ci.org/elastic/apm-agent-nodejs)
+[![Build Status](https://apm-ci.elastic.co/buildStatus/icon?job=apm-agent-nodejs%2Fapm-agent-nodejs-mbp%2F3.x)](https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/3.x/)
+[![Build status](https://travis-ci.org/elastic/apm-agent-nodejs.svg?branch=3.x)](https://travis-ci.org/elastic/apm-agent-nodejs)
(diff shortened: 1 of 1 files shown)


mdelapenya:
> Will you create a separate PR to update the Jenkins badge to `3.x` as well?

I was thinking about keeping master branch for Jenkins and 3.x for Travis as per your previous comments. From my perspective, and taking apart this specific use case where 3.x is a clone-ish of master, providing information (the CI badge) in a branch (master) from other branch (3.x) could cause confusion, so each branch should contain its own metadata (that was the rationale for sending #1530 against the 3.x branch).

On the other hand, having dynamism in the badges, it forces us to send a PR updating the markdown (as we did here: https://github.com/elastic/apm-agent-nodejs/pull/370/files and https://github.com/elastic/apm-agent-nodejs/pull/682/files). Which could be convenient, or be toil.

If the purpose of the badge in master is to show the state of the project in the latest released version (3.x?), we have the badge for the git tag (v3.2.0), which is the most accurate status for the release. But again, we should update the badge with every release.

watson:
I probably didn't express myself clearly. I'd prefer that the badges [shown on npm](https://www.npmjs.com/package/elastic-apm-node) show the latest version that is released and not the current state of `master`. Hence why I previously had pointed the badges to those branches. No matter what, I don't think it makes sense to have the Travis badge and the Jenkins badge point to two different branches, so they should either point to `master` or to `3.x`. I personally prefer that they point to `3.x` as that's the easiest to maintain and works well with how npm shows the content of the README as it looked when the package was published.

What's the reason why you want them to point to `master`?

mdelapenya:
The main reason to use `master` is because we are in the master branch :)

Another topic is we are using `master` as a `develop` branch, that's why we use a 3.x branch which is _stable_, and we release (_git tag and npm stuff_) from there.

Not going deep in what process NPM uses to fetch and publish the readme in their site (_I guess they get the code from the branch and stores in their system -they keep a different version of the readme per release_), I'd go with your solution: having 3.x in master branch, but then it's important not to forget updating the badge as we did last time, or otherwise NPM will point to a broken branch as of now (pointing to a broken release branch: https://www.npmjs.com/package/elastic-apm-node)

As a side note, attaching a placeholder such as 3.x to a release, instead of the git tag itself (v3.2.0), it would mean that any minor/patch release of the v3 family (v3.2.0, v3.1.0, ...) will point to the same build: 3.x HEAD, which would be incorrect in most of the cases.

watson:
We can fix the commit message lint error when we merge

mdelapenya:
> We can fix the commit message lint error when we merge

I can push-force my branch to avoid the overhead of a new review. The error is related to the uppercased sentence, right?

watson:
> The error is related to the uppercased sentence, right?

Correct - First letter the `:` should be lowercase

watson:
When squashing we should name the commit `docs: use 3.x branch for CI badges` as this is really a docs thing and not a CI thing

mdelapenya:
squashed and pushed with the name you suggested, thanks!

watson:
Thanks 👍 Btw, I realized that we hadn't really documented these commit message guidelines, so I made a PR to fix that: #1541

mdelapenya:
Ready to merge!

image

PR #1540, opened 2019-11-20
ci(travis): don't update npm to latest version

### LINKED ISSUE #1532: CI fails for Node.js 6.0.x on 2.x branch
CI fails because it's trying to use the latest npm version on Node.js 6.0.x which isn't supported. Instead of depending on `npm@latest`, I suggest finding a suitable version and hardcoding that for the `6.0.x` tests on the `2.x` branch.

watson:
Fixes elastic/apm-agent-nodejs#1532

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,6 @@ install:
   - for i in {1..3}; do travis_wait 5 npm install && break; rm -rf node_modules; done
 
 before_script:
-  - npm install -g npm@latest
   - wait-on tcp:9200
 
 node_js:


watson:
We could have a special step for Node.js 6.0.x only, but I don't actually think that we _technically_ need the latest npm on Travis now that we don't run the TAV tests there anymore. So this seemed like the simpler solution. We'll know ones the tests complete...

watson:
It did the job!

watson:
Test failures are unrelated (see elastic/apm-integration-testing#672)

PR #1555, opened 2019-11-29
feat(config): extend central config support

axw:
Extend central config support to accept `transaction_max_spans` and `capture_body`.

Relates to https://github.com/elastic/apm/issues/138

### Checklist

- [x] Implement code
- [x] Add tests
~- [ ] Update TypeScript typings~
~- [ ] Update documentation~

### DIFF
--- a/test/central-config-enabled.js
+++ b/test/central-config-enabled.js
@@ -9,7 +9,17 @@ const http = require('http')
 const test = require('tape')
 
 test('remote config enabled', function (t) {
-  t.plan(2)
+  const updates = {
+    transaction_sample_rate: '0.42',
+    transaction_max_spans: '99',
+    capture_body: 'all'
+  }
+  const expect = {
+    transactionSampleRate: 0.42,
+    transactionMaxSpans: 99,
+    captureBody: 'all'
+  }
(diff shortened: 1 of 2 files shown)


test/central-config-enabled.js
```diff
@@ -36,15 +44,28 @@ test('remote config enabled', function (t) {
...
-      configurable: true
-    })
+      } else {
+        Object.defineProperty(agent._conf, key, {
+          set (value) {
+            const expectValue = expect[key]
+            if (expectValue !== undefined) {
+              t.equal(value, expectValue)
+              delete expect[key]
+              if (expect === {}) {
```
watson: This if-statement will never evaluate true as `expect` is one object and `{}` is another - and hence never the same object. The reason why this test works anyway is that the `t.plan()` statement above ends the test automatically after 4 assertions.

You can choose to either leave it at that and remove this if-block, or you can update the if-statement to check th amount of keys left in `expect` instead, e.g:

```suggestion
              if (Object.keys(expect).length === 0) {
```

axw: oops, and thank you :man_facepalming:

PR #1560, opened 2019-12-09
Make server CA certificate configurable

### LINKED ISSUE #1508: add configuration option to specify a certificate authority file ( SSL )
nodejs agents and are required to disable SSL verification as described [here](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#validate-server-cert). The key statement is

Disabling validation is normally required if using self-signed certificates.

Implement an option to define a configuration option to specify a certificate authority file.

**Additional context**
https://nodejs.org/api/cli.html#cli_node_extra_ca_certs_file

axw:
Add a new `serverCaCert` config, which accepts a path to a file containing a CA certificate. If specified, this will be used for verifying the server's TLS certificate instead of the default trusted CAs.

The config name was chosen to align with the Ruby's agents: https://www.elastic.co/guide/en/apm/agent/ruby/current/configuration.html#config-ssl-ca-cert.

Closes #1508

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


docs/configuration.asciidoc
```diff
@@ -73,7 +73,20 @@ The URL to where the APM Server is deployed.
...
 You can switch this behavior off by setting this option to `false`.
-Disabling validation is normally required if using self-signed certificates.
+
+[[server-ca-cert]]
+==== `serverCaCert`
+
+* *Type:* String
+* *Env:* `ELASTIC_APM_SERVER_CA_CERT`
+
+By default the agent will validate the TLS/SSL certificate of the APM Server using the
```
vigneshshanmugam: can we mention somewhere as `path to the cert location` or `file path`?  Just by going over the docs, I was confused with whether user had to read the file and pass the contents.

docs/configuration.asciidoc
```diff
@@ -73,7 +73,20 @@ The URL to where the APM Server is deployed.
...
+well-known CAs curated by Mozilla, as described at https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options.
```
watson: I would prefer if the full URL wasn't displayed in the docs, but instead something like:

```suggestion
well-known CAs curated by Mozilla, as described in the Node.js docs for https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options[`tls.createSecureContext()`].
```

test/integration/verify-server-ca-cert.js
```diff
@@ -0,0 +1,50 @@
...
+    var cert = fs.readFileSync(path.join(__dirname, 'cert.pem'))
+    var key = fs.readFileSync(path.join(__dirname, 'key.pem'))
+
+    var server = https.createServer({ cert, key }, function (req, res) {
+      t.pass('server received client request')
+      res.end()
+    })
+
+    server.listen(port, function () {
+      agent.captureError(new Error('boom!'), function () {
```
watson: I would check if the callback is called with an error as this could be an indication that something with the server certificate was wrong:

```suggestion
      agent.captureError(new Error('boom!'), function (err) {
        t.error(err)
```

Then you could either ajust `t.plan` to `3`, or just drop the extra `t.pass` below.

docs/configuration.asciidoc
watson: If we want users to only be able to provide a file-path and not the content of the certificate, we might consider renaming the config option to `serverCaCertFile` to be more explicit. This would also allow us to add another config option in the future called simply `serverCaCert` which you could use to provide the actual cert body (a feature which I would find quite useful in certain scenarios as you might get the cert body from a configuration system or an environment variable and not actually have it on disk). This could of course also be achieved in the same config option by detecting the cert `BEGIN` signature.

axw: Yeah, I was on the fence with this. My instinct was to call it `serverCaCertFile`, but then there's the existing Ruby agent config. I'll open an issue on the Ruby agent to deprecate/change its config name.

test/integration/verify-server-ca-cert.js
axw: Done. I copied the test from `test/integration/allow-invalid-cert.js`, so I've made the change there also.

docs/configuration.asciidoc
axw: Opened https://github.com/elastic/apm-agent-ruby/issues/650, and I've renamed `serverCaCert` to `serverCaCertFile`.

@vigneshshanmugam the docs already said

> You can set this option to the path of a file containing a CA certificate that will be used instead.

But I guess part of the confusion came from the attribute name, so hopefully your concern is also addressed. Let me know if not.

lib/config.js
```diff
@@ -548,3 +550,15 @@ function loadConfigFile (configFile) {
...
   return null
 }
+
+function loadServerCaCertFile (opts) {
+  if (opts.serverCaCertFile) {
+    try {
+      return fs.readFileSync(opts.serverCaCertFile)
+    } catch (err) {
+      opts.logger.error('Elastic APM initialization error: Can\'t read server CA cert file %s', opts.serverCaCertFile)
+      opts.logger.error('APM Server transport error:', err.stack)
```
watson: The text here is probably just due to a copy-paste error. And since the stack trace isn't important here, I would personally just merge the two lines into something like this:

```js
      opts.logger.error('Elastic APM initialization error: Can\'t read server CA cert file %s (%s)', opts.serverCaCertFile, err.message)
```

Or do as in the `loadConfigFile` function above and just output the stack trace standalone.

lib/config.js
```diff
@@ -548,3 +550,15 @@ function loadConfigFile (configFile) {
...
+    }
+  }
+  return undefined
```
watson: Not important, but just FYI. In JavaScript, a `return undefined` is implied in functions if either no `return` is called, or `return` without any value is called - so we normally don't explicitly do this.

docs/configuration.asciidoc
vigneshshanmugam: Ahh, i missed the docs part. I just glimpsed over and was wondering about wide vareity of options that we can pass here like string, string[] or buffer. But yeah the config option change clearly conveys it. Thanks 👍

lib/config.js
axw: yikes :)
done, thanks!

lib/config.js
axw: thanks!

PR #1566, opened 2019-12-11
fix(elasticsearch): ensure requests can be aborted

### LINKED ISSUE #1565: Elasticsearch instrumentation breaks returned promise
In the [elasticsearch](https://www.npmjs.com/package/elasticsearch) module instrumentation, any request can return a promise, eg:

```js
client.ping().then(() => {
  console.log('ping successful')
})
```

The returned promise has an extra function not normally on promises: `abort()`. The abort function is used to abort the request to Elasticsearch. As part of the instrumentation, this agent returns a different promise which doesn't have the expected `abort()` function. This, in turn, causes errors like elastic/kibana#52617.

watson:
Fixes #1565

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/test/instrumentation/modules/elasticsearch.js
+++ b/test/instrumentation/modules/elasticsearch.js
@@ -68,6 +68,7 @@ test('client.ping with unhandled promise', function userLandCode (t) {
 
   function unhandledRejection (reason, promise) {
     t.equal(promise, unhandledPromise)
+    t.strictEqual(typeof promise.abort, 'function', 'promise.abort should be a function')
     agent.endTransaction()
     agent.flush()
     client.close()
--- a/test/instrumentation/modules/elasticsearch.js
+++ b/test/instrumentation/modules/elasticsearch.js
@@ -48,6 +48,32 @@ test('client.ping with promise', function userLandCode (t) {
   })
 })
 
+test('client.ping with unhandled promise', function userLandCode (t) {
+  resetAgent(2, doneNoHTTP(t, 'HEAD', '/'))
+
+  agent.startTransaction('foo')
+
+  const client = new elasticsearch.Client({
+    host: 'thiswillneverwork',
+    log: {
+      type: 'file',
+      level: 'error',
+      path: '/dev/null'
+    }
+  })
+
+  const unhandledPromise = client.ping()
+
+  process.once('unhandledRejection', agent._instrumentation.bindFunction(unhandledRejection))
+
+  function unhandledRejection (reason, promise) {
+    t.equal(promise, unhandledPromise)
+    agent.endTransaction()
+    agent.flush()
+    client.close()
+  }
+})
+
 test('client.search with callback', function userLandCode (t) {
   resetAgent(done(t, 'POST', '/_search', '{"q":"pants"}'))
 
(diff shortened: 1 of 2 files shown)


test/instrumentation/modules/elasticsearch.js
```diff
@@ -220,14 +237,22 @@ function done (t, method, path, query) {
     }
 
     t.ok(span1.timestamp > span2.timestamp, 'http span should start after elasticsearch span')
-    t.ok(span1.timestamp + span1.duration * 1000 < span2.timestamp + span2.duration * 1000, 'http span should end before elasticsearch span')
+    if (abort) {
+      t.ok(span1.timestamp + span1.duration * 1000 > span2.timestamp + span2.duration * 1000, 'http span should end after elasticsearch span when req is aborted')
```
vigneshshanmugam, line 241: @watson This is due to how the abort is called under the hood in the library. Is this okay from users perspective? Are we can somehow end all the underlying spans when http request is aborted?

watson: I can't currently remember precisely why the order here is reverse (it's been a while since I looked at this issue). I think it's because we actively intercept the call to `abort()`, at which time we end the span. So at this point in time, the underlying http request hasn't been notified about the abort yet, and so the http span is still active - however, hopefully not for very long.

You could also argue, that in a non-aborted situation, the elasticsearch span is ended because the underlying http request ends, where in this case, it's ended because the "user" makes an abort request, in which case the elasticsearch client forcefully ends the http request. So the order is also kind of reverse in real life.

Regarding ending of child-spans: The agent doesn't currently have a way to walk all the child-spans. In some cases, this would not even be possible because the precise relationship might not even be known (all we might know is that they all belong to the same transaction). However, I can't remember if the agent in this case does have the necessary information for us to be able to do that _if_ the information was exposed.

vigneshshanmugam: > So at this point in time, the underlying http request hasn't been notified about the abort yet, and so the http span is still active - however, hopefully not for very long.

Yes i thought the same and we are actually aborting the Elasticsearch span immediately on the user's abort call https://github.com/elastic/apm-agent-nodejs/blob/d22f59453e448db6cf09d25caee5ba953bfa5b80/lib/instrumentation/modules/elasticsearch.js#L73-L81

which would notify the http and abort event fires later which explains the behaviour.  

Agree with you on the non-aborted situation.

watson: I think it's ok to ship. You could move the `span.end()` call down below the call to `originalAbort.apply(this, args)`. However, I doubt that it would change much since the HTTP request span most likely would be ended asynchronously anyway. But the code might show the intent better this way?

vigneshshanmugam: Yeah sure that would show the intent better. I will make the change.

vigneshshanmugam: Done, ready for review now.

watson:
jenkins run the module tests for elasticsearch

test/instrumentation/modules/elasticsearch.js
watson: LGTM 👍  I can't technically approve it on GitHub because I'm the author though

vigneshshanmugam:
jenkins run the module tests for elasticsearch

Issue #994, opened 2019-04-10
Add full support for TraceContext


watson:
See elastic/apm#71 for details

PR #1587, opened 2020-01-10
feat: support W3C TraceContext traceparent header

jahtalab:
This PR adds support for W3C TraceContext `traceparent` header while still keeping the `elastic-apm-traceparent` header. The use of the elastic header is controlled by `ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER` config option. For more see: #994

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] ~~Update TypeScript typings~~
- [ ] ~~Update documentation~~

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -73,7 +73,8 @@ var DEFAULTS = {
   transactionMaxSpans: 500,
   transactionSampleRate: 1.0,
   usePathAsTransactionName: false,
-  verifyServerCert: true
+  verifyServerCert: true,
+  useElasticTraceparentHeader: true
 }
 
 var ENV_TABLE = {
@@ -126,7 +127,8 @@ var ENV_TABLE = {
   transactionMaxSpans: 'ELASTIC_APM_TRANSACTION_MAX_SPANS',
   transactionSampleRate: 'ELASTIC_APM_TRANSACTION_SAMPLE_RATE',
   usePathAsTransactionName: 'ELASTIC_APM_USE_PATH_AS_TRANSACTION_NAME',
-  verifyServerCert: 'ELASTIC_APM_VERIFY_SERVER_CERT'
+  verifyServerCert: 'ELASTIC_APM_VERIFY_SERVER_CERT',
+  useElasticTraceparentHeader: 'ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER'
 }
 
 var CENTRAL_CONFIG = {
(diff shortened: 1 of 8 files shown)


lib/config.js
```diff
@@ -73,7 +73,8 @@ var DEFAULTS = {
   transactionMaxSpans: 500,
   transactionSampleRate: 1.0,
   usePathAsTransactionName: false,
-  verifyServerCert: true
+  verifyServerCert: true,
+  useElasticTraceparentHeader: true
```
watson: We used to sort these lists alphabetically - probably best to keep doing that :)

test/instrumentation/modules/http/basic.js
```diff
@@ -82,23 +82,41 @@ test('new http.Server', function (t) {
     server.on('request', onRequest(t))
     sendRequest(server)
   })
+
+  t.test('support elastic header', function (t) {
```
watson: I would probably name this test a little more descriptive than "elastic header" as it's not clear which header this refers to, eg:

```suggestion
  t.test('support elastic-apm-traceparent header', function (t) {
```

test/instrumentation/modules/http/outgoing.js
```diff
@@ -135,8 +135,12 @@ function echoTest (type, handler) {
         res.resume()
       })
 
-      var traceparent = req.getHeader('elastic-apm-traceparent')
-      t.ok(traceparent, 'should have elastic-apm-traceparent header')
+      var traceparent = req.getHeader('traceparent')
+      t.ok(traceparent, 'should have traceparent header')
+      if (agent._conf.useElasticTraceparentHeader) {
```
watson: This will as far as I can see never be set in this test right? How about creating a test where you specifically test that this config option has the desired effect?

jahtalab: The default for `useElasticTraceparentHeader` config is `true` so it will be set, unless I'm missing something. I'm not against having a separate test though, just want to make sure we're on the same page.

watson: I was more thinking of testing that if you set it to `false`, then it works as expected. In its current state, the if-statement will _always_ execute as the config-option is _always_ true. But we don't have a test to test that the header is not present if the config option is set to `false` as far as I can see

jahtalab: I've added the tests, please have another look.

jahtalab:
@watson , I've rebased to include the babel test fix. Let me know what do you think.

PR #1613, opened 2020-02-05
feat: parse error stacks manually

### LINKED ISSUE #1590: Agent doesn't report GraphQLError
> it seems that the agent gets a bit confused because that object appears not to be a native Error type, and so the agent represents it as a log message rather than an exception

thanks @axw 

if you call apm.captureError(err.originalError), that does have a stack trace. 

have a look at : 
https://discuss.elastic.co/t/capturing-stack-traces-from-graphqlerrors-in-apollo-server/212563/1

vigneshshanmugam:
+ fix #1590 
+ Currently our error stack construction depends on v8 API stack generation function hook which would be called whenever error stack is created, but when the error stack property is changed somewhere in the code/modules, we wont be able to access the callsite information. 
+ This PR provides a sensible fallback when we cannot access the callsite information for errors and also for errors that have additional `originalError` property which reveals the true error (graphqljs adds this property and creates GraphqlError which has this originalError property). 

An example error in Apollo server would look like this in our UI. 
Screenshot 2020-02-05 at 15 35 22

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -41,15 +41,14 @@ exports.parseMessage = function (msg) {
 }
 
 exports.parseStackTrace = function (err) {
+  // graphqljs adds the `originalError` property which represents the original
+  // error thrown within the resolver
   err = err.originalError || err
   let stackframes = []
   if (err.stack == null) {
     return stackframes
   }
-  try {
-    stackframes = stacktrace.parse(err)
-  } catch (_) {}
-  const frames = stackframes.map((frame) => {
+  return stacktrace.parse(err).map((frame) => {
     const filename = frame.getFileName() || ''
     return {
       abs_path: filename,
@@ -59,7 +58,6 @@ exports.parseStackTrace = function (err) {
       library_frame: !isApp(frame)
     }
   })
-  return frames
 }
 
 exports.parseError = function (err, agent, cb) {
@@ -95,11 +93,13 @@ exports.parseError = function (err, agent, cb) {
       // they were passed on, we would want to suppress them here anyway
 
       if (frames.length === 0) {
-        /**
-         * If we cannot able to extract callsite information from the error, then
-         * we fallback to parsing the error manually
-         */
-        frames = exports.parseStackTrace(err)
+         // If we are not able to extract callsite information from the error, then
+         // we fallback to parsing the error manually
+        try {
+          frames = exports.parseStackTrace(err)
+        } catch (parseError) {
+          agent.logger.debug('error parsing the stack: %s', parseError.message)
+        }
       }
       var culprit = getCulprit(frames)
       var module = getModule(frames)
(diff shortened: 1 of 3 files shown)


vigneshshanmugam:
@watson Can i get review on this?

lib/parsers.js
```diff
@@ -38,6 +40,28 @@ exports.parseMessage = function (msg) {
   return error
 }
 
+exports.parseStackTrace = function (err) {
+  err = err.originalError || err
```
watson: I'd prefer a comment saying which framework we normally expect `originalError` to come from as this isn't a standard thing

lib/parsers.js
```diff
@@ -70,14 +94,18 @@ exports.parseError = function (err, agent, cb) {
...
 
-      if (frames) {
-        var culprit = getCulprit(frames)
-        var module = getModule(frames)
-        if (culprit) error.culprit = culprit // TODO: consider moving culprit to exception
-        if (module) error.exception.module = module // TODO: consider if we should include this as it's not originally what module was intended for
-        error.exception.stacktrace = frames
+      if (frames.length === 0) {
+        /**
+         * If we cannot able to extract callsite information from the error, then
```
watson: Multiline comments should be written with regular `//`

lib/parsers.js
```diff
@@ -252,3 +280,23 @@ function tryJsonStringify (obj) {
...
 }
+
+function getRelativeFileName (filename) {
+  var root = process.cwd()
+  if (root[root.length - 1] !== path.sep) root += path.sep
+  return !~filename.indexOf(root) ? filename : filename.substr(root.length)
+}
+
+/**
+ * Below two functions expects the passed in object to be of type stackframe
```
watson: Multiline comments should be written with regular `//`. I'd also be careful with one comment concerting two functions as things might get refactored. In cases like this, I think it's better to just repeat the comment above each function 🙂

lib/parsers.js
```diff
@@ -252,3 +280,23 @@ function tryJsonStringify (obj) {
...
+ * https://github.com/stacktracejs/stackframe
```
watson: If I understand this correctly, it's just an implementation of the native V8 stack trace API. Better link to that as it's the real source: https://v8.dev/docs/stack-trace-api

test/parsers.js
```diff
@@ -457,7 +457,37 @@ test('#parseError()', function (t) {
...
+      t.equal(parsed.culprit, 'original stack')
+      t.notOk('log' in parsed)
+      t.ok('exception' in parsed)
+      t.equal(parsed.exception.message, 'error with originalError')
+      t.equal(parsed.exception.type, 'Error')
+      t.notOk('code' in parsed.exception)
+      t.notOk('handled' in parsed.exception)
+      t.notOk('attributes' in parsed.exception)
+      t.ok('stacktrace' in parsed.exception)
+      t.ok(parsed.exception.stacktrace.length > 0)
```
watson: Wouldn't it be safest to also test that the `stacktrace` contains what we expect?

vigneshshanmugam: yeah, will add them to the test

brianstebar2:
👍 Thank you!

PR #1637, opened 2020-02-18
[docs] Add EOL table

bmorelli25:

This PR defines end of life and end of maintenance dates for the Node.js Agent. EOL dates were determined by going through old releases, finding the GA release date of the major Node.js Agent versions, and adding 18 months.

## Why is it important?

We love all our products, but sometimes we must say goodbye to a release so that we can continue moving forward on future development and innovation.

## What does EOL/Maintenance mean?

> When a version goes out of maintenance, it is no longer guaranteed to get bug fixes. Between the end of maintenance and EOL, bug fixes will be discretionary, e.g. critical security issues. Once it's EOL, it's no longer given any support attention; issues related to it will be closed with WONTFIX.

## Related issues

For elastic/apm#173.

## Screenshots

Screen Shot 2020-02-18 at 1 39 17 PM


docs/upgrading.asciidoc
```diff
@@ -7,12 +7,33 @@ endif::[]
 
 == Upgrading
 
+Upgrades between minor versions of the agent, like from 1.1 to 1.2 are always backwards compatible.
+Upgrades that involve a major version bump often come with some backwards incompatible changes.
```
watson: I would probably add the following sentence before the two you just added:

> The Elastic APM Node.js Agent uses [semantic versioning](https://semver.org/), and therefore:

Is there a reason you say "often" and not "always" in the last sentence?

bmorelli25: Will do!

> Is there a reason you say "often" and not "always" in the last sentence?

Just trying to be safe. I can update to _always_ if you'd like.

bmorelli25:
The EOL table was incorrect. I've updated it to include minors

docs/upgrading.asciidoc
watson: Nah, it's a more soft wording to use "often" (though technically there's no reason to make a new major unless there's a breaking change, so technically there will always be one in that case - but I'm fine with "often")

watson:
If you rebase with `master`, the unrelated CI errors should now go away as they have been fixed in #1636

bmorelli25:
Everything passed! Can you review so I can merge?

PR #1660, opened 2020-02-27
chore: pin @types/node to 8.5.x

watson:
3.x of this module needs to be able to run on Node.js 8.6.0. By pinning @types/node to this version (there's no version 8.6.x), we ensure developers will not see APIs from newer versions that they can't use anyway.

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@commitlint/travis-cli": "^8.3.5",
     "@hapi/hapi": "^18.4.1",
     "@koa/router": "^8.0.8",
-    "@types/node": "^13.7.4",
+    "@types/node": "8.5.x",
     "apollo-server-express": "^2.10.1",
     "aws-sdk": "^2.622.0",
     "backport": "^4.9.0",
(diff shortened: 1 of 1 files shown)


watson:
@sqren hmm do you know what's going on here? Did the TypeScript syntax change between `@types/node@8.5.10` and now since [it's complaining](https://travis-ci.org/elastic/apm-agent-nodejs/jobs/655753924#L11404)?

package.json
```diff
@@ -115,7 +115,7 @@
     "@commitlint/travis-cli": "^8.3.5",
     "@hapi/hapi": "^18.4.1",
     "@koa/router": "^8.0.8",
-    "@types/node": "^13.7.4",
+    "@types/node": "8.5.x",
```
sqren: Shouldn't this be 8.5.10?

sqren:
```suggestion
    "@types/node": "~8.5.10",
```

watson: 8.5.x will currently resolve to 8.5.10

watson: Ah forgot about the tidle. Much better 👍

watson:
TypeScript 🙄

PR #1661, opened 2020-02-27
[2.x] chore: pin @types/node to 6.0.x

watson:
2.x of this module needs to be able to run on Node.js 6.0.0. By pinning @types/node to this version, we ensure developers will not see APIs from newer versions that they can't use anyway.


watson:
TypeScript 🙄

PR #1666, opened 2020-03-05
refactor!: drop support for Node 8

watson:
Due to how quickly the community is dropping support for Node.js 8, it gets harder and harder for us to maintain support for it. So let's get ready to release version 4.0.0 of the agent with removed support for Node.js 8.

Note however that we'll still have to support it in the `3.x` branch for the sake of security/bug fixes, but this also means that we can stop backporting to the `2.x` branch which means no longer having to deal with Node.js 6 😅


watson:

trentm:
We will do this at some point, but given all the current conflicts, I think it'll be easier to start over.

PR #1685, opened 2020-03-19
feat: enrich spans with destination info

### LINKED ISSUE #1684: add destination fields to all spans for Service Maps and SIEM
+ APM cross agent feature - Implement elastic/apm#180 and https://github.com/elastic/apm/issues/115

## Subtasks
+ [x] Destination info for HTTP Spans #1685 
+ [ ] Destination info for DB Spans

vigneshshanmugam:
+ part of #1684 
+ Add new destination fields to all spans for Service maps feature and SIEM. 
+ More details about the feature request is on https://github.com/elastic/apm/issues/115 and https://github.com/elastic/apm/issues/180

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


lib/instrumentation/modules/http2.js
```diff
@@ -5,6 +5,7 @@ var eos = require('end-of-stream')
 var shimmer = require('../shimmer')
 var symbols = require('../../symbols')
 var { parseUrl } = require('../../parsers')
+var { getDestination } = require('../context')
```
vigneshshanmugam: Not sure if its best to keep it on `context` file. I am not sure we follow any convention on the Node.js agent. I am open to suggestions.

vigneshshanmugam:
We should not need to worry about truncation for these fields if i understand correctly from here.. 
https://github.com/elastic/apm-nodejs-http-client/blob/master/lib/truncate.js#L128

lib/instrumentation/context.js
```diff
@@ -0,0 +1,43 @@
...
+
+// Get the port number including the default port for a protocols
+function getPortNumber (port, protocol) {
+  if (port === '') {
+    port = protocol === 'http:' ? '80' : protocol === 'https:' ? '443' : ''
+  }
+  return port
+}
+
+function getDestination (url, spanType) {
```
axw: Perhaps rename as `getURLDestination`, or `getHTTPDestination`? This is really only appropriate for http(s) spans.

vigneshshanmugam:
@axw Totally forgot about the `db` case, will create a new PR for that.

vigneshshanmugam:
@watson Can i have your review on this one?

lib/instrumentation/context.js
```diff
@@ -0,0 +1,43 @@
...
+      name: origin,
+      resource: hostname + ':' + portNumber,
+      type: spanType
+    },
+    address,
+    port: Number(portNumber)
+  }
+}
+
+module.exports = { getHTTPDestination }
```
watson, line 43: To enhance visibility into what's being exported from a module, I usually either move these export declarations up above the functions ore declare the functions directly on the existing `exports` object (e.g. `exports.getHTTPDestination = function (url, spanType) {...}`)

lib/instrumentation/context.js
```diff
@@ -0,0 +1,43 @@
...
+  // since address and port are recorded separately, we are recording the
+  // info in canonical form without square brackets
+  const ipv6Hostname =
+    hostname.charCodeAt(0) === LEFT_SQUARE_BRACKET &&
+    hostname.charCodeAt(hostname.length - 1) === RIGHT_SQUARE_BRACKET
+
+  let address = hostname
+  if (ipv6Hostname) {
+    address = hostname.slice(1, -1)
+  }
```
watson, line 30: To avoid re-setting the variable, you could write this as a ternary expression instead:

```suggestion
  const address = ipv6Hostname ? hostname.slice(1, -1) : hostname
```

lib/instrumentation/modules/http2.js
```diff
@@ -178,6 +179,8 @@ module.exports = function (http2, agent, { enabled }) {
           url
         })
 
+        span.setDestinationContext(getHTTPDestination(url, span.type))
```
watson, line 182: 🤔 I'm normally inclined to set stuff like this as early in the span lifecycle as possible. In this case just after the `url` variable as been created. That might help debugging later as the state of the span is as complete as possible as early as possible. On the other hand, this doesn't change the end result, so I'm 🤷‍♂️

lib/instrumentation/span.js
```diff
@@ -87,6 +88,11 @@ Span.prototype.setHttpContext = function (context) {
   this._http = Object.assign(this._http || {}, context)
 }
 
+Span.prototype.setDestinationContext = function (context) {
+  if (!context) return
```
watson, line 92: What's the reason for this guard? As far as I can see based on the current code, `context` should always be set, right?

test/instrumentation/context.js
```diff
@@ -0,0 +1,62 @@
...
+        type: 'external'
+      },
+      address: 'testing.local',
+      port: 1234
+    })
+    t.end()
+  })
+
+  t.test('https default port', (t) => {
+    const url = 'https://www.elastic.co:443/products/apm'
```
watson, line 22: Shouldn't this url not have a port if we were to test the default port fallback?

```suggestion
    const url = 'https://www.elastic.co/products/apm'
```

test/instrumentation/context.js
```diff
@@ -0,0 +1,62 @@
...
+        resource: 'testing.local:1234',
+        type: 'external'
+      },
+      address: 'testing.local',
+      port: 1234
+    })
+    t.end()
+  })
+
+  t.test('https default port', (t) => {
```
watson, line 21: Probably a good idea to have a test for regular http we all

test/instrumentation/context.js
```diff
@@ -0,0 +1,62 @@
...
+        resource: '[::1]:80',
+        type: 'external'
+      },
+      address: '::1',
+      port: 80
+    })
+    t.end()
+  })
+
+  t.test('ipv6 default scheme and port', (t) => {
```
watson, line 49: What do you mean by default scheme?

lib/instrumentation/context.js
```diff
@@ -0,0 +1,43 @@
...
+function getHTTPDestination (url, spanType) {
+  const { port, protocol, hostname, origin } = parseUrl(url)
+  const portNumber = getPortNumber(port, protocol)
+
+  // If hostname begins with [ and ends with ], assume that it's an IPv6 address.
+  // since address and port are recorded separately, we are recording the
+  // info in canonical form without square brackets
+  const ipv6Hostname =
+    hostname.charCodeAt(0) === LEFT_SQUARE_BRACKET &&
+    hostname.charCodeAt(hostname.length - 1) === RIGHT_SQUARE_BRACKET
```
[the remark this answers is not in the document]
vigneshshanmugam: Just followed the convention from Node internals. Sure I can update that.

lib/instrumentation/modules/http2.js
vigneshshanmugam: Dont have any strong preference either. Let me know what you think. I can update accordingly. Just followed the similarity with http context.

test/instrumentation/context.js
vigneshshanmugam: meant it as URI scheme

test/instrumentation/context.js
vigneshshanmugam: Yeah, I will add a test with and without default port to show the diff.

vigneshshanmugam:
@watson can i get another round of eyes on this?

vigneshshanmugam:
Merging this, Will incorporate changes if necessary on the other PR.

PR #1687, opened 2020-03-20
feat: add destination metadata for db spans

vigneshshanmugam:
+ fix https://github.com/elastic/apm-agent-nodejs/issues/1684
+ Add destination information to all database connectionss. 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


lib/instrumentation/modules/memcached.js
```diff
@@ -32,6 +50,8 @@ module.exports = function (memcached, agent, { version, enabled }) {
             arguments[0] = function queryCompiler () {
               return query
             }
+            // Hack to ensure the span is propagated correctly
+            agent._instrumentation.activeSpan = span
```
vigneshshanmugam, line 54: @watson  except for this change, I am satisfied about other changes. 

Without this change, activeSpan would always point to the previous span. Had to update it to add the destination info with correct span. Let me know if i am missing something here or any other way we can trigger the update.

lib/instrumentation/modules/memcached.js
```diff
@@ -13,17 +14,34 @@ module.exports = function (memcached, agent, { version, enabled }) {
...
   agent.logger.debug('shimming memcached.prototype.command')
+  const [type, subtype] = ['db', 'memcached']
+
   shimmer.wrap(memcached.prototype, 'command', wrapCommand)
+  shimmer.wrap(memcached.prototype, 'connect', wrapConnect)
   return memcached
 
+  function wrapConnect (original) {
+    return function wrappedConnect () {
+      const activeSpan = agent._instrumentation.activeSpan
```
watson, line 25: I think what you really want here is `currentSpan` and not `activeSpan`.

watson: Come to think of it, that might actually explain why you have to do the "hack" down below

vigneshshanmugam: Yeah, that totally did the job. Thanks for pointing it out.

lib/instrumentation/context.js
```diff
@@ -33,3 +33,23 @@ exports.getHTTPDestination = function (url, spanType) {
     port: Number(portNumber)
   }
 }
+
+exports.getDBDestination = function (type, subType, host, port) {
+  const destination = {
+    service: {
+      name: subType,
+      resource: subType,
+      type
```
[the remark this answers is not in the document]
vigneshshanmugam: Yeah I like this approach, seems cleaner to me. Don't know why i went this path in first place. Will address it and also remove the array assignment as its not necessary.

test/instrumentation/modules/tedious.js
```diff
@@ -19,11 +19,12 @@ const mockClient = require('../../_mock_http_client')
 const version = require('tedious/package').version
 
 let connection
+const hostname = process.env.APPVEYOR ? 'localhost' : (process.env.MSSQL_HOST || '127.0.0.1')
```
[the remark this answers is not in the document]
vigneshshanmugam: Took the liberty of cleaning up the code altogether and removed the dead code.

lib/instrumentation/context.js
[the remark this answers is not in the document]
vigneshshanmugam: Done.

lib/instrumentation/modules/tedious.js
```diff
@@ -63,7 +65,8 @@ module.exports = function (tedious, agent, { version, enabled }) {
       }
 
       makeRequest (request) {
-        const span = agent.startSpan(null, 'db', 'mssql', 'query')
+        const [type, subType] = ['db', 'mssql']
```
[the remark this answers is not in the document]
vigneshshanmugam: Hah, I went over a way ahead on that one 😅. I have not evaluate the performance of the above code but as you have pointed out there might be a hidden cost to allocating an array and accessing it. I will post some details on the bench soonish.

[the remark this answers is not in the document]
vigneshshanmugam: You are right, Array version as expected was slower 
Screenshot 2020-05-14 at 23 05 30

I have removed those code branches. so we should be good here.

lib/instrumentation/modules/memcached.js
```diff
@@ -14,16 +15,31 @@ module.exports = function (memcached, agent, { version, enabled }) {
...
+
   // Wrap the generic command that is used to build touch, get, gets etc
   function wrapCommand (original) {
     return function wrappedCommand () {
       if (typeof arguments[0] === 'function') {
         var query = arguments[0]()
         // If the callback is not a function the user doesn't care about result
         if (query && typeof query.callback === 'function') {
-          var span = agent.startSpan(`memcached.${query.type}`, `db.memcached.${query.type}`)
+          var span = agent.startSpan(`memcached.${query.type}`, 'db', 'memcached', `${query.type}`)
```
watson, line 42: No need for the double wrapped string right? Or is this in case it's not a string?

```suggestion
          var span = agent.startSpan(`memcached.${query.type}`, 'db', 'memcached', query.type)
```

vigneshshanmugam: Yes indeed 👍

PR #1694, opened 2020-03-27
ci(jenkins): disable stages for only docs

v1v:
### What

- Any builds for any branch or PR that only contains changes related to the asciidoc then it will skip all the stages but the `checkout`

### Why

Docs validation happens with the `elasticsearch-ci/docs` therefore there is no need to validate anything in our end.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -54,8 +54,8 @@ pipeline {
             ]
             env.TAV_UPDATED = isGitRegionMatch(patterns: regexps)
 
-            // Skip all the stages except docs for PR's with asciidoc changes only
-            env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.asciidoc' ], comparator: 'regexp', shouldMatchAll: true)
+            // Skip all the stages except docs for PR's with asciidoc or md changes only
+            env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.(asciidoc|md)' ], shouldMatchAll: true)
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -53,6 +53,9 @@ pipeline {
               "^test/instrumentation/modules/"
             ]
             env.TAV_UPDATED = isGitRegionMatch(patterns: regexps)
+
+            // Skip all the stages except docs for PR's with asciidoc changes only
+            env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.asciidoc' ], comparator: 'regexp', shouldMatchAll: true)
```
watson: We should include `.md` as well

v1v: are those files hosted in the same folder?

watson: No, all the markdown files are currently in the root of the repo (where there's also one asciidoc file). The rest of the asciidoc files are all in the `docs` directory). I thought this pattern check was for all files no matter where in the repo they were located?

watson: Come to think of it, for markdown changes, there's no reason to run any CI checks at all (besides to check if the committer have signed the CLA). For asciidoc changes, we should still run the `elasticsearch/docs` step

v1v: got it, I'll then use a pattern like `'.*\\.(asciidoc|md)'` then all the changes in the PR should be markdown or asciidoc files.

PR #1702, opened 2020-03-31
ci(jenkins): use nodejs lts in the windows stage

v1v:
Use the chocolatey nodejs-lts to see if the 12.16.1 version is solved there.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update TypeScript typings
- [ ] Update documentation
- [ ] Commit message follows [commit guidelines](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#commit-message-guidelines)

### Tests

```
[2020-03-31T12:47:34.649Z] Getting latest nodejs-lts version for 12 ...
[2020-03-31T12:47:37.975Z] 
[2020-03-31T12:47:37.975Z] nodejs-lts 12.16.1 [Approved]
[2020-03-31T12:47:37.975Z] nodejs-lts 12.16.0 [Approved]
[2020-03-31T12:47:37.975Z] nodejs-lts 12.15.0 [Approved]
[2020-03-31T12:47:37.975Z] nodejs-lts 12.14.1 [Approved]
[2020-03-31T12:47:37.975Z] nodejs-lts 12.14.0 [Approved]
[2020-03-31T12:47:37.975Z] nodejs-lts 12.13.1 [Approved]
[2020-03-31T12:47:37.975Z] nodejs-lts 12.13.0 [Approved]
[2020-03-31T12:47:42.187Z] Installing nodejs-lts version: 12.16.1 ...
[2020-03-31T12:47:42.187Z] Chocolatey v0.10.11
[2020-03-31T12:47:42.449Z] Installing the following packages:
[2020-03-31T12:47:42.449Z] nodejs-lts
[2020-03-31T12:47:42.449Z] By installing you accept licenses for the packages.
[2020-03-31T12:47:46.660Z] 
[2020-03-31T12:47:46.661Z] nodejs-lts v12.16.1 [Approved]
[2020-03-31T12:47:46.661Z] nodejs-lts package files install completed. Performing other installation steps.
[2020-03-31T12:47:51.963Z] Installing 64 bit version
[2020-03-31T12:47:51.963Z] Installing nodejs-lts...
[2020-03-31T12:48:13.919Z] nodejs-lts has been installed.
[2020-03-31T12:48:13.919Z]   nodejs-lts may be able to be automatically uninstalled.
[2020-03-31T12:48:13.919Z] Environment Vars (like PATH) have changed. Close/reopen your shell to
[...]

PR #1704, opened 2020-03-31
[3.x] ci(jenkins): use nodejs lts in the windows stage (#1702) (7b933627)

kuisathaverat:
Backports the following commits to 3.x:
 - ci(jenkins): use nodejs lts in the windows stage (#1702) (7b933627)


watson:
FYI: We normally shouldn't backport into the `3.x` branch as it's "autogenerated" every time we make a new release. So any commits that are added there will be overwritten at release-time. However, since this commit is also in `master` it will be functionally equivalent. The only downside is that we'll have to force-push and so history will be re-written, which at release time will make the person making the release unsure about what's going on.

PR #1714, opened 2020-04-10
Fully-automated release process

cachedout:
Hi all,

I've been playing around with various ways to automate the release process and I am opening this PR as a point of discussion to clarify a few things before we move forward.

## Background

To begin, let's look at the current release process:

 1. Be sure you have checked out the master branch and have pulled latest changes
 1. Update the version in package.json according to the scale of the change. (major, minor or patch)
 1. Add commit messages to CHANGELOG.asciidoc (You may skip non-user-visible changes)
 1. If a major or minor release, update the EOL table in docs/upgrading.asciidoc. EOL is 18 months after release date.
 1. Commit changes with message x.y.z where x.y.z is the version in package.json
 1. Tag the commit with git tag vx.y.x, for example git tag v1.2.3
 1. Reset the latest major branch (1.x, 2.x etc) to point to the current master, e.g. git branch -f 3.x master
 1. Run tests with npm test
 1. Push commits and tags upstream with git push upstream master && git push upstream --tags (and optionally to your own fork as well)
 1. Update the latest major branch on upstream with git push upstream <major_branch>
 1. Publish to npm with npm publish

(There is a second process which is slightly different for minor versions but I'm making it out of scope for this discussion and we'll come back to it after decisions are made here.)

## Proposal and discussion

I propose that we make a couple of high-level decisions.

1. Steps 1-6 should continue to be performed by a developer prior to the release process starting. The reason this seems prudent to me is that we want to start the release process from a git tag. Although we could easily automate step 2 and maybe even 3 and 4, we'd still want those changes to be a part of a tag.

2. Do we want to automate step 7? I personally tend to be leery of CI systems pushing code but we can do this if we want to.

3. For step 8, the way this is written right now, the release step can run after tests run. So, this would be implied for a normal CI-based release.

4. Step 9 becomes irrelevant because it is assumed to have been done prior to the CI job being started.

5. That pretty much leaves step 11 as the step we'd be left automating. I've taken a high-level stab at that in this PR, using work that @v1v has already done for the rumjs project as a template. I will mark areas in-line where I think there are points to discuss further.

6. Instead of being tag-based, I set this to require that a specific box is ticked in the CI before the release stage is run. We could go back to tag-based if that's easier for the development team. From the CI perspective, it doesn't matter much to us so let us know what you'd prefer.

Let me know your thoughts. Thanks!


.ci/Jenkinsfile
```diff
@@ -281,32 +282,69 @@ pipeline {
       }
     }
     stage('Release') {
-      options { skipDefaultCheckout() }
+      options {
+        skipDefaultCheckout()
+        timeout(time: 12, unit: 'HOURS')
```
cachedout, line 287: This might be unneeded. We can remove it or shorten it.

.ci/Jenkinsfile
```diff
@@ -392,6 +430,21 @@ def generateStep(Map params = [:]){
   }
 }
 
+def prepareRelease(String nodeVersion='node:lts', Closure body){
+  withNpmrc(secret: "${env.NPMRC_SECRET}", path: "${env.WORKSPACE}/${env.BASE_DIR}") {
+    withCredentials([usernamePassword(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken',
```
cachedout, line 435: @v1v I need some direction in getting the right token here.

.ci/Jenkinsfile
```diff
@@ -392,6 +430,21 @@ def generateStep(Map params = [:]){
...
+                                      passwordVariable: 'GITHUB_TOKEN', usernameVariable: 'GITHUB_USER')]) {
+      docker.image(nodeVersion).inside(){
+        withEnv(["HOME=${env.WORKSPACE}/${env.BASE_DIR}"]) {
+          withTotpVault(secret: "${env.TOTP_SECRET}", code_var_name: 'TOTP_CODE'){
```
cachedout, line 439: I don't know if we need to go to the same degree as we did with rum-js here. I assume this is needed for npm in this case as well but I'm not positive.

.ci/Jenkinsfile
v1v: This will help to run the release with an input approval and enough time. IMO, If no input approval then it's unnecessary

.ci/Jenkinsfile
v1v: I think that's the correct one. There is only one token for the `apm` user, I'd rather keep the name obfuscated to avoid any disclosure.

.ci/Jenkinsfile
v1v: I'm not sure how the npm publish works. In Lerna we were able to use the environment variable `TOTP_CODE` in conjunction with the flag `--otp=`.

What are your thoughts?

.ci/Jenkinsfile
```diff
@@ -281,32 +282,72 @@ pipeline {
...
-        stage('Opbeans') {
-          environment {
-            REPO_NAME = "${OPBEANS_REPO}"
+      stages {    
+        stage('Release CI') {
+          options {
+            skipDefaultCheckout()
           }
-          steps {
+
```
v1v, line 304: NIT, cosmetic line

.ci/Jenkinsfile
```diff
@@ -281,32 +282,72 @@ pipeline {
       }
     }
     stage('Release') {
-      options { skipDefaultCheckout() }
+      options {
+        skipDefaultCheckout()
+        timeout(time: 12, unit: 'HOURS')
+        }
```
v1v, line 288: NIT
```suggestion
      }
```

.ci/Jenkinsfile
```diff
@@ -281,32 +282,72 @@ pipeline {
...
+                currentBuild.description = "${currentBuild.description?.trim() ? currentBuild.description : ''} released"
+              }
             }
           }
         }
       }
     }
+
+    stage('Opbeans') {
+
```
v1v, line 326: NIT, cosmetic line
```suggestion
```

.ci/Jenkinsfile
```diff
@@ -281,32 +282,72 @@ pipeline {
...
+      options { skipDefaultCheckout() }
+      when {
+        beforeAgent true
+        tag pattern: '@elastic/apm-rum@\\d+\\.\\d+\\.\\d+$', comparator: 'REGEXP'
+      }
+
```
v1v, line 332: NIT: cosmetic line

```suggestion
```

.ci/Jenkinsfile
```diff
@@ -281,32 +282,72 @@ pipeline {
...
+          // It's required to transform the tag value to the artifact version
+          sh script: ".ci/bump-version.sh ${env.BRANCH_NAME.replaceAll('^v', '')}", label: 'Bump version'
+          // The opbeans pipeline will trigger a release for the master branch
+          gitPush()
+          // The opbeans pipeline will trigger a release for the release tag
+          gitCreateTag(tag: "${env.BRANCH_NAME}")
+        }
+      }
+    }
+
```
v1v, line 350: NIT: cosmetic line

```suggestion
```

.ci/Jenkinsfile
```diff
@@ -281,32 +282,72 @@ pipeline {
...
+        allOf {
+          branch 'master'
+          expression { return params.release }
+        }
       }
-      stages {
-        stage('Opbeans') {
-          environment {
-            REPO_NAME = "${OPBEANS_REPO}"
+      stages {
```
v1v, line 299: NIT: cosmetic space
```suggestion
      stages {
```

.ci/Jenkinsfile
```diff
@@ -392,6 +433,21 @@ def generateStep(Map params = [:]){
   }
 }
 
+def prepareRelease(String nodeVersion='node:lts', Closure body){
+  withNpmrc(secret: "${env.NPMRC_SECRET}", path: "${env.WORKSPACE}/${env.BASE_DIR}") {
```
v1v, line 437: env variables are missing in the global environments

.ci/Jenkinsfile
```diff
@@ -392,6 +433,21 @@ def generateStep(Map params = [:]){
   }
 }
 
+def prepareRelease(String nodeVersion='node:lts', Closure body){
```
v1v, line 436: We might need to double check if `node:lts` can be used for the release or it's required some other version.

.ci/Jenkinsfile
cachedout: OK, cool. Wasn't sure what was what but this sounds like it should work. Thanks!

.ci/Jenkinsfile
cachedout: Thanks. Removed.

watson:
Because this project [uses the Conventional Commits standard](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#commit-message-guidelines), we can automate releases based on the commit message of each commit landing in a release branch. This will automate step 1-6. Is there any reason why you don't want to do that? (for inspiration, see the [semantic-release](https://github.com/semantic-release/semantic-release) project).

.ci/Jenkinsfile
cachedout: @watson Can you give us insight on this point? What are the requirements for the version of node that we use when running `npm publish`? Is using LTS your recommendation?

.ci/Jenkinsfile
cachedout: It doesn't _seem_ like this would be required unless 2FA is required via a configuration flag. I don't see that being the case so far as I can tell, but I would like to confirm with @watson whether or not this is typically a part of the release workflow since the [current release documentation does not make any mention of it](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#releasing).

cachedout:
> This will automate step 1-6. Is there any reason why you don't want to do that

No reason from an automation perspective that this shouldn't be feasible. I will start writing scripts to manage these steps. Thanks for the feedback!

.ci/Jenkinsfile
watson: It doesn't really matter much. LTS is fine 👍

cachedout: Sounds good. We'll stick with that for now. It's easy to change in the future. :) Thanks!

cachedout:
@v1v The changelog automation script has been added. If you want to re-use it, we can pull it out of here and find another home for it. Let me know what you think.

v1v:
> @v1v The changelog automation script has been added. If you want to re-use it, we can pull it out of here and find another home for it. Let me know what you think.

I think for other cases we have created a docker image, we could potentially use a similar approach, what do you think? Maybe a follow up to be refactored if we don't wanna block this PR

.ci/scripts/generate_cc_changelog/generate.py
```diff
@@ -0,0 +1,346 @@
...
+    pr : github.PullRequest
+
+    Returns
+    -------
+    bool
+        True if the pull request is one that we consider as being appropriate
+        for the changelog. See the implementation for details on how this
+        selection process operates.
+    """
+    if re.match("fix(\(.*\)){0,1}:|\
```
v1v, line 192: I see there are more types defined in https://github.com/conventional-changelog/commitlint , is this intended to be the only ones to be searched?

cachedout:
> I think for other cases we have created a docker image, we could potentially use a similar approach, what do you think? Maybe a follow up to be refactored if we don't wanna block this PR

It is designed to be containerized and a Dockerfile is included. If we wanted, we could pre-build the image but it's so lightweight that it shouldn't be too much trouble to build it on-the-fly during the pipeline execution.

.ci/scripts/generate_cc_changelog/generate.py
cachedout: This is drawn from the list of previous changelog entries for this project. I had to speculate about what is (and isn't) included. Perhaps @watson could shed some more light on what kinds of changelogs entries they wish to include and exclude?

.ci/scripts/generate_cc_changelog/templates/entry.jinja2
```diff
@@ -0,0 +1,7 @@
+[[release-notes-{{release_number}}]]
+==== {{release_number}} - {{today}}
+{% for pr in prs %}
+* {{ pr.title -}}
+{%-endfor %}
+
+
```
cachedout, line 7: These newlines are necessary to conform with the current spacing in the file.

cachedout:
> Build with the release checkbox, It should only run a minimal set of stages, as it is exactly the same commit than the above build but with a release stage.

I built this with the ability to just run a single release pipeline from start to finish. It seems OK, but let's talk more about this if you think we should go to multiple runs like you suggest.

.ci/Jenkinsfile
```diff
@@ -282,31 +288,90 @@ pipeline {
...
-              gitPush()
-              // The opbeans pipeline will trigger a release for the release tag
-              gitCreateTag(tag: "${env.BRANCH_NAME}")
+            unstash 'source'
+            // Bail out if we don't have the right parameters
+            isValidReleaseVer("${params.release_ver}")
+            // Prepare our release toolchain
+            sh(script: ".ci/scripts/prepare-release-containers.sh")
+            // Prepare GitHub environment
+            githubEnv()
```
v1v, line 314: This is already called internally with the gitcheckout step
```suggestion
```

.ci/Jenkinsfile
```diff
@@ -282,31 +288,90 @@ pipeline {
...
+              withCredentials([string(credentialsId: "${credentialsId}", variable: 'GITHUB_TOKEN')]) {
+                def preview = sh(script: "./.ci/scripts/generate_changelog.sh --token ${GITHUB_TOKEN} --version ${params.release_ver} --preview true", returnStdout: true)
+              }
+              input(message: 'Do you accept these proposed entries for the CHANGELOG.asciidoc file?', ok: 'Yes, proceed with writing them.',
+                    parameters: [text(description: 'IMPORTANT: Proposed changes CANNOT be edited here. Abort the build if not acceptable.',
+                    defaultValue: "${preview}", name: 'preview')])
+              withCredentials([string(credentialsId: "${credentialsId}", variable: 'GITHUB_TOKEN')]) {
+                // Re-run without preview flag and rewrite the file
+                sh(script: "./.ci/scripts/generate_changelog.sh --token ${GITHUB_TOKEN} --version ${params.release_ver} --preview false > ./CHANGELOG.asciidoc", returnStdout: true)
+              }
```
v1v, line 327: is `params.get('credentialsId', '2a9602aa-ab9f-4e52-baf3-b71ca88469c7')` configured?, what do you think to use the hardcoded credentials instead and reuse the withCredentials clousure?

```suggestion
              withCredentials([string(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7', variable: 'GITHUB_TOKEN')]) {
                def preview = sh(script: "./.ci/scripts/generate_changelog.sh --token ${GITHUB_TOKEN} --version ${params.release_ver} --preview true", returnStdout: true)
                input(message: 'Do you accept these proposed entries for the CHANGELOG.asciidoc file?', ok: 'Yes, proceed with writing them.',
                      parameters: [text(description: 'IMPORTANT: Proposed changes CANNOT be edited here. Abort the build if not acceptable.',
                      defaultValue: "${preview}", name: 'preview')])
                // Re-run without preview flag and rewrite the file
                sh(script: "./.ci/scripts/generate_changelog.sh --token ${GITHUB_TOKEN} --version ${params.release_ver} --preview false > ./CHANGELOG.asciidoc", returnStdout: true)
              }
```

.ci/Jenkinsfile
```diff
@@ -282,31 +288,90 @@ pipeline {
...
+            }
+            // Update the EOL table in the upgrading.asciidoc file
+            sh(script: "./.ci/scripts/generate_eol.sh --version ${params.release_ver} > ./docs/upgrading.asciidoc")
+            // Update package.json
+            sh ".ci/scripts/update_package_version.sh ${params.release_ver}"
```
v1v, line 332: NIT: use the label with the comment:
```suggestion
            sh(label: 'Update package.json', script: ".ci/scripts/update_package_version.sh ${params.release_ver}")
```

.ci/Jenkinsfile
```diff
@@ -282,31 +288,90 @@ pipeline {
...
+            // Tag and push
+            sh (script: "./.ci/scripts/tag_release.sh --version ${params.release_ver}")
```
v1v, line 334: NIT and suggestion:
- add docs to the label, 
- if exec privileges in the script then it's not required `./`

```suggestion
            sh(label: 'Tag and push', script: ".ci/scripts/tag_release.sh --version ${params.release_ver}")
```

.ci/Jenkinsfile
```diff
@@ -282,31 +288,90 @@ pipeline {
...
+              input(message: 'Do you accept these proposed entries for the CHANGELOG.asciidoc file?', ok: 'Yes, proceed with writing them.',
+                    parameters: [text(description: 'IMPORTANT: Proposed changes CANNOT be edited here. Abort the build if not acceptable.',
+                    defaultValue: "${preview}", name: 'preview')])
+              withCredentials([string(credentialsId: "${credentialsId}", variable: 'GITHUB_TOKEN')]) {
+                // Re-run without preview flag and rewrite the file
+                sh(script: "./.ci/scripts/generate_changelog.sh --token ${GITHUB_TOKEN} --version ${params.release_ver} --preview false > ./CHANGELOG.asciidoc", returnStdout: true)
+              }
+            }
+            // Update the EOL table in the upgrading.asciidoc file
+            sh(script: "./.ci/scripts/generate_eol.sh --version ${params.release_ver} > ./docs/upgrading.asciidoc")
```
v1v, line 330: NIT and suggestion:
- add docs to the label, 
- if exec privileges in the script then it's not required `./`

```suggestion
            sh(label: 'Update EOL table', script: ".ci/scripts/generate_eol.sh --version ${params.release_ver} > ./docs/upgrading.asciidoc")
```

.ci/Jenkinsfile
```diff
@@ -282,31 +288,90 @@ pipeline {
...
-              sh script: ".ci/bump-version.sh ${env.BRANCH_NAME.replaceAll('^v', '')}", label: 'Bump version'
-              // The opbeans pipeline will trigger a release for the master branch
-              gitPush()
-              // The opbeans pipeline will trigger a release for the release tag
-              gitCreateTag(tag: "${env.BRANCH_NAME}")
+            unstash 'source'
+            // Bail out if we don't have the right parameters
+            isValidReleaseVer("${params.release_ver}")
+            // Prepare our release toolchain
+            sh(script: ".ci/scripts/prepare-release-containers.sh")
```
v1v, line 312: NIT:
```suggestion
            sh((label: 'Prepare release toolchain', script: '.ci/scripts/prepare-release-containers.sh')
```

.ci/Jenkinsfile
```diff
@@ -502,3 +577,9 @@ def generateStepForWindows(Map params = [:]){
     sh label: 'Stop services', script: ".ci/scripts/windows/stop-test.sh ${version}"
   }
 }
+
+def isValidReleaseVer(String ver){
+    if (ver == "0.0.0") {
+        error("Must supply version number to release")
+    }
```
v1v, line 584: The version could be a typo, if semversion in the format, what about using a few more validations:
- no empty values
- no default value
- no sem version format (major.minor.patch)

NOTE: I have not tested the below snippet


```suggestion
    if (!ver?.trim() || ver == "0.0.0" || !(ver ==~ /(\d+)\.(\d+).(\d+)/) ) {
        error("Must supply version number to release")
    }
```

.ci/scripts/generate_eol.sh
```diff
@@ -0,0 +1,41 @@
...
+    ;;
+    *)    # unknown option
+    POSITIONAL+=("$1") # save it in an array for later
+    shift # past argument
+    ;;
+esac
+done
+
+docker run \
+-v "$(pwd)":${DOCKER_MOUNT_POINT} \
```
v1v, line 38: is it required to run with the flags `-t` or `-ti`?

.ci/scripts/generate_changelog.sh
```diff
@@ -0,0 +1,65 @@
...
+    ;;
+    *)    # unknown option
+    POSITIONAL+=("$1") # save it in an array for later
+    shift # past argument
+    ;;
+esac
+done
+
+docker run \
+-v "$(pwd)":${DOCKER_MOUNT_POINT} \
```
v1v, line 58: ditto with the `-t` or `-ti` flag?

.ci/scripts/tag_release.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
```
v1v, line 2: In order to use this particular script is required to configure the workspace with the git context as the current git context is a git detached commit, for such see:
- https://github.com/elastic/apm-agent-rum-js/blob/master/.ci/scripts/prepare-git-context.sh

v1v:
Regarding the 2FA, I think this project has been configured with 2FA in the npmjs registry, and likely the user defined in the env variable `NPMRC_SECRET`, @watson , can you confirm when you released this project in the past, were you forced to use the 2FA?

.ci/scripts/generate_changelog.sh
cachedout: The `-t` flag is added [on L59.](https://github.com/elastic/apm-agent-nodejs/pull/1714/files/f7171b04a30361f5101edb2ff3c8d0fad190ea48#diff-586c5516b74e314fe2a3d5ab56496ba8R59)

.ci/scripts/generate_eol.sh
cachedout: The `-t` flag is appended [on L39](https://github.com/elastic/apm-agent-nodejs/pull/1714/files/f7171b04a30361f5101edb2ff3c8d0fad190ea48#diff-ea60c515fa2f2c7eefd57bb0912f4491R39)

v1v: I meant `--tty `, what's `${TAG}` required for?

.ci/scripts/generate_changelog.sh
v1v: I meant `--tty `, what's `${TAG}` required for?

.ci/scripts/tag_release.sh
cachedout: Ah, OK. Interesting. I'll borrow from that script and add it to the top of this one. Good catch here. I wasn't aware of that.

.ci/scripts/generate_changelog.sh
v1v: aha! I said nothing, for some reason I got confused with the `docker build -t ${TAG}`, could we rename the TAG with `IMAGE`?

.ci/scripts/generate_changelog.sh
```diff
@@ -0,0 +1,65 @@
...
+# The mount point internal to the Docker container where we want to operate on this repo
+DOCKER_MOUNT_POINT=/src
+
+# The name of the CHANGELOG file to search for relative to the mount point
+CHANGELOG=CHANGELOG.asciidoc
+
+# The repo to query for entries to be made to the log
+REPO=elastic/apm-agent-nodejs
+
+# The tag designating the container that was built from the `generate_cc_changelog` project
```
v1v, line 22:
```suggestion
# The tag designating the docker image that was built from the `generate_cc_changelog` project
```

.ci/scripts/generate_eol.sh
```diff
@@ -0,0 +1,41 @@
...
+# This is a wrapper script to encapsulate calls like the following:
+# docker run -v \"$(pwd)\":/src -t update_eol_doc --release {$env.RELEASE_VER} --doc /src/docs/upgrading.asciidoc > ./docs/upgrading.asciidoc"
+
+# The mount point internal to the Docker container where we want to operate on this repo
+DOCKER_MOUNT_POINT=/src
+
+# The upgrading file, relative to the root of the repo.
+UPGRADING=/docs/upgrading.asciidoc
+
+# The tag designating the container that was built from the `generate_cc_changelog` project
```
v1v, line 12:
```suggestion
# The tag designating the docker image that was built from the `generate_cc_changelog` project
```

.ci/scripts/generate_eol.sh
v1v: see https://github.com/elastic/apm-agent-nodejs/pull/1714/files/f7171b04a30361f5101edb2ff3c8d0fad190ea48#r411966586

.ci/scripts/tag_release.sh
cachedout: @v1v Could you please take a look at my latest commit and see if that lines up with how your script expects to operate?

.ci/Jenkinsfile
watson: I'd limit the string start and end using `^` and `$` in regular expressions. For the official semver regex, see https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string - but this is if you also want to support the build metadata, which I'm not sure if the automated release needs

.ci/scripts/generate_cc_changelog/generate.py
watson: We tend to only put user-facing commits in the changelog because we had the philosophy that if we put all commits there, we might as well just link to the commit log instead.

User-facing commits are always `feat`, `fix`, and `perf` and sometimes `chore` and `refactor`. But never `test`, `ci`, or `docs`. Previously we've had the luxury of being able to cherry-pick which of the `chore` and `refactor` commits to include, but we can't do that if we automate the process. There's a ton of `chore` commits in most releases that are just noise - e.g. when dev-dependency packages are upgraded. It would be sad to include those. So I propose the following:

1. We always include `feat`, `fix`, and `perf`
1. We always include any commit that contains a breaking change, no matter the type.

Rule 2 would mean that `refactor!: drop support for Node 6` would be included in the changelog even though `refactor` normally isn't. I think this would solve our problems.

Speaking of breaking changes: Recently Conventional Commits released v1.0.0 of their spec, which added the ability to denote breaking changes with an exclamation mark (`!`) before the colon, e.g. `refactor!: drop support for Node 6`. So there are two ways to mark a commit as containing a breaking change:

1. Using `!` in the first line (see [rule 1](https://www.conventionalcommits.org/en/v1.0.0/#specification))
1. Using `BREAKING CHANGE: ...` in the footer (see [rule 12](https://www.conventionalcommits.org/en/v1.0.0/#specification))

.gitignore
```diff
@@ -13,3 +13,6 @@ node_modules
 test/benchmarks/.tmp
 coverage.lcov
 test-suite-output.tap
+.pytest_cache/
+**/__pycache__/
+*.pyc
```
watson, line 18: We have previously put files and folders in separate sections in this file. If `*.pyc` is a file, it should go up into the file-section above.

I don't have any hard feelings whether or not we should continue this practice, but until/if we change it, we should probably continue it 🙂

.ci/scripts/generate_changelog.sh
cachedout: Sure thing. Done. :)

.ci/scripts/update_package_version.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -exo pipefail
+
+echo "Now updating package.json for version: ${1}"
+
+# This is a 3-stage process because jq seems to stream data and
+# will create an empty file if the write is not done atomically
+mydir=$(mktemp -d "${TMPDIR:-/tmp/}$(basename $0).XXXXXXXXXXXX")
+jq '.version = '\"${1}\" package.json > $mydir/package.json
```
watson, line 9: Will this keep the layout of the rest of the `package.json` file intact or will it be completely re-written in a new style?

.ci/scripts/generate_eol.sh
cachedout: I've removed the `-t` which is unneeded and changed to using `IMAGE` as the variable name to make the meaning more clear.

watson:
Does this also create a proper release on the [GitHub Releases page](https://github.com/elastic/apm-agent-nodejs/releases)? That would be super useful for users who are subscribed to GitHub releases, where GitHub will send you an e-mail whenever a project whos releases you watch make a new release. The e-mail will then contain a list of changes in that release.

.gitignore
cachedout: No problem at all. :) I'll change it. Thanks for letting me know.

.ci/scripts/update_package_version.sh
cachedout: The file is entirely re-written using the [standard output defaults of jq](https://stedolan.github.io/jq/). If you have a particular style you would like to adhere to, we can try to get `jq` to adhere to it. If we can't get that to happen, we can try other approaches.

watson: TL;DR: I just did a test with my version of `jq` and it seems like it doesn't rewrite anything else than what we ask it to so all is good 👍👍👍 ☺️

This, of course, isn't a deal-breaker, but there's a defacto standard in the Node.js community to have a certain order of the important fields in `package.json`. From a technical perspective, it doesn't matter, but if a user opens the file, it makes the readability a bit harder if it doesn't keep the current structure (e.g. first field is always `name`, 2nd is `version` etc.).

.ci/Jenkinsfile
```diff
@@ -282,31 +288,89 @@ pipeline {
...
+                sh(script: "./.ci/scripts/generate_changelog.sh --token ${GITHUB_TOKEN} --version ${params.release_ver} --preview false > ./CHANGELOG.asciidoc", returnStdout: true)
+              }
+            }
+            // Update the EOL table in the upgrading.asciidoc file
+            sh(label: 'Update EOL table', script: ".ci/scripts/generate_eol.sh --version ${params.release_ver} > ./docs/upgrading.asciidoc")
+            sh(label: 'Update package.json', script: ".ci/scripts/update_package_version.sh ${params.release_ver}")
+                withCredentials([usernamePassword(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken',
+                                      passwordVariable: 'GITHUB_TOKEN', usernameVariable: 'GITHUB_USER')]) {
+                                        sh(label: 'Tag and push', script: ".ci/scripts/tag_release.sh --version ${params.release_ver}")
+                                      }
```
v1v, line 332: NIT: cosmetic ident

```suggestion
            withCredentials([usernamePassword(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken',
                                         passwordVariable: 'GITHUB_TOKEN', usernameVariable: 'GITHUB_USER')]) {
              sh(label: 'Tag and push', script: ".ci/scripts/tag_release.sh --version ${params.release_ver}")
            }
```

.ci/Jenkinsfile
```diff
@@ -282,31 +288,89 @@ pipeline {
...
+            
+            dir("${BASE_DIR}") {
```
v1v, line 334: this particular `dir` should be added to the very beginning [L311](https://github.com/elastic/apm-agent-nodejs/pull/1714/files#diff-8db37e8fcea0f1a8f2f39667e94ebcc4R311)

.ci/Jenkinsfile
v1v: Groovy is a bit tricky, and actually. you got a good point, the `==~` operator is the exact match based, therefore `^` and `$` are not required. If it is not intuitive, then we could force the match with the find operator `=~` instead and the `^` and `$`.

.ci/scripts/generate_cc_changelog/generate.py
cachedout: Hi @watson 

Thanks for this excellent write-up. I have modified the detection script to handle breaking changes by searching for breaking changes using both the rules that you mention.

One question, though. Let's say that (hypothetically), you had a PR which had to separate commits which would qualify for the CHANGELOG. Would you prefer to have (a) a single entry for the CHANGELOG which give the title of the PR in question or would you prefer to have (b) two separate entries, one for each qualifying commit, both of point to the same PR?

The script currently uses the behavior described in (a) but I can modify it if you think the second is preferable. Let me know what you think.

cachedout:
> Does this also create a proper release on the GitHub Releases page?

[It does now](https://github.com/elastic/apm-agent-nodejs/pull/1714/commits/e1987ea567c802216981558a8841313ed70bd770). :)

.ci/scripts/tag_release.sh
cachedout: Switched to using the [new git context manager](https://github.com/elastic/apm-pipeline-library/pull/481).

cachedout:
OK, this is getting pretty close. The last items are waiting on a response from @watson for the following:

1. [A question about the possible use of 2FA](https://github.com/elastic/apm-agent-nodejs/pull/1714#issuecomment-617013358).

1. [A question about CHANGELOG behavior](https://github.com/elastic/apm-agent-nodejs/pull/1714#discussion_r412162930).

Pending the results of those questions and any additional changes, we should be ready for final reviews. Once those are done, I'll do a CC-compliant rebase and squash down to one commit and hopefully we should be all set.

.ci/scripts/generate_cc_changelog/generate.py
watson: Good question... How will you currently know which PR title to pick in case of (a)? Do you rely on the PR ID to be in the commit message? If we use "Rebase and merge" to merge a PR so that the individual commits are retained, the PR ID will _not_ be added to the commit message. This only happens when you use "Squash and merge" - which is one reason why I don't like "Rebase and merge". So are you sure that your current implementation works in that case?

cachedout: In the case of (a) the PR title is used and it is fetched by querying the GitHub API. In the case of (b), we examine the git history directly instead of querying GitHub.

watson: Ah cool - I didn't know GitHub retained that information 👍 

This project has rarely used the old merge nor the rebase-and-merge in the past. Normally we've always used squash-and-merge. But I just went back to see an examples of one of the few times we did use rebase-and-merge:

PR: #1636
PR title: Fix CI
PR description: _\<blank\>_
Commits:
- test(pg): fix pg TAV tests
- ci(tav): fix @hapi/hapi TAV tests
- ci(tav): fix got TAV tests
- test: fix APM Server cert test to work with 13.9.0+
- fix: support promisifying setTimeout and friends …
- ci(jenkins): ensure postgres container doesn't need password
- fix(cassandra): improve support for cassandra-driver v4.4.0+ …
- test: fix APM Server schema tests

In this case option (b) would have been best as we didn't want the title "Fix CI", but would have liked the following to be added to the changelog:

- fix: support promisifying setTimeout and friends
- fix(cassandra): improve support for cassandra-driver v4.4.0+

### Conclusion

In this light I think option (b) would be best.

watson: Btw, some of the commits on purpose have a body with a detailed explanation that can't fit in the commit message headline. I think those bodies should go into the release notes as well if we can (previously we have not done that, but I think it would be worth investigating).

For example, the two above "fix" commits actually both had bodies, eg:

```
fix: support promisifying setTimeout and friends

When using `util.promisify` on functions where the callback is not the
last argument, a custom promisify function is normally provided on a
symbol on the main function. This is also the case with `setTimeout` and
`setImmediate`.

Our patch meant that `setTimeout` and friends would no longer have this
symbol on them and as such would not be promisified correctly.

This commit looks for the presence of such a promisifier and copies it
over from the original function to the wrapper function.

This commit also applies the same logic for patching functions where the
callback isn't the first argument. Currently we don't know of a case
where there is a similar issue with these kind of functions, but we
might as well protect those functions from this sort of issue while
we're at it.
```

cachedout: Sounds good! Thank you for the clear analysis. I'll make the necessary changes.

cachedout: OK, I have pushed a change which allows us to use either the GitHub strategy (a) or the git-based strategy (b) and defaults to git-based.

As a part of this, I also ensured that the git-strategy pushes the entire body of the commit message into the changelog [as requested](https://github.com/elastic/apm-agent-nodejs/pull/1714#discussion_r414356442). Therefore, I think the requirements of this request are now satisfied and I'm going to go ahead and resolve this thread.

cachedout:
@watson No hurry whatsoever, but just so you're aware -- the last item on our list to get this deployed is [the question about 2FA](https://github.com/elastic/apm-agent-nodejs/pull/1714#issuecomment-617013358). Once we have that answered, we should be ready for final reviews. Thanks!

cachedout:
Hi @watson. No hurry but this is just a polite bump in case you missed the above request. Thanks!

kuisathaverat:
@cachedout heads-up, we will have the same issue we have had with Ruby agent to checkout tags, this is the workaround https://github.com/elastic/apm-agent-ruby/commit/b371515b72f1a1450a89fac3407b0df5e6d6404c

watson:
@cachedout sorry, I misunderstood your first comment as you were still working out the details of how to do 2FA and that I should wait with a review until the PR was updated with that?

cachedout:
@watson No worries. We're actually not sure if the current process uses 2FA or not. Right now, the PR assumes that it is not required but if it is required, we'll need to update the PR to reflect that and then we'll ping you for a review after that. So, if you could just let us know if it's part of the current release process, we can move forward from there. Thanks!

watson:
Ah ok, yes we do use 2FA and it's required on the org level.

cachedout:
Hi @watson and @v1v 

This should be ready for a final review.

.ci/Jenkinsfile
```diff
@@ -282,31 +289,89 @@ pipeline {
...
+              sh(label: 'Update package.json', script: ".ci/scripts/update_package_version.sh ${params.release_ver}")
+              withCredentials([usernamePassword(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken',
+                                          passwordVariable: 'GITHUB_TOKEN', usernameVariable: 'GITHUB_USER')]) {
+                withGitRelease(credentialsId: '2a9602aa-ab9f-4e52-baf3-b71ca88469c7-UserAndToken') {
+                  sh(label: 'Tag and push', script: ".ci/scripts/tag_release.sh --version ${params.release_ver}")
+                }
+              }
+              prepareRelease() {
+                sh 'npm publish'
+              }
```
v1v, line 336: When running `npm publish` in a vanilla workspace, will it download all the dependencies?  If so, I'd say it might be a bit slow that the TOTP that has been queried, might be expired

watson: No, publish will just take whatever is in the CWD and publish it. It doesn't care about what's checked into git or not, or the state of `node_modules`. It literally just takes the files (respecting `.npmignore`) and bundles them up into a tar-ball which is pushed to npm.

watson:
Due to the size of this PR and that it involves Jenkins, it's a bit hard to review properly for me, given how much time I have. I assume y'all have done a great job, but if you want a proper review from me, it would be easier if there was a way that I could test this to see what happened without actually publishing - Or maybe by modifying it to publish to a tag other than `latest` which can then always be deleted later. Can I run this locally somehow?

cachedout:
>  if you want a proper review from me

Hi @watson. We've had a review of this code by members on the @elastic/observablt-robots team and in additional, we just generally just like to give the repo owners a chance to look over changes if they want to. 

Since this only very minimally touches the current Jenkinsfile and the rest is simply additive, my proposal would be similar to what we've done with other teams -- we go ahead and get this in and when you're ready to do your next release, we schedule some time to sit with you and run through the automated process. If issues are discovered in the new functionality for the release process, we'll address them in a follow-up PR.

Please let me know how that sounds. Thanks!

cachedout:
Hi @watson. I haven't heard from you on the above, so I'll likely rebase this if you aren't opposed, I can get it merged soon. Let me know if that sounds all right. Thanks!

cachedout:
I haven't heard back from the above so unless there are any objections, I'll rebase and merge this tomorrow (June 3). Thanks!

cachedout:
Update on this. I am waiting for the branch failures to be resolved before rebasing this a final time and merging it.

watson:
Hi, it's been hard for me to keep on top of things in this repo lately, but I'm super happy that y'all run with it and keep on improving things around here.

Regarding the current CI failures in `master`, I haven't looked at them yet, so I'm not yet sure how complicated they will be to fix. But I'll try to find some time ASAP to get it all to green.

Once they are fixed and we want to release a new version, I'll happily join a call where we can merge this 👍 Would love to learn more about this from y'all 🙂

trentm:
Victor, Alan, and I had a call today and we decided to close this for now. After Alan and I have some more experiences with releases of this (and https://github.com/elastic/apm-nodejs-http-client/) we will revisit automating parts of the release process.

(Part of the value of using a new issue later will also be to reset the discussion on changelog generation, because that has changed a lot in this repo since Alan and I started maintenance.)

PR #1727, opened 2020-04-23
ci(dependabot): configuration file

v1v:
If we would like to enable the dependabot automation process to bump the version of the packages automatically, then it's required this particular file.

### How does it work?

There is a weekly pipeline that will care for all the projects with this particular file `ci/dependabot.yml` and if so it will run the validation and PR creations.

### For instance

I ran the below command locally

```bash
docker run -ti \
    --rm -e GITHUB_ACCESS_TOKEN=${GITHUB_ACCESS_TOKEN} \
    -e PROJECT_PATH=v1v/apm-agent-nodejs \
    -e PACKAGE_MANAGER=npm_and_yarn \
    docker.elastic.co/observability-ci/dependabot
```

and caused the below PRs in my forked repo:

https://github.com/v1v/apm-agent-nodejs/pulls?q=is%3Apr+is%3Alabel%3Adependencies

### Actions

- [ ] Verify if this particular dependabot is a good candidate to be enabled in this project.
- [ ] Assign?

### DIFF
--- a/.ci/dependabot.yml
+++ b/.ci/dependabot.yml
@@ -1,0 +1,5 @@
+### The list of GitHub accounts that will be used to assign the PRs to (no space and comma separated)
+assign: ''
+
+### The package manager to be used (no space and comma separated if more than one package)
+manager: npm_and_yarn


watson:
This project already uses a different service for the same thing ([Greenkeeper](https://greenkeeper.io/)). However, this service is shutting down, so we need to find a different one and Dependabot is a possibility.

Another option that I have been exploring is Snyk, as they are collaborating with Greenkeeper to take over their old users. The benefit of Snyk would be that we had one less 3rd-party bot to worry about (since we already use Snyk for security monitoring). But Dependabot is also a good alternative, so I think that would be a fine too.

Here are the use-cases/requirements for one such alternative:

- **We need to monitor both dev and prod dependencies:** We need to know when a dev-dependency is breaking the build since a lot of those actually are modules that we are patching (e.g. `mysql`, `mongodb` etc). If one of these releases a new version that our patch no longer works with, we need to know ASAP so we can fix the patch - our tests will detect that.
- **No need to bump versions within our already allowed ranges:** We usually use `^x.y.z` semver ranges for dependencies. We don't need to bump these numbers every time a new version of a module is released if it's within that range. We only need to either pin the version if a new version within the current range suddenly breaks something
- **Keeping up to date with new majors:** When a new major is released of a prod dependency, it would be wise, to update to this version if possible. This can either be done by just changing the range in `package.json` in case we don't need to make any code changes, or it involves changes to our code in case a breaking change is introduced and we rely on the old behavior. This however, is not that time-critical and should just be done when we have time.
- **Ignore new versions of certain modules:** In some cases, we don't want to upgrade a specific module, and we need to tell the bot to ignore any new versions of that particular module. E.g: https://github.com/elastic/apm-agent-nodejs/blob/db0959808c36c16d57171158c9992fda25213189/package.json#L183-L187

.ci/dependabot.yml
```diff
@@ -0,0 +1,5 @@
+### The list of GitHub accounts that will be used to assign the PRs to (no space and comma separated)
+assign: ''
+
+### The package manager to be used (no space and comma separated if more than one package)
+manager: npm_and_yarn
```
watson, line 5: This project uses npm

v1v:
I'll close this PR now for the time being, as we are still gathering details about what's the tool to be used. Thanks so much for your well detailed list of requirements, in fact, we are gonna use it :)

Issue #1718, opened 2020-04-20
License headers 


lreuven:
We need to add headers for all source code files. If it isn't a source code file, no need.

```
/*
 * Copyright Elasticsearch B.V. and other contributors where applicable.
 * Licensed under the BSD 2-Clause License; you may not use this file except in
 * compliance with the BSD 2-Clause License.
 */
```

Update: Changed license header to one (agreed upon with legal) and for the BSD-2-Clause license this repo is using.

PR #1748, opened 2020-05-13
chore: rule to enforce license header on source

vigneshshanmugam:
+ fix elastic/apm-agent-nodejs#1718 
+ Uses `eslint` to enforce the standard style as we cannot extend rules with standard. 
+ Project specific rules are applied with `--rulesdir` option when on lint runs. 

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,15 @@
+module.exports = {
+  env: {
+    node: true,
+    es6: true
+  },
+  extends: [
+    'standard'
+  ],
+  parserOptions: {
+    ecmaVersion: 2018
+  },
+  rules: {
+    'require-license-header': 2
+  }
+}
--- a/test/script/run_tests.sh
+++ b/test/script/run_tests.sh
@@ -38,7 +38,7 @@ setup_env () {
 }
 
 run_test_suite () {
-  standard
+  eslint .
   npm run test:deps
   npm run lint:commit
   
--- a/dev-utils/eslint-rules/require-license-header.js
+++ b/dev-utils/eslint-rules/require-license-header.js
@@ -1,0 +1,104 @@
+/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
+ * license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Elasticsearch B.V. licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+function normalizeWhitespace (string) {
+  return string.replace(/\s+/g, ' ').trim()
+}
+
+// License to be appended to every source file
+const LICENSE = `/*
+ * Licensed to Elasticsearch B.V. under one or more contributor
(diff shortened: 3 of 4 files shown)


vigneshshanmugam:
@watson Let me know what you think about this approach, RUM agent uses something similar and thought we could reuse some of the code. 

I have not applied the fix as the PR would introduce lots of changes. Wanted to get a bake in and we can just run lint:fix to add the license for all files.

test/script/run_tests.sh
```diff
@@ -38,7 +38,7 @@ setup_env () {
 }
 
 run_test_suite () {
-  standard
+  eslint .
```
watson, line 41: How come this doesn't need the `--ruledir` argument like we do in `package.json`?

.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  env: {
+    node: true,
+    es6: true
+  },
+  extends: [
+    'standard'
+  ],
+  parserOptions: {
+    ecmaVersion: 2018
```
watson, line 10: What's the difference between setting this and setting `es6: true`? If I understand correctly `es6: true` sets `ecmeVersion: 6`, but `2018` is equal to `ecmeVersion: 8` - so isn't there some sort of conflict here?

test/script/run_tests.sh
vigneshshanmugam: Yes we need it, forgot to update it here. May be i can do `npm run lint` instead?

.eslintrc.js
vigneshshanmugam: If i remember correctly, this option is used by `espree` which is the default parser of eslint to understand the new syntax. While `env: 6` would allow using global variables without causing any lint errors. 

Some details here - https://eslint.org/docs/user-guide/configuring#specifying-parser-options

test/script/run_tests.sh
watson: Yeah good idea, then you don't need to repeat the same line 👍

.eslintrc.js
watson: Ok, as long as it's on purpose 👍 

I tried to look up what's needed to support Node.js 8 and nothing newer than that, but couldn't easily figure it out. Is this the right setup for Node.js 8?

vigneshshanmugam: I guess yes `node`: true should pull in necessary globals. Unless we want to use new ones like `SharedArrayBuffer`, we may need to add them manually on globals property. 

Also, I am using eslint 7.0 which dropped support for Node 8 - should we need support it? https://eslint.org/blog/2020/05/eslint-v7.0.0-released#dropping-support-for-node-js-v8

watson: Using the latest eslint is fine with me - it doesn't need to run on Node.js 8 as long as it runs on one of our supported Node.js versions. The linting will be the same no matter which version it runs on.

What I think would be really cool would be if we could get it to lint for exact ECMAScript features we support. For example, we can't use the RegExp [`dotAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/dotAll) feature as that's not supported in all the versions of Node.js we support.

We need to target V8 6.0.287.53 which is the version that ships with Node.js v8.6.0. If I understand correctly, any version of V8 will not implement 100% of all ECMAScript editions, but say 100% of ES2015, 98% of ES2016, and so on. If eslint somehow was able to lint against this, that would be really nice so that we don't ship code that's not compatible with Node.js v8.6.0 by a mistake.

I'm not sure if `node: true` does this? After a bit of googling, I can see that there's an [eslint-plugin-node](https://www.npmjs.com/package/eslint-plugin-node) plugin which seems to use the `engines` field in `package.json` to lint for the minimum supported Node.js version?

astorm:
@vigneshshanmugam We're doing some backlog cleanup -- is this still something you're interested in doing.

trentm:
Closing in favour of https://github.com/elastic/apm-agent-nodejs/pull/2775

PR #1755, opened 2020-05-18
chore(package): upgrade to tape v5

watson:
List of breaking changes in tape v5 ([source](https://github.com/substack/tape/commit/cd81a404d7a8308b9aad50d2aa2c92e330351d1e)) - I've **highlighted** changes that required changes to our code:

- **only `looseEqual`/`deepEqual, and their inverses, are now non-strict.**
- make equality functions consistent:
- **`equal`: use `==`, not `===`, to match `assert.equal`**
- `strictEqual`: bring `-0`/`0`, and `NaN` into line with `assert`
- update `deep-equal` to v2
- **fail any assertion after `.end()` is called**
- `error` should not emit `expected`/`actual` diags
- support passing in an async function for the test callback
- tests with no callback are failed TODO tests
- equality functions: throw when < 2 arguments are provided
- add "exports" to restrict public API
- `throws`: bring into line with node’s `assert.throws`
- use default `require.extensions` collection instead of the magic Array `['.js']`


test/instrumentation/modules/http2.js
```diff
@@ -155,10 +158,11 @@ isSecure.forEach(secure => {
   })
 
   test(`http2.${method} stream respondWithFile`, t => {
+    t.plan(14)
```
vigneshshanmugam, line 161: this API is really nice

watson:
The CI error is unrelated. Will merge

PR #1792, opened 2020-07-09
feat: add support for amqplib

### LINKED ISSUE #1791: Add support for RabbitMQ through amqplib

`apm-agent-nodejs` supports databases and Redis but not AMQP-related messaging services like RabbitMQ.

Add a wrapper for `amqplib`.

**Additional context**

IamBlueSlime:
This PR adds an implementation for the `amqplib` package. It creates a span when a message is sent to an AMQP server (like RabbitMQ).

I've tested all the versions of the package and it fails starting from `0.5.4`. That's why I've required at least the version `0.5.5`.

The simple test is passing.

Fixes elastic/apm-agent-nodejs#1791 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


IamBlueSlime:
Travis's build is failing due to the `apollo-server-express` module.

lib/instrumentation/modules/amqplib.js
```diff
@@ -0,0 +1,54 @@
...
+
+  function wrapConnect (original) {
+    return function wrappedConnect (url) {
+      let protocol, host, port
+      if (typeof url === 'object') {
+        protocol = (url.protocol || 'amqp') + ':'
+        host = url.hostname
+        port = url.port || ((protocol === 'amqp:') ? 5672 : 5671)
+      } else {
+        var parts = URL(url, true)
```
vigneshshanmugam, line 42: Can you change this to use the default parser? https://github.com/elastic/apm-agent-nodejs/blob/62fc6acb0d03f874a3a28b9fc9768822faad14df/lib/parsers.js#L246-L248

lib/instrumentation/modules/amqplib.js
```diff
@@ -0,0 +1,54 @@
...
+    })
+  }
+
+  function wrapConnect (original) {
+    return function wrappedConnect (url) {
+      let protocol, host, port
+      if (typeof url === 'object') {
+        protocol = (url.protocol || 'amqp') + ':'
+        host = url.hostname
+        port = url.port || ((protocol === 'amqp:') ? 5672 : 5671)
```
vigneshshanmugam, line 40: nit: you can move the default port logic inside here to avoid doing two times
https://github.com/elastic/apm-agent-nodejs/blob/7d1c43b1986f112c686a4a0c0c95f5c7640a8405/lib/instrumentation/modules/amqplib.js#L24

lib/instrumentation/modules/amqplib.js
```diff
@@ -0,0 +1,54 @@
...
+      } else {
+        var parts = URL(url, true)
+        protocol = parts.protocol
+        host = parts.hostname
+        port = parseInt(parts.port) || ((protocol === 'amqp:') ? 5672 : 5671)
+      }
+
+      return original.apply(this, arguments).then(function (channelModel) {
+        patchChannelModel(channelModel, host, port)
```
vigneshshanmugam, line 49: Question: I am not too familiar with the amqplib, Does this code handle both the callback and promise version of the API? 

Can we add tests for the same?

package.json
```diff
@@ -105,7 +105,8 @@
     "sql-summary": "^1.0.1",
     "stackman": "^4.0.1",
     "traceparent": "^1.0.0",
-    "unicode-byte-truncate": "^1.0.0"
+    "unicode-byte-truncate": "^1.0.0",
+    "url-parse": "^1.4.7"
```
vigneshshanmugam, line 109: removed with using the default parser.

lib/instrumentation/modules/amqplib.js
```diff
@@ -0,0 +1,54 @@
...
+  agent.logger.debug('shimming amqplib.connect')
+  shimmer.wrap(amqplib, 'connect', wrapConnect)
+
+  return amqplib
+
+  function patchChannelModel (channelModel, host, port) {
+    agent.logger.debug('shimming amqplib.ChannelModel.connection.sendMessage')
+    shimmer.wrap(channelModel.connection, 'sendMessage', function (orig) {
+      return function (channel, method, fields, properties, props, content) {
+        var span = enabled && agent.startSpan(null, 'messaging', 'amqp', 'publish')
```
vigneshshanmugam, line 18: shouldn't `publish` be the actual method that gets called?

test/instrumentation/modules/amqplib.js
```diff
@@ -0,0 +1,65 @@
...
+  metricsInterval: 0,
+  centralConfig: false
+})
+
+var amqplib = require('amqplib')
+var test = require('tape')
+
+var mockClient = require('../../_mock_http_client')
+
+test('instrument simple message', function (t) {
```
vigneshshanmugam, line 16: may be add few more tests with different methods?

lib/instrumentation/modules/amqplib.js
IamBlueSlime: The callback version is just a wrapping of the promise version so it will work as it.
I can add a test for the callback version if you want.

lib/instrumentation/modules/amqplib.js
IamBlueSlime: `sendMessage` is the method needed to be patched but I've thought that `publish` is more universal. That's why I've named the span like this. Are you ok with that?

package.json
IamBlueSlime: Fixed in https://github.com/elastic/apm-agent-nodejs/pull/1792/commits/38fdf0b00bc64ff5933f58c194a74a4b54765f88

lib/instrumentation/modules/amqplib.js
IamBlueSlime: Fixed in https://github.com/elastic/apm-agent-nodejs/pull/1792/commits/38fdf0b00bc64ff5933f58c194a74a4b54765f88

lib/instrumentation/modules/amqplib.js
IamBlueSlime: Fixed in https://github.com/elastic/apm-agent-nodejs/pull/1792/commits/38fdf0b00bc64ff5933f58c194a74a4b54765f88

lib/instrumentation/modules/amqplib.js
vigneshshanmugam: Yeah adding a test should be good here.

IamBlueSlime: You were right, some issues with the callback version. I'm digging into.

IamBlueSlime:
One question @vigneshshanmugam, do you know how I can shim a submodule of one library? I need to shim `amqplib/callback_api`. It's not directly exposed from the `amqplib`. I need to replicate my patches to this file too for the callback version to work properly.

vigneshshanmugam:
@IamBlueSlime Have a look at this one https://github.com/elastic/apm-agent-nodejs/pull/1266/

IamBlueSlime:
I can't manage to find anything useful for my case on your PR. This is not another module but one file of it. I know `amqplib` is strange on its implementation but we don't have the choice haha.

Here is an exemple snipper showing what I'm telling: https://github.com/squaremo/amqp.node/blob/master/examples/tutorials/callback_api/send.js#L3

I think that this needs a refactor on the agent internals which is surely not the scope of this PR. Maybe we can go as far to support only the promise version for now?

vigneshshanmugam:
@IamBlueSlime What i meant is , you can add the callback version to the MODULES list and those calls will also be intercepted in the same way. 

https://github.com/elastic/apm-agent-nodejs/blob/6c81e4afcec1b502c1b52fbb59726efa76540c2f/lib/instrumentation/index.js#L13
```js
var MODULES = ['amqplib/callback_api', 'amqplib', ...]
```

Did you try it? Are you saying it did not work?

IamBlueSlime:
Yup, I've tested this but it loads only the `amqplib` module. I mean, with setting the log level to DEBUG it's saying that only the `amqplib` module is shimmed. I'm pretty sure that it's because it's a subfile and not a module under a `@` package group...

JCMais:
From looking at the instrumentation code, looks like taht if you add the `amqplib/callback_api` to the `MODULES` array like @vigneshshanmugam mentioned, you must also create a separated instrumentation at `amqplib/callback_api.js`, this is due to the way the instrumentation file is required here https://github.com/elastic/apm-agent-nodejs/blob/6c81e4afcec1b502c1b52fbb59726efa76540c2f/lib/instrumentation/index.js#L79-L84

test/instrumentation/modules/amqplib.js
```diff
@@ -0,0 +1,65 @@
...
+    t.strictEqual(trans.result, 'success')
+
+    var span = data.spans[0]
+
+    // t.strictEqual(span.name, null)
+    t.strictEqual(span.type, 'messaging')
+    t.strictEqual(span.subtype, 'amqp')
+    t.strictEqual(span.action, 'publish')
+    t.deepEqual(span.context.destination, {
+      service: { name: 'amqp', resource: 'amqp', type: 'messaging' },
```
imsergiobernal, line 115: I suggest type to be `message` instead of `messaging` in order to follow `request` type, which is not `requesting`. What do you think?

lib/instrumentation/modules/amqplib.js
imsergiobernal, line 18: I suggest type to be message instead of messaging in order to follow request type, which is not requesting. What do you think?

imsergiobernal, line 18: Why starting an span and not a transaction? I'm not pretty sure but I think that Express, for instance, starts with a transaction because if not, you should create a transaction before express handles the route request. I'm right?

Issue #1815, opened 2020-08-25
Collect cloud metadata


felixbarny:
see https://github.com/elastic/apm/pull/290

see also _current_ spec doc: https://github.com/elastic/apm/blob/3f224a9cbacb196bbb196b0868d9fa90a0bbaf6c/specs/agents/metadata.md#cloud-provider-metadata

JCMais:
How would someone go about implementing this?

From what I can see the metadata itself will need to be added to https://github.com/elastic/apm-agent-nodejs/blob/cde2dde8c896e3f6f8313fa3db66da88106d6397/lib/config.js#L239-L249 (which means adding it to the apm-nodejs-http-client package)

`net.createConnection()` is probably the fastest way to check if the metadata server exists, if it does a connection can be made to retrieve the information. This is basically what was done on the reference implementation: https://github.com/basepi/apm-agent-python/blob/f13110156e57eec99134a7f9342243416b969698/elasticapm/utils/cloud.py#L79-L115

But the biggest "issue" so far, is that the config generation is synchronous. An alternative would be to create the cloud metadata fetcher as a function that accepts a callback and then pass that to `ElasticAPMHttpClient`. Writing to the stream could then be deferred until the callback has been called: https://github.com/elastic/apm-nodejs-http-client/blob/1526dedc26392bd92f20b5079f848c7e77567e7c/index.js#L496-L500

felixbarny:
That sounds like a good plan and is in line with what the spec says:

> Any intake API requests to the APM server should be delayed until this metadata is available.

Any events that are created before the metadata is ready would be buffered in the meantime.

JCMais:
I've started working on this at https://github.com/elastic/apm-agent-nodejs/pull/1816

astorm:
With https://github.com/elastic/apm-agent-nodejs/pull/1937 and https://github.com/elastic/apm-nodejs-http-client/pull/129 landed this work is done, and we be available after the next http client and agent release.

PR #1816, opened 2020-08-26
Cloud Metadata

JCMais:
Still WIP, but implementation is ready to be reviewed.

Reference implementation:
- https://github.com/elastic/apm-agent-python/pull/826
- https://github.com/elastic/apm-agent-python/blob/4bef64b96aab441ef192ffc6faa0513edb440578/elasticapm/utils/cloud.py

See elastic/apm-agent-nodejs#1815

Depends on https://github.com/elastic/apm-nodejs-http-client/pull/109 being merged.

### Checklist

- [X] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -116,6 +116,10 @@ function ensureUrl (v) {
   }
 }
 
+function getInternalIgnoreHeader (res) {
+  return res.getHeader ? res.getHeader('x-elastic-apm-agent-nodejs-ignore') : res._headers.host
+}
+
 function getSafeHost (res) {
   return res.getHeader ? res.getHeader('Host') : res._headers.host
 }
@@ -163,6 +167,9 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
 
       if (getSafeHost(req) === agent._conf.serverHost) {
         agent.logger.debug('ignore %s request to intake API %o', moduleName, { id: id })
+        return req
+      } else if (getInternalIgnoreHeader(req) === '1') {
+        agent.logger.debug('ignore %s request with ignore header flag %o', moduleName, { id: id })
         return req
       } else {
         var protocol = req.agent && req.agent.protocol
--- a/lib/cloud.js
+++ b/lib/cloud.js
@@ -19,7 +19,7 @@ const httpRequest = (url, method, headers, cb, { isJsonBody = true } = {}) => {
     timeout: 3000
   }
 
-  const req = http.request(options, (res) => {
+  const req = http.request(url, options, (res) => {
     let json = ''
 
     res.setEncoding('utf8')
--- a/lib/cloud.js
+++ b/lib/cloud.js
@@ -19,7 +19,7 @@ const httpRequest = (url, method, headers, cb, { isJsonBody = true } = {}) => {
     timeout: 3000
   }
 
-  const req = https.request(options, (res) => {
+  const req = http.request(options, (res) => {
     let json = ''
 
     res.setEncoding('utf8')
--- a/lib/cloud.js
+++ b/lib/cloud.js
@@ -1,43 +1,239 @@
+'use strict'
+
 const net = require('net')
+const http = require('http')
 
 const VALID_CLOUD_PROVIDERS = ['aws', 'gcp', 'azure']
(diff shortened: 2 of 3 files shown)


JCMais:
I think the implementation is finished, also opened PR at https://github.com/elastic/apm-nodejs-http-client/pull/109, which needs to be merged before this one.

I still need to test this on the cloud providers themselves, I don't know if I will be able to test on AWZ and Azure, as I have easy access only to GCP.

It's also missing tests, would love some feedback on how to test this feature.

lib/instrumentation/http-shared.js
```diff
@@ -164,6 +168,9 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       if (getSafeHost(req) === agent._conf.serverHost) {
         agent.logger.debug('ignore %s request to intake API %o', moduleName, { id: id })
         return req
+      } else if (getInternalIgnoreHeader(req) === '1') {
+        agent.logger.debug('ignore %s request with ignore header flag %o', moduleName, { id: id })
+        return req
```
JCMais, line 173: This is basically a new feature inside this feature, it would allow us to ignore some http requests if they contain a specific header. I don't know if this header should be configured by the client or not, as it's being used only internally for now.

JCMais:
Tested on GCP and seems to be working.

Here are the steps to test the PR: https://gist.github.com/JCMais/f46099b245401bc8da7d095840a63422

image

basepi (commented):
This looks excellent to my untrained eyes (I don't really know Node, I work on the Python agent).

As far as testing strategies go, this is a problem across all the agents, since we don't have cross-cloud testing automation set up (yet).

For Python, I harvested some real payloads from each of the metadata services (during my manual testing) and set up tests against those payloads using a mocked http request. This prevents regressions in the parsing code, but doesn't do anything for changes in the upstream API behavior. Luckily, these metadata services are *very* stable, since they're used by everyone all the time. We'll still want to do integration tests at some point, though.

My only other comment was that we need to add documentation for the new config value, but it looks like you already have a checkbox for that, so 👍

JCMais:
I should be able to get back to this next week

astorm:
@JCMais :wave: hello there!  We're doing a bit of open source gardening and just wanted to check in with you here. Is this PR something you're still interested in persuing?  If so is there any feedback you need from us?

No rushing or pressure intended -- we just wanted to make sure you have what you need to succeed here at your own pace.

JCMais:
@astorm I do have time to take a look into this, but I need help with the tests, no idea how to test this feature correctly.

astorm:
@JCMais 

Hello again!  Cloud Metadata's currently our next up feature, so we should be able to come back to this this week (in a limited fashion, given the quiet last week of the year) and next.  Also -- check your email! :) 

To that end

Re: testing -- we've put together a PR that implements a mock testing server: https://github.com/elastic/apm-agent-nodejs/pull/1932 -- in order to use this you'd need to refactor things so that the URL isn't hard coded to `169.254.169.254` and could accept an arbitrary metadata URL and port.  In practice this will always be `169.254.169.254` since that's the defacto standard, but if the URL is injectable that would let us test this code using the above test server

Re: general feedback -- First, thank you for making this initial PR -- it really helped up focus our thoughts about how we want to approach this in the Node JS Agent. 

After reviewing this PR, we realized we want to approach things slightly differently that what you've prototyped here. 

1. We'd like to get the "do send/don't send" logic out of the `apm-nodejs-http-client` client.  i.e. we're leaning towards a philosophy where `apm-nodejs-http-client` is just a "dumb" client, and any logic about whether to send or not should live in the agent itself. 

2. Given that -- we want to take an approach where the agent attempts to fetch this metadata (possibly synchronously) and passes the fetched data along (or, if no metadata servers were detected, passes no data) when instantiating the `ElasticAPMHttpClient` object

This should fulfill the requirements of not sending any data until the metadata is fetched.  Once we land this functionality, we can consider approaches on what to do with data that's collected in the brief window between agent startup and having fetched the completed metadata.

All of this is also open to discussion, so let's coordinate here (and/or elsewhere( to figure out what we're doing.

astorm:
Having landed this feature, we're going to close this issue out, but if there's more to discuss here don't hesitate to keep the conversation going.

PR #1823, opened 2020-10-01
fix: add missing dependency parseurl

merceyz:
Adds a missing dependency `parseurl`

Fixes https://github.com/yarnpkg/berry/issues/1911
Fixes https://github.com/elastic/apm-agent-nodejs/issues/1822

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,6 +180,14 @@
     "wait-on": "^3.3.0",
     "ws": "^7.2.1"
   },
+  "peerDependencies": {
+    "parseurl": "^1.3.3"
+  },
+  "peerDependenciesMeta": {
+    "parseurl": {
+      "optional": true
+    }
+  },
   "greenkeeper": {
     "ignore": [
       "inquirer"


merceyz:
In the ICLA document it asks for `Title`, what exactly is it after?

astorm:
Hey @merceyz, nice to meet you and thanks for making this initial PR.  

Regarding your question, the folks at `cla-questions@elastic.co` are the ultimate authority when it comes to questions about the CLA, but Re: _asks for `Title`, what exactly is it after?_, I believe it's looking for your job title.  "Senior Software Engineer, CEO, President, Vice Co-President of the Americas Region, etc."  That sort of thing.  Does that make sense?

Regarding this PR generally, looking over our source it seems like the only place we're still referencing the `parseurl` library is over in [`express-utils.js`](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/express-utils.js).  If the `parseurl` module is _not_ present we define our own polyfill, and then the `getPathFromRequest` uses whatever version is defined to parse the URL and return the `pathname`.

Put another way -- the fact that this module is missing from `package.json` looks like a deliberate choice.  Don't have it installed?  You get our URL parsing.  Have it installed?  You get `parseurl` parsing. 

Thinking out loud about this -- it may be that having this conditional `parseurl` module behavior may have outlived whatever usefulness it's had in the past. it may be time to drop it and rely solely on the parsing in `lib/parsers.js`.  To do this we'd want to 

1. Confirm that the elastic parsing functionality is equivalent to the `parseurl` functionality (and enshrine this in a test)

2. Remove the `require('parseurl')` try/catch block from `express-utils.js`

Do you think you'd be up for that @merceyz

merceyz:
> Regarding your question, the folks at cla-questions@elastic.co are the ultimate authority when it comes to questions about the CLA, but Re: asks for Title, what exactly is it after?, I believe it's looking for your job title. "Senior Software Engineer, CEO, President, Vice Co-President of the Americas Region, etc." That sort of thing. Does that make sense?

That makes perfect sense if I was signing the CLA for "A company wanting its employees to contribute" but as an individual it doesn't really make sense to me.

> Put another way -- the fact that this module is missing from package.json looks like a deliberate choice. Don't have it installed? You get our URL parsing. Have it installed? You get parseurl parsing.

Sounds like it should be an optional peer dependency then

> Do you think you'd be up for that

I'm afraid I don't have the time to do that, i've changed it to a optional peer dependency and i'll leave it up to you what to do next

astorm:
> I'm afraid I don't have the time to do that, i've changed it to an optional peer dependency and i'll leave it up to you what to do next

Totally reasonable -- you've already gone above and beyond.   Thanks again for bringing this to our attention. 

We're going to close this out in favor on [this new issue to remove the `parseurl` dependency from the agent](https://github.com/elastic/apm-agent-nodejs/issues/1884). 

If anyone has additional conversation around this we can move it over there.

Issue #1814, opened 2020-08-25
Add outcome to transactions and spans


felixbarny:
see https://github.com/elastic/apm/pull/299

astorm:
Hamid's PR https://github.com/elastic/apm-agent-nodejs/pull/1824. gives transactions and spans an outcome property.  Spans that represent outgoing HTTP transactions  will have their outcome set.  The Transaction

Remaining work/questions

- Are there spans other than those representing an HTTP transactions that represent work with an outcome?  

- Or should we attempt to set outcome on a span if _any_ error occurred during the span of time it represents?

- We need an agent level API for setting the transaction outcome manually?

- Do we want/need to tackle an agent level API for setting the span outcome manually -- what would that even look like given that a span can represent any arbitrary unit of time? (would you always be setting the outcome of the current span? Or would you be walking the span tree to find the span object you wanted? etc.)

astorm:
Merged https://github.com/elastic/apm-agent-nodejs/pull/1968 should bring this to a close.

PR #1824, opened 2020-10-02
feat: add span and transaction outcome

jahtalab:
Part of #1814
### Checklist

- [x] Implement code
- [x] Add tests


index.d.ts
```diff
@@ -120,6 +120,8 @@ declare class Agent implements Taggable, StartSpanFn {
   registerMetric(name: string, labels: Labels, callback: Function): void;
```
astorm, line 120: I see here we've created a new type, `Outcome`, that can be one of three string values.  We've also used the type to add a new `outcome` property to the `GenericSpan` class.

lib/instrumentation/http-shared.js
```diff
@@ -206,6 +206,15 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
...
         })
+
+        if (typeof statusCode !== 'undefined') {
+          if (statusCode >= 400) {
+            span.outcome = 'failure'
+          } else {
+            span.outcome = 'success'
+          }
+        }
+
```
astorm, line 217: Here I see we're setting the `outcome` property of a span based on the the HTTP status code -- anything greater than 400 is a `failure`. This matches the behavior described in the spec.  The inline type checking seems prudent given how the lifecycle of the `statusCode` variable is a little weird in this module.

lib/instrumentation/modules/http.js
```diff
@@ -41,6 +41,11 @@ module.exports = function (http, agent, { enabled }) {
         // It shouldn't be possible for the statusCode to be falsy, but just in
         // case we're in a bad state we should avoid throwing
         trans.result = 'HTTP ' + (this.statusCode || '').toString()[0] + 'xx'
+        if (this.statusCode >= 500) {
+          trans.outcome = 'failure'
+        } else {
+          trans.outcome = 'success'
+        }
```
astorm, line 49: Here I see that during our handling/wrapping of the `http.ServerResponse` method we're examining the status code used in the response, and if it's greater than or equal to 500 we set the outcome as a failure -- otherwise we set the outcome to success.  This matches the behavior described in the spec.

lib/instrumentation/modules/http2.js
```diff
@@ -80,6 +80,11 @@ module.exports = function (http2, agent, { enabled }) {
       var status = headers[':status'] || 200
       trans.result = 'HTTP ' + status.toString()[0] + 'xx'
       trans.res.statusCode = status
+      if (status >= 500) {
+        trans.outcome = 'failure'
+      } else {
+        trans.outcome = 'success'
+      }
       trans.res.headers = mergeHeaders(trans.res.headers, headers)
```
astorm, line 88: Here I see that during our handling of an http2 response we're setting the transaction's outcome to `failure` if the http status code is greater than or equal to 500.  This behavior matches the spec.

lib/instrumentation/modules/http2.js
```diff
@@ -179,6 +184,14 @@ module.exports = function (http2, agent, { enabled }) {
...
         })
 
+        if (typeof statusCode !== 'undefined') {
+          if (statusCode >= 400) {
+            span.outcome = 'failure'
+          } else {
+            span.outcome = 'success'
+          }
+        }
+
```
astorm, line 194: Here I see that during our handling of calls the the `request` method of of objects returned by http.connect we're examining the HTTP2 response code and setting the outcome property to `failure` if the status code is greater than or equal to 400.  This behavior matches the spec.

lib/instrumentation/span.js
```diff
@@ -153,7 +154,8 @@ Span.prototype._encode = function (cb) {
       duration: self.duration(),
       context: undefined,
       stacktrace: frames,
-      sync: self.sync
+      sync: self.sync,
+      outcome: self.outcome
     }
```
astorm, line 159: Here I see we're adding `outcome` to the serialized span payload.

lib/instrumentation/span.js
```diff
@@ -36,6 +36,7 @@ function Span (transaction, name, ...args) {
 
   this.transaction = transaction
   this.name = name || 'unnamed'
+  this.outcome = 'unknown'
```
astorm, line 40: 👍 for a default outcome of unknown.

lib/instrumentation/span.js
```diff
@@ -153,7 +154,8 @@ Span.prototype._encode = function (cb) {
       duration: self.duration(),
       context: undefined,
       stacktrace: frames,
-      sync: self.sync
+      sync: self.sync,
+      outcome: self.outcome
```
astorm, line 158: @jahtalab What is the `sync` property here for?  Is this related to the `outcome` functionality?

lib/instrumentation/transaction.js
```diff
@@ -129,7 +130,8 @@ Transaction.prototype.toJSON = function () {
     sync: this.sync,
     span_count: {
       started: this._builtSpans
-    }
+    },
+    outcome: this.outcome
   }
```
astorm, line 135: I see here we're adding the outcome to the serialized transaction.

astorm (commented):
@jahtalab this looks like a great and also showed me where a few of the important agent hooking/listening points were with regards to incoming and outgoing HTTP transactions.  TIL :) 

I had one question about the `sync` property below -- the rest of my review is just stating out loud what I see in order to make sure we're on the same page.  

In addition to that sync question -- how would you feel about creating a `./lib/errors` (or similar) module so logic for handling status code is in one central place?  We don't need to do the full API portions of the spec right now to land this, but unless there's a compelling reason not to I'd rather have the logic for identifying what status codes relate to which outcomes in one place.  It'll save us a bunch of time/headache when we get around to implementing an API and configuration based approach for setting this value.

lib/instrumentation/span.js
```diff
@@ -36,6 +36,7 @@ function Span (transaction, name, ...args) {
 
   this.transaction = transaction
   this.name = name || 'unnamed'
+  this.outcome = 'unknown'
```
vigneshshanmugam, line 39: can you move this to `GenericsSpan` class as `outcome` belongs to both Transaction and Span.

test/instrumentation/modules/http/outgoing.js
```diff
@@ -230,6 +231,10 @@ function abortTest (type, handler) {
         t.equal(data.transactions.length, 1, 'has one transaction')
         t.equal(data.spans.length, 1, 'has one span')
         t.equal(data.spans[0].name, `${req.method} localhost:${port}/`, 'has expected span name')
+        if (typeof httpContext.status_code === 'undefined') {
```
vigneshshanmugam, line 234: can you add else case which also checks for partial response? 

https://github.com/elastic/apm-agent-nodejs/blob/5bd3eb79fd806befb52c87bed1e1883229156918/test/instrumentation/modules/http/outgoing.js#L59-L71

vigneshshanmugam (approved):
LGTM 👍  with moving the `outcome` to GenericSpan.

jahtalab:
@astorm , Thanks for the review, I agree with refactoring the logic related to the outcome into a central place 👍 . I've now added `setOutcome` function on Span/Transaction that contains this logic. I think since the `outcome` logic depends on event type (span/transaction), and also it changes its internal state, it would be better to keep the logic in their respective events. Let me know WDYT?

In regards to the other comments, your observations are aligned with what I intended 👍, (I'll mark them as resolved)

lib/instrumentation/http-shared.js
jahtalab: I've add a comment in `span.setOutcome` with regards to the type checking to make it clear

lib/instrumentation/span.js
jahtalab: The `sync` property is not related to outcome in any way. The change here only adds a comma `,`, is that what you meant?

astorm: Ah -- I had misread the diff.  We're all good here.

astorm (approved):
💥 Looks good, thanks for making those changes.  Approving and merging.

PR #1828, opened 2020-10-13
feat: tracestate

### LINKED ISSUE #1797: Record sampling weight on transactions and spans
See elastic/apm#307

We need to record a sampling weight on transactions and spans, propogate the value through `tracestate` header to downstream services.

astorm:
This PR closes #1797.

This pull request implements the tracestate functionality for the elastic node agent.  The new `TraceState` class is capable of getting/setting values under a single list-member of the tracestate string while preserving the values of the other list-members.  It can also produce output in either the W3C string format or the W3C binary format (via a hex string). 

The pull request also uses this new tracestate functionality to add an `es` list-member to the root trace node with an `s` key that records the transaction sample rate.  This header is initially set on the root trace node.  If the transaction is the child of another transaction the priority is not explicitly set, and instead the tracestate header is simply propagated on. 

This pull request also replaces the `TraceParent` class with a `TraceContext` class that includes both a `TraceParent` and `TraceState` functionality. 

The `tracestate` headers is not added to http2 transactions, as http2 transaction do not currently support distributed tracing. See: https://github.com/elastic/apm-agent-nodejs/issues/1830

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/instrumentation/generic-span.js
+++ b/lib/instrumentation/generic-span.js
@@ -4,8 +4,7 @@ const truncate = require('unicode-byte-truncate')
 
 const config = require('../config')
 const Timer = require('./timer')
-const TraceParent = require('traceparent')
-
+const TraceContext = require('../tracecontext')
 module.exports = GenericSpan
 
 function GenericSpan (agent, ...args) {
@@ -14,7 +13,10 @@ function GenericSpan (agent, ...args) {
     : {}
 
   this._timer = new Timer(opts.timer, opts.startTime)
-  this._context = TraceParent.startOrResume(opts.childOf, agent._conf) // _context is used by the OT bridge, and should unfortunately therefore be considered public
+
+  // _context is used by the OT bridge, and should unfortunately therefore be considered public
+  this._context = TraceContext.startOrResume(opts.childOf, agent._conf, opts.tracestate)
+
   this._agent = agent
   this._labels = null
   this._ids = null // Populated by sub-types of GenericSpan
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -24,8 +24,10 @@ exports.instrumentRequest = function (agent, moduleName) {
           agent._instrumentation.currentTransaction = null
         } else {
           var traceparent = req.headers['elastic-apm-traceparent'] || req.headers.traceparent
+          var tracestate = req.headers.tracestate
           var trans = agent.startTransaction(null, null, {
-            childOf: traceparent
+            childOf: traceparent,
+            tracestate: tracestate
           })
           trans.type = 'request'
           trans.req = req
@@ -151,8 +153,11 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       // Use the transaction context as the parent, in this case.
       var parent = span || agent.currentTransaction
       if (parent && parent._context && shouldPropagateTraceContext(options)) {
-        const headerValue = parent._context.toString()
+        const headerValue = parent._context.toTraceParentString()
+        const traceStateValue = parent._context.toTraceStateString()
+
         options.headers.traceparent = headerValue
+        options.headers.tracestate = traceStateValue
         if (agent._conf.useElasticTraceparentHeader) {
           options.headers['elastic-apm-traceparent'] = headerValue
         }
--- a/lib/tracecontext/index.js
+++ b/lib/tracecontext/index.js
@@ -1,0 +1,74 @@
+'use strict'
+const TraceParent = require('traceparent')
+const TraceState = require('./tracestate')
+
+const definePassthroughProp = (objMe, objToPassTo, propName) => {
+  // properties that call through to traceparent
+  Object.defineProperty(objMe, propName, {
+    configurable: true,
+    enumerable: true,
+    get () {
+      return objToPassTo[propName]
+    }
+  })
+}
+
+class TraceContext {
+  constructor (traceparent, tracestate) {
+    this.traceparent = traceparent
+    this.tracestate = tracestate
+
+    definePassthroughProp(this, this.traceparent, 'recorded')
+    definePassthroughProp(this, this.traceparent, 'version')
+    definePassthroughProp(this, this.traceparent, 'traceId')
+    definePassthroughProp(this, this.traceparent, 'id')
+    definePassthroughProp(this, this.traceparent, 'flags')
+    definePassthroughProp(this, this.traceparent, 'parentId')
+  }
+
+  static startOrResume (childOf, conf, tracestateString) {
+    if (childOf && childOf._context instanceof TraceContext) return childOf._context.child()
+    const traceparent = TraceParent.startOrResume(childOf, conf)
+    const tracestate = TraceState.fromStringFormatString(tracestateString)
+    tracestate.setValue('s', conf.transactionSampleRate)
+
+    return new TraceContext(traceparent, tracestate)
+  }
+
+  static fromString (header) {
+    return TraceParent.fromString(header)
+  }
+
+  ensureParentId () {
+    return this.traceparent.ensureParentId()
+  }
+
+  child () {
+    const childTraceParent = this.traceparent.child()
+    const childTraceContext = new TraceContext(
+      childTraceParent, this.tracestate
+    )
+    return childTraceContext
+  }
+
+  /**
+   * Returns traceparent string only
+   *
+   * @todo legacy -- can we remove to avoid confusion?
+   */
+  toString () {
+    return this.traceparent.toString()
+  }
+
+  toTraceStateString () {
+    return this.tracestate.toW3cString()
+  }
+
+  toTraceParentString () {
+    return this.traceparent.toString()
+  }
+}
+
+TraceContext.FLAGS = TraceParent.FLAGS
+
+module.exports = TraceContext
--- a/lib/tracecontext/tracestate.js
+++ b/lib/tracecontext/tracestate.js
@@ -1,0 +1,346 @@
+/**
+ * Class for Managing Tracestate
+ *
+ * Class that creates objects for managing trace state.
+ * This class is capable of parsing both tracestate strings
+ * and tracestate binary representations, allowing clients
+ * to get and set values in a single list-member/namespace
+ * while preserving values in the other namespaces.
+ *
+ * Capable of working with either the binary of string
+ * formatted tracestate values.
+ *
+ * Usage:
+ *   const tracestate = TraceState.fromStringFormatString(headerTracestate, 'es')
+ *   tracestate.setValue('s',1)
+ *   const newHeader = tracestate.toW3cString()
+ */
+class TraceState {
+  constructor (sourceBuffer, listMemberNamespace = 'es', defaultValues = {}, logger) {
+    if (!this._validateVendorKey(listMemberNamespace)) {
+      throw new Error('Vendor namespace failed validation.')
+    }
+
+    // buffer representation of the trace state string.
+    // The initial value of this.buffer will keep the
+    // values set in the listMemberNamespace list-member,
+    // but as soon as an initial value is set (via setValue)
+    // then the listMemberNamespace values will be removed
+    // from this.buffer and stored in the this.values.  While
+    // slightly more complicated, this allows us to maintain
+    // the order of list-member keys in an un-mutated tracestate
+    // string, per the W3C spec
+    this.buffer = sourceBuffer
+
+    this.listMemberNamespace = listMemberNamespace
+
+    // values for our namespace, set via setValue to
+    // ensure names conform
+    this.values = {}
+    for (const [key, value] of Object.entries(defaultValues)) {
+      this.setValue(key, value)
+    }
+
+    // if no logger, then fallback to agent singleton
+    logger = logger || require('../..').logger
+    this.logger = logger
+  }
+
+  setValue (key, value) {
+    const strKey = String(key)
+    const strValue = String(value)
+    if (!this._validateElasicKeyAndValue(strKey, strValue)) {
+      this.logger.trace('could not set tracestate key, invaliad characters detected')
+      return false
+    }
+    const isFirstSet = (Object.keys(this.values).length === 0)
+    const oldValue = this.values[strKey]
+    this.values[strKey] = value
+
+    // per: https://github.com/elastic/apm/blob/d5b2c87326548befcfec6731713932a00e430b99/specs/agents/tracing-distributed-tracing.md
+    // If adding another key/value pair to the es entry would exceed the limit
+    // of 256 chars, that key/value pair MUST be ignored by agents.
+    // The key/value and entry separators : and ; have to be considered as well.
+    const serializedValue = this._serializeValues(this.values)
+    if (serializedValue.length > 256 && (typeof oldValue === 'undefined')) {
+      delete this.values[strKey]
+      return false
+    }
+    if (serializedValue.length > 256 && (typeof oldValue !== 'undefined')) {
+      this.values[strKey] = oldValue
+      return false
+    }
+
+    // the first time we set a value, extract the mutable values from the
+    // buffer and set this.values appropriately
+    if (isFirstSet && Object.keys(this.values).length === 1) {
+      const [buffer, values] = TraceState._removeMemberNamespaceFromBuffer(
+        this.buffer,
+        this.listMemberNamespace
+      )
+      this.buffer = buffer
+      this.values = values
+      values[strKey] = value
+    }
+
+    return true
+  }
+
+  getValue (keyToGet) {
+    const allValues = this.toObject()
+    const rawValue = allValues[this.listMemberNamespace]
+    const values = TraceState._parseValues(rawValue)
+    return values[keyToGet]
+  }
+
+  toHexString () {
+    const newBuffer = Buffer.alloc(this.buffer.length)
+    let newBufferOffset = 0
+    for (let i = 0; i < this.buffer.length; i++) {
+      const byte = this.buffer[i]
+      if (byte === 0) {
+        const indexOfKeyLength = i + 1
+        const indexOfKey = i + 2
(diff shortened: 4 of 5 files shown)


lib/instrumentation/generic-span.js
```diff
@@ -14,7 +13,10 @@ function GenericSpan (agent, ...args) {
     : {}
 
   this._timer = new Timer(opts.timer, opts.startTime)
-  this._context = TraceParent.startOrResume(opts.childOf, agent._conf) // _context is used by the OT bridge, and should unfortunately therefore be considered public
+
+  // _context is used by the OT bridge, and should unfortunately therefore be considered public
+  this._context = TraceContext.startOrResume(opts.childOf, agent._conf, opts.tracestate)
+
```
astorm, line 19: Here's where the old TraceParent object was added to a `Transaction` or `Span`  object.  We changed this to use a `TraceContext` object instead. `TraceContext` behaves the same as `TraceParent`, but all accepts a `tracestate` header to parse.

lib/instrumentation/http-shared.js
```diff
@@ -24,8 +24,10 @@ exports.instrumentRequest = function (agent, moduleName) {
           agent._instrumentation.currentTransaction = null
         } else {
           var traceparent = req.headers['elastic-apm-traceparent'] || req.headers.traceparent
+          var tracestate = req.headers['elastic-apm-tracestate'] || req.headers.tracestate
           var trans = agent.startTransaction(null, null, {
-            childOf: traceparent
+            childOf: traceparent,
+            tracestate: tracestate
           })
```
astorm, line 31: Here's where we parse either the `elastic-apm-tracestate` or `tracestate` header (preferring the former) looking for `tracestate` values, and then pass them along to the `startTranasction` method.

lib/instrumentation/http-shared.js
```diff
@@ -151,8 +153,11 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       // Use the transaction context as the parent, in this case.
       var parent = span || agent.currentTransaction
       if (parent && parent._context && shouldPropagateTraceContext(options)) {
-        const headerValue = parent._context.toString()
+        const headerValue = parent._context.toTraceParentString()
+        const traceStateValue = parent._context.toTraceStateString()
+
         options.headers.traceparent = headerValue
+        options.headers.tracestate = traceStateValue
         if (agent._conf.useElasticTraceparentHeader) {
```
astorm, line 161: Here's where we _set_ the `tracestate` header on outgoing http transactions.  We are _not_ setting or giving an option to set the old `elastic-apm-tracestate` header in an attempt to keep things simple.  If a customer ends up needed this functionality we can add it later with relative simplicity.

lib/tracecontext/index.js
```diff
@@ -0,0 +1,74 @@
+'use strict'
```
astorm, line 1: This is our new `TraceContext` class, replacing the `TraceParent` class.  This object composes together a `TraceParent` and `TraceState` class to do its work (parsing of the header, starting a new trace if necessary, re-serializing the headers as strings for outgoing transactions)

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,303 @@
...
+ * while preserving values in the other namespaces.
+ *
+ * Capable of working with either the binary of string
+ * formatted tracestate values.
+ *
+ * Usage:
+ *   const tracestate = TraceState.fromStringFormatString(headerTracestate, 'es')
+ *   tracestate.setValue('s',1)
+ *   const newHeader = tracestate.toW3cString()
+ */
```
astorm, line 17: The class is simpler than it seems -- the above usage is all an end user of the class really needs to know.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+      this.values
+    )
+    return json
+  }
+
+  toString () {
+    return this.toW3cString()
+  }
+
+  toW3cString () {
```
jahtalab, line 167: The w3c spec states that the order of key/value pairs need to be preserved and that the modified keys need to be moved to the start:

> Vendors receiving a tracestate request header MUST send it to outgoing requests. It MAY mutate the value of this header before passing to outgoing requests. When mutating tracestate, the order of unmodified key/value pairs MUST be preserved. Modified keys MUST be moved to the beginning (left) of the list.

I think using `toJson` here might loose that order, since the `Object.entries` doesn't guarantee the order of the keys.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+        const lengthKey = buffer[indexOfKeyLength]
+
+        const indexOfValueLength = indexOfKey + lengthKey
+        const indexOfValue = indexOfValueLength + 1
+        const lengthValue = buffer[indexOfValueLength]
+
+        const key = buffer.slice(indexOfKey, indexOfKey + lengthKey).toString()
+
+        // if this is our mutable namespace extract
+        // and set the value in vlaues, otherwise
```
jahtalab, line 279: typo: `values`

jahtalab (commented):
Thanks @astorm , Generally looks good, I've made some comments. 
I think we don't need to check for `elastic-apm-tracestate` header, unless any of our agents use that header before (The RUM agent doesn't, but I'm not sure about others)
Also, added a comment with regards to the order of key/values pairs in the tracestate header.

Let me know if you want to discuss these further.

lib/instrumentation/http-shared.js
```diff
@@ -24,8 +24,10 @@ exports.instrumentRequest = function (agent, moduleName) {
           agent._instrumentation.currentTransaction = null
         } else {
           var traceparent = req.headers['elastic-apm-traceparent'] || req.headers.traceparent
+          var tracestate = req.headers['elastic-apm-tracestate'] || req.headers.tracestate
```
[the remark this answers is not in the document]
vigneshshanmugam: ++ to dropping the `elastic-apm-tracestate`. We can reintroduce if required, but right now we don't need it.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+    if (!this._validateElasicKeyAndValue(strKey, strValue)) {
+      this.logger.trace('could not set tracestate key, invaliad characters detected')
+      return false
+    }
+
+    const oldValue = this.values[strKey]
+    this.values[strKey] = value
+
+    // if new length is greater than 256, undo the setting
+    const serializedValue = this._serializeValues(this.values)
```
vigneshshanmugam, line 49: I am kind of confused with this interpretation. https://www.w3.org/TR/trace-context/#tracestate-limits

Do we need to account for this during setting or during serialisation?

vigneshshanmugam (commented):
@astorm  I would use `Map` to preserve the order to make it easier.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+    if (serializedValue.length > 256 && (typeof oldValue === 'undefined')) {
+      delete this.values[strKey]
+    }
+    if (serializedValue.length > 256 && (typeof oldValue !== 'undefined')) {
+      this.values[strKey] = oldValue
+    }
```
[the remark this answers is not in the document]
astorm: > Shouldn't this method return false if the new value is not set due to the length?

It should, and this probably helped confused @vigneshshanmugam -- I'll get that fixed

lib/tracecontext/tracestate.js
astorm: _nod_ -- this isn't straight forward, but [the elastic distributed tracing spec](https://github.com/elastic/apm/blob/d5b2c87326548befcfec6731713932a00e430b99/specs/agents/tracing-distributed-tracing.md) includes the following.

> If adding another key/value pair to the es entry would exceed the limit of 256 chars, that key/value pair MUST be ignored by agents. The key/value and entry separators : and ; have to be considered as well.

If the serialized value is greater than 256 characters, then the value needs to be ignored.  We _could_ handle this purely at serialization time by using a `Map` to store keys (to ensure their order), and then serialize until things got too long.  However, doing so would mean `setValue` would return `true` for a value that ends up not being set (which @jahtalab pointed out this code fails to do, which I intend to fix)   

That's why I chose this implementation -- it's _slightly_ more work each time a value is set, but we're able to signal out that a value reached the limit.  

My preference would be to keep this implementation, but it's only a slight preference.  If you think we need something else let's talk offline/slack about it.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+      if (!value) { continue }
+      chars.push(key)
+      chars.push('=')
+      chars.push(value)
+      chars.push(',')
+    }
+    return chars.join('')
+  }
+
+  _serializeValues (string) {
```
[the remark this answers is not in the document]
astorm: 👍

lib/tracecontext/tracestate.js
astorm: 👍

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+      newBuffer.length + ourBuffer.length
+    ).toString('hex')
+  }
+
+  /**
+   * Returns JSON reprenstation of tracestate key/value pairs
+   *
+   * Does not parse the mutable list namespace
+   */
+  toJson () {
```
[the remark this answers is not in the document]
astorm: fair -- a bad habit from an old team where we refereed to plain javascript objects `{}` as json.  Changing this method name to `toObject` to better reflect what it does.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,309 @@
...
+
+  toW3cString () {
+    const json = this.toJson()
+    const chars = []
+    for (const [key, value] of Object.entries(json)) {
+      if (!value) { continue }
+      chars.push(key)
+      chars.push('=')
+      chars.push(value)
+      chars.push(',')
```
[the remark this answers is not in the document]
astorm: _nod_ -- best to be tidy here.  I'll skip the last comma.

lib/instrumentation/http-shared.js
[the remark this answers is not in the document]
astorm: Good call -- no need to introduce legacy, pre-spec code unless there's a specific need to (and it will be relativly easy to add if there is).  Changed this to check `req.headers.tracestate` only.

lib/tracecontext/tracestate.js
astorm: Excellent catch -- I had not registered that portion of the specification. 

Also, in addition to the problem of plain objects not having a guaranteed order, this made me realize that we wern't moving our mutated `es` list-member to the left to the left side of the new `tracestate` header AND that we weren't maintaining the order of an unmutated `es` header. 

I've introduced a `toMap` method that ensure that the original order of the `tracestate` list-members is maintained, and added some new logic to ensure we don't remove `es` from the raw buffer until the first value is set.  This should ensure we conform to the letter of the W3C spec. There are also new tests to confirm the order is maintained/changed as specified.

jahtalab (commented):
@astorm , just one comment.

Also have you seen the [trace-context python tests](https://github.com/w3c/trace-context/tree/master/test). I think the tests we have in this PR are already good but it might be worth having a quick look at those as well.

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,346 @@
...
+    // the order of list-member keys in an un-mutated tracestate
+    // string, per the W3C spec
+    this.buffer = sourceBuffer
+
+    this.listMemberNamespace = listMemberNamespace
+
+    // values for our namespace, set via setValue to
+    // ensure names conform
+    this.values = {}
+    for (const [key, value] of Object.entries(this.values)) {
```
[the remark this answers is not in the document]
astorm: Good catch here -- the previous rejiggering of code means the `defaultValues` parameter isn't be exercised at all, so this wasn't caught by tests. Thinking out loud -- even though it's not being exercised I think keeping this param around makes sense, if only to give future us a way to create TraceState objects without needing to use populated buffer objects.  I've fixed the code and added some tests to make sure it works.  Thanks again for catching this.

jahtalab (dismissed):

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,346 @@
...
+
+    // if no logger, then fallback to agent singleton
+    logger = logger || require('../..').logger
+    this.logger = logger
+  }
+
+  setValue (key, value) {
+    const strKey = String(key)
+    const strValue = String(value)
+    if (!this._validateElasicKeyAndValue(strKey, strValue)) {
```
vigneshshanmugam, line 52: Just for my understanding, Is this applicable only to elastic namespace or every other vendor key - value?

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,346 @@
...
+    const isFirstSet = (Object.keys(this.values).length === 0)
+    const oldValue = this.values[strKey]
+    this.values[strKey] = value
+
+    // per: https://github.com/elastic/apm/blob/d5b2c87326548befcfec6731713932a00e430b99/specs/agents/tracing-distributed-tracing.md
+    // If adding another key/value pair to the es entry would exceed the limit
+    // of 256 chars, that key/value pair MUST be ignored by agents.
+    // The key/value and entry separators : and ; have to be considered as well.
+    const serializedValue = this._serializeValues(this.values)
+    if (serializedValue.length > 256 && (typeof oldValue === 'undefined')) {
```
vigneshshanmugam, line 65: May be merge the below two if statements?

test/tracecontext/tracestate.js
```diff
@@ -0,0 +1,289 @@
...
+    'foo=34f067aa0ba902b7,bar=0.25,es=a:b;cee:de,;34@ree=xxxy,',
+    'foo=34f067aa0ba902b7,bar=0.25,es=a:b;cee:de;,34@ree=xxxy,',
+    'foo=34f067aa0ba902b7,,bar=0.25,es=a:b;;cee:de,34@ree=xxxy;',
+    'foo=34f067aa,0ba902b7,,bar=0.25,es=a:b;;cee:de,34@ree=xxxy;',
+    'foo=34f067aa,0ba9,02b7,,bar=0.25,es=a:b;;;cee:de,,,34@ree=xxxy;',
+    'foo=34f067aa0ba902b7,bar=0.25,es=a::b;cee:de,34@ree=xxxy',
+    'foo=34f067aa0ba902b7,bar=0.25,es=a:b;cee:de,es=xxxy'
+  ]
+  for (const [, format] of stringFormats.entries()) {
+    this.ok(TraceState.fromStringFormatString(format))
```
vigneshshanmugam, line 240: nit: `t.ok`

lib/tracecontext/tracestate.js
```diff
@@ -0,0 +1,346 @@
...
+  static fromBinaryFormatHexString (string, listMemberNamespace = 'es') {
+    return new TraceState(Buffer.from(string, 'hex'), listMemberNamespace)
+  }
+
+  static fromStringFormatString (string = '', listMemberNamespace = 'es') {
+    // converts string format to byte format
+    const bytes = []
+
+    const parts = string.split(',')
+    for (const [, part] of parts.entries()) {
```
vigneshshanmugam, line 269: May be just use part.values?

vigneshshanmugam (approved):
Looks good @astorm 🎉 , code wise looks good to me, Haven't really verified 1-1 with the official spec though.

lib/tracecontext/tracestate.js
astorm: Good question -- it's applicable only to the elastic namespace.  The intention of the `setValue` method is to allow users to set values in the elastic namespace.  The are deliberately no methods for setting values in the other namespaces (other than the initial value set when creating the tracestate object)

test/tracecontext/tracestate.js
astorm: Looks like there's more than on occurrence of this :facepalm: -- surprised that `this.` actually works -- I guess `tape` binds `this` to -- a? the current? -- test.  

Good catch -- better to be explicit here.   This is fixed now, all tests use `t` object.

lib/tracecontext/tracestate.js
astorm: _nod_ that seems like an OK change.  I generally prefer stand alone `if` statements (no `else/else if` etc.) to avoid super complicated conditional logic -- but this isn't that complicated _and_ with the `else` we avoid a few extra instructions on ever `set`.  Made this change.

lib/tracecontext/tracestate.js
astorm: I presume that's a suggestion to use `parts.values()` in the `for` -- if I've misread that let me know :) 

Good callout -- that said the choice to use `entries` here (and elsewhere) was deliberate/explicit.  I've found it's simpler to 

1. _Always_ use either `entries()` or `Object.entries()`  
2. with the const `[...]` form
3. with a `for of` loop

The main benefit is with your loops more consistent/explicit you don't need to stop and think about what's happening in a particular loop. i.e. "What's being looped over" or "is a single variable the index or the value", etc.

Going to resolve this, but my preference is a minor one.  If there's a compelling reason to use `values()` (perf., etc)  I'm happy to switch this up.

astorm:
> Also have you seen the trace-context python tests. I think the tests we have in this PR are already good but it might be worth having a quick look at those as well.

Thanks for pointing these out @jahtalab -- I manually validated some of the base tracecontext functionality using these tests.  However -- it looks like the Agent's traceparent implementation doesn't pass these tests at full strictness level, which makes the tests that rely on both traceparent and tracestate functionality difficult to break apart into just tracestate functionality.  Backporting this suite to the current traceparent functioanlity feels out of scope for this specific issue, so I've opened a new one so we don't lose track of this: https://github.com/elastic/apm-agent-nodejs/issues/1837

vigneshshanmugam (dismissed):

PR #1843, opened 2020-10-28
fix(tests): skip testing knex >=0.21 with node v8

### LINKED ISSUE #1842: ci: knex@0.21.8 cannot be tested with node v8
**Describe the bug**

The knex package has dropped support for Node 8 and introduced code that breaks Node 8.  This has broken the build.  No CI can pass until this is resolved.

    /home/travis/build/elastic/apm-agent-nodejs/node_modules/knex/lib/dialects/postgres/schema/columncompiler.js:7
    const commentEscapeRegex = /(?<!')'(?!')/g;
    SyntaxError: Invalid regular expression: /(?<!')'(?!')/: Invalid group

trentm:
Fixes #1842

### DIFF
--- a/test/instrumentation/modules/pg/knex.js
+++ b/test/instrumentation/modules/pg/knex.js
@@ -13,6 +13,11 @@ var agent = require('../../../..').start({
 var knexVersion = require('knex/package').version
 var semver = require('semver')
 
+if (semver.gte(knexVersion, '0.21.0') && semver.lt(process.version, '10.0.0')) {
+  // Skip these tests: knex@0.21.x dropped support for node v8.
+  process.exit(0)
+}
+
 var Knex = require('knex')
 var test = require('tape')
 


test/instrumentation/modules/pg/knex.js
```diff
@@ -13,6 +13,11 @@ var agent = require('../../../..').start({
 var knexVersion = require('knex/package').version
 var semver = require('semver')
 
+if (semver.gte(knexVersion, '0.21.0') && semver.lt(process.version, '10.0.0')) {
+  // Skip these tests: knex@0.21.x dropped support for node v8.
+  process.exit(0)
+}
+
```
astorm, line 20: I see a conditional that will exit with a success code of `0` if the knexVersion is greater than or equal to 0.21, and if the node version is less than 10. 👍

astorm (approved):
Approving -- looks like it does exactly what we want/need it to do w/r/t the new version of `ioredis` that dropped support for version 8 of NodeJS.

test/instrumentation/modules/pg/knex.js
trentm: If we switched (someday) from `tape` to `tap` we could use its more reliable TAP `skip()` handling to get actual TAP output that showed a message about skipping tests.  But obviously that's for another day.

PR #1852, opened 2020-11-03
fix: redis@2.x instrumentation was broken

### LINKED ISSUE #1850: "redis@2.x" TAV test failures
"redis@2.x" TAV tests are failing.

```
% TAV=redis ./node_modules/.bin/tav --compat
Testing compatibility with redis:
✔ 3.0.2
✔ 3.0.1
✔ 3.0.0
✖ 2.8.0
✖ 2.7.1
✖ 2.7.0
✖ 2.6.5
✖ 2.6.4
✖ 2.6.3
✖ 2.6.2
✖ 2.6.1
✖ 2.6.0
✖ 2.5.3
✖ 2.5.2
✖ 2.5.1
✖ 2.5.0
✖ 2.4.2
✖ 2.4.1
✖ 2.4.0
✖ 2.3.1
✖ 2.3.0
✖ 2.2.5
✖ 2.2.4
✖ 2.2.3
✖ 2.2.2
✖ 2.2.1
✖ 2.2.0
✖ 2.1.0
✖ 2.0.1
✖ 2.0.0
```

Output from `TAV=redis ./node_modules/.bin/tav --quiet | tee tav-redis.tap`: https://gist.github.com/trentm/eed181febf2f8ea9555ed6c9af2144ee

trentm:
- instrumentation of redis 2.x was accidentally in v3.7.0
- before the addition of destination metadata in commit aa8eb56 (v3.6.0)
  didn't work for redis 2.3.0 .. 2.5.3

Fixes #1850

### DIFF
--- a/lib/instrumentation/modules/redis.js
+++ b/lib/instrumentation/modules/redis.js
@@ -6,9 +6,23 @@ var shimmer = require('../shimmer')
 var { getDBDestination } = require('../context')
 
 module.exports = function (redis, agent, { version, enabled }) {
-  if (!semver.satisfies(version, '^3.0.0')) {
+  if (!semver.satisfies(version, '>=2.0.0 <4.0.0')) {
     agent.logger.debug('redis version %s not supported - aborting...', version)
     return redis
+  }
+
+  // The undocumented field on a RedisClient instance on which connection
+  // options are stored has changed a few times.
+  let connOptsFromRedisClient
+  if (semver.gte(version, '2.4.0')) {
+    connOptsFromRedisClient = (rc) => rc.connection_options
+  } else if (semver.gte(version, '2.3.0')) {
+    connOptsFromRedisClient = (rc) => rc.connection_option
+  } else if (semver.gte(version, '0.12.0')) {
+    connOptsFromRedisClient = (rc) => rc.connectionOption
+  } else {
+    // *Maybe* available on `client.{host|port}`, but not verified.
+    connOptsFromRedisClient = () => {}
   }
 
   var proto = redis.RedisClient && redis.RedisClient.prototype
@@ -41,8 +55,10 @@ module.exports = function (redis, agent, { version, enabled }) {
 
       if (span) {
         let host, port
-        if (typeof this.connection_options === 'object') {
-          ({ host, port } = this.connection_options)
+        let connOpts = connOptsFromRedisClient(this)
+        if (connOpts) {
+          host = connOpts.host
+          port = connOpts.port
         }
         span.setDestinationContext(getDBDestination(span, host, port))
       }
@@ -67,8 +83,10 @@ module.exports = function (redis, agent, { version, enabled }) {
 
       if (span) {
         let host, port
-        if (typeof this.connectionOption === 'object') {
-          ({ host, port } = this.connectionOption)
+        let connOpts = connOptsFromRedisClient(this)
+        if (connOpts) {
+          host = connOpts.host
+          port = connOpts.port
         }
         span.setDestinationContext(getDBDestination(span, host, port))
       }
--- a/lib/instrumentation/modules/redis.js
+++ b/lib/instrumentation/modules/redis.js
@@ -13,17 +13,13 @@ module.exports = function (redis, agent, { version, enabled }) {
 
   // The undocumented field on a RedisClient instance on which connection
   // options are stored has changed a few times.
-  let connOptsFromRedisClient
-  if (semver.gte(version, '2.4.0')) {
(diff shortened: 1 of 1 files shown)


trentm:
Here is a redis local TAV run with this patch:

```
% TAV=redis ./node_modules/.bin/tav --compat
Testing compatibility with redis:
✔ 3.0.2
✔ 3.0.1
✔ 3.0.0
✔ 2.8.0
✔ 2.7.1
✔ 2.7.0
✔ 2.6.5
✔ 2.6.4
✔ 2.6.3
✔ 2.6.2
✔ 2.6.1
✔ 2.6.0
✔ 2.5.3
✔ 2.5.2
✔ 2.5.1
✔ 2.5.0
✔ 2.4.2
✔ 2.4.1
✔ 2.4.0
✔ 2.3.1
✔ 2.3.0
✔ 2.2.5
✔ 2.2.4
✔ 2.2.3
✔ 2.2.2
✔ 2.2.1
✔ 2.2.0
✔ 2.1.0
✔ 2.0.1
✔ 2.0.0
```

lib/instrumentation/modules/redis.js
```diff
@@ -6,11 +6,25 @@ var shimmer = require('../shimmer')
...
+    connOptsFromRedisClient = (rc) => rc.connection_options
+  } else if (semver.gte(version, '2.3.0')) {
+    connOptsFromRedisClient = (rc) => rc.connection_option
+  } else if (semver.gte(version, '0.12.0')) {
+    connOptsFromRedisClient = (rc) => rc.connectionOption
+  } else {
+    // *Maybe* available on `client.{host|port}`, but not verified.
+    connOptsFromRedisClient = () => {}
+  }
+
```
astorm, line 27: Instead of using version sniffing here, what do you think about using feature sniffing?  I've been generally sympathetic to the idea that keeping track of features (especially undocumented or private APIs) by versions leads to more complex code over time, and can sometimes bite you when/if a future version reverts back to an old prop name.  So instead of this version checking, we might do something like this 

```
const connOptsFromRedisClient = (rc) => {    
    if(rc.connection_options) {         
        return rc.connection_options
    }
    if(rc.connection_option) {          
        return rc.connection_option
    }
    if(rc.connectionOption) {
        return rc.connectionOption
    }
    if(rc.connection_options) {
        return rc.connection_options
    }       
    
    return {} 
}
```

lib/instrumentation/modules/redis.js
```diff
@@ -41,8 +55,10 @@ module.exports = function (redis, agent, { version, enabled }) {
 
       if (span) {
         let host, port
-        if (typeof this.connection_options === 'object') {
-          ({ host, port } = this.connection_options)
+        let connOpts = connOptsFromRedisClient(this)
```
astorm, line 58: nit: `connOpts` could probably be `const` here.

lib/instrumentation/modules/redis.js
```diff
@@ -67,8 +83,10 @@ module.exports = function (redis, agent, { version, enabled }) {
 
       if (span) {
         let host, port
-        if (typeof this.connectionOption === 'object') {
-          ({ host, port } = this.connectionOption)
+        let connOpts = connOptsFromRedisClient(this)
```
astorm, line 86: nit: `connOpts` could probably be `const` here.

astorm (dismissed):
Overall looks good -- nice job tracking down the failures.  

I left a few comments about a potential different approach to version sniffing and some nits/tics w.r.t to `let` vs. `const`.  Nothing that would stop a merge though, so approving and leaving them both to dealer's choice :)

lib/instrumentation/modules/redis.js
trentm: I can generally get behind biasing towards feature-sniffing over version-switching. Initially I'd used this approach to try to do the detection in the init, rather than having to feature-sniff for each call of the wrapped send command. However that is likely premature optimization.

I doubt node-redis will revert back to an old prop name. One (again unlikely) benefit of feature detection is if someone is playing with a personal build of node-redis that isn't one of the tagged releases.

Objections to shortening the above to:

```
const connOptsFromRedisClient = (rc) => rc.connection_options || rc.connection_option || rc.connectionOption || {}
```

?

trentm:
jenkins run the module tests for redis

lib/instrumentation/modules/redis.js
astorm: > premature optimization

fair and reasonable. 

> Objections to shortening the above to:

zero objections, have at it.

astorm (approved):

Issue #1851, opened 2020-10-30
"ws@1.x" TAV test failures


trentm:
`TAV=ws ./node_modules/.bin/tav --quiet | tee tav-ws.tap` output: https://gist.github.com/trentm/6a92a1fcf63ab9f116bb9f01cbdd1297

```
% TAV=ws ./node_modules/.bin/tav --compat
Testing compatibility with ws:
✔ 7.3.1
✔ 7.3.0
✔ 7.2.5
✔ 7.2.3
✔ 7.2.2
✔ 7.2.1
✔ 7.2.0
✔ 7.1.2
✔ 7.1.1
✔ 7.1.0
✔ 7.0.1
✔ 7.0.0
✔ 6.2.1
✔ 6.2.0
✔ 6.1.4
✔ 6.1.3
✔ 6.1.2
✔ 6.1.1
✔ 6.1.0
✔ 6.0.0
✔ 5.2.2
✔ 5.2.1
✔ 5.2.0
✔ 5.1.1
✔ 5.1.0
✔ 5.0.0
✔ 4.1.0
✔ 4.0.0
✔ 3.3.3
✔ 3.3.2
✔ 3.3.1
✖ 1.1.5
✔ 3.3.0
✔ 3.2.0
✔ 3.1.0
✔ 3.0.0
✔ 2.3.1
✔ 2.3.0
✔ 2.2.3
✔ 2.2.2
✔ 2.2.1
✖ 1.1.4
✖ 1.1.3
✔ 2.2.0
✔ 2.1.0
✖ 1.1.2
✔ 2.0.3
✔ 2.0.2
✔ 2.0.1
✔ 2.0.0
✖ 1.1.1
✖ 1.1.0
✖ 1.0.1
✖ 1.0.0
```

PR #1855, opened 2020-11-05
ci: fix ws@1.x TAV tests

astorm:
Short: The PR modifies this test's use of the the `mockClient` ___ parameter, and instead relies on the default mock client test ending behavior.  This fixes the TAV tests for the `ws@1.x` module, at the cost of no longer testing that there were exactly two writes to the HTTP client.

Long:  The `ws@1.x` failures were introduced [in this PR](https://github.com/elastic/apm-agent-nodejs/commit/0d0eeaa7a46ba721e509ac3c4c4d9b05e4b09893#diff-7cb2e6576b8a433a53a30518c4f99782e993baecc8bcdcdb1086f22f7c248798), when we moved to tape 5.  

Tape 5 fixed [a long standing bug](https://github.com/substack/tape/commit/e925bb15b5f21eac357a3f86170da1241cca3439) where tape didn't enforce `t.end()`.  This `tape` fix revealed that the `ws.js` tests have haven been incorrectly written, in that test assertions are made after `t.end()` is called, but only in the ws@1.x branch.  This has been going on for an unknown amount of time and for unknown reasons.

Attempting to bisect back with older `ws.js` tests running against tape 5 [led to this commit](https://github.com/elastic/apm-agent-nodejs/commit/e14a07e651daf238d116a81b2a49cde89c11f05e), where the `ws.js` test fails against the 1.x branch due to some error in the HTTP client (likely a different version at the time this commit landed) 

    % node test/instrumentation/modules/ws.js
    /path/to/apm-agent-nodejs/node_modules/elastic-apm-http-client/index.js:354
      if (missing.length > 0) throw new Error('Missing required option(s): ' + missing.join(', '))
                              ^

    Error: Missing required option(s): agentName, agentVersion, serviceName
        at normalizeOptions (/path/to/apm-agent-nodejs/node_modules/elastic-apm-http-client/index.js:354:33)
        at new Client (/path/to/apm-agent-nodejs/node_modules/elastic-apm-http-client/index.js:52:23)
        at Agent.start (/path/to/apm-agent-nodejs/lib/agent.js:134:22)
        at Object.<anonymous> (/path/to/apm-agent-nodejs/test/instrumentation/modules/ws.js:3:33)
        at Module._compile (internal/modules/cjs/loader.js:1063:30)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
        at Module.load (internal/modules/cjs/loader.js:928:32)
        at Function.Module._load (internal/modules/cjs/loader.js:769:14)
        at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
        at internal/main/run_main_module.js:17:47

This does not look like our culprit but does block us from using git bisect further to track this down. 

It feels like we're running up against diminishing returns here, so the tradeoff of giving up the "exactly two writes" test seems like a reasonable one, especially since the other assertions are passing.

### DIFF
--- a/test/instrumentation/modules/ws.js
+++ b/test/instrumentation/modules/ws.js
@@ -67,6 +67,6 @@ function done (t) {
 
 function resetAgent (cb) {
   agent._instrumentation.currentTransaction = null
-  agent._transport = mockClient(2, cb)
+  agent._transport = mockClient(cb)
   agent.captureError = function (err) { throw err }
 }


trentm:
Note to self: This is to fix https://github.com/elastic/apm-agent-nodejs/issues/1851

trentm:
I *think* this is a race in test/instrumentation/modules/ws.js between (a) the ` ws.on('message',` for "pong" and (b) the APM agent sending the second message (the span) to its transport (our mockClient). It wasn't trivial to diff the 1.x and 2.x lines of `ws` so I'm timeboxing digging in as well just for fixing a test.  IOW, your workaround sounds good to me.

trentm (approved):

Issue #1859, opened 2020-11-06
allow download-json-schemas.sh to work on macOS without needing gtar


trentm:
Currently a first run `npm run test` will fail on macOS unless you happen to have `gtar` (GNU tar which supports `--wildcards`) installed -- e.g. via homebrew.

```
% npm run test
...
running (cwd: ./): node --require /Users/trentm/el/apm-agent-nodejs/test/_promise_rejection.js test/integration/api-schema/basic.js
downloading schemas from GitHub to /var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache...
download failed. cleaning up...
/Users/trentm/el/apm-agent-nodejs/test/integration/api-schema/basic.js:20
  if (err) throw err
           ^
Error: Command failed: "/Users/trentm/el/apm-agent-nodejs/test/integration/api-schema/download-json-schemas.sh" "/var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache"
++ '[' -n /var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache ']'
++ download_schema /var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache master
++ rm -rf /var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache
++ mkdir -p /var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache
++ for run in 1 2 3 4 5
+++ command -v gtar
++ '[' -x '' ']'
++ curl --silent --fail https://codeload.github.com/elastic/apm-server/tar.gz/master
++ tar xzvf - --wildcards --directory=/var/folders/kj/pd34mxx52575q85q6x13nmm40000gn/T/.schemacache --strip-components=1 '*/docs/spec/*'
tar: Option --wildcards is not supported
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
    at ChildProcess.exithandler (child_process.js:294:12)
    at ChildProcess.emit (events.js:198:13)
    at maybeClose (internal/child_process.js:982:16)
    at Socket.stream.socket.on (internal/child_process.js:389:11)
    at Socket.emit (events.js:198:13)
    at Pipe._handle.close (net.js:607:12)
[...]

PR #1862, opened 2020-11-07
fix: remove gtar dep

astorm:
Fixes: #1859

MacOS `tar` has `--wildcards` behavior by default.  This PR sniffs for Darwin (I don't know of a an easy/clean feature sniff for a tar flag) and then invokes the same tar command, just without the `--wildcards`.

Tested locally, but since none of our CI a second set of eyes/fingers would be appreciated here. 

# Checklist

- [x] Implement code

### DIFF
--- a/test/integration/api-schema/download-json-schemas.sh
+++ b/test/integration/api-schema/download-json-schemas.sh
@@ -10,7 +10,7 @@ if ! [ -n "$1" ]; then
     exit 1
 fi
 
-# download_schema [/path/to/download/to] [branch]
+# download_schema </path/to/download/to> <branch>
 download_schema()
 {
     rm -rf ${1} && mkdir -p ${1}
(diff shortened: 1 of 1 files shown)


astorm:
jenkins run the tests

test/integration/api-schema/download-json-schemas.sh
```diff
@@ -1,21 +1,24 @@
...
 if ! [ -n "$1" ]; then
     echo "USAGE: ./download-json-schemas.sh /path/to/folder"
     echo ""
     echo "Downloads the APM Server Schema files to the specified folder"
+    echo ""
+    echo "Sources from: https://codeload.github.com/elastic/apm-server/"
     exit 1
 fi
 
+# download_schema [/path/to/download/to] [branch]
```
trentm, line 13: nit: This function is written to *require* both those args, right. Could we write that as:

```suggestion
# download_schema </path/to/download/to> <branch>
```

It may be I'm showing a bias to use `[...]` syntax for optional args.

trentm (approved):
WFM with an optional nit.

Issue #1866, opened 2020-11-10
pg TAV test failures


trentm:
Per https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/master/656/pipeline these TAV tests failed:

Node.js-8-pg
Node.js-10-pg
Node.js-12-pg
Node.js-13-pg

Note that "Node.js-14-pg" passed.

From the node v10 pg test log:

```
[2020-11-10T06:35:50.839Z] node_tests_1  | -- running test "node test/instrumentation/modules/pg/pg.js" with pg
[2020-11-10T06:35:50.839Z] node_tests_1  | /app/test/instrumentation/modules/pg/pg.js:469
[2020-11-10T06:35:50.839Z] node_tests_1  |           .then(function () {
[2020-11-10T06:35:50.839Z] node_tests_1  |            ^
[2020-11-10T06:35:50.839Z] node_tests_1  | 
[2020-11-10T06:35:50.839Z] node_tests_1  | TypeError: client.query(...).then is not a function
[2020-11-10T06:35:50.839Z] node_tests_1  |     at /app/test/instrumentation/modules/pg/pg.js:469:12
[2020-11-10T06:35:50.839Z] node_tests_1  |     at /app/node_modules/pg/lib/pool.js:84:11
[2020-11-10T06:35:50.839Z] node_tests_1  |     at elasticAPMCallbackWrapper (/app/lib/instrumentation/index.js:309:27)
[2020-11-10T06:35:50.839Z] node_tests_1  |     at /app/node_modules/pg/node_modules/generic-pool/lib/generic-pool.js:347:9
[2020-11-10T06:35:50.839Z] node_tests_1  |     at /app/node_modules/pg/lib/pool.js:58:20
[2020-11-10T06:35:50.839Z] node_tests_1  |     at Connection.<anonymous> (/app/node_modules/pg/lib/client.js:149:7)
[2020-11-10T06:35:50.839Z] node_tests_1  |     at Object.onceWrapper (events.js:286:20)
[2020-11-10T06:35:50.839Z] node_tests_1  |     at Connection.emit (events.js:203:15)
[2020-11-10T06:35:50.839Z] node_tests_1  |     at Socket.<anonymous> (/app/node_modules/pg/lib/connection.js:109:12)
[2020-11-10T06:35:50.839Z] node_tests_1  |     at Socket.emit (events.js:198:13)
[...]

PR #1846, opened 2020-10-30
fix(pg): prevent unhandled promise rejection

glasseyes42:
Fixes: elastic/apm-agent-nodejs#1829

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Commit message follows [commit guidelines](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#commit-message-guidelines)

PR #1868, opened 2020-11-10
fix: tests correct codepath for pg span ending

astorm:
Fixes: https://github.com/elastic/apm-agent-nodejs/issues/1866

It looks like there was an incorrect assumption in the the test that was submitted along with https://github.com/elastic/apm-agent-nodejs/pull/1846  

This PR adds guard clauses that match the behavior of the guard clauses in the `pg` instrumentation 

https://github.com/elastic/apm-agent-nodejs/blob/8a5e908b8e9ee83bb1b828a3bef980388ea6e08e/lib/instrumentation/modules/pg.js#L93

to ensure this test only runs when/where it needs to. 

The tests were failing in older versions of `pg` without promise support.  Putting the tests behind the `pg@5.1` guard clause resulted in errors in the 6.x branch because the agent is adding `then` and `catch` callbacks to the promise in versions of `pg` without an `on` method (i.e. the 6.x line of `pg`)

This PR adds a guard clause to ensure the test is run only when the agent is manipulating the returned promise.  This test will not run on versions of `pg` without promise support, (<5.1) or for versions of `pg` where the agent is using event handlers instead of promise-manipulation (anything in the 6.x line)

### Checklist

- [x] Implement code

### DIFF
--- a/test/instrumentation/modules/pg/pg.js
+++ b/test/instrumentation/modules/pg/pg.js
@@ -447,38 +447,47 @@ if (global.Promise || semver.satisfies(pgVersion, '<6')) {
     })
   })
 
-  test.only('handles promise rejections from pg', function (t) {
-    function unhandledRejection (e) {
-      t.fail('had unhandledRejection')
-    }
-    process.once('unhandledRejection', unhandledRejection)
-    t.on('end', function () {
-      process.removeListener('unhandledRejection', unhandledRejection)
-      teardown()
-    })
+  // The same guard logic as from the instrumentation module
+  //
+  // https://github.com/elastic/apm-agent-nodejs/blob/8a5e908b8e9ee83bb1b828a3bef980388ea6e08e/lib/instrumentation/modules/pg.js#L91
+  //
+  // ensures this tests runs when ending a span via promise.then
+  if( typeof pg.Client.prototype.query.on !== 'function' &&
+      typeof pg.Client.prototype.query.then === 'function') {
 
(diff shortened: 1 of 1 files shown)


astorm:
@glasseyes42 :wave: Hello there!  We have some nightly tests that failed on older versions of the `pg` module after your PR.  This PR should fix it (no new code for you), but would you mind checking my assumptions above to make sure they're inline with your intent for the test that failed?

trentm:
Pro-tip: Look at the diff without whitespace changes: https://github.com/elastic/apm-agent-nodejs/pull/1868/files?w=1

trentm:
That integration test failure feels like an unrelated flaky fault:

```
[2020-11-10T21:45:21.463Z] subprocess.CalledProcessError: Command '['docker-compose', '-f', '/var/lib/jenkins/workspace/ration-tests-selector-mbp_master/src/github.com/elastic/apm-integration-testing/docker-compose.yml', '--no-ansi', '--log-level', 'ERROR', 'build', '--pull', '--no-cache', '--parallel', 'agent-nodejs-express', 'opbeans-node']' returned non-zero exit status 1.
```

trentm:
jenkins run the module tests for pg

trentm (approved):
LGTM

glasseyes42:
I left a `test.only` 🤦

glasseyes42:
Had thought the block the test was in being next to other promise related would have guarded 🤷 . Test change looks fine to me 👍

astorm:
Thanks for the extra feedback @glasseyes42 and no worries on the `t.only` -- we should have spotted that as well.  Trust the System™ -- it's eventually consistent ;)

Issue #1689, opened 2020-03-25
Implement transaction_ignore_urls


lreuven:
See [elastic/apm#144](https://github.com/elastic/apm/issues/144)

trentm:
IIUC, this needs the kibana changes under https://github.com/elastic/kibana/tree/master/x-pack/plugins/apm/common/agent_configuration/

trentm:
kibana changes:
- master: https://github.com/elastic/kibana/pull/85655
- 7.x: https://github.com/elastic/kibana/pull/85798

PR #1876, opened 2020-11-17
feat: transaction_ignore_urls wildcard matching

astorm:
Will Fix: elastic/apm-agent-nodejs#1689

Implements `ELASTIC_TRANSACTION_IGNORE_URL`/`transactionIgnoreUrls` configuration for wildcard URL ignoring.  The intent of this new configuration is to gives users the ability to skip generating transactions for a request to a URL that matches an exact string, or a URL that matches simple wildcard pattern.  This wildcard pattern will behave consistently across language agents.

This PR leaves the existing `ignoreUrls` configuration in place to allow users access to more powerful regular expression matching _and_ to preserve existing functionality for users using this older configuration setting. 

This PR also introduces a new dependency, [`escape-string-regexp`](https://www.npmjs.com/package/escape-string-regexp).  In order to implement the suggested wildcard matching algorithm we needed the ability to escape a string for use in a regular expression.  Since Node.js doesn't have this functionality natively, I pulled in `escape-string-regexp` rather than attempt to write (easy) and maintain (harder) this functionality ourselves.  Happy to revisit this decision if we'd rather implement this functionality ourselves. 

### Open items for consideration/completion
- [x] is `escape-string-regexp` an acceptable dependency for ignore regular expressions?
- [x] Docs
- [x] Fixture downloading
- [x] central config handling? How does this work in Node Agent, do we need more here?
- [x] should leading `/` be assumed? (we'll prepend this, meaning a configured value of 'foo*' and '/foo*' will match the same)
- [x] need clarity on `(?-i)` and `(?+i)` prefixes across agents
- [x] Does this apply only to transactions? (yes, `isRequestBlacklisted` is only called in `instrumentRequest`, and `instrumentRequest` is only used when wrapping http(s).Server's `emit` method (i.e. it's used in transaction starting)

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -12,6 +12,10 @@ endif::[]
 
 [float]
 ===== Features
+* feat: transactionIgnoreUrl wildcard matching {pull}1870[#1870]
+  Allows users to ignore URLs using simple wildcard matching patterns that behave
+  the same across language agents. See https://github.com/elastic/apm/issues/144
+
 * Cool new feature: {pull}2526[#2526]
 
 [float]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -185,6 +185,42 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
 
 Note that not all core Node.js API's can be instrumented without the use of Async Hooks if running Node.js 8 or above.
 
+[[transaction-ignore-urls]]
+==== `transactionIgnoreUrls`
+
+* *Type:* Array
+* *Default:* `[]`
+* *Env:* `ELASTIC_TRANSACTION_IGNORE_URLS`
+
+Array or comma-separated string used to restrict requests for certain URLs from being instrumented.
+
+This property should be set to an array containing one or more string patterns.
+
+When an incoming HTTP request is detected, its URL pathname will be tested against each element
+in this list.  The `transactionIgnoreUrls` property supports exact string matches,
+simple wildcard (`*`) matches, and may not include commas.  Wildcard matches are
+case insensative by default. You may make wildcard searches case sensative by
+using the `(?-i)` prefix.
+
+Note that all errors that are captured during a request to an ignored URL are still sent
+to the APM Server regardless of this setting.
+
+If you need full regular expression pattern matching, see <<ignore-urls>>.
+
+Example usage:
+
+[source,js]
+----
+require('elastic-apm-node').start({
+  transactionIgnoreUrls: [
+    '/ping',
+    '/fetch/*',
+    '(?-i)/caseSensitiveSearch'
+  ]
+})
+----
+
+
 [[ignore-urls]]
 ==== `ignoreUrls`
 
@@ -202,6 +238,8 @@ If an element in the array is a `RegExp` object,
 its test function will be called with the URL being tested.
 
 Note that all errors that are captured during a request to an ignored URL are still sent to the APM Server regardless of this setting.
+
+If you'd prefer simple wildcard pattern matching, see <<transaction-ignore-urls>>.
 
 Example usage:
 
--- a/lib/config.js
+++ b/lib/config.js
@@ -10,6 +10,9 @@ var truncate = require('unicode-byte-truncate')
 
 var version = require('../package').version
 var packageName = require('../package').name
+
+const { WildcardMatcher } = require('./wildcard-matcher')
+
 // Standardize user-agent header. Only use "elasticapm-node" if it matches "elastic-apm-node".
 if (packageName === 'elastic-apm-node') {
   packageName = 'elasticapm-node'
@@ -70,6 +73,7 @@ var DEFAULTS = {
   sourceLinesSpanAppFrames: 0,
   sourceLinesSpanLibraryFrames: 0,
   stackTraceLimit: 50,
+  transactionIgnoreUrls: [],
   transactionMaxSpans: 500,
   transactionSampleRate: 1.0,
   useElasticTraceparentHeader: true,
@@ -124,6 +128,7 @@ var ENV_TABLE = {
   sourceLinesSpanAppFrames: 'ELASTIC_APM_SOURCE_LINES_SPAN_APP_FRAMES',
   sourceLinesSpanLibraryFrames: 'ELASTIC_APM_SOURCE_LINES_SPAN_LIBRARY_FRAMES',
   stackTraceLimit: 'ELASTIC_APM_STACK_TRACE_LIMIT',
+  transactionIgnoreUrls: 'ELASTIC_TRANSACTION_IGNORE_URLS',
   transactionMaxSpans: 'ELASTIC_APM_TRANSACTION_MAX_SPANS',
   transactionSampleRate: 'ELASTIC_APM_TRANSACTION_SAMPLE_RATE',
   useElasticTraceparentHeader: 'ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER',
@@ -134,7 +139,8 @@ var ENV_TABLE = {
 var CENTRAL_CONFIG = {
   transaction_sample_rate: 'transactionSampleRate',
   transaction_max_spans: 'transactionMaxSpans',
-  capture_body: 'captureBody'
+  capture_body: 'captureBody',
+  transaction_ignore_urls: 'transactionIgnoreUrls'
 }
 
 var VALIDATORS = {
@@ -186,7 +192,8 @@ var MINUS_ONE_EQUAL_INFINITY = [
 ]
 
 var ARRAY_OPTS = [
-  'disableInstrumentations'
+  'disableInstrumentations',
+  'transactionIgnoreUrls'
 ]
 
 var KEY_VALUE_OPTS = [
@@ -204,7 +211,7 @@ class Config {
     this.ignoreUrlRegExp = []
     this.ignoreUserAgentStr = []
     this.ignoreUserAgentRegExp = []
-
+    this.transactionIgnoreUrlRegExp = []
     // If we didn't find a config file on process boot, but a path to one is
     // provided as a config option, let's instead try to load that
     if (confFile === null && opts && opts.configFile) {
@@ -369,17 +376,31 @@ function readEnv () {
 }
 
 function normalize (opts) {
-  normalizeIgnoreOptions(opts)
   normalizeKeyValuePairs(opts)
   normalizeNumbers(opts)
   normalizeBytes(opts)
   normalizeArrays(opts)
   normalizeTime(opts)
   normalizeBools(opts)
+  normalizeIgnoreOptions(opts)
   truncateOptions(opts)
 }
 
 function normalizeIgnoreOptions (opts) {
+  if (opts.transactionIgnoreUrls) {
+    // We can't guarantee that opts will be a Config so set a
+    // default value. This is to work around CENTRAL_CONFIG tests
+    // that call this method with a plain object `{}`
+    if (!opts.transactionIgnoreUrlRegExp) {
+      opts.transactionIgnoreUrlRegExp = []
+    }
+    const wildcard = new WildcardMatcher()
+    for (const ptn of opts.transactionIgnoreUrls) {
+      const re = wildcard.compile(ptn)
+      opts.transactionIgnoreUrlRegExp.push(re)
+    }
+  }
+
   if (opts.ignoreUrls) {
     for (const ptn of opts.ignoreUrls) {
       if (typeof ptn === 'string') opts.ignoreUrlStr.push(ptn)
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -78,6 +78,9 @@ function isRequestBlacklisted (agent, req) {
   for (i = 0; i < agent._conf.ignoreUrlRegExp.length; i++) {
     if (agent._conf.ignoreUrlRegExp[i].test(req.url)) return true
   }
+  for (i = 0; i < agent._conf.transactionIgnoreUrlRegExp.length; i++) {
+    if (agent._conf.transactionIgnoreUrlRegExp[i].test(req.url)) return true
+  }
 
   var ua = req.headers['user-agent']
   if (!ua) return false
--- a/lib/wildcard-matcher.js
+++ b/lib/wildcard-matcher.js
@@ -1,0 +1,48 @@
+'use strict'
+/**
+ * Implements a Wildcard matcher
+ *
+ * Exports a function that implements a simple wildcard matcher
+ * per: https://github.com/elastic/apm/issues/144
+ */
+const escapeStringRegexp = require('escape-string-regexp')
+
+// coverts elastic-wildcard pattern into a
+// a javascript regular expression.
+const starMatchToRegex = (pattern) => {
+  // case insensative by default
+  let regexOpts = ['i']
+  if (pattern.startsWith('(?-i)')) {
+    regexOpts = []
+    pattern = pattern.slice(5)
+  }
+
+  const patternLength = pattern.length
+  const reChars = ['^']
+  for (let i = 0; i < patternLength; i++) {
+    const char = pattern[i]
+    switch (char) {
+      case '*':
+        reChars.push('.*')
+        break
+      default:
+        reChars.push(
+          escapeStringRegexp(char)
+        )
+    }
+  }
+  reChars.push('$')
+  return new RegExp(reChars.join(''), regexOpts.join(''))
+}
+
+class WildcardMatcher {
+  compile (pattern) {
+    return starMatchToRegex(pattern)
+  }
+
+  match (string, pattern) {
+    const re = this.compile(pattern)
+    return string.search(re) !== -1
+  }
+}
+module.exports = { WildcardMatcher }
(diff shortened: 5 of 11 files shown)


lib/config.js
```diff
@@ -369,17 +376,40 @@ function readEnv () {
...
 function normalize (opts) {
-  normalizeIgnoreOptions(opts)
   normalizeKeyValuePairs(opts)
   normalizeNumbers(opts)
   normalizeBytes(opts)
   normalizeArrays(opts)
   normalizeTime(opts)
   normalizeBools(opts)
+  normalizeIgnoreOptions(opts)
   truncateOptions(opts)
```
astorm, line 386: Move `normalizeIgnoreOptions` down in the list in order to ensure `normalizeArrays` had run, and that the `transactionIgnoreUrls` value would already be converted from a comma,delimited,string into an array (if necessary)

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
...
+
+Example usage:
+
+[source,js]
+----
+require('elastic-apm-node').start({
+  transactionIgnoreUrls: [
+    '/ping',
+    '/fetch/*',
+    '(?+i)/caseSensativeSearch'
```
bmorelli25, line 217:
```suggestion
    '(?+i)/caseSensitiveSearch'
```

bmorelli25 (commented):
Nice! One small typo, otherwise [docs](https://apm-agent-nodejs_1876.docs-preview.app.elstc.co/guide/en/apm/agent/nodejs/master/configuration.html#transaction-ignore-urls) LGTM 🎉

astorm:
jenkins run lint

lib/config.js
```diff
@@ -369,17 +376,40 @@ function readEnv () {
...
+    // default value. This is to work around CENTRAL_CONFIG tests
+    // that call this method with a plain object `{}`
+    if (!opts.transactionIgnoreUrlRegExp) {
+      opts.transactionIgnoreUrlRegExp = []
+    }
+    const wildcard = new WildcardMatcher()
+    for (let ptn of opts.transactionIgnoreUrls) {
+      // if configured without a leading '/', be nice
+      // and add this to pattern for the user
+      if (ptn[0] !== '/') ptn = '/' + ptn
```
trentm, line 407: I don't think we should do this "favour". 

- I suspect other agents don't do it, which would slightly/subtly break commonality. The Python agent at least doesn't do this.
- Combined with a possible empty string (which could come from a dangling comma like `ELASTIC_TRANSACTION_IGNORE_URLS=a,b,`) this would result in a pattern to ignore the root URL, `'/'`, which would be a surprise.

lib/instrumentation/http-shared.js
```diff
@@ -78,7 +78,9 @@ function isRequestBlacklisted (agent, req) {
   for (i = 0; i < agent._conf.ignoreUrlRegExp.length; i++) {
     if (agent._conf.ignoreUrlRegExp[i].test(req.url)) return true
   }
-
+  for (i = 0; i < agent._conf.transactionIgnoreUrlRegExp.length; i++) {
+    if (agent._conf.transactionIgnoreUrlRegExp[i].test(req.url)) return true
+  }
```
trentm, line 83: nit: restore the blank line before the `ua` block?

lib/wildcard-matcher.js
```diff
@@ -0,0 +1,62 @@
+const escapeStringRegexp = require('escape-string-regexp')
```
trentm, line 1: Should we `'use strict'` at the top of all our JS files?

nit: A relatively brief top-comment at the top of each file giving its general intent would be helpful. E.g. could mention this implements a common "WildcardMatcher" spec from apm.git (though that repo doesn't have a good link for that spec, just the JSON file of tests).

test/tmp-fixtures.js
```diff
@@ -0,0 +1,194 @@
+module.exports = {
```
trentm, line 1: Should this be dropped now that test/fixtures/json-specs/wildcard_matcher_tests.json is available?

test/wildcard-matcher.js
```diff
@@ -0,0 +1,17 @@
+var test = require('tape')
```
trentm, line 1: `'use strict'` ?

test/config.js
```diff
@@ -455,6 +456,27 @@ test('should separate strings and regexes into their own blacklist arrays', func
...
+    agent._conf.transactionIgnoreUrlRegExp.length,
+    3,
+    'was everything added?'
+  )
+
+  t.strictEquals(
+    '/foo'.search(agent._conf.transactionIgnoreUrlRegExp[0]),
+    0,
+    'was leading / added automatically to "foo"'
+  )
```
trentm, line 475: Would want to drop this test if dropping the "auto prefix with '/'" feature above.

test/central-config-enabled.js
```diff
@@ -36,7 +26,8 @@ test('remote config enabled', function (t) {
   })
 
   server.listen(function () {
-    agent = require('..').start({
+    // agent = require('..').start({
```
trentm, line 29: Drop this commented line?

lib/wildcard-matcher.js
```diff
@@ -0,0 +1,62 @@
...
+  // remove prefix
+  const parts = pattern.split(')')
+  parts.shift()
+  return { finalPattern: parts.join(''), finalOpts: opts }
+}
+
+// coverts elastic-wildcard pattern into a
+// a javascript regular expression.
+const starMatchToRegex = (pattern, opts) => {
+  const { finalPattern, finalOpts } = parsePatternAndOptsFromPattern(pattern, opts)
```
trentm, line 33: The spec doesn't mention `(?+i)` at all. I think you could simplify to:

```
if (pattern.startsWith('(?-i)')) {
    opts = []     // or perhaps change `opts` to be a specific caseSensitive bool instead of generic regexp flags
    pattern = pattern.slice(5)
}
```

and drop `parsePatternAndOptsFromPattern`.

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
...
+* *Env:* `ELASTIC_TRANSACTION_IGNORE_URLS`
+
+Used to restrict requests for certain URLs from being instrumented.
+
+This property should be set to an array containing one or more string patterns.
+
+When an incoming HTTP request is detected, its URL will be tested against each element 
+in this list.  The `transactionIgnoreUrls` property supports exact string matches, and
+simple wildcard (`*`) matches.  You may turn on case sensitivity by using the 
+`(?+i)` prefix.
```
trentm, line 202: It took me a LONG time to realize that this is case-*insensitive* by default.  

1. Eww, but that's the way it was specified.
2. I think we should have a clear "Matching is case-sensitive by default. Prefix your pattern with `(?-i)` to make the matching case-sensitive."
3. Do you really want to extend the spec by supporting `(?+i)`? At least the Python agent doesn't support that. I argue lower in this review that the code can be made simpler by only supporting `(?-i)` and not `(?+i)`.

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
...
+
+Example usage:
+
+[source,js]
+----
+require('elastic-apm-node').start({
+  transactionIgnoreUrls: [
+    '/ping',
+    '/fetch/*',
+    '(?+i)/caseSensitiveSearch'
```
trentm, line 217:
```suggestion
    '(?-i)/caseSensitiveSearch'
```

trentm (changes requested):

test/wildcard-matcher.js
astorm: _nod_ `use strict` at the top of every JS file should be the defacto choice.  I'm used to ESLint catching this -- maybe another bit of evidence to move away from `standard` or at least enhance our CI it with some additional ESLint rules

lib/wildcard-matcher.js
astorm: _nod_ -- seems reasonable -- will add.

CHANGELOG.asciidoc
```diff
@@ -12,6 +12,10 @@ endif::[]
 
 [float]
 ===== Features
+* feat: transaction_ignore_url wildcard matching {pull}1870[#1870]
```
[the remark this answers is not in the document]
astorm: will change to `transactionIgnoreUrls`, and I'm up for giving spec links (when available) a try with our newer, more verbose changelog overlords :)   Linked to https://github.com/elastic/apm/issues/144, which is the closest we have to a spec for this feature.

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
 
 Note that not all core Node.js API's can be instrumented without the use of Async Hooks if running Node.js 8 or above.
 
+[[transaction-ignore-urls]]
+==== `transactionIgnoreUrls`
+
```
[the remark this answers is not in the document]
astorm: Good call out.  If I'm reading that right "Dynamic" configuration means users can confg this via central config. 

How about we hold off this for now? My thinking is this configuration isn't available yet in the 7.10 release, and agents release off cycle from mainline elastic.  Once the `TRANSACTION_IGNORE_URL` setting makes it into Kibana/APM-UI proper we can come back and mark this as dynamic. 

Thoughts? (if so, just unresolved and we can keep talking)

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
...
+* *Type:* Array
+* *Default:* `[]`
+* *Env:* `ELASTIC_TRANSACTION_IGNORE_URLS`
+
+Used to restrict requests for certain URLs from being instrumented.
+
+This property should be set to an array containing one or more string patterns.
+
+When an incoming HTTP request is detected, its URL will be tested against each element
```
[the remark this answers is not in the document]
astorm: 👍

lib/config.js
```diff
@@ -369,17 +376,40 @@ function readEnv () {
...
   normalizeTime(opts)
   normalizeBools(opts)
+  normalizeIgnoreOptions(opts)
   truncateOptions(opts)
 }
 
 function normalizeIgnoreOptions (opts) {
+  if (opts.transactionIgnoreUrls) {
+    // if users accidently configure a string instead of an
+    // array with a single string, handle it for them
```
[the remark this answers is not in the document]
astorm: 👍 Good call out -- `normalizeArrays` was a late discovery, will change.

lib/config.js
```diff
@@ -369,17 +376,40 @@ function readEnv () {
...
+    if (!Array.isArray(opts.transactionIgnoreUrls)) {
+      opts.transactionIgnoreUrls = [opts.transactionIgnoreUrls]
+    }
+
+    // We can't guarentee that opts will be a Config so set a
```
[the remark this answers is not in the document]
astorm: 👍

docs/configuration.asciidoc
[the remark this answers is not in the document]
trentm: @bmorelli25 Thoughts on this? Is there a way we could say "in APM server version 7.11" alongside the "dynamic configuration" badge/link?

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
 
 Note that not all core Node.js API's can be instrumented without the use of Async Hooks if running Node.js 8 or above.
 
+[[transaction-ignore-urls]]
+==== `transactionIgnoreUrls`
+
+* *Type:* Array
+* *Default:* `[]`
+* *Env:* `ELASTIC_TRANSACTION_IGNORE_URLS`
```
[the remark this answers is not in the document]
astorm: Hmmm -- looks like there's precedent for calling out the comma separation in the `ELASTIC_APM_DISABLE_INSTRUMENTATIONS`, so I'm game.  The "no commas" is trickier -- it's only there as a side effect of the the way APM Server will be sending us that data AND URLs don't seem common in URL Paths.  On the flip side though -- truth :)   I've added a comment about commas.  I'll let @bmorelli25 be the final arbiter once we've got this locked in.

docs/configuration.asciidoc
```diff
@@ -185,6 +185,41 @@ please https://github.com/elastic/apm-agent-nodejs/issues[open an issue].
...
+
+When an incoming HTTP request is detected, its URL will be tested against each element 
+in this list.  The `transactionIgnoreUrls` property supports exact string matches, and
+simple wildcard (`*`) matches.  You may turn on case sensitivity by using the 
+`(?+i)` prefix.
+
+Note that all errors that are captured during a request to an ignored URL are still sent 
+to the APM Server regardless of this setting.
+
+If you need full regular expression pattern matching, see <<ignore-urls>>.
```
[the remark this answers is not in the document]
astorm: Agreed -- once this is actually available as a Kibana central config option we can/should come back and make mention of it.  

Similar to "Added in version x.y.z" -- that would be nice to have as a user -- on the flip side that's a burden on us to start adding when every configuration option was added.  I'm kind of 50/50 on this -- maybe table that until we know what the actual next version of the agent is? (something to add to our release check list maybe)

lib/config.js
astorm: That's fair -- I can back off this one.  (Likely came from years of trauma with apache handling the path differently in the main config vs. `.htaccess` files :))

lib/instrumentation/http-shared.js
astorm: 👍

test/config.js
astorm: 👍

test/tmp-fixtures.js
astorm: 🤦 👍

test/wildcard-matcher.js
astorm: 👍

test/central-config-enabled.js
astorm: we regret the error -- we also regret the next occurrence of this error -- I can be comment blind towards the end of a feature 🤦

docs/configuration.asciidoc
astorm: Hmm Re: `(?+i)` -- my memory is I saw that used in the Java Agent's wildcard implementation, and I included parsing it out in case someone was copying rules between agents. Going back now though I don't see it anywhere -- so good catch and a reminder for me meditate on the fallibility of memory  

i.e. I'll remove the code for handling the + case and be explicit in the language about what is and isn't supported. 

And yes -- I'm also looking forward to participating in spec discussions instead of implementing specs decided on before our time. :)

docs/configuration.asciidoc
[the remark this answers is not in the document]
trentm: Personally I'd make mention of it now, referencing future "Kibana 7.11", but I'm also fine doing it later.

> maybe table that until we know what the actual next version of the agent is?

Sounds good.

lib/wildcard-matcher.js
astorm: agreed -- dropped the extra function in favor the single conditional.

docs/configuration.asciidoc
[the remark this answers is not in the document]
bmorelli25: Good catch. Whether we wait on adding this badge to the docs or not, it's probably important either way to mention that stack version 7.11+ is required. I think text next to the badge is fine. We could add it to the badge too, but that might be overkill. . . 

Screen Shot 2020-11-18 at 3 52 58 PM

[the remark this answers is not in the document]
trentm: @bmorelli25 Ooh, I like that. What's the asciidoc syntax for that?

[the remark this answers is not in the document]
bmorelli25: Badges are a bit of a hack in our build as we don't have native support for them. In this case, the new [badge](https://img.shields.io/badge/Dynamic-Requires%20APM%20Server%207.11%2B-blue) needs to be added to the images directory and then referenced:

`<<dynamic-configuration, image:./images/dynamic-config-stack-711.svg[] >>`

I will say-- after thinking some more on this, I don't think it makes to add this badge to the docs yet; I think it will only confuse users.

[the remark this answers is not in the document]
astorm: Thanks @bmorelli25 -- we've captured our intent to update this label once the feature lands in kibana: https://github.com/elastic/apm-agent-nodejs/issues/1883

lib/wildcard-matcher.js
```diff
@@ -0,0 +1,48 @@
+'use strict'
+/**
+ * Implements a Wildcard matcher
+ *
+ * Exports a function that implements a simple wildcard matcher
+ * per: https://github.com/elastic/apm/issues/144
+ */
+const escapeStringRegexp = require('escape-string-regexp')
+
+// coverts elastic-wildcard pattern into a
```
trentm, line 10: nit:

```suggestion
// Converts elastic-wildcard pattern into a
```

trentm (approved):
Just a spelling typo in the docs and a typo. Soft approving.

astorm:
Merging, and will handle the docs typos in a quick follow-on PR.

PR #1877, opened 2020-11-17
feat(elasticsearch): support @elastic/elasticsearch client

### LINKED ISSUE #968: Instrument the @elastic/elasticsearch module

trentm:
Based on earlier work in:
https://github.com/elastic/apm-agent-nodejs/pull/1266

Fixes #968


trentm:
This is ready to review now. There are a few "TODO" comments in there:

```
% git diff master.. | grep TODO
+// TODO: ask Tomas about apiName addition to RequestEvent
+        // TODO set "span.outcome" (from err and result.statusCode)
+          // TODO: rebuild error object to avoid serialization issues?
+    // TODO test that httpSpan is child of esSpan? currently it isn't
```

that I don't think *all* need to be done now.

@delvedor You are welcome to review and/or play with this, but I don't expect you to.

trentm:
review note: the linting "error" will go away when these commits are squashed

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,124 @@
...
+// Instrument the @elastic/elasticsearch module.
+//
+// This uses to 'request' and 'response' events from the Client (documented at
+// https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/observability.html)
+// to hook into all ES server interactions.
+
+const { getDBDestination } = require('../../context')
+
+// TODO: ask Tomas about apiName addition to RequestEvent
+const pathIsAQuery = /\/_((search|msearch)(\/template)?|count)$/
```
astorm, line 12: Calling out some offline discussions here -- this test means that we'll only capture the full request bodies for REST queries whose paths match the following _exactly_. 

```
_search
_msearch
_search/template
_msearch/template
_count
```

We may want to expand this after discussions with stakeholders.

astorm (commented):
I took an initial pass at this -- it looks like a solid start and captures the functionality of the  previous instrumentation and @delvedor's first pass at this well.  The use of the `WeakMap` is a nice call-out as well -- Trust your ~Mechanic~ Garbage Collector :) 

It doesn't look like sub-classing is the common approach for instrumentation in the agent -- normally we'd identify the request method, wrap it, start our business, call the original method, and then clean up our business.  

I think the sub-classing/event approach can work though -- and in some ways seems less intrusive than the direct monkey-patching we normally do. There's a few additional defensive measures we'll want to take, but otherwise I'm down with this approach.

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    span.name = `Elasticsearch: ${params.method} ${params.path}`
+    return span
+  }
+
+  class ApmClient extends elasticsearch.Client {
+    constructor (opts) {
+      super(opts)
+
```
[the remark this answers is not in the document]
trentm: Heh, I'd had a comment for that myself, then removed it. I doubt this signature will change, but it doesn't hurt, so sounds good.

lib/instrumentation/modules/@elastic/elasticsearch.js
trentm: I've also asked @delvedor about perhaps changing elasticsearch-js to provide a field in the 'request' event that is better than just regex matching the path (which can have user data in it).

I've also got this change coming:

```
// Allow the following URL path components:
// - _search
// - _msearch
// - _search/template
// - _msearch/template
// - _count
// - _async_search
const pathIsAQuery = /\/(_search|_msearch|_count|_async_search)(\/|$)/
```

to add `_async_search` because Kibana is using that.

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,124 @@
...
+    const span = agent.startSpan(null, 'db', 'elasticsearch', 'request')
+    if (span === null) {
+      return null
+    }
+
+    span.name = `Elasticsearch: ${params.method} ${params.path}`
+    return span
+  }
+
+  class ApmClient extends elasticsearch.Client {
```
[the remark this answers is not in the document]
trentm: Thanks. I'll add a guard for that to fail gracefully:

```
  if (!elasticsearch.Client) {
    agent.logger.debug('@elastic/elasticsearch@%s is not supported (no `elasticsearch.Client`) - aborting...', version)
    return elasticsearch
  }
```

In testing it looks like this:

```
$ ELASTIC_APM_LOG_LEVEL=trace node my-test-app.js
...
shimming https@10.23.0 module
shimming https.Server.prototype.emit function
shimming https.request function
shimming https.get function
shimming @elastic/elasticsearch@7.10.0 module
@elastic/elasticsearch@7.10.0 is not supported (no `elasticsearch.Client`) - aborting...   <---- HERE
shimming elasticsearch@16.7.1 module
shimming elasticsearch.Transport.prototype.request
...
```

trentm:
> I think the sub-classing/event approach can work though -- and in some ways seems less intrusive than the direct monkey-patching we normally do.

My understanding is this was intentional. The 'request' and 'response' events described at https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/observability.html were, at least in part, intended for this kind of thing.

trentm:
Remaining TODOs before merging:

- [ ] Look into getting the http span(s) to be a *child* of the Elasticsearch span
- [ ] Decide on error serialization. From the [original PR](https://github.com/elastic/apm-agent-nodejs/pull/1266#issue-305457143): "we should figure out how to serialize the errors to avoid the loss of important information"


Related TODOs that can be done separate from this PR (IMHO):

- [ ] Set "span.outcome" (from err and result.statusCode). I'll take a stab at this, but might defer it to https://github.com/elastic/apm-agent-nodejs/issues/1814
- [ ] Trent to possibly open a ticket/PR on elasticsearch-js to have an `apiName` field on the "request" event data to identify what Elasticsearch API endpoint is being called. This would make the current `pathIsAQuery` filtering (to decide if `context.db` is added to an APM span) more efficient and precise.
- [ ] Trent to open a separate ticket/PR to look at adding a "prepare-request" (or whatever name) event from elasticsearch-js to include request setup time. This was explored in the original PR: https://github.com/elastic/apm-agent-nodejs/pull/1266/files#diff-8b8cdefabed92970b05a687c766e3c51678ea17686b219d93ff63043f40a6538R27 and ... I can't find the "elasticsearch-js/update-events" branch anymore (perhaps since deleted).

trentm:
I am moving the TODO item `Look into getting the http span(s) to be a child of the Elasticsearch span` to a *separate* ticket, because handling that is a bigger change to agent behaviour: https://github.com/elastic/apm-agent-nodejs/issues/1889

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,140 @@
...
+
+// TODO: ask Tomas about apiName addition to RequestEvent
+// Allow the following URL path components:
+// - _search
+// - _msearch
+// - _search/template
+// - _msearch/template
+// - _count
+// - _async_search
+const pathIsAQuery = /\/(_search|_msearch|_count|_async_search)(\/|$)/
```
delvedor, line 19: The `POST /_sql` path can contain a query as well. ([documentation](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference.html#_sql_query))
```suggestion
const pathIsAQuery = /\/(_search|_msearch|_count|_async_search|_sql)(\/|$)/
```

delvedor (commented):
Great work! I love the use of the WeakMap.

Regarding the `preRequest` event I'll work on it soon. I'm wondering if it makes more sense to have a `serialization` and `deserialization` events tho.
So the order would be:
```
serialization
  │ (serialization and compression happens between those two events)
  └─▶ request
	  │ (actual time spent over the wire)
	  └─▶ deserialization
		  │ (deserialization and decompression happens between those two events)
		  └─▶ response
```
What do you think?

delvedor:
I've opened https://github.com/elastic/elasticsearch-js/pull/1364, which adds the api name to the request metadata so you can avoid using the regex and get better performances out of the box :)
If you are ok with the addition, it will land in 7.11.

trentm:
> I'm wondering if it makes more sense to have a `serialization` and `deserialization` events tho.

From our discussion earlier, this sounds great. Thanks for working on this.

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -0,0 +1,117 @@
...
+            // https://www.elastic.co/guide/en/elasticsearch/reference/current/common-options.html#common-options-error-options
+            errOpts.custom = {
+              type: err.body.error.type,
+              reason: err.body.error.reason,
+              caused_by: err.body.error.caused_by,
+              status: err.body.status
+            }
+          }
+
+          delete err.data // possibly large or private user data
```
vigneshshanmugam, line 102: might trigger a deopt, cant we just discard it or ignore specific fields from getting captured as part of error instead of deleting them and triggering deopt.

vigneshshanmugam (commented):
Great work and love the use of WeakMap.

lib/instrumentation/modules/@elastic/elasticsearch.js
trentm: @vigneshshanmugam Thanks for bringing it up. I haven't played with tracing deopts at all. Currently the node.js agent's `captureError` doesn't provide a way to guide which error fields should be captured. I can look into adding that capability.

astorm (approved):
👍 I see my comments from the previous review have been addressed.  Also, Re:

> I am moving the TODO item Look into getting the http span(s) to be a child of the Elasticsearch span to a separate ticket, because handling that is a bigger change to agent behaviour: #1889

This feels like the correct approach, and that it would be better to get this PR landed and then address the span parenting in a separate issue.

lib/instrumentation/modules/@elastic/elasticsearch.js
trentm: https://github.com/elastic/apm-agent-nodejs/pull/1912 is my hoped-for solution for this.

trentm: That is merged now and this PR moved to using it: https://github.com/elastic/apm-agent-nodejs/pull/1877/commits/20984f22cc073a08d85a551a1b8ddbdf0b312bfd

trentm:
jenkins run the module tests for elasticsearch,@elastic/elasticsearch

trentm:
jenkins run the module tests for elasticsearch,@elastic/elasticsearch

trentm:
jenkins run the module tests for elasticsearch,@elastic/elasticsearch

trentm:

trentm:
Since the last time y'all reviewed:

- I moved to using `apm.captureError(..., {captureAttributes: false})` to avoid deopt (per @vigneshshanmugam comment): https://github.com/elastic/apm-agent-nodejs/pull/1877/files#diff-8b8cdefabed92970b05a687c766e3c51678ea17686b219d93ff63043f40a6538R92-R97
- I fixed a side issue where `error.exception.module` was not handling "@ns/package" namespaced packages: https://github.com/elastic/apm-agent-nodejs/pull/1877/files#diff-9ceb00d437ddb7c1de1041eea52561f5e7166af4375d35d769590caace91ed31R282-R307
- I removed races from the added tests.

This is ready for re-review when you have the time. Thanks.

lib/instrumentation/modules/elasticsearch.js
```diff
@@ -2,8 +2,7 @@
 
 var shimmer = require('../shimmer')
 var { getDBDestination } = require('../context')
-
-var queryRegexp = /_((search|msearch)(\/template)?|count)$/
+const { setElasticsearchDbContext } = require('../elasticsearch-shared')
```
astorm, line 5: 👍 to pulling this into a shared lib across both versions of the client.

lib/parsers.js
```diff
@@ -279,13 +279,32 @@ function getCulprit (frames) {
...
-  var match = frame.filename.match(/node_modules\/([^/]*)/)
+  var match = frame.filename.match(MODULE_NAME_REGEX)
   if (!match) return
-  return match[1]
+  var moduleName = match[1]
+  if (moduleName && moduleName[0] === '@' && match[3]) {
+    // Normalize the module name separator to '/', even on Windows.
+    moduleName += '/' + match[3]
+  }
+  return moduleName
```
astorm, line 307: I see here we're modifying `getModule` so it works with `@namespaced/module-names` as well as with windows based filesystems. I presume this is to capture `@elastic/elasticsearch` things.

Do you think this changes requires any additional test fixtures in `test/parsers.js`, or is the existing coverage we have sufficient?

.tav.yml
```diff
@@ -242,6 +242,15 @@ koa-router:
 elasticsearch:
   versions: '>=8.0.0'
   commands: node test/instrumentation/modules/elasticsearch.js
+
+# @elastic/elasticsearch
```
astorm, line 246: 👍 on the context for the `versions` choices.

astorm (approved):
Looks good from here -- previous feedback's been address and it gets us instrumentation for the modern `@elastic/elasticsearch` client. One dealer's choice comment on some additional test coverage, but I think we're good to go as it.  Approving.

lib/instrumentation/index.js
```diff
@@ -11,6 +11,7 @@ var shimmer = require('./shimmer')
 var Transaction = require('./transaction')
 
 var MODULES = [
+  '@elastic/elasticsearch',
```
delvedor, line 14: In the coming days I'll publish another module named `@elastic/elasticsearch-canary`, which is identical to `@elastic/elasticsearch`, and it will be used mainly by Kibana.
I'll ping you here as soon as it's available on npm.

.ci/docker/docker-compose.yml
```diff
@@ -71,7 +71,7 @@ services:
       retries: 30
 
   elasticsearch:
-    image: docker.elastic.co/elasticsearch/elasticsearch:6.6.0
+    image: docker.elastic.co/elasticsearch/elasticsearch:7.10.1
     environment:
```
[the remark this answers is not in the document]
trentm: Yes, that was intentional. I could separate it if we want, but I'd wanted to test this integration against a recent ES.

lib/parsers.js
trentm: Yah, fair, I could definitely improve the test coverage here. Actually before this change there wasn't a single test case checking `error.exception.module`. I'll open a separate ticket to do this in a followup: https://github.com/elastic/apm-agent-nodejs/issues/1918

PR #1879, opened 2020-11-17
fix: obviously sync spans are not sent as sync

### LINKED ISSUE #1878: Unexpected Asynchronous Spans
I ran into a situation where the Elastic Node.js Agent is marking span payloads with the `sync=false` key/value pair even when the spans themselves contain no asynchronous work. 

Consider the following program

    const http = require('http');

    const requestListener = function (req, res) {
      const span1 = apm.startSpan('This is span 1')
      span1.end()
      const span2 = apm.startSpan('This is span 2')
      span2.end()
      const span3 = apm.startSpan('This is span 3')
      span3.end()
      res.writeHead(200);
      res.end('Hello, World!');
    }

    const server = http.createServer(requestListener);
    server.listen(8080);

This program handles an HTTP request, and then manually starts and stops three spans.  Payloads are generated for each span, and those payload have the value `sync=false`.  As a user, my first instinct would be that these spans would be considered synchronous work.

It's unclear whether this is a bug, or a deliberate tradeoff decision to treat all NodeJS spans as async.

Next Steps: Investigate where and why these spans have their default `sync=true` property set to `false`.

trentm:
Fixes #1878

### DIFF
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -93,12 +93,12 @@ module.exports = function (ins) {
 
   function before (asyncId) {
     const span = activeSpans.get(asyncId)
-    if (span) {
+    if (span && !span.ended) {
       debug(`setting span.sync=false on span.id==${span.id} with span.ended==${span.ended}`)
       span.sync = false
     }
     const transaction = span ? span.transaction : activeTransactions.get(asyncId)
-    if (transaction) {
+    if (transaction && !transaction.ended) {
       transaction.sync = false
     }
     ins.bindingSpan = null
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -71,6 +71,9 @@ module.exports = function (ins) {
     // type, which will init for each scheduled timer.
     if (type === 'TIMERWRAP') return
 
+    // XXX
+    // debug(`${type}(${asyncId}): trigger: ${triggerAsyncId}`);
+
     const transaction = ins.currentTransaction
     if (!transaction) return
 
@@ -91,6 +94,7 @@ module.exports = function (ins) {
   function before (asyncId) {
     const span = activeSpans.get(asyncId)
     if (span) {
+      debug(`setting span.sync=false on span.id==${span.id} with span.ended==${span.ended}`)
       span.sync = false
     }
     const transaction = span ? span.transaction : activeTransactions.get(asyncId)
@@ -115,4 +119,10 @@ module.exports = function (ins) {
     activeTransactions.delete(asyncId)
     activeSpans.delete(asyncId)
   }
+}
+
+var fs = require('fs');
+function debug(msg) {
+  fs.writeSync(process.stdout.fd,
+    `[executionAsyncId=${asyncHooks.executionAsyncId()}] ${msg}\n`);
 }
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -71,9 +71,6 @@ module.exports = function (ins) {
     // type, which will init for each scheduled timer.
     if (type === 'TIMERWRAP') return
 
-    // XXX
-    // debug(`${type}(${asyncId}): trigger: ${triggerAsyncId}`);
-
     const transaction = ins.currentTransaction
     if (!transaction) return
 
@@ -94,7 +91,6 @@ module.exports = function (ins) {
   function before (asyncId) {
     const span = activeSpans.get(asyncId)
     if (span && !span.ended) {
-      debug(`setting span.sync=false on span.id==${span.id} with span.ended==${span.ended}`)
       span.sync = false
     }
     const transaction = span ? span.transaction : activeTransactions.get(asyncId)
@@ -119,10 +115,4 @@ module.exports = function (ins) {
     activeTransactions.delete(asyncId)
     activeSpans.delete(asyncId)
   }
-}
-
-var fs = require('fs');
-function debug(msg) {
-  fs.writeSync(process.stdout.fd,
-    `[executionAsyncId=${asyncHooks.executionAsyncId()}] ${msg}\n`);
 }
--- a/test/instrumentation/async-hooks.js
+++ b/test/instrumentation/async-hooks.js
@@ -135,7 +135,7 @@ test('sync/async tracking', function (t) {
   })
 })
 
-test.only('sync/async tracking 2', function (t) {
+test('sync/async tracking 2', function (t) {
   var trans = agent.startTransaction()
 
   var span = agent.startSpan()
--- a/test/instrumentation/async-hooks.js
+++ b/test/instrumentation/async-hooks.js
@@ -125,28 +125,20 @@ test('sync/async tracking', function (t) {
   var trans = agent.startTransaction()
   t.strictEqual(trans.sync, true)
 
-  var span = agent.startSpan()
-  t.strictEqual(span.sync, true)
+  var span1 = agent.startSpan()
+  t.strictEqual(span1.sync, true)
+
+  // This span will be *ended* synchronously. It should stay `span.sync=true`.
+  var span2 = agent.startSpan()
+  t.strictEqual(span2.sync, true, 'span2.sync=true immediately after creation')
+  span2.end()
+  t.strictEqual(span2.sync, true, 'span2.sync=true immediately after end')
 
   setImmediate(() => {
     t.strictEqual(trans.sync, false)
-    t.strictEqual(span.sync, false)
-    t.end()
-  })
-})
-
-test('sync/async tracking 2', function (t) {
-  var trans = agent.startTransaction()
-
-  var span = agent.startSpan()
-  t.strictEqual(span.sync, true, 'span.sync=true immediately after creation')
-
-  span.end()
-  t.strictEqual(span.sync, true, 'span.sync=true immediately after end')
-
-  setImmediate(() => {
-    t.strictEqual(span.sync, true,
-      'span.sync=true later after having ended sync')
+    t.strictEqual(span1.sync, false)
+    t.strictEqual(span2.sync, true,
+      'span2.sync=true later after having ended sync')
     t.end()
   })
 })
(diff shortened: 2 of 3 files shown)


trentm:
So far this is just a test case to show the error case. To run it (using test.only so this obviously can't be committed as is):

```
node test/instrumentation/async-hooks.js
```

trentm:
My test file (only slight tweak on Alan's from the issue):

```
% cat syncspans.js
  serverUrl: 'http://localhost:8200'
})

const http = require('http')

const requestListener = function (req, res) {
  const span1 = apm.startSpan('This is span 1')
  span1.end()
  const span2 = apm.startSpan('This is span 2')
  span2.end()
  const span3 = apm.startSpan('This is span 3')
  span3.end()
  res.writeHead(200)
  res.end('Hello, World!')
}

const server = http.createServer(requestListener)
server.listen(8080, function () {
  console.log('listening http://localhost:8080')
})
[15:41:46 trentm@pink:~/el/apm-agent-nodejs2 (git:trentm/fix-sync-span)]
% curl http://localhost:8080
Hello, World!%
```

And the transaction and spans sent to APM server for that curl:

```
    {"transaction":{"name":"GET unknown route","type":"request","result":"HTTP 2xx","id":"c7d0b944cb848078","trace_id":"229244c41ce5059ca688e542db46c171","subtype":null,"action":null,"duration":9.682,"timestamp":1605656518924070,"sampled":true,"context":{"user":{},"tags":{},"custom":{},"request":{"http_version":"1.1","method":"GET","url":{"raw":"/","protocol":"http:","hostname":"localhost","port":"8080","pathname":"/","full":"http://localhost:8080/"},"socket":{"remote_address":"::1","encrypted":false},"headers":{"host":"localhost:8080","user-agent":"curl/7.64.1","accept":"*/*"}},"response":{"status_code":200,"headers":{"date":"Tue, 17 Nov 2020 23:41:58 GMT","connection":"keep-alive","transfer-encoding":"chunked"}}},"sync":false,"span_count":{"started":3},"outcome":"success"}}
    {"span":{"name":"This is span 1","type":"custom","id":"4fef0ca1a9c177eb","transaction_id":"c7d0b944cb848078","parent_id":"c7d0b944cb848078","trace_id":"229244c41ce5059ca688e542db46c171","subtype":null,"action":null,"timestamp":1605656518927589,"duration":2.246,"stacktrace":[{"filename":"lib/instrumentation/transaction.js","lineno":112,"function":"Transaction.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/transaction.js"},{"filename":"lib/instrumentation/index.js","lineno":283,"function":"Instrumentation.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/index.js"},{"filename":"lib/agent.js","lineno":107,"function":"Agent.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/agent.js"},{"filename":"syncspans.js","lineno":8,"function":"requestListener","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/syncspans.js"},{"filename":"events.js","lineno":198,"function":"emit","library_frame":true,"abs_path":"events.js"},{"filename":"lib/instrumentation/http-shared.js","lineno":67,"function":"Server.emit","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/http-shared.js"},{"filename":"_http_server.js","lineno":695,"function":"parserOnIncoming","library_frame":true,"abs_path":"_http_server.js"},{"filename":"_http_common.js","lineno":111,"function":"parserOnHeadersComplete","library_frame":true,"abs_path":"_http_common.js"}],"sync":true,"outcome":"unknown"}}
    {"span":{"name":"This is span 2","type":"custom","id":"9625f8a75437755e","transaction_id":"c7d0b944cb848078","parent_id":"c7d0b944cb848078","trace_id":"229244c41ce5059ca688e542db46c171","subtype":null,"action":null,"timestamp":1605656518930645,"duration":0.282,"stacktrace":[{"filename":"lib/instrumentation/transaction.js","lineno":112,"function":"Transaction.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/transaction.js"},{"filename":"lib/instrumentation/index.js","lineno":283,"function":"Instrumentation.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/index.js"},{"filename":"lib/agent.js","lineno":107,"function":"Agent.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/agent.js"},{"filename":"syncspans.js","lineno":10,"function":"requestListener","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/syncspans.js"},{"filename":"events.js","lineno":198,"function":"emit","library_frame":true,"abs_path":"events.js"},{"filename":"lib/instrumentation/http-shared.js","lineno":67,"function":"Server.emit","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/http-shared.js"},{"filename":"_http_server.js","lineno":695,"function":"parserOnIncoming","library_frame":true,"abs_path":"_http_server.js"},{"filename":"_http_common.js","lineno":111,"function":"parserOnHeadersComplete","library_frame":true,"abs_path":"_http_common.js"}],"sync":true,"outcome":"unknown"}}
    {"span":{"name":"This is span 3","type":"custom","id":"1b4a5cfb660f9af4","transaction_id":"c7d0b944cb848078","parent_id":"c7d0b944cb848078","trace_id":"229244c41ce5059ca688e542db46c171","subtype":null,"action":null,"timestamp":1605656518931031,"duration":0.173,"stacktrace":[{"filename":"lib/instrumentation/transaction.js","lineno":112,"function":"Transaction.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/transaction.js"},{"filename":"lib/instrumentation/index.js","lineno":283,"function":"Instrumentation.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/index.js"},{"filename":"lib/agent.js","lineno":107,"function":"Agent.startSpan","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/agent.js"},{"filename":"syncspans.js","lineno":12,"function":"requestListener","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/syncspans.js"},{"filename":"events.js","lineno":198,"function":"emit","library_frame":true,"abs_path":"events.js"},{"filename":"lib/instrumentation/http-shared.js","lineno":67,"function":"Server.emit","library_frame":false,"abs_path":"/Users/trentm/el/apm-agent-nodejs2/lib/instrumentation/http-shared.js"},{"filename":"_http_server.js","lineno":695,"function":"parserOnIncoming","library_frame":true,"abs_path":"_http_server.js"},{"filename":"_http_common.js","lineno":111,"function":"parserOnHeadersComplete","library_frame":true,"abs_path":"_http_common.js"}],"sync":true,"outcome":"unknown"}}
```

astorm:
Just talking the async-hooks logic through out loud -- no conclusions yet

- Node will call the hook's `init` method whenever an object that _might_ produce an asynchronous event is instantiated.  When this happens —
    - We write the current transaction to the `activeTransactions` Map, indexed by the unique async_hook ID.
    - We write _a span_ (`ins.bindingSpan || ins.activeSpan` -- let's assume this means "the current span" for now) to the `activeSpan` Map, indexed by the unique async_hook ID.
    - We keep track of a list of every async ID for the transaction via the `contexts` variable, which is a `WeakMap` (an array of ids indexed by transaction)
    
- Node will call the hook's `before` method right before initiating an async operation, passing in the unique async hook id we saw in `init`. When this happens —
    - We fetch the span from `activeSpans` by that ID -- this span gets marked as `sync=false` even if it's ended. 
    
So a span gets added to `activeSpans` if it happens to be "the current" span when some async-y thing _might_  happen in the future.  When the async-y thing finally does happen, we mark the span as `sync=false` regardless of whether the span is still active or not, and seemingly regardless of whether it was the span that initiated that async work.  

So saying all that out loud, 

- _What_ async operations are triggering this -- especially for the case where we start, then stop, immediately. 
- If async work is _initiated_ during an active span that seems like something we might want to capture -- but the async hook's `init` method doesn't represent the work being initiated, it just represents that work _might_ be initiated.

astorm:
> What async operations are triggering this -- especially for the case where we start, then stop, immediately.

Looks like the async operation type is a `TickObject`.

lib/instrumentation/async-hooks.js
```diff
@@ -90,11 +90,11 @@ module.exports = function (ins) {
 
   function before (asyncId) {
     const span = activeSpans.get(asyncId)
-    if (span) {
+    if (span && !span.ended) {
       span.sync = false
```
astorm, line 94: I see here that we are checking to see if the span had ended before marking is as an asynchronous span.  To repeat discussions we've had elsewhere, this is the correct behavior. 👍  This `before` handler will fire before a specific asyncronous resource's callback runs. If a span has ended before this callback runs, it means the span completed without "crossing an asynchronous boundary", and is a sync span.

test/instrumentation/async-hooks.js
```diff
@@ -125,12 +125,20 @@ test('sync/async tracking', function (t) {
   var trans = agent.startTransaction()
   t.strictEqual(trans.sync, true)
 
-  var span = agent.startSpan()
-  t.strictEqual(span.sync, true)
+  var span1 = agent.startSpan()
```
astorm, line 128: I see in this test we're explicitly starting and stopping `span2` in the same execution/async context, and then making sure its `sync` value remains `true`.  I also see we're testing `span1`'s `sync` value _inside_ the `setImmediate` async context to make sure it's been identified as an asynchronous span.

astorm (approved):
_does the thing it says it does_ -- approving.

PR #1835, opened 2020-10-23
Closes #1834 fix: Treat set-cookie in response headers as sensitive data

StoraH:

Issue #1811, opened 2020-08-19
[META 319] Align sanitize_field_names option + central config


felixbarny:
See elastic/apm#319

trentm:
Should this point to https://github.com/elastic/apm/pull/334 now and be on the 7.11 milestone?

felixbarny:
that's correct

trentm:
From @AlexanderWert on #1871:
"Note that the implementation of this also includes adding the option to [Kibana](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts)."

astorm:
https://github.com/elastic/apm-agent-nodejs/pull/1898. implements this functionality, and closes this out.

PR #1886, opened 2020-11-20
Closes #1834 fix: Treat set-cookie in response headers as sensitive data

### LINKED ISSUE #1834: Auto filter sensitive header set-cookie not filtered for response headers
**Describe the bug**

When filterHttpHeaders is set to true (ELASTIC_APM_FILTER_HTTP_HEADERS) it's expected to redact sensitive cookie and set-cookie information from the request and response headers. This is done in http-headers.js but only the request headers and not the response headers is getting filtered. This is mainly an issue for the set-cookie in the response header.

**To Reproduce**

1. Setup the APM and set ELASTIC_APM_FILTER_HTTP_HEADERS to true (Should be true by default)
2. Do a call that have a set-cookie with a sensitive value in the response. Ex. A login request
3. Check the APM logs and check the response headers and the set-cookie shows up in clear text.

**Expected behavior**

The expected behavior is that the headers with sensitive values should be masked by the string [REDACTED]. Right now to check what is a sensitive value the node package [is-secret](https://www.npmjs.com/package/is-secret) is used. It's a good starting point but to help this along it would be desirable to be able to set additional headers and cookies that contains sensitive information.

Something like this as environment variable could be a solution:

- ELASTIC_APM_SENSETIVE_HTTP_HEADERS = auth,secreat
- ELASTIC_APM_SENSETIVE_HTTP_COOKIES = auth,private_token

- OS: OSX
- Node.js version: 12.18.3
- APM Server version: 7.9.2
- Agent version: 3.4.0

- [X] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)

StoraH:
Re-submission of https://github.com/elastic/apm-agent-nodejs/pull/1835

### DIFF
--- a/lib/filters/http-headers.js
+++ b/lib/filters/http-headers.js
@@ -9,10 +9,16 @@ const SetCookie = require('set-cookie-serde')
 module.exports = httpHeaders
 
 function httpHeaders (obj) {
-  const headers = obj.context && obj.context.request && obj.context.request.headers
+  const requestHeaders = obj.context && obj.context.request && obj.context.request.headers
+  const responseHeaders = obj.context && obj.context.response && obj.context.response.headers
 
-  if (!headers) return obj
+  if (requestHeaders) filterSensitiveHeaders(requestHeaders)
+  if (responseHeaders) filterSensitiveHeaders(responseHeaders)
 
+  return obj
+}
+
+function filterSensitiveHeaders (headers) {
   for (const key in headers) {
     switch (key.toLowerCase()) {
       case 'authorization':
@@ -41,8 +47,6 @@ function httpHeaders (obj) {
         break
     }
   }
-
-  return obj
 }
 
 function stringify (value) {
--- a/test/filters.js
+++ b/test/filters.js
@@ -10,14 +10,23 @@ function makeTransactionWithHeaders (headers) {
     context: {
       request: {
         headers
+      },
+      response: {
+        headers
       }
     }
   }
 }
 
-function getHeaders (result) {
+function getRequestHeaders (result) {
   try {
     return result.context.request.headers
+  } catch (err) {}
+}
+
+function getResponseHeaders (result) {
+  try {
+    return result.context.response.headers
   } catch (err) {}
 }
 
@@ -35,7 +44,14 @@ test('set-cookie', function (t) {
     })
   )
 
-  t.deepEqual(getHeaders(result), {
+  t.deepEqual(getRequestHeaders(result), {
+    'set-cookie': [
+      'password=%5BREDACTED%5D',
+      'card=%5BREDACTED%5D; Secure'
+    ]
+  })
+
+  t.deepEqual(getResponseHeaders(result), {
     'set-cookie': [
       'password=%5BREDACTED%5D',
       'card=%5BREDACTED%5D; Secure'


astorm:
jenkins run the tests please

astorm:
jenkins run the tests please

astorm:
Thanks for the re-submission @StoraH -- it looks like one of our internal CI jobs is hanging.  I'm working with a team here to get that sorted out.  Once we do that, we should be to accept.  Thanks again for finding the bug, making the PR, and jumping through the hoops to get this submitted :)

astorm:
@StoraH I'm not sure if something changed in the branch or if I missed this the first time around, but I saw some odd behavior with your test in my final "pre-merge" manual testing and wanted to get your opinion on it. 

If I take your [`test/filters.js`](https://github.com/elastic/apm-agent-nodejs/blob/aa6b91ff97a9d601a23f998b8b9ef772c224d056/test/filters.js) file, and run it against the _master_ branch -- it doesn't produce a failure.  I would expect it to produce a failure on `master`, and pass on this feature branch.  

I _think_ the problem is the creation of the test fixture.  

```
function makeTransactionWithHeaders (headers) {
  return {
    context: {
      request: {
        headers
      },
      response: {
        headers
      }
    }
  }
}
```

When we create the test fixture, we're using **same** instance object for the `request` and the `response` header.  This means when we filter this fixture and `request` gets filtered that the `response` object is also filtered -- because `response` and `request` both reference the same object.  This means it appears as though `response` has been filtered when, in the real world, `response` wouldn't be filtered because it was be a different object.

I'm guessing something like this would fix it

```
function makeTransactionWithHeaders (headersRequest, headersResponse) {
  return {
    context: {
      request: {
        headers:headersRequest
      },
      response: {
        headers:headersResponse
      }
    }
  }
}

//...

const headers = {
      'set-cookie': [
        'password=this-is-a-password',
        'card=1234%205678%201234%205678; Secure'
      ]
    }
const result = filters.process(
    makeTransactionWithHeaders(
        headers,
        Object.assign({}, headers) // clone the object
    )
  )
  
```

Do you think this is something we should fix before landing this?

astorm:
jenkins run the tests please

astorm:
@StoraH Hello!  We're about to begin work on [Align sanitize_field_names option + central config](https://github.com/elastic/apm-agent-nodejs/issues/1811), which is going to mean touching `lib/filters/http-headers.js`.  Because of that we'd like to get your work landed.  We'll merge the PR, and **_we'll take care_** of that one final nit with the test not working as  expected.  It's still worth your looking at though -- it's a nice lesson in how objects are passed around and referenced in javascript. 

Thanks again for finding and fixing this bug!

astorm (approved):
re approving from accidental close

StoraH:
@astorm Hi! That's super good to hear! I'm glad this could move forward without me helping out at the last bit.
I've been home sick with covid and on top of that I replaced my computer with the new Macbook PRO with the M1 chip and sadly that can't run docker yet so weren't able to run the tests😅 I read thought the code again tho, and yes, that is a nice lesson :) I would say this whole experience have been a good lesson and I'm very glad how I was greeted into this community. I will gladly to more pull requests in the future🚀

All the best!

PR #1890, opened 2020-11-25
fix: spans started in same async op should be children

### LINKED ISSUE #1889: `startSpan('A'); startSpan('B')` in the same async op: B should be a child of A, but isn't
If you start a span (B) in **the same async operation** as a preceding span (A), the current agent (v3.8.0) will make A and B siblings. I think B should be a child span of A.

If you'd like to test your own expectations, here is a sort of survey: https://gist.github.com/trentm/63e5dbdeded8b568e782d1f24eab9536

**Works as expected: each span in separate async ops**

The common case (at least I assume) in node.js is that code significant enough to be a traced span are in separate async ops on the node.js event loop. E.g.:

```
app.get('/my-endpoint', (req, res) => {
  var A = apm.startSpan('this is span A')
  setTimeout(function () {
    var B = apm.startSpan('this is span B')
    setTimeout(function () {
      B.end()
      A.end()
    }, 10)
  }, 10)
})
```

For this case, the APM agent generates a trace like this (where indentation denotes a parent/child relationship):

```
- transaction: GET /my-endpoint
    - span: this is span A
        - span: this is span B
```

This works as I'd expect.

**The contested case: multiple spans started in the same async op**

```
app.get('/my-endpoint', (req, res) => {
  var A = apm.startSpan('this is span A')

  var B = apm.startSpan('this is span B')
  B.end()

  A.end()
})
```

For this code, the current APM agent generates a trace like this:

```
- transaction: GET /my-endpoint
    - span: this is span A
    - span: this is span B
```

I think this is a bug, or at least should be considered. My expectation for the above code is a trace like this:

```
- transaction: GET /my-endpoint
    - span: this is span A
        - span: this is span B
```

**The patch**

If you want to try it out, here is a patch to lib/instrumentation/index.js that changes to my expected behaviour: https://gist.github.com/trentm/63e5dbdeded8b568e782d1f24eab9536#file-the-diff
I'll start a draft PR in a little bit with that.

**Proposal**

- [ ] Get a feel for the test suite fall out.
- [ ] Discuss with others whether their expectations are the same (so far it is just @astorm and I).
- [ ] If considering changing this behaviour, do we consider it a *breaking changing*? I'm not aware of specific documentation on this scenario, but I haven't looked very hard. The patch *does* result in at least a few failing tests.

trentm:
This commit is just a start. There are still tests and discussion to be
had.

Fixes #1889

### DIFF
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -17,7 +17,14 @@ module.exports = Span
 util.inherits(Span, GenericSpan)
 
 function Span (transaction, name, ...args) {
-  var childOf = transaction._agent._instrumentation.activeSpan || transaction
+  var childOf = transaction._agent._instrumentation.activeSpan ||
+    transaction._agent._instrumentation.bindingSpan ||
+    transaction
+  console.log('XXX new Span: childOf set to',
+    'activeSpan?', !!transaction._agent._instrumentation.activeSpan,
+    'bindingSpan?', !!transaction._agent._instrumentation.bindingSpan,
+    'transaction?', !!transaction,
+    'childOf.id:', childOf.id)
   const opts = typeof args[args.length - 1] === 'object'
     ? (args.pop() || {})
     : {}
@@ -71,6 +78,7 @@ Span.prototype.end = function (endTime) {
 
   this._timer.end(endTime)
   this._agent._instrumentation._recoverTransaction(this.transaction)
+  this._agent._instrumentation.bindingSpan = null
 
   this.ended = true
   this._agent.logger.debug('ended span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type, subtype: this.subtype, action: this.action })


trentm:
Here is a case, from @dgieselaar, that shows this patch is insufficient.

```js
// ... apm and app (express) setup ...
app.get('/nspans-dario', (req, res) => {
  var firstSpan = apm.startSpan('first-span', 'type', 'subtype', 'action')
  process.nextTick(function ( ) {
    firstSpan.end()
    var secondSpan = apm.startSpan('second-span', 'type', 'subtype', 'action')
    secondSpan.end()
    res.end('done')
  })
})
```

The transaction/span hierarchy resulting from a call to this is:

```
    transaction
        span first-span
            span second-span
```

We are no longer in the same async op here, so technically wasn't what this ticket was first considering. However, this is likely a more common case and we should consider it.

Take a look at Dario's draft PR here: https://github.com/elastic/apm-agent-nodejs/pull/1964

trentm:
This can be closed if/when https://github.com/elastic/apm-agent-nodejs/pull/2181 goes in.

PR #1891, opened 2020-11-25
feat: support fastify 3

Marsup:
Hi,

This module seems to break fastify v3 when used with typescript as `default` and `fastify` are not exposed on the exported module anymore, so the imports are invalid. Also, starting from fastify 3.3 a few properties are exposed directly on the request but I don't think it's worth an entirely different wrapper.

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -8,6 +8,10 @@ endif::[]
 ==== x.x.x - YYYY/MM/DD
 
 * fix: Synchronous spans would never have `span.sync == true`. {pull}1879[#1879]
+* feat: support fastify 3
+  Adds .default and .fastify module.exports to instrumented fastify function
+  for 3.x line, and prefers req.routerMethod and req.routerPath for
+  transaction name
 
 [float]
 ===== Breaking changes
--- a/lib/instrumentation/modules/fastify.js
+++ b/lib/instrumentation/modules/fastify.js
@@ -9,7 +9,36 @@ module.exports = function (fastify, agent, { version, enabled }) {
 
   agent.logger.debug('wrapping fastify build function')
 
-  return semver.gte(version, '2.0.0-rc') ? wrappedBuild2Plus : wrappedBuildPre2
+  if (semver.gte(version, '3.0.0')) {
+    wrappedBuild3Plus.default = wrappedBuild3Plus
+    wrappedBuild3Plus.fastify = wrappedBuild3Plus
+    return wrappedBuild3Plus
+  }
+
+  if (semver.gte(version, '2.0.0-rc')) return wrappedBuild2Plus
+
+  return wrappedBuildPre2
+
+  function wrappedBuild3Plus () {
+    const _fastify = fastify.apply(null, arguments)
+
+    agent.logger.debug('adding onRequest hook to fastify')
+    _fastify.addHook('onRequest', (req, reply, next) => {
+      const method = req.routerMethod || req.raw.method // Fallback for fastify >3 <3.3.0
+      const url = req.routerPath || reply.context.config.url // Fallback for fastify >3 <3.3.0
+      const name = method + ' ' + url
+      agent._instrumentation.setDefaultTransactionName(name)
+      next()
+    })
+
+    agent.logger.debug('adding onError hook to fastify')
+    _fastify.addHook('onError', (req, reply, err, next) => {
+      agent.captureError(err, { request: req.raw })
+      next()
+    })
+
+    return _fastify
+  }
 
   function wrappedBuild2Plus () {
     const _fastify = fastify.apply(null, arguments)


astorm:
jenkins run the tests please

astorm:
@Marsup Thanks for the bug report AND the proactive PR. That's always appreciated.  

Quick question:  When you say

> seems to break fastify v3 when used with typescript

Can you provide a little more context (and maybe a set of steps to reproduce?)

I ask because I just [installed the fastify-typescript-starter](https://github.com/matschik/fastify-typescript-starter), installed the agent, and [imported the agent](https://www.elastic.co/guide/en/apm/agent/nodejs/current/es-modules.html#esm-typescript) and was able to get up and running. It sounds like your experience was different.

If you can let us know the broken behavior you're seeing we'll be able to confirm it on this side, confirm your PR provides a fix, and then work with you to get this merged.

Marsup:
Sure, this might be somewhat specific to how we are set up.
To follow along on your reproduction:
- in `tsconfig.json`, set `esModuleInterop` to `false`
- in `src/app.ts`, at the very beginning of the file, before importing fastify, put this:
```ts
import * as apm from 'elastic-apm-node'
apm.start({ ... })
```
- `npm run build && npm start` should get you this error `TypeError: fastify_1.default is not a function`

astorm (approved):
Root cause analysis is below, TL;DR; -- solving this here seems like right thing and we have existing tests that should ensure the transaction naming doesn't cause a regression. Approving

One question and one request before we merge.

@Marsup More out of general typescript curiosity than anything else: Is there a reason you'all are using a configuration with `esModuleInterop` set to `false` instead of `true`?  Was this a deliberate choice or just a default you ran with?

@Marsup One request before we merge: would you mind adding a changelog entry to the [CHANGELOG.asciidoc](https://github.com/elastic/apm-agent-nodejs/blob/master/CHANGELOG.asciidoc) file in the `==== x.x.x - YYYY/MM/DD` section?  Something like the following should do.

* feat: support fastify 3
  Adds `.default` and `.fastify` module.exports to instrumented fastify function
  for 3.x line, and prefers `req.routerMethod` and `req.routerPath` for
  transaction name

## Root Cause

Looking at the root cause of this -- without `esModuleInterop`, the following typescript

```
    import fastify from "fastify";
    import router from "./router";

    const server = fastify({
      // Logger only for production
      logger: !!(process.env.NODE_ENV !== "development"),
    });

    // Middleware: Router
    server.register(router);

    export default server;
```
    
Compiles down to this

```
    "use strict";
    Object.defineProperty(exports, "__esModule", { value: true });
    const fastify_1 = require("fastify");
    const router_1 = require("./router");
    const server = fastify_1.default({
        logger: !!(process.env.NODE_ENV !== "development"),
    });
    server.register(router_1.default);
    exports.default = server;
    //# sourceMappingURL=app.js.map%                                                                            
```
    
That `.default` property on modules appears [to be a Babel thing](https://babeljs.io/docs/en/babel-plugin-transform-modules-commonjs).  When Babel compiles an ECMAScript module to CommonJS, something like

```
    export default 42;
``` 

becomes

```
    exports.default = 42;
```
    
It looks like fastify 3+ is [doing some clever shenanigans](https://github.com/fastify/fastify/blob/ab1707f68300abb7ed9da2404bcad4cac5a45535/fastify.js#L633) in order to support their CommonJS module being imported in a variety of different ways. This includes defining a `default` property themselves.  For the agent, this means when we replace the `fastify` module via [`require-in-the-middle`](https://www.npmjs.com/package/require-in-the-middle) with a new object/function, our new object/function doesn't have a `default` property.  However, the TypeScript compiler can't know this, so the TypeScript compiler still produces the same output. 

I took a deeper look at the modules we're instrumenting and our dev dependencies, only `got` (which we don't instrument) was doing something similar with `default`  i.e. this doesn't seem like a widespread enough pattern for us to fix generally and makes sense for us to fix directly in the fastify instrumentation.

Marsup:
Changelog updated, am I supposed to add that to the commit as well ?

This is a module re-used by others, since a module built with `esModuleInterop: true` can only be used if you do so as well apparently, we didn't want to impose that on the consumer side if we could help it.

Regardless, I think it's mandatory that wrappers have an identical public interface to that of the original module, and it seems quite hard (if at all possible) to do it in a generic way for any module out there, this could be done for the standard `default` export but it wouldn't solve it if for some reason you use the `import { fastify } from 'fastify'` format, which by the way would equally break without typescript if you did `const { fastify } = require('fastify')` without this patch.

Thanks for the detailed write-up :slightly_smiling_face:

astorm:
jenkins run the tests

astorm:
jenkins run the tests

astorm:

astorm (approved):
> Changelog updated, am I supposed to add that to the commit as well ?

Yes, -- we're starting to ask that every substantial PR include a human readable changelog entry.  We've only [just added this to the checklist](https://github.com/elastic/apm-agent-nodejs/pull/1894) though, and your PR predates this change in process.  Thanks for being flexible and making that change.

> This is a module re-used by others, since a module built with esModuleInterop: true can only be used if you do so as well apparently, we didn't want to impose that on the consumer side if we could help it.

_nod_ -- this makes sense.  We're approaching typescript slowly and trying to learn from what others are doing.  Thank you for sharing your experience. 

> Regardless, I think it's mandatory that wrappers have an identical public interface to that of the original module, and it seems quite hard (if at all possible) to do it in a generic way for any module out there

Agreed.  Just musing out loud for the future -- I wonder if we ought to have a set of tests that compares the properties on a naturally exported module with the properties on module that's wrapped.  

PR looks good -- approving and merging.

PR #1898, opened 2020-11-30
feat: sanitize field names

astorm:
**Update**

Subtitles abound here.  

The end goal of all this is to get version of this feature that is suitable for a minor version bump while setting us up for 100% spec and agent compliance in our next major version.

Our current spec discussion say that the `SANITIZE_FIELD_NAMES` should fully _READACT_ the value of a header, post field, or cookie field if they're listed in the `SANITIZE_FIELD_NAMES` array.  These _might_ be captured by agents in the following fields (see also [the intake API/Schema](https://github.com/elastic/apm-server/blob/d18c559e4136b2213a4f6e206258c5d303ec1b0f/docs/spec/v2/transaction.json#L129-L135))

- `transaction.context.request.cookies`
- `transaction.context.request.headers`
- `transaction.context.request.body` (if form/url encoded)
- `transaction.context.response.headers`

The node agent currently captures `transaction.context.request.headers`, `transaction.context.request.body`, and `transaction.context.response.headers`, but does not capture `transaction.context.request.cookies`.  

Also worth noting -- the Java agent also has a behavior where it blanks out the values of `transaction.context.request.headers.cookies` when it parses values into `transaction.context.request.cookies`.

The Node.js agent also has a preexisting `http-filter` module that works slightly differently from the behavior of `SANITIZE_FIELD_NAMES`.  Prior to this PR the `http-filter` module was hard-coded to redact the `authorization` HTTP header.  The filter also looked for both `set-cookie` and `cookie` HTTP headers, and would parse them into name/value pairs in order to redact individual cookie fields.  Once redacted, the name/value pairs would be re-serialized as a cookie string and **captured in the `.headers` properties**. Which individual cookie names got redacted was/is controlled by a list of fields in `redact-secrets`.

The `redact-secrets` module also looks for values to redact.  Specifically, it looks for things that match a credit card regular expression. 

There's a few important subtleties here. 

1. The new `SANITIZE_FIELD_NAMES` feature, when combined with the default configuration values, indicates that we should _completely redact_ the `set-cookie` **header**.

2. The feature, as spec-ed, does not speak to parsing the values in the HTTP headers and redacting them.  It only speaks to parsing values collected in `transaction.context.request.cookies`

3. The Node.js agent does not appear to have used `redact-secrets` to control which HTTP headers or post-body fields/values are redacted.  The `redact-secrets` list only controls which "cookie field names encoded in the headers" were redacted.

**This PR**

In order to avoid a major version bump  this PR 

1. Removes the hard coded `authorization` redacting, and instead relies on `authorization` being one of the default values of `SANITIZE_FIELD_NAMES`

2. Maintains the "cookie in header" parsing via `redact-secrets`, but also includes the `SANITIZE_FIELD_NAMES` wildcard patterns as part of `redact-secrets` 

3. If either `cookie` or `set-cookie` is listed in the `SANITIZE_FIELD_NAMES` configuration array, per the spec this value will be completely redacted. 

4. The default value of `SANITIZE_FIELD_NAMES` will include the extra configuration values `['pw','pass','connect.sid']` (which we have currently proposed in the spec as `MAY` include values)

The preserves as much of the previous Node.js agent behavior as possible in order to prevent a major version bump.  The one "controversial" behavior I see here is with the new feature behavior, `set-cookie` will be completely redacted by default -- whereas in the previous versions of the agent only specific fields would be redacted.  I'm choosing to interpret this as a new enhanced behavior vs. a breaking change.  

**Future Next Major Version PR**

In order to be fully spec compliant I also propose that we [revisit this when 4.0 rolls around](https://github.com/elastic/apm-agent-nodejs/issues/1924).  At this time we'll

1. Remove `['pw','pass','connect.sid']` from the `SANITIZE_FIELD_NAMES` defaults
2. Conform to the java agent behavior of capturing cookies in `transaction.context.request.cookies` and blanking out `transaction.context.request.headers.cookies`
3. Redact/sanitize `transaction.context.request.cookies` per spec
4. Remove the individual redacting of the `set-cookie` http  header

---

**Original Description**

This PR implements the [Data sanitization spec](https://github.com/elastic/apm/blob/master/specs/agents/sanitization.md).

This PR

1. Adds configuration handling for the new `sanitizeFeatureName` value
2. Implements filtering of payload request/response headers and the request body for `application/x-www-form-urlencoded` based on `sanitizeFeatureName`

Fields are "removed" from the payload object by setting their values to `undefined` in the javascript payload object.  This will result in them being skipped in their JSON serialization 

    > JSON.stringify({foo:"bar",zip:undefined})
    '{"foo":"bar"}'

and allows us to avoid any perf. issues with `delete`. 

During testing we discovered that the agent does not capture `application/x-www-form-urlencoded`  request body payloads for  [fastify](https://github.com/elastic/apm-agent-nodejs/issues/1906), [hapi](https://github.com/elastic/apm-agent-nodejs/issues/1905), and [koa](https://github.com/elastic/apm-agent-nodejs/issues/1904).  This is why those tests skip those fixtures.

**Tests**:

Additional packages (middleware body parsers for web frameworks) were added to our dev dependencies to allow integration testing of the `application/x-www-form-urlencoded` handling.

The `test/sanitize-field-names/main.js` file contains unit-ish tests for the new functions.

The rest of the tests are integration-ish tests where we make HTTP POSTs with each of our supported frameworks and different configured values to determine if the right items are removed from the payloads.  Test fixtures are shared between all the different framework integration tests. Fixtures contain input values and expected output values, as well as which style middleware should be used for parsing.

### To Do

- [x] Integration tests -- with full agent config loading lifecycle _and_ with various web frameworks handling of req.body
- [x] Document new configuration value
- [x] tests hanging in startTransaction 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


.tav.yml
```diff
@@ -339,3 +339,8 @@ finalhandler:
 memcached:
   versions: '>=2.2.0'
   commands: node test/instrumentation/modules/memcached.js
+
+body-parser:
```
astorm, line 343: Express has an external body parser middleware module for capturing `application/x-www-form-urlencoded` form bodies, so we run it's tests through the version runner. 

Restify and HAPI do not use external middleware for capturing `application/x-www-form-urlencoded` form bodies, so there's no module to run through tav.

Fastify and koa _do_ use external middleware for capturing `application/x-www-form-urlencoded` bodies, but the agent doesn't currently capture their request bodies, so we don't test those module (yet!)

docs/configuration.asciidoc
```diff
@@ -692,6 +692,18 @@ The timeout is applied once the agent has sent the entire request body to the AP
 If the response from the server takes longer than allowed by this timeout,
 the HTTP request is terminated and the TCP socket closed.
 
+[[sanitize-field-names]]
+==== `sanitizeFieldNames`
+* *Type:* Array
+* *Default:* ['password', 'passwd', 'pwd', 'secret', '\*key', '*token*', '\*session*', '\*credit*', '\*card*', 'authorization', 'set-cookie']
```
astorm, line 698: @bmorelli25 do you know how asciidoc `*` escaping works? I tried escaping each individual item but that ended up with a mismash of bolding.  The above seems to work, but seems odd that `token` doesn't need escaping.

lib/config.js
```diff
@@ -66,6 +66,9 @@ var DEFAULTS = {
   logUncaughtExceptions: false, // TODO: Change to `true` in the v4.0.0
```
astorm, line 66: Everything in this file is just configuration handling (including converting the wildcards into regular expressions)

lib/filters/http-headers.js
```diff
@@ -20,6 +20,11 @@ function httpHeaders (obj) {
```
astorm, line 20: This is weird one. Because of the way our pre-existing header **redacting** (not filtering) interacted with the descision to use `undefined`, without this change out `undefined` header would be changed to have a values of `REDACTED`, and would not be removed from the payload.  This change fixes that.

lib/filters/sanitize-field-names.js
```diff
@@ -0,0 +1,85 @@
...
+
+const HEADER_FORM_URLENCODED = 'application/x-www-form-urlencoded'
+
+/**
+ * Handles req.body as object, string, or Buffer
+ *
+ * Express provides multiple body parser middlewares with x-www-form-urlencoded
+ * handling.  See http://expressjs.com/en/resources/middleware/body-parser.html
+ */
+function removeKeysFromPostedFormVariables (body, requestHeaders, regexes) {
```
astorm, line 12: This function is probably more complicated than it needs to be right now.  I originally thought we would need to be able to handle the parsed request body in all its possible formats. The way feature development ending up going the agent had already normalized these values by the time we wanted to handle them.  This could probably be simplified to remove the raw/buffer handling, but I decided to leave this code in place as it may be useful when/if we handle the request body parsing for fastify, hapi, and koa.   I was 52%/48% on this -- not a strongly held decision.

lib/parsers.js
```diff
@@ -142,9 +150,16 @@ exports.getContextFromRequest = function (req, conf, type) {
 
   if (haveBody) {
     if (captureBody) {
+      body = removeKeysFromPostedFormVariables(body, req.headers, conf.sanitizeFieldNamesRegExp)
+
+      if (Buffer.isBuffer(body)) {
```
astorm, line 155: This conditional changes how the agent handles the raw/Buffer body parsing.  Previously, if a customer user the raw/buffer middleware, we'd send up data that looked like this

    '{"type":"Buffer","data":[49,50,51,52,53,54,55]}'

i.e. we weren't really handling this middleware in a way that was useful to users.  This conditional turns the payload Buffer into its string representation.

package.json
```diff
@@ -117,8 +117,10 @@
     "@babel/cli": "^7.8.4",
```
astorm, line 117: The body parers are here for the test suites.  Even though we discovered that out `koa` and `fastify` instrumentation does not capture the body correly (i.e doesn't need to be tested), the test harness I came up with becomes awkward if this middleware isn't there.  The was a waffling decision -- happy to reconsider it.

docs/configuration.asciidoc
trentm: Youch, that's unfortunate.  Frankly that feels like a bug in asciidoc design (or at least a wart), but I don't know it well enough to say.

How about this:

```
* *Default:* +['password', 'passwd', 'pwd', 'secret', '*key', '*token*', '*session*', '*credit*', '*card*', 'authorization', 'set-cookie']+
```

https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#escaping-text

index.d.ts
```diff
@@ -216,6 +216,7 @@ interface AgentConfigOptions {
   metricsInterval?: string; // Also support `number`, but as we're removing this functionality soon, there's no need to advertise it
   payloadLogFile?: string;
   centralConfig?: boolean;
+  sanitizeFieldNames?:Array<string>;
```
trentm, line 219:
```suggestion
  sanitizeFieldNames?: Array<string>;
```

docs/configuration.asciidoc
```diff
@@ -692,6 +692,18 @@ The timeout is applied once the agent has sent the entire request body to the AP
...
+* *Env:* `ELASTIC_SANITIZE_FIELD_NAMES`
+
+Sometimes it is necessary to sanitize the data sent to Elastic APM, e.g. remove sensitive data.
+
+The `sanitizeFieldNames` configuration values allows you to configure a list of wildcard patterns of field names which should be _removed_ from agent payloads. These patterns apply to the request and response HTTP headers, as well as any form field capturing during an `application/x-www-form-urlencoded data` request.
```
trentm, line 703: Does the spec necessarily say these headers should be *removed*? Or just that the sensitive data in them should be cleared. E.g. As a user I might want to know that my form data included a "passwd", even if the value of that field is "REDACTED" or whatever.

XXX

trentm, line 703: This section should briefly explain the "wildcard" pattern syntax: specifically that it is case-insensitive by default and the `(?-i)` flag.

lib/config.js
```diff
@@ -383,9 +390,20 @@ function normalize (opts) {
   normalizeTime(opts)
   normalizeBools(opts)
   normalizeIgnoreOptions(opts)
+  normalizesanitizeFieldNames(opts)
```
trentm, line 393:
```suggestion
  normalizeSanitizeFieldNames(opts)
```

lib/config.js
```diff
@@ -383,9 +390,20 @@ function normalize (opts) {
...
   truncateOptions(opts)
 }
 
+function normalizesanitizeFieldNames (opts) {
```
trentm, line 397:
```suggestion
function normalizeSanitizeFieldNames (opts) {
```

lib/filters/http-headers.js
```diff
@@ -20,6 +20,11 @@ function httpHeaders (obj) {
 
 function filterSensitiveHeaders (headers) {
   for (const key in headers) {
+    // undefined headers will be dropped when serialized as
+    // json, and don't need to be redacted
+    if (typeof headers[key] === 'undefined') {
```
trentm, line 25: nit: This works as well and theoretically should require less work from the JS engine:

```suggestion
    if (headers[key] === undefined) {
```

lib/filters/http-headers.js
```diff
@@ -20,6 +20,11 @@ function httpHeaders (obj) {
 
 function filterSensitiveHeaders (headers) {
   for (const key in headers) {
```
trentm, line 22: Unrelated to this change: Wouldn't it typically be faster to iterate through the set of sensitive headers, rather than iterating through the set of given headers? I say "typically" because my guess is request and response headers generally number greater than 3. Anyway, if you agree, I can open a ticket to consider changing this. Low prio, this is possible premature optimization.

trentm (changes requested):
I'm not done reviewing yet, but I wanted to get this in to discuss my main Q: whether the spec requires *removing* fields rather than just censoring field values.

docs/configuration.asciidoc
bmorelli25: If you turn the entire array into a code snippet, there's no escaping required:

```suggestion
* *Default:* `['password', 'passwd', 'pwd', 'secret', '*key', '*token*', '*session*', '*credit*', '*card*', 'authorization', 'set-cookie']`
```

Screen Shot 2020-12-07 at 11 20 56 AM

CHANGELOG.asciidoc
```diff
@@ -25,10 +25,26 @@ Notes:
...
+
+[float]
+===== Features
+* feat: implemented sanitize_feature_names specification
+
+        Allows users to configure a list of wildcard patterns to _remove_ items
+        from the agent's HTTP header and `application/x-www-form-urlencoded` payloads.
+
+        spec: https://github.com/elastic/apm/blob/master/specs/agents/sanitization.md
+        docs: https://github.com/elastic/apm-agent-nodejs/blob/master/docs/configuration.asciidoc#sanitize-field-names
```
bmorelli25, line 44: Spacing like this doesn't render well in asciidoc. For example, our build  assumes these are undefined code blocks:
Screen Shot 2020-12-07 at 11 24 37 AM

If you want to maintain this spacing, what about something like this?

```suggestion
* feat: implemented sanitize_feature_names specification +
  Allows users to configure a list of wildcard patterns to _remove_ items
  from the agent's HTTP header and `application/x-www-form-urlencoded` payloads.
  ** https://github.com/elastic/apm/blob/master/specs/agents/sanitization.md[spec]
  ** https://github.com/elastic/apm-agent-nodejs/blob/master/docs/configuration.asciidoc#sanitize-field-names[docs]
```

Screen Shot 2020-12-07 at 11 28 37 AM

docs/configuration.asciidoc
```diff
@@ -692,6 +692,18 @@ The timeout is applied once the agent has sent the entire request body to the AP
 If the response from the server takes longer than allowed by this timeout,
 the HTTP request is terminated and the TCP socket closed.
 
+[[sanitize-field-names]]
+==== `sanitizeFieldNames`
+* *Type:* Array
+* *Default:* ['password', 'passwd', 'pwd', 'secret', '\*key', '*token*', '\*session*', '\*credit*', '\*card*', 'authorization', 'set-cookie']
+* *Env:* `ELASTIC_SANITIZE_FIELD_NAMES`
+
+Sometimes it is necessary to sanitize the data sent to Elastic APM, e.g. remove sensitive data.
```
bmorelli25, line 701: When possible, use something like `for example` instead of `e.g.`. If you want to use `e.g.`, a comma should follow. In this case, it sounds like we're saying "in other words", so we should swap this to id est (i.e).
As I type all of this, the sentence still doesn't feel right though. Sanitize is in the name of the field, so we can probably get away with something like this:

```suggestion
Remove sensitive data sent to Elastic APM.
```

docs/configuration.asciidoc
trentm: Brandon, are you sure that works with asciidoc in general? Here is my attempt using GH's rendering of asciidoc showing all the forms we've tried here:
https://gist.github.com/trentm/3bf4bd9473c40dfae7d2221058a26bf1

Note how the "Default (with backticks per Brandon)" case seems to "use" the asterisk on "key" to start bolding -- at least with GH's rendering.

bmorelli25: > Brandon, are you sure that works with asciidoc in general?

Nope, but I'm not surprised that it doesn't. I only tested in our doc build; our implementation of asciidoc doesn't always behave. . . as expected?

Here's what I tried and how it renders in our system:

```
* *Default:* `['password', 'passwd', 'pwd', 'secret', '*key', '*token*', '*session*', '*credit*', '*card*', 'authorization', 'set-cookie']`

* *Default:* ['password', 'passwd', 'pwd', 'secret', '\*key', '*token*', '\*session*', '\*credit*', '\*card*', 'authorization', 'set-cookie']

* *Default:* +['password', 'passwd', 'pwd', 'secret', '*key', '*token*', '*session*', '*credit*', '*card*', 'authorization', 'set-cookie']+

* *Default:* `[+password+, +passwd+, +pwd+, +secret+, +\*key+, +\*token*+, +\*session*+, +\*credit*+, +\*card*+, +authorization+, +set-cookie+]`
```

Screen Shot 2020-12-07 at 2 28 09 PM

bmorelli25: Oh! Also, our build system uses Asciidoctor, not Asciidoc. The syntax is mostly the same, but there are quite a few [differences](https://asciidoctor.org/docs/asciidoc-asciidoctor-diffs/). That's probably the main reason why these are rendering so different.

trentm: Whoa. TIL those are two different things.

bmorelli25: Yeah, that's likely my fault, sorry! I tend to use the two terms interchangeably, which isn't correct.

trentm: Those diffs don't mention the difference in behaviour of `+foo+` for "passthrough", https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#escaping-text

And the https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/#formatted-text section seems to indicate that backticks shouldn't be escaping, only styling as monospace. 

https://asciidoctor.org/docs/migration/#new-syntax refers to changes in backtick and plus handling in asciidoctor v1.5.0. I'm a little confused, but likely not a big deal. :)

bmorelli25: So, yeah, good point. While asciidoctor and asciidoc are different, that doesn't appear to be the reasoning behind these inconsistencies. Our implementation of asciidoctor is highly customized and while some things work the same way, some don't. Consider the https://asciidoctor.org/ reference a great starting point, but not a source of truth. Our source of truth is the https://github.com/elastic/docs readme file. The only downside is it's not as exhaustive as the asciidoctor reference.

trentm: Thanks for the link. That helps.

CHANGELOG.asciidoc
astorm: Ah, good to know.  My heart's still letting go of markdown :). Formatting changed to match the suggestion.

docs/configuration.asciidoc
astorm: @bmorelli25 _nod_ thanks for catching this, I will change it.  FWIW -- I lifted this language [from some of the other agent docs](https://www.google.com/search?client=firefox-b-1-d&ei=oFHZX_6hOuTi9APh_YzQBA&q=site%3Awww.elastic.co+%22e.g.+remove+sensitive+data%22&oq=site%3Awww.elastic.co+%22e.g.+remove+sensitive+data%22&gs_lcp=CgZwc3ktYWIQAzoECAAQR1CKXlivhwFg-4gBaAJwAngAgAFliAGgC5IBBDIwLjGYAQCgAQGqAQdnd3Mtd2l6yAEIwAEB&sclient=psy-ab&ved=0ahUKEwj-8biunNHtAhVkMX0KHeE-A0oQ4dUDCAw&uact=5) -- if you're looking for a break from React land it's probably worth chasing those down.

docs/configuration.asciidoc
astorm: @bmorelli25 The "treat this as code" meta-charaters worked great _and_ match our handling of the other configuration values.  👍

index.d.ts
astorm: ✅

docs/configuration.asciidoc
astorm: Rejiggered things so that `sanitizeFieldNames` redacts the headers instead of removing them.

lib/config.js
astorm: 👍 made this change in a recent commit

lib/config.js
astorm: 👍 made this change in a recent commit

lib/filters/http-headers.js
astorm: This is good optimization to consider, but I agree it may be premature (especially given the unexpected gymnastic to conform to the spec and maintain 3.x compatibility).  I've opened an issue to consider this optimization here: https://github.com/elastic/apm-agent-nodejs/issues/1923.

test/sanitize-field-names/fixtures.js
```diff
@@ -0,0 +1,309 @@
+'use strict'
```
trentm, line 1: This filename should be prefixed with '_' as well to avoid being considered a file with tests.

lib/parsers.js
trentm: I'm uncomfortable making the assumption that the POST'd body *is* meaningfully representable as a string. E.g. here is me with a simple file upload endpoint (using express):

```
var rawParser = bodyParser.raw()
app.post('/see-my-body', rawParser, (req, res) => {
  console.log('XXX req.headers', req.headers);
  console.log('XXX req.body', req.body);
  res.end('I see you')
}
```

and POSTing a small ~60kB [kitty image](https://www.litter-robot.com/blog/wp-content/uploads/2016/08/shutterstock_191770991-scared-cat-900x600.jpg) to it:

```
% curl   -Sis http://localhost:3000/see-my-body --data-binary @kitty.jpg -H content-type:application/octet-stream
HTTP/1.1 100 Continue

HTTP/1.1 200 OK
X-Powered-By: Express
Date: Fri, 18 Dec 2020 20:14:15 GMT
Connection: keep-alive
Content-Length: 9

I see you
```

The log from my app's endpoint is:

```
XXX req.headers { host: 'localhost:3000',
  'user-agent': 'curl/7.64.1',
  accept: '*/*',
  'content-type': 'application/octet-stream',
  'content-length': '69930',
  expect: '100-continue' }
XXX req.body <Buffer ff d8 ff e0 00 10 4a 46 49 46 00 01 01 01 01 2c 01 2c 00 00 ff db 00 43 00 06 04 04 05 04 04 06 05 05 05 06 06 06 07 09 0e 09 09 08 08 09 12 0d 0d 0a ... >
```

*with* this `body = body.toString()` the data sent to APM server looks like this:

Screen Shot 2020-12-18 at 12 13 47 PM

What about considering the capturing of a *binary* body (leading to the current `'{"type":"Buffer","data":[49,50,51,52,53,54,55]}'`) to be a bug? Instead we change the definition of `captureBody: true` to either (a) exclude a binary (i.e. Buffer.isBuffer) body, or explicitly encode it as something like `<Buffer>`:

```js
if (haveBody) {
    if (!captureBody) {
        // Aside: I'm not sure why the agent calls this "REDACTED" here. If
        // the config is `captureBody: false`, then I would just leave
        // context.body = undefined.
        context.body = '[REDACTED]'
    } else if (Buffer.isBuffer(body)) {
        context.body = '<Buffer>'
    } else {
        ...
    }
}
```

trentm (changes requested):
I have one main Q regarding the Buffer handling, and then mostly nits. I haven't yet reviewed all the test files.

test/sanitize-field-names/express.js
```diff
@@ -0,0 +1,91 @@
+'use strict'
+const {
+  createAgentConfig,
+  resetAgent,
+  assertRequestHeadersWithFixture,
+  assertResponseHeadersWithFixture,
+  assertFormsWithFixture
+} = require('./shared')
```
trentm, line 8: Are these intentionally imported before the agent is imported and started?

test/sanitize-field-names/express.js
```diff
@@ -0,0 +1,91 @@
...
+  } else if (type === 'raw') {
+    return bodyParser.raw({ type: '*/*' })
+  }
+
+  throw new Error(`I don't know how to create a ${type} middleware`)
+}
+
+test('Running fixtures with express', function (t) {
+  for (const [, fixture] of fixtures.entries()) {
+    test(fixture.name, function (t2) {
```
trentm, line 78: Perhaps use [subtest](https://github.com/substack/tape#ttestname-opts-cb) syntax here:

```diff
--- a/test/sanitize-field-names/express.js
+++ b/test/sanitize-field-names/express.js
@@ -73,11 +73,11 @@ function createMiddleware (type) {
   throw new Error(`I don't know how to create a ${type} middleware`)
 }

-test('Running fixtures with express', function (t) {
+test('Running fixtures with express', function (suite) {
   for (const [, fixture] of fixtures.entries()) {
-    test(fixture.name, function (t2) {
+    suite.test(fixture.name, function (t) {
       runTest(
-        t2,
+        t,
         fixture.expected,
         createAgentConfig(fixture.agentConfig),
         fixture.input.requestHeaders,
@@ -87,5 +87,5 @@ test('Running fixtures with express', function (t) {
       )
     })
   }
-  t.end()
+  suite.end()
 })
```

I'm not married to "suite", but it is the name I've tended to use for a wrapper with subtests.  Because there are no subsequent top-level `test(...)` usages it probably doesn't make a difference here, but in general it can make a diff.

test/sanitize-field-names/fixtures.js
```diff
@@ -0,0 +1,309 @@
...
+/**
+ * Description of the test fixture's structure
+ *
+ * const fixture = {
+ *   name: '', // the name of the test,
+ *   agentConfig: '', // additional default configuration values for agent
```
trentm, line 7: nit:

```suggestion
 *   agentConfig: {}, // additional default configuration values for agent
```

test/sanitize-field-names/fixtures.js
```diff
@@ -0,0 +1,309 @@
...
+        ],
+        defined: { keepmeForm: 'formFields' }
+      }
+    }
+  },
+
+  {
+    name: 'tests configured wildcard handling, with urlencode bodyparsing',
+    agentConfig: {
+      sanitizeFieldNames: ['thi*isa']
```
trentm, line 274: Could perhaps include a `sanitizeFieldNames` entry that disables case-insensitivity.

test/sanitize-field-names/express.js
```diff
@@ -0,0 +1,91 @@
...
+    request.post(
+      url,
+      {
+        form: formFields,
+        headers: requestHeaders
+      },
+      function (error, response, body) {
+        if (error) {
+          t.fail(error)
+        }
```
trentm, line 52: Any reason to not use [`t.error(error)`](https://github.com/substack/tape#terrorerr-msg) here?

test/sanitize-field-names/main.js
```diff
@@ -0,0 +1,128 @@
...
+test('removeKeysFromObject unit tests', function (t) {
+  t.ok(removeKeysFromObject, 'can import function')
+
+  const obj1 = {
+    one: 'two',
+    three: 'four',
+    five: 'six'
+  }
+  removeKeysFromObject(obj1, [/th.*ee/])
+  // t.ok(!obj1.three, 'key three removed from object')
```
trentm, line 19: Debug line to remove?

test/sanitize-field-names/main.js
```diff
@@ -0,0 +1,128 @@
...
+  }
+  removeKeysFromObject(obj2, [/th.*ee/, /three/, /.*five/])
+  t.equals(obj2.three, '[REDACTED]', 'key three redacted')
+  t.equals(obj2.one, 'two', 'key one remains in ibject')
+  t.equals(obj2.five, '[REDACTED]', 'key five redacted')
+
+  t.end()
+})
+
+test('removeKeysFromObject tests', function (t) {
```
trentm, line 37: Is this whole test a dupe of the one above?

lib/config.js
```diff
@@ -195,7 +201,8 @@ var MINUS_ONE_EQUAL_INFINITY = [
 
 var ARRAY_OPTS = [
   'disableInstrumentations',
-  'transactionIgnoreUrls'
+  'transactionIgnoreUrls',
+  'sanitizeFieldNames'
```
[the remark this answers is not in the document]
astorm: 👍

lib/filters/http-headers.js
```diff
@@ -18,12 +18,22 @@ function httpHeaders (obj) {
   return obj
 }
 
+/**
+ * Filters authorization header and cookie _values_
+ *
+ * The filterSensitiveHeaders method filters individual
+ * cookie values
+ */
 function filterSensitiveHeaders (headers) {
```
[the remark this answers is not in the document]
astorm: 👍 to both.

lib/filters/http-headers.js
```diff
@@ -18,12 +18,22 @@ function httpHeaders (obj) {
...
   for (const key in headers) {
+    // undefined headers will be dropped when serialized as
+    // json, and don't need to be redacted.  If a cookie
+    // header is already redacted there's no need to parse its
+    // values.
+    if (headers[key] === undefined || REDACTED === headers[key]) {
```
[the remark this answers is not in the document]
astorm: I was thinking about a `lib/constants.js` when I was writing that so, overkill or not I added it -- if only because once it's there it'll make future decisions about "should this be a constant" easier if we have a place for them.  For now I'd like to keep the two module separate with an eye towards unifying them when we do the 4.0 work.

lib/parsers.js
astorm: Agreed -- but I'll tackle this in a separate PR: https://github.com/elastic/apm-agent-nodejs/issues/1925

test/sanitize-field-names/express.js
astorm: 👍 changed it.

test/sanitize-field-names/fastify.js
```diff
@@ -0,0 +1,91 @@
...
+  t.on('end', done)
+}
+
+function createMiddleware (type) {
+  // fastify only has the one body parsing middleware
+  // there's no text or raw/Buffer to worry about
+  return fastifyFormbody
+}
+
+test('Running fixtures with express', function (t) {
```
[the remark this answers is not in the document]
astorm: 👍 corrected, the copy/paste mØØse have been sacked.

test/sanitize-field-names/fixtures.js
astorm: 👍 changed

test/sanitize-field-names/main.js
astorm: 👍

test/sanitize-field-names/main.js
astorm: Other than the _ibject_ type it is -- very strange.  I'm unsure how that happened. Removing. 👍

test/sanitize-field-names/restify.js
```diff
@@ -0,0 +1,92 @@
...
+  assertResponseHeadersWithFixture,
+  assertFormsWithFixture
+} = require('./shared')
+const agent = require('../..').start(createAgentConfig())
+const test = require('tape')
+const request = require('request')
+const restify = require('restify')
+const fixtures = require('./fixtures')
+
+test('Running fixtures with koa', function (t1) {
```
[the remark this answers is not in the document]
astorm: 👍

test/sanitize-field-names/express.js
astorm: I like it -- `suite`'s a good name for it.  I've updated these tests as well as the similar invocation in the other files.

lib/filters/sanitize-field-names.js
```diff
@@ -0,0 +1,86 @@
...
+function removeKeyFromObject (obj, regex) {
+  for (const [key] of Object.entries(obj)) {
+    if (regex.test(key)) {
+      obj[key] = REDACTED
+    }
+  }
+  return obj
+}
+
+function removeKeysFromObject (obj, regexes) {
```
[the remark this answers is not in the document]
astorm: Good call -- I did `redactKeysFromPostedFormVariables` instead of `removeKeysFromPostedFormVariables`

lib/redact-secrets/is-secret.js
```diff
@@ -39,12 +42,14 @@ var KEYS = [
...
-
 exports.key = key
 exports.value = value
 
 function key (str) {
-  return KEYS.some(function (regex) {
+  // grab current value of agent._conf.sanitizeFieldNames
+  // and include those as keys
+  const keys = KEYS.concat(agent._conf.sanitizeFieldNamesRegExp)
+  return keys.some(function (regex) {
```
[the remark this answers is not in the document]
astorm: neat!  The syntax is a little denser than I'd normally like -- but I also think it's worth it as this is a medium warm path and avoiding the allocation makes sense.

test/sanitize-field-names/fixtures.js
```diff
@@ -0,0 +1,309 @@
...
+          'password', 'passwd', 'pwd', 'secret', 'Somethingkey', 'FOOtokenBAR',
+          'ZIPsessionZAP', 'FOOsessionBAR', 'FULLOFcreditBEES', 'SINGcardSONG',
+          'authorization', 'set-cookie'
+        ],
+        defined: { keepmeForm: 'formFields' }
+      }
+    }
+  },
+  {
+    name: 'tests deault wildcard handling, with text bodyparsing',
```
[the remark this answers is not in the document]
astorm: Got the last of these 

```
% ack 'deault'
% 

```

test/sanitize-field-names/fixtures.js
astorm: fair -- it's a bit of a retest of the existing Wildcard functionality that we're reusing BUT with the harness in place adding a new test is trivial.  Done.

docs/configuration.asciidoc
astorm: 👍

astorm:
PR feedback's addressed, look for the big old DONE stamp unless there's more work to be done here -- putting out for re-review

test/sanitize-field-names/express.js
```diff
@@ -0,0 +1,92 @@
...
+    request.post(
+      url,
+      {
+        form: formFields,
+        headers: requestHeaders
+      },
+      function (error, response, body) {
+        if (error) {
+          t.error(error)
+        }
```
trentm, line 53: nit:
```suggestion
        t.error(error)
```

With `t.error` you can drop the `if (error) {` guard.

trentm (commented):
Just the one nit, and the follow up from the Buffer change we discussed and this all LGTM.
Oh, and the test failure to figure out.

astorm:
Adjustments made (removed handling of and related tests) for new handling of raw body parsering

lib/parsers.js
```diff
@@ -160,6 +168,12 @@ function getContextFromRequest (req, conf, type) {
     } else if (Buffer.isBuffer(body)) {
       context.body = '<Buffer>'
     } else {
+      body = redactKeysFromPostedFormVariables(body, req.headers, conf.sanitizeFieldNamesRegExp)
+
+      if (Buffer.isBuffer(body)) {
+        body = body.toString()
+      }
```
trentm, line 175: This block should be removed.

astorm: agreed and removed.

trentm (approved):

PR #1899, opened 2020-11-30
chore: have a visible 'Unreleased' section in changelog

trentm:
We're in the process of moving dev process to have PRs include an update to CHANGELOG.asciidoc so that we are authoring the changelog as part of making changes.  We started by adding unreleased changes to the *block comment* at the top of the changelog file. One pain point with that is that you cannot see those in the GitHub rendering.

I asked Brandon and he says that having an "Unreleased" section will not cause a problem for doc rendering for releases. In retrospect, I understand why because we have the release branches for those.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -19,23 +19,31 @@ endif::[]
 [float]
 ===== Features
 
-* feat: Set "destination" context on spans for "mongodb". {pull}1893[#1893] +
-  This allows Kibana APM Service Maps to show a "mongodb" node for services using
-  the https://www.npmjs.com/package/mongodb[mongodb] package (which includes
-  mongoose and mongojs).
-
-* feat: transactionIgnoreUrl wildcard matching {pull}1870[#1870] +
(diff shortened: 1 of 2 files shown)


bmorelli25 (dismissed):
Looks great!

trentm:
From Brandon in DMs:

> And feel free to remove the block comment entirely if it just gets in the way. Some of the teams like to have it as a reminder of the correct header levels, but it’s your call

astorm (approved):
👍

CHANGELOG.asciidoc
```diff
@@ -4,38 +4,57 @@ https://www.elastic.co/guide/en/apm/agent/nodejs/current/release-notes.html[elas
...
+        [float]
+        ===== Features
+
+        [float]
+        ===== Bug fixes
+
+2. When making a release, change the "==== Unreleased" section header to:
+
+        [[release-notes-x.x.x]]
+        ==== x.x.x - YYYY/MM/DD
```
bmorelli25, line 25: Love this

PR #1902, opened 2020-11-30
chore: use merge=union for changelog

trentm:
One (possibly minor) concern with having PRs include updates to CHANGELOG.asciidoc is more common merge conflicts. Time will tell if that proves to be a PITA and barrier.

Gil suggested using:

```
% cat .gitattributes
CHANGELOG.asciidoc merge=union
```

from the docs:
  Run 3-way file level merge for text files, but take lines from both
  versions, instead of leaving conflict markers. This tends to leave the
  added lines in the resulting file in random order and the user should
  verify the result. Do not use this if you do not understand the
  implications.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+CHANGELOG.asciidoc merge=union


astorm (approved):
Huh -- I didn't realize you could do that sort of thing an a per file basis.  TIL today.  I'm game for giving this a try. Talking out loud I suppose there's a chance of some awkward merging  happening if two people are editing an older section of the file -- but that seems rare and as long as we're reading the actual changelog entries before merging I don't think it's a real problem.  Approving.

astorm:
@trentm integration tests failed here.  I don't think it's this change, but I'm going to re-kickoff jenkins tests just to be sure.

astorm:
jenkins run the tests please

astorm:
jenkins run the tests

PR #1903, opened 2020-12-01
fix(config): improve parsing of APM Server port

watson:
The `serverPort` property is currently only used to validate the
configuration to ensure that it's within an acceptable range of 1-65535.

However, in case the APM Server port was either 443 or 80, its value
would be an empty string when parsed. This in turn would result in
`serverPort` having the value `NaN`.

This would technically still work as the old range check would still be
valid. However, it could make debugging harder, so it's better to fix
it.


astorm:
jenkins run the tests

trentm:
@watson Would you please be able to rebase on upstream master? I think this is based off of a change from 7 weeks ago. At least one test failure (regarding ".schemacache") is already fixed.

lib/agent.js
```diff
@@ -110,12 +110,14 @@ Agent.prototype.startSpan = function (name, type, subtype, action, { childOf } =
...
 
-  const { host, port } = this._conf.serverUrl
+  const { host, port, protocol } = this._conf.serverUrl
     ? parsers.parseUrl(this._conf.serverUrl)
     : { host: 'localhost:8200', port: '8200' }
 
   this._conf.serverHost = host
-  this._conf.serverPort = parseInt(port, 10)
+  this._conf.serverPort = port === ''
+    ? (protocol === 'https:' ? 443 : 80)
```
trentm, line 119: I don't suspect we need to worry about a protocol other than "https" or "http" here. :)

trentm (approved):
LGTM

trentm:
Thanks!

PR #1907, opened 2020-12-03
fix: fixes intent of filter test

astorm:
This PR fixes the intent of the test in `test/filters.js`.

The intent of this test is to make sure that both the request response objects are filtered correctly. 

The problem with the old test was that `context.request.headers` and `context.response.headers` referenced _the same_ object.  

This means when `context.request.headers` was filtered that `context.response.headers` also appeared filtered regardless of whether code explicitly filtered it or not. 

This PR ensures that `context.request.headers` and `context.response.headers` are two separate object instances. 

This is a follow on PR to https://github.com/elastic/apm-agent-nodejs/pull/1886

### DIFF
--- a/test/filters.js
+++ b/test/filters.js
@@ -5,14 +5,14 @@ const test = require('tape')
 const Filters = require('object-filter-sequence')
 const filterHttpHeaders = require('../lib/filters/http-headers')
 
-function makeTransactionWithHeaders (headers) {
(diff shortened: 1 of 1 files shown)


astorm:
@StoraH took me a bit longer to get to this fix -- our `sanitize_filter_name` work ended up going in a slightly different direction.  

Does this seem like it meets the original intent of your changes to that test?

trentm (approved):

PR #1908, opened 2020-12-03
feat: add log_level central config support

trentm:


trentm:

trentm:

lib/agent.js
```diff
@@ -44,12 +46,6 @@ function Agent () {
   this.lambda = lambda(this)
 }
 
-Object.defineProperty(Agent.prototype, 'logger', {
```
trentm, line 47: I've updated the only two places `agent.logger` needs to be set. This avoids runtime code needing to go through an getter function for every `agent.logger...` call. Theoretically that should be an ever so slight perf win. :)  
(This originated from an earlier stab at this that removed even having `.logger` on the "Config" object -- since backed out.)

lib/agent.js
```diff
@@ -170,7 +167,7 @@ Agent.prototype.start = function (opts) {
       startTrace: stackObj.stack.split(/\n */).slice(1),
       main: pkg && pkg.main,
       dependencies: pkg && pkg.dependencies,
-      conf: this._conf
+      conf: this._conf.toJSON()
```
trentm, line 170: This and the `toJSON()` implementation are to avoid a large log.trace dump to the console. I'm happy to pull out this change to a separate PR if you prefer.  This gist shows the the before/after of that log dump: https://gist.github.com/trentm/59864b57fb91388e061e5a64c9e8e80e

lib/config.js
```diff
@@ -303,13 +302,21 @@ class Config {
           }
 
           if (Object.keys(conf).length > 0) {
-            normalize(conf, agent._conf)
+            normalize(conf)
```
trentm, line 305: This change is unrelated. The second arg to `normalize()` has never been used.

lib/tracecontext/tracestate.js
```diff
@@ -16,7 +23,7 @@
  *   const newHeader = tracestate.toW3cString()
  */
 class TraceState {
-  constructor (sourceBuffer, listMemberNamespace = 'es', defaultValues = {}, logger) {
+  constructor (sourceBuffer, listMemberNamespace = 'es', defaultValues = {}) {
```
trentm, line 26: Alan, that `logger` arg is never used, AFAICT. Let me know if you think it should still be a feature.

lib/tracecontext/tracestate.js
```diff
@@ -40,17 +47,13 @@ class TraceState {
     for (const [key, value] of Object.entries(defaultValues)) {
       this.setValue(key, value)
     }
-
-    // if no logger, then fallback to agent singleton
-    logger = logger || require('../..').logger
-    this.logger = logger
```
trentm, line 46: In the *extreme* case, it is possible for the `agent.logger` to be replaced with a new logger object from a central config update *while* a TraceState instance is holding a ref to it.

trentm:

lib/tracecontext/tracestate.js
astorm: @trentm No objection to this change -- but a question to make sure I understand what we're fixing here.  

Is the problem with the previous code that 

1. A TraceState object is instantiated
2. We receive a message from APM Server that there's a new log level, so a new logger is instantiated
3. However, the TraceState object will still hold a reference to the old logger, meaning it will continue to log at the "wrong"/"old" level for the remainder of the trace's life?

i.e. is the problem (in the extreme case) a trace state header will log at the wrong level?  Or are there other problems with the header holding on to the old object reference?

test/central-config-enabled.js
```diff
@@ -117,3 +119,46 @@ test('remote config enabled: receives non delimited string', function (t) {
...
+    const SHORT_TIME_LATER = 10 // Hopefully this isn't a race we lose.
+    setTimeout(function () {
+      // 4. agent.logger should be updated from central config.
+      t.equal(agent.logger.level, 'error',
+        'shortly after fetching central config, agent.logger level should be updated')
+
+      agent.destroy()
+      server.close()
+      t.end()
+    }, SHORT_TIME_LATER)
```
astorm, line 146: Would an event listener (`agent._transport.on('config',`?) be less prone to a race condition than this `setTimeout` function?

astorm (approved):
Approving, as it does the thing and looks good as is. There's an english language nit and an untested testing suggestion I'll leave to your discretion.

lib/tracecontext/tracestate.js
trentm: Yes, that (your 1. 2. 3.) is the (theoretical) problem.  With the current logger lib and only logging to stdout/stderr, yes the only issue is potentially logging at the wrong level.  Going *more* extreme, if we were to support a new logging framework (e.g. the potential move to `pino`) **and** we were to support logging to a file or some stream, the TraceState instance holding on to that logger could delay resource cleanup. I'm stretching now. :)

test/central-config-enabled.js
trentm: I'll give it a try, thanks.

lib/agent.js
```diff
@@ -109,6 +105,7 @@ Agent.prototype.startSpan = function (name, type, subtype, action, { childOf } =
 
 Agent.prototype._config = function (opts) {
   this._conf = config(opts)
+  this.logger = this._conf.logger
```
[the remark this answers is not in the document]
trentm: I'll add a `this.logger = null`. It might not make a diff, but no reason not to follow the best practice.

docs/configuration.asciidoc
```diff
@@ -414,9 +414,9 @@ require('elastic-apm-node').start({
 Set the verbosity level for the agent.
 Note that this does not have any influence on the types of errors that are sent to the APM Server.
 This only controls how chatty the agent is in your logs.
-This only applies when not using a custom logger.
+This only applies when not using a custom `logger`.
```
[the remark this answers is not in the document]
trentm: Thanks, I'll update with your phrasing.

test/central-config-enabled.js
trentm: Much nicer. Thanks! With your suggestion, I'm unable to get that sequencing to fail with a number of dozen quick tries:

```
while true; do if ! node test/central-config-enabled.js; then break; fi; done
```

whereas with the possibly-race setTimeout and setting the timeout to 0ms or 1ms I was able to occasionally  get a failure.

astorm (approved):

PR #1909, opened 2020-12-06
Supports boolean and number labels. Fixes #1598

### LINKED ISSUE #1598: Add support for numeric and boolean labels

Since APM Server 6.7, labels can have numerical or boolean values. We should add support for these in the agent's API, which currently supports only string values (non-string values are stringified).

[`setLabel`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/transaction-api.html#transaction-set-label) and [`addLabels`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/transaction-api.html#transaction-add-labels) should accept numeric and boolean values, and should not stringify them.

Just changing the behaviour would be a breaking change (particularly because they would end up as different Elasticsearch field types, causing indexing to fail), so that change would have to be done with a major version bump.

If waiting for a major version bump is too much, then we could add parameters to the functions that instruct them to not stringify values. Those parameters would be removed in a future major version, and the behaviour would be switched so stringification is not applied by default.

**Additional context**
https://www.elastic.co/guide/en/apm/get-started/current/metadata.html#labels-fields

ddolcimascolo:
Support boolean and numeric labels.

Fixes #1598 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,14 @@ Notes:
 [[release-notes-3.x]]
 === Node.js Agent version 3.x
 
+[float]
+===== Features
+
+* feat: support numeric and boolean labels {pull}1909[#1909] +
+  Add an optional `stringify` option to `apm.setLabel(name, version, stringify = true)`
+  and `apm.addLabels(labels, stringify = true)`, which can be set `false` to
+  allow numeric and boolean labels. Stringify defaults to true for backwards
+  compatibility -- stringification will be removed in a future major version.
 
 [[release-notes-3.10.0]]
 ==== 3.10.0 - 2021/01/11
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -199,18 +199,19 @@ TIP: Before using custom context, ensure you understand the different types of
 {apm-overview-ref-v}/metadata.html[metadata] that are available.
 
 [[apm-set-label]]
-==== `apm.setLabel(name, value)`
+==== `apm.setLabel(name, value[, stringify = true])`
 
-[small]#Added in: v0.1.0#
-
-[small]#Renamed from `apm.setTag()` to `apm.setLabel()`: v2.10.0#
+[small]#Added in: v0.1.0# +
+[small]#Renamed from `apm.setTag()` to `apm.setLabel()`: v2.10.0# +
+[small]#Added `stringify` argument in: v3.10.0#
 
 * `name` +{type-string}+
 Any periods (`.`), asterisks (`*`), or double quotation marks (`"`) will be replaced by underscores (`_`),
 as those characters have special meaning in Elasticsearch
-* `value` +{type-string}+
-If a non-string data type is given,
-it's converted to a string before being sent to the APM Server
+* `value` +{type-string}+ | +{type-number}+ | +{type-boolean}+
+* `stringify` +{type-boolean}+
+Defaults to `true`. When true, if a non-string `value` is given, it is
+converted to a string before being sent to the APM Server.
 
 Set a label on the current transaction.
 You can set multiple labels on the same transaction.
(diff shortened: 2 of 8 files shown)


trentm:
Thanks! I'll take a look soon.

trentm:
jenkins run the tests

docs/configuration.asciidoc
```diff
@@ -834,6 +834,16 @@ Example:
...
+[[extended-label-types]]
+==== `extendedLabelTypes`
+
+* *Type:* Boolean
+* *Default:* `false`
+* *Env:* `ELASTIC_APM_EXTENDED_LABEL_TYPES`
+
+Whether numeric and boolean labels should be sent as-is to the APM server.
+When this option is `false`, all labels are sent as strings, no matter their original type.
+
```
trentm, line 846: Did you consider adding an optional argument to `setLabel` and `addLabels`, something like:

```diff
-GenericSpan.prototype.setLabel = function (key, value) {
+// TODO: in v4, remove the stringify option and never stringify
+GenericSpan.prototype.setLabel = function (key, value, stringify = true) {
```

Pros that I see:
- The user of `setLabel()` needs to explicitly work through each usage in their code to verify that changing 'stringify' won't break them.
- One isn't led to infer that `extendedLabelTypes=false` might affect the `globalLabels` config var (it won't, globalLabels already passes through bools and numbers untouched).

Cons:
- The user may need to slog through multiple usages. :)

ddolcimascolo: I didn't TBH. I very much like the way you can configure the agent with env vars, like most containing apps do anyway... I thought users could
1. Update the app environment, restart the app
2. Let the agent send a few transactions
3. Revert the environment, restart the app

Then they can safely test their label-specific ES queries on the new transactions and validate that the type changes work.

But I'm OK to move to an explicit third parameter and drop the config flag, you tell me 👍

trentm: > But I'm OK to move to an explicit third parameter and drop the config flag

Yes please, and thanks. My initial gut reaction is to prefer the extra param to `setLabel` and `addLabels`. I asked a colleague and he preferred the extra param as well (more strongly than I).

CHANGELOG.asciidoc
```diff
@@ -29,6 +29,20 @@ Notes:
...
+
+[float]
+===== Features
+
+* feat: support numeric and boolean labels {pull}1909[#1909] +
+  When `extendedLabelTypes` config option is set to `true`, boolean and numeric
+  labels are sent as-is to the APM server, wihtout stringification.
+
+[float]
+===== Bug fixes
```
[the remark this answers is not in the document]
ddolcimascolo: Done

trentm:
jenkins run the tests

trentm (changes requested):
Sorry again for the slow response (and ahead of time for after this because I'll be on break for the holidays for a bit).

1. I've made a couple suggestions below for a comment and the changelog.
2. I've updated the docs for this change. Can you please apply this patch to your branch: https://gist.github.com/trentm/1f1afdaa42661a05a6f9141864d2f718
3. Can you merge in changes from master?

ddolcimascolo:
Hi @trentm 

I've applied your suggestions, rebased on master and merged the doc patch. Sorry I missed this :astonished: 

Regards,
David

trentm:
jenkins run the tests

trentm:

trentm (approved):
Because I just *did* a 3.10 release the changelog and a couple refs to 3.10 are out of date. However, I'm just going to merge and fix those up in a follow up commit to avoid the "dropping the ball"-potential of another back and forth.

Thanks!

trentm:
There will be a minor release of the agent in the next two weeks that will include this.

ddolcimascolo:
Thx for merging!

Issue #1883, opened 2020-11-19
Document new 7.11 central config vars as Central/Dynamic Configuration 


astorm:
When 7.11 of the Elastic Stack is released, update the agent docs for the following config vars to mark them as "Dynamic" (i.e. available in central config). Also possibly label the configuration as dynamic in Kibana X.Y.  

- `transaction_ignore_urls` https://github.com/elastic/apm-agent-nodejs/pull/1876
- `log_level` https://github.com/elastic/apm-agent-nodejs/pull/1908
- `sanitize_field_names` https://github.com/elastic/apm-agent-nodejs/issues/1811

trentm:
This landed in Kibana in https://github.com/elastic/kibana/pull/84678
@astorm @bmorelli25 Does that mean we add the badge to the docs now? Or when 7.11 is in freeze or released?

astorm:
@trentm I'll ultimately defer to @bmorelli25, but my thinking on this is we wait for for 7.11 to be officially released.  My thinking around this is based on the following use case

1. User comes to docs site, sees that this is a dynamic/central config
2. User goes to the most recently _released_ version of Kibana (7.10) to look for config
3. User doesn't see this in central config.  User is sad.

trentm:
I defer to Brandon as well. :)

My preference in reading tech docs is to have "New in version X.Y"-like prose next to things for when they are added. That allows:
1. A user of an older version of something to read *current* docs to see what is coming and to be able to tell if something they are reading doesn't apply. E.g. these let me know about things in Python 3.8 and 3.9 even if I'm not quite using that version yet: https://docs.python.org/3/library/zoneinfo.html#module-zoneinfo https://docs.python.org/3/library/sys.html#sys.__breakpointhook__

2. It also lets us the maintainers add the docs in the same commit as the functionality, which is nice.

That said, it isn't a biggie either way.

bmorelli25:
Generally in `current`, we try to avoid adding references to stack features before that version of the stack releases. But for a lot of different reasons, that's not always possible. In the past, for small changes like this, I've kept a release checklist of doc updates that need to be made before a release. I'm happy to add this one to my list for 7.11.

> My preference in reading tech docs is to have "New in version X.Y"-like prose next to things for when they are added

I agree that the "New in version X.Y" or "Added in: vx.x.x", like we currently have in the Node.js APM API reference, is useful -- especially because we only publish a new documentation branch per major.

[...]

PR #1911, opened 2020-12-09
docs: document the central config var name for appropriate vars

trentm:
This allows one to search for central config var names in these docs
when starting from central config. It also allows searching the
configuration doc page for "central" to find all centrally controllable
vars -- helpful given that the "Dynamic" badge isn't searchable.

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -289,6 +289,7 @@ require('elastic-apm-node').start({
 * *Type:* String
 * *Default:* `off`
 * *Env:* `ELASTIC_APM_CAPTURE_BODY`
+* *Central config name:* `capture_body`
 
 The HTTP body of incoming HTTP requests is not recorded and sent to the APM Server by default.
 
@@ -349,6 +350,7 @@ If the `errorOnAbortedRequests` property is `false`, this property is ignored.
 * *Type:* Number
 * *Default:* `1.0`
 * *Env:* `ELASTIC_APM_TRANSACTION_SAMPLE_RATE`
+* *Central config name:* `transaction_sample_rate`
 
 Specify the sampling rate to use when deciding whether to trace a request.
 
@@ -470,7 +472,7 @@ Normally only `Error` objects have a stack trace associated with them.
 This stack trace is stored along with the error message when the error is sent to the APM Server.
 The stack trace points to the place where the `Error` object was instantiated.
 
-But sometimes its valuable to know,
+But sometimes it's valuable to know,
 not where the `Error` was instantiated,
 but where it was detected.
 For instance,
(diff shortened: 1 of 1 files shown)


trentm:
@bmorelli25 @astorm Thoughts on this?

Screen Shot 2020-12-09 at 9 02 37 AM


Note: I haven't done this for the central config vars coming in 7.11. We can do that as part of https://github.com/elastic/apm-agent-nodejs/issues/1883 separately.

bmorelli25:
Neat idea. I like it.

bmorelli25:
Could even move the dynamic badge into the beginning of the central config bullet point to group it all together.

bmorelli25 (dismissed):

trentm:
> Could even move the dynamic badge into the beginning of the central config bullet point to group it all together.

Ooh, that sounds good. It looks (a) like this:

Screen Shot 2020-12-09 at 9 12 03 AM

or (b) like this:

Screen Shot 2020-12-09 at 9 13 53 AM


@bmorelli25 Do you have a preference for (a) vs (b), or another suggestion?

bmorelli25:
It looks like there's a (slightly annoying) vertical-align mismatch in both options, but I do like it better with the badge in the bullet. No preference.

trentm:
Or there is this (badge up with the config var header name):

Screen Shot 2020-12-09 at 10 02 13 AM

Pro: Looks nicer.
Con: Breaks the association between "Dynamic" and "Central".

trentm:
> It looks like there's a (slightly annoying) vertical-align mismatch in both options

I think they are actually vertically aligned on the baseline:

Screen Shot 2020-12-09 at 10 07 59 AM

It is just a font size thing.

bmorelli25:
Another downside of option c is that it makes it harder to copy the config value from the header. If you don't think that's a problem, then I agree, it does look nicer.

trentm:
>  it makes it harder to copy the config value from the header.

Ew, yah. That disqualifies (c) for now.

(Aside: FWIW, I'd like to be about the double-click on the headers to select the config values. Opinion alert: Having the headers work like GitHub-rendered Markdown headers would be nice. I.e. the header itself is not a link, only the on-`:hover` link icon is a clickable link.)


I'm choosing (a) for now.

bmorelli25:
> (Aside: FWIW, I'd like to be about the double-click on the headers to select the config values. Opinion alert: Having the headers work like GitHub-rendered Markdown headers would be nice. I.e. the header itself is not a link, only the on-:hover link icon is a clickable link.)

> I'm choosing (a) for now.

🎉 
Totally agree. We've tried to fix this... https://github.com/elastic/docs/pull/1967

bmorelli25 (approved):
✅

PR #1912, opened 2020-12-10
feat: add `captureAttributes` option to `apm.captureError()`

trentm:
This adds a `captureAttributes` boolean option to `apm.captureError()` to
allow *disabling* the automatic capture of Error object properties. This
is useful for cases where those properties should not be sent to the APM
Server, e.g. for performance (large string fields) or security (PII data).

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,6 +38,10 @@ Notes:
 [float]
 ===== Features
 
+* feat: Add `captureAttributes` boolean option to `apm.captureError()` to
+  allow *disabling* the automatic capture of Error object properties. This
+  is useful for cases where those properties should not be sent to the APM
+  Server, e.g. for performance (large string fields) or security (PII data).
 * feat: Add `log_level` central config support. {pull}1908[#1908] +
   Spec: https://github.com/elastic/apm/blob/master/specs/agents/logging.md
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -264,6 +264,7 @@ Agent.prototype.captureError = function (err, opts, cb) {
   var res = opts && opts.response instanceof ServerResponse
     ? opts.response
     : trans && trans.res
+  var captureAttributes = opts && opts.captureAttributes
   var _isError = isError(err)
 
   if ((!opts || opts.handled !== false) &&
@@ -277,7 +278,7 @@ Agent.prototype.captureError = function (err, opts, cb) {
   if (!_isError) {
     prepareError(parsers.parseMessage(err))
   } else {
-    parsers.parseError(err, agent, function (_, error) {
+    parsers.parseError(err, captureAttributes, agent, function (_, error) {
       // As of now, parseError suppresses errors internally, but even if they
       // were passed on, we would want to suppress them here anyway
       prepareError(error)
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -59,7 +59,19 @@ exports.parseStackTrace = function (err) {
   })
 }
 
-exports.parseError = function (err, agent, cb) {
+// `parseError` starts the serialization of the given Error instance into the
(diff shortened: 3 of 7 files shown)


trentm:
The initial motivator for this change was a hack on the in-progress #1877 work to avoid having `apm.captureError(err)` automatically include `err.data` on some errors from the elasticsearch JS client... because that data could be large and might include user data that shouldn't be sent to the APM server.

Specifically this code and comment from @vigneshshanmugam https://github.com/elastic/apm-agent-nodejs/pull/1877#discussion_r532604652

lib/parsers.js
```diff
@@ -83,9 +95,11 @@ exports.parseError = function (err, agent, cb) {
...
     }
 
-    var props = stackman.properties(err)
-    if (props.code) delete props.code // we already have it directly on the exception
-    if (Object.keys(props).length > 0) error.exception.attributes = props
+    if (captureAttributes) {
+      var props = stackman.properties(err)
+      if (props.code) delete props.code // we already have it directly on the exception
+      if (Object.keys(props).length > 0) error.exception.attributes = props
+    }
```
astorm, line 102: my personal value of "don't have single line brace-less conditionals" is battling it out with my personal value of "leave code as you found it" with [no clear winner](https://www.youtube.com/watch?v=t-dK9996Ks4&has_verified=1&t=0m37s)

lib/agent.js
```diff
@@ -277,7 +278,7 @@ Agent.prototype.captureError = function (err, opts, cb) {
   if (!_isError) {
     prepareError(parsers.parseMessage(err))
   } else {
-    parsers.parseError(err, agent, function (_, error) {
+    parsers.parseError(err, captureAttributes, agent, function (_, error) {
       // As of now, parseError suppresses errors internally, but even if they
```
astorm, line 282: I see here we're changing the function signature of the `parseError` method.  I also see elsewhere that we've updated every invocation of this method _in this package_ to use this new method signature.  We don't expose `lib/parsers` as a public api, so this method change should be ok without requiring a major version semver update.

CHANGELOG.asciidoc
```diff
@@ -38,6 +38,10 @@ Notes:
 [float]
 ===== Features
 
+* feat: Add `captureAttributes` boolean option to `apm.captureError()` to
+  allow *disabling* the automatic capture of Error object properties. This
+  is useful for cases where those properties should not be sent to the APM
+  Server, e.g. for performance (large string fields) or security (PII data).
 * feat: Add `log_level` central config support. {pull}1908[#1908] +
```
astorm, line 45: Do we think it's worth mentioning the signature change to `parseError` in the changelog?  Or better to leave it out since this is, technically speaking, a private API that folks shouldn't be calling directly, or should be ready to accept the risks of relying on a private(-ish?) API?  (I'm split down the middle on this personally -- no compelling clear case for either approach)

lib/parsers.js
astorm, line 102: Just saying outloud that `error.exception.attributes` won't be at all if `captureAttributes` is false.  Is is probably OK since this would already be the case if `Object.keys(props)` was zero-length.

astorm (approved):
Seems good as is -- small PR that does the thing. 

Review comments are the usual talking out loud and hypothetical questions.  Approving.

lib/parsers.js
trentm: Heh, I went through the same debate, albeit without a star trek reference.

CHANGELOG.asciidoc
trentm: Unless I missed something, it is a completely private API (the whole lib/parsers.js is private) -- as private as you can get with JS without going to lengths -- so I'm not inclined to update the changelog.

trentm:
Thanks for the quick review!

Issue #1925, opened 2020-12-19
Buffer/Binary Capturing Middlewares Results in Large Blob-y not useful Data Being Reported


astorm:
If a user is using a body parsers module that captures a raw Buffer (such as [the Express raw body parser](http://expressjs.com/en/resources/middleware/body-parser.html#bodyparserrawoptions)) this can result in large blobs of binary data being reported via the agent.  This seems like a bug.

To fix this, we should

1. Refine the definition of `captureBody:true` to _not_ include binary data and
2. Instead, when `captureBody:true` and the agent encounters a raw buffer, it will capture the string `<Buffer>` to indicate there was binary data present. 

from @trentm 

```
if (haveBody) {
    if (!captureBody) {
        // Aside: I'm not sure why the agent calls this "REDACTED" here. If
        // the config is `captureBody: false`, then I would just leave
        // context.body = undefined.
        context.body = '[REDACTED]'
    } else if (Buffer.isBuffer(body)) {
        context.body = '<Buffer>'
    } else {
        ...
    }
}
```

trentm:
@astorm Correct me if I'm wrong that this is now resolved.

PR #1929, opened 2020-12-21
feat: improve/change processing of Buffer objects in post body

astorm:
PR for #1925.  If a middleware processes `request.body` as Buffer, rather than report the potentially large binary object we'll report the string `<Buffer>` to indicate binary data was detected. 

Also updates public docs to reflect this. 

- [x] Implement code
- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -314,6 +314,8 @@ Possible options are: `off`, `all`, `errors`, and `transactions`.
 
 The recorded body will be truncated if larger than 2 KiB.
 
+If the body parsing middleware captures the body as raw `Buffer` data, the request body will be represtend as the string `"<Buffer>"`.
+
 For the agent to be able to access the body,
 the body needs to be available as a property on the incoming HTTP https://nodejs.org/api/http.html#http_class_http_incomingmessage[`request`] object.
 The agent will look for the body on the following properties:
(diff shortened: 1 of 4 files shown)


astorm:
jenkins run the tests

docs/configuration.asciidoc
```diff
@@ -314,6 +314,8 @@ Possible options are: `off`, `all`, `errors`, and `transactions`.
 
 The recorded body will be truncated if larger than 2 KiB.
 
+If the body parsing middleware captures the body as raw `Buffer` data, the request body will be represtend as the string `"<Buffer>"`.
```
trentm, line 317: typo

```suggestion
If the body parsing middleware captures the body as raw `Buffer` data, the request body will be represented as the string `"<Buffer>"`.
```

trentm (approved):
It would be good to add a changelog entry for this.

docs/configuration.asciidoc
trentm: I think you resolved this without taking the fix. ;)

trentm (approved):

docs/configuration.asciidoc
astorm: Fixed it locally, but didn't push 🤦

trentm (approved):

Issue #1922, opened 2020-12-17
s/Travis CI/(GitHub Actions)?/g


trentm:
This is the master ticket for moving this repo and repos related to the Node.js APM Agent away from Travis CI, given the coming shutdown of travis-ci.org.

## tl;dr

- Some repos currently use Travis **in addition** to [Elastic's internal jenkins CI](https://apm-ci.elastic.co/job/apm-agent-nodejs/). Sometimes Travis ran only a subset of the full tests. These repos will drop Travis usage and *possibly* use GitHub Actions as a replacement.
- Some repos only use Travis currently and will likely be switched to use GitHub Actions.

## Why both Travis and Jenkins?

Why would we use both Travis and Jenkins for CI for some repos? To allow a subset of tests to be automatically run on externally contributed PRs -- something we can't enable for our internal Jenkins CI for security reasons. (Note that the typical review process *does* involve an Elastic employee manually triggering a complete test suite run on Jenkins via a comment on the PR.

## Repos to update

- [x] https://github.com/elastic/apm-agent-nodejs
- [x] https://github.com/elastic/apm-nodejs-http-client
- [x] https://github.com/elastic/apm-agent-nodejs-opentracing
- [x] https://github.com/elastic/node-traceparent
- [x] https://github.com/elastic/require-in-the-middle
- [x] https://github.com/elastic/object-filter-sequence
- [x] https://github.com/elastic/sql-summary
- [x] https://github.com/elastic/opbeans-node

trentm:
See https://github.com/elastic/observability-dev/issues/1278 for disabling some of these repos in travis-ci.org settings.

PR #1930, opened 2020-12-23
s/Travis CI/GitHub Actions

trentm:
Compared to the Travis CI config, this drops the test runs (of a subset of node versions) with ELASTIC_APM_ASYNC_HOOKS=false. Those are could be re-added later if we want (this case is
covered by the Jenkins tests).

Refs: https://github.com/elastic/apm-agent-nodejs/issues/1922


.github/workflows/test.yml
```diff
@@ -0,0 +1,120 @@
...
+        volumes:
+          - nodemysqldata:/var/lib/mysql
+        env:
+          MYSQL_ALLOW_EMPTY_PASSWORD: 1
+
+      mssql:
+        image: microsoft/mssql-server-linux
+        env:
+          ACCEPT_EULA: 'Y'
+          SA_PASSWORD: 'Very(!)Secure'
```
astorm, line 80: Presuming these are not real secrets, and instead are defaults for the SQL Server provided by microsoft/mssql-server-linux

astorm (approved):
👍 -- removing the `ELASTIC_APM_ASYNC_HOOKS=false` tests from the GHA tests is fine, since, (as you mentioned) Jenkins is covering these cases. 

I didn't spend much time reviewing/investigating the resource allocations (i.e. `HEAP_NEWSIZE: '400m' `) -- but if anything is off there we can easily revisit.  

approving.

.github/workflows/test.yml
trentm: That is my understanding yes. Copied this from test/docker-compose.yml.

PR #1936, opened 2021-01-11
3.10.0

trentm:
Another release will be coming up for the stack 7.11 release, but it is useful to get a release out with the `@elastic/elasticsearch` new instrumentation to help those working with Kibana+APM at least. As well, there are a number of other fixes and features here.


astorm (approved):
release :allthethings:

PR #1939, opened 2021-01-12
a small 'elRequest()' fn to support HTTP reqs with a connectTimeout

trentm:
A small lib with a single function to make HTTP requests with a connectTimeout. This from discussion with Alan on his cloud metadata work.

### Checklist

- [x] Implement code
- [ ] Add tests

### DIFF
--- a/lib/elrequest.js
+++ b/lib/elrequest.js
@@ -1,0 +1,187 @@
+'use strict'
+
+// A tiny "El"astic HTTP "request" library.
+
+const { URL } = require('url')
+
+// A wrapper around `{http|https}.request()` that adds support for a connection
+// timeout separate from the existing `options.timeout`.
+//
+// The existing `options.timeout` to `http.request()` sets `socket.setTimeout()`
+// which will emit the 'timeout' event if there is an period of socket idleness
+// that is this long. In practice for short-lived requests, it is a timeout on
+// getting the start of response data back from the server.
+//
+// The new `opts.connectTimeout` is a number of milliseconds count from socket
+// creation to socket 'connect'. If this time is reached a 'connectTimeout'
+// event will be emitted on the request object. As with 'timeout', it is up
+// to the caller to handle destroying the request. See "Usage" below.
+// In pratice this allows for a shorter timeout to see if the remote server
+// is handling connections in a timely manner. To be useful, a `connectTimeout`
+// is typically shorter than a given `timeout`.
+//
+// Note on call signature:
+//
+// Currenly this only supports the one signature:
+//      elRequest(url: String, opts?: Object, cb?: Function)
+// rather than the three signatures (and 25 lines of argument normalization)
+// that core `http.request()` supports. If we need the other signatures we
+// can add them later.
+//
+// Usage:
+//     const { elRequest } = require('./elrequest')
+//
+//     var req = elRequest(url, {
+//         connectTimeout: connectTimeout,
+//         // Any {http|https}.request options...
+//         timeout: timeout
+//     }, function onRes(res) {
+//         // Handling of the response...
+//     })
+//
+//     req.on('timeout', function () {
+//         // ...
+//         req.destroy()
+//     });
+//
+//     req.on('connectTimeout', function () {
+//         // ...
+//         req.destroy()
+//     });
+//
+//     req.on('error', function (err) {
+//         // ...
+//     })
+//
+//     req.end()
+//
+function elRequest (url, opts, cb) {
+  const {
+    connectTimeout,
+    ...reqOpts
+  } = opts || {}
+
+  // http or https
+  const protocol = reqOpts.protocol || (new URL(url)).protocol
+  let proto
+  if (protocol === 'http:') {
+    proto = require('http')
+  } else if (protocol === 'https:') {
+    proto = require('https')
+  } else {
+    throw new Error(`unsupported protocol: "${protocol}"`)
+  }
+
+  const req = proto.request(url, reqOpts, cb)
+
+  if (connectTimeout) {
+    // Handle a connection timeout with a timer starting when the request
+    // socket is *created* ("socket" event) and ending when the socket
+    // is connected.
+    req.on('socket', function (socket) {
+      // log.trace({url: url}, 'start connectTimeout')
+      var connectTimer = setTimeout(function onConnectTimeout () {
+        // log.trace({url: url}, 'connectTimeout')
+        req.emit('connectTimeout')
+      }, connectTimeout)
+
+      socket.on('connect', function () {
+        // log.trace({url: url}, 'socket connected, clear connectTimeout')
+        clearTimeout(connectTimer)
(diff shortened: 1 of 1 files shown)


trentm:
- [ ] Perhaps add a `$method` to the mainline so can test PUT/POST. Low prio.

lib/elrequest.js
```diff
@@ -0,0 +1,187 @@
+'use strict'
+
+// A tiny "El"astic HTTP "request" library.
+
```
astorm, line 4: It made me smile -- but with so much elastic provided javascript floating around publicly  I'd worry the name might create an impression that this is somehow a generic elastic.co library for making HTTP requests (vs. a local-to-this-project library for making requests with a socket timeout).  As far as replacement names go I don't have any strong suggestions, but with it being a local project module a plain `request.js` or `request-wrapper.js` would be fine with me.

lib/elrequest.js
```diff
@@ -0,0 +1,187 @@
...
+      })
+    })
+  }
+
+  return req
+}
+
+module.exports = {
+  elRequest: elRequest
+}
```
astorm, line 107: nit: could be

```
{
    elRequest
}
```

lib/elrequest.js
```diff
@@ -0,0 +1,187 @@
...
+  if (connectTimeout) {
+    // Handle a connection timeout with a timer starting when the request
+    // socket is *created* ("socket" event) and ending when the socket
+    // is connected.
+    req.on('socket', function (socket) {
+      // log.trace({url: url}, 'start connectTimeout')
+      var connectTimer = setTimeout(function onConnectTimeout () {
+        // log.trace({url: url}, 'connectTimeout')
+        req.emit('connectTimeout')
+      }, connectTimeout)
```
astorm, line 86: Question: Is this `connectTimeout` event is an arbitrary name we chose for signaling a socket timeout?  Or is there some magic internal behavior here as well?

lib/elrequest.js
```diff
@@ -0,0 +1,187 @@
...
+
+  var req = elRequest(url, {
+    timeout: timeout,
+    connectTimeout: connectTimeout
+    // TODO: log support
+  }, function onRes (res) {
+    res.pipe(process.stdout)
+  })
+
+  req.on('timeout', function () {
```
astorm, line 165: Question: Is listening for the timeout event required in Node.js -- or will the request object can/destroy itself when the timeout is reached?

lib/elrequest.js
```diff
@@ -0,0 +1,187 @@
...
+    console.warn(`elrequest: response timeout (${timeout}ms): destroying request`)
+    req.destroy()
+    process.exitCode = 28 // using cURL's errno for a timeout
+  })
+
+  req.on('connectTimeout', function () {
```
astorm, line 171: Dovetailing with the previous question about `connectTimeout` -- I presume that listening for this event _is_ required in order to end the full request due to our "socket ping" failing?

astorm (commented):
Overall I like it -- I was able to drop it into my work and it simplified the code greatly.  It also made me realize I was testing a "socket connection refused" case and not a socket connection timeout :) I have some questions below, but with some test coverage this looks like a worth addition.

lib/elrequest.js
trentm: Sure, I'll think of some other name.

lib/elrequest.js
trentm: Nothing magic. It is the chosen name, so far, by this function for the event to indicate a connection timeout. `http.request` already has `options.timeout -> "timeout" event`, so I matched suit.

lib/elrequest.js
trentm: https://nodejs.org/api/all.html#http_event_timeout 

> Emitted when the underlying socket times out from inactivity. This only notifies that the socket has been idle. The request must be aborted manually.

So, if one passes in a `options.timeout` for the request, then one must manually `req.destroy()`. 

Also note: https://nodejs.org/api/all.html#http_request_abort   hence why I'm using `req.destroy()` rather than `req.abort()`.

lib/elrequest.js
trentm: Yes, if the user passes in a `opts.connectTimeout`, then they'll have to respond to that event to cancel/abort/destroy the request. (This roughly mirrors the behaviour of the "timeout" option and event.)

astorm (commented):
@trentm Another testing datapoint.  In Node 8.6 the library will throw a _"listener" argument must be a function_ error. 

```
    % nvm use 8
    Now using node v8.17.0 (npm v6.13.4)
    % node lib/elrequest.js http://elastic.co 1000 5000
    events.js:340
        throw new TypeError('"listener" argument must be a function');
        ^

    TypeError: "listener" argument must be a function
        at ClientRequest.once (events.js:340:11)
        at new ClientRequest (_http_client.js:174:10)
        at Object.request (http.js:39:10)
        at elRequest (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/lib/elrequest.js:75:21)
        at main (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/lib/elrequest.js:157:13)
        at Object.<anonymous> (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/lib/elrequest.js:186:3)
        at Module._compile (module.js:653:30)
        at Object.Module._extensions..js (module.js:664:10)
        at Module.load (module.js:566:32)
        at tryModuleLoad (module.js:506:12)
```

I did not see this in nodes 10, 12, or 14

lib/elrequest.js
```diff
@@ -0,0 +1,187 @@
...
+
+  if (connectTimeout) {
+    // Handle a connection timeout with a timer starting when the request
+    // socket is *created* ("socket" event) and ending when the socket
+    // is connected.
+    req.on('socket', function (socket) {
+      // log.trace({url: url}, 'start connectTimeout')
+      var connectTimer = setTimeout(function onConnectTimeout () {
+        // log.trace({url: url}, 'connectTimeout')
+        req.emit('connectTimeout')
```
astorm, line 85: Should we also send along an error object with the `connectTimeout`?
```
const error = new Error('some useful message about the socket timing out')
req.emit('connectTimeout', error) 
```

or is that better left to the client listening for the `connectTimeout` event?

lib/elrequest.js
trentm: -> `require('http-request').httpRequest`

lib/elrequest.js
trentm: Node core does not include any event data with the "timeout event: https://github.com/nodejs/node/blob/948f5152d7e90451439cb8040fd15237decbc229/lib/_http_client.js#L723
I am inclined to do the same.

trentm:
> In Node 8.6 the library will throw a _"listener" argument must be a function_ error.

Thanks. Should be fixed now.

astorm (approved):
👍 LGTM

PR #1949, opened 2021-01-21
feat: add apm.getServiceName()

trentm:
This will be used by ecs-logging-js libs to set "service.name" in log
records.

### Checklist

- [x] Implement code
- [x] Discuss APM standardization of this (possibly overkill) at https://github.com/elastic/apm/pull/404
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -60,6 +60,14 @@ Use `isStarted()` to check if the agent has already started.
 Returns `true` if the agent has started,
 otherwise returns `false`.
 
+[[apm-get-service-name]]
+==== `apm.getServiceName()`
+
+[small]#Added in: v3.11.0#
+
+Get the configured <<service-name,`serviceName`>>. If a service name was not
+explicitly configured, this value may have been automatically determined.
+
 [[apm-set-framework]]
 ==== `apm.setFramework(options)`
 
--- a/docs/agent-api.asciidoc
+++ b/docs/agent-api.asciidoc
@@ -68,6 +68,10 @@ otherwise returns `false`.
 Get the configured <<service-name,`serviceName`>>. If a service name was not
 explicitly configured, this value may have been automatically determined.
 
+This may be called before `agent.start()`, but the values before and after
+might differ as config is reloaded during and can be set from options given to
+`.start()`. A misconfigured agent can have an `undefined` service name.
+
 [[apm-set-framework]]
 ==== `apm.setFramework(options)`
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,6 +10,7 @@ declare class Agent implements Taggable, StartSpanFn {
   // Configuration
   start (options?: AgentConfigOptions): Agent;
   isStarted (): boolean;
+  getServiceName (): string;
   setFramework (options: {
     name?: string;
     version?: string;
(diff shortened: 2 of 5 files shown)


docs/agent-api.asciidoc
```diff
@@ -60,6 +60,14 @@ Use `isStarted()` to check if the agent has already started.
 Returns `true` if the agent has started,
 otherwise returns `false`.
 
+[[apm-get-service-name]]
+==== `apm.getServiceName()`
+
+[small]#Added in: v3.11.0#
```
astorm, line 66: +1 -- hadn't realized we marked all our new API methods with an _Added in_ until I saw this here.

test/agent.js
```diff
@@ -18,6 +18,13 @@ var inContainer = 'containerId' in (containerInfo() || {})
 process.env.ELASTIC_APM_METRICS_INTERVAL = '0'
 process.env.ELASTIC_APM_CENTRAL_CONFIG = 'false'
 
+test('#getServiceName()', function (t) {
+  var agent = Agent()
+  agent.start()
```
astorm, line 23: nit: since this is a public API and users are going to user, we could also capture how we expect `getServiceName` to behave prior to `start`being called.

index.d.ts
```diff
@@ -10,6 +10,7 @@ declare class Agent implements Taggable, StartSpanFn {
   // Configuration
   start (options?: AgentConfigOptions): Agent;
   isStarted (): boolean;
+  getServiceName (): string;
   setFramework (options: {
```
astorm, line 14: nit: Can also return _undefined_ if the configuration value isn't set.  _makes noises about the herculean task typescript has set for itself in typing the world's most dynamic mainstream programming language_

astorm (dismissed):
👍 Does the thing it says it does -- approving with some bonus `nit`s in the comments.

index.d.ts
trentm: Thanks, I'll update. You have to try hard, but yes it can be undefined.  The Agent actually loads config once before `.start()` and then again on `.start()`.  That first load was added a long while back just to be able to get create an internal logger to pass around.

That first "load config before .start()" will load `serviceName` from the following:
- the `ELASTIC_APM_SERVICE_NAME` env var, if it exists
- looking in $CWD for a "elastic-apm-node.js" file to `require()`
- looking up $CWD for a "package.json"

If none of those, then `serviceName` is undefined.

Unrelated to this change, but a fun breakage example:

```
% pwd
/

% ls ~/el/apm-agent-nodejs3/tmp
bar.js              elastic-apm-node.js

% cat ~/el/apm-agent-nodejs3/tmp/bar.js
var apm = require('../')
console.warn('XXX before', apm.getServiceName())
apm.start({
  serviceName: 'foo'
})
console.warn('XXX after', apm.getServiceName())

% node ~/el/apm-agent-nodejs3/tmp/bar.js
XXX before undefined
XXX after foo

% ELASTIC_APM_SERVICE_NAME="fromenv" node ~/el/apm-agent-nodejs3/tmp/bar.js
XXX before fromenv
XXX after fromenv

% ELASTIC_APM_SERVICE_NAME="" node ~/el/apm-agent-nodejs3/tmp/bar.js
XXX before undefined
Elastic APM isn't correctly configured: Missing serviceName
XXX after undefined
```

I found it surprising that a defined but empty-string envvar overrides the `serviceName` provided in the options to `.start()`.

test/agent.js
trentm: I've updated tests and docs.

astorm (approved):

PR #1963, opened 2021-02-03
Enforce parent-child relationship for ES/HTTP span

dgieselaar:
No idea if this makes sense.

### DIFF
--- a/lib/instrumentation/modules/@elastic/elasticsearch.js
+++ b/lib/instrumentation/modules/@elastic/elasticsearch.js
@@ -25,6 +25,8 @@ module.exports = function (elasticsearch, agent, { version, enabled }) {
     }
 
     span.name = `Elasticsearch: ${params.method} ${params.path}`
+    agent._instrumentation.activeSpan = span
+
     return span
   }
 
--- a/test/instrumentation/modules/@elastic/elasticsearch.js
+++ b/test/instrumentation/modules/@elastic/elasticsearch.js
@@ -455,6 +455,8 @@ function checkDataAndEnd (t, method, path, dbStatement) {
     t.strictEqual(httpSpan.subtype, 'http')
     t.strictEqual(httpSpan.action, 'http')
 
+    t.equal(httpSpan.parent_id, esSpan.id, 'http span should be a child of the elasticsearch span')
+
     t.equal(httpSpan.name, method + ' ' + host + path, 'http span should have expected name')
     t.equal(esSpan.name, 'Elasticsearch: ' + method + ' ' + path, 'elasticsearch span should have expected name')
 


trentm:
Interesting. 

Compare to https://github.com/elastic/apm-agent-nodejs/pull/1890/files FWIW, which is my proposed patch for solving this general issue for all cases -- albeit with the requirement to grasp possible impact of that wider behaviour change.

dgieselaar:
@trentm I've moved this out of draft so we can make a decision at some point. Feel free to close it if you feel it's not the right approach.

trentm:
jenkins run the tests

Trying the jenkins run again to rule out a transient.

dgieselaar:
jenkins run the tests please

trentm:
> I'm also wondering if we can use this strategy in our custom instrumentation in Kibana (instead of awaiting process.nextTick or Promise.resolve.

Ah, do you hit the same issue with custom spans? Something like:

```js
// ...
const span = apm.createSpan("doingThatThing")
await esClient.makeACall(...)
// ...
span.end()
```

Obviously that underscore in `agent._instrumentation` is a big hint that we'd rather not have usage like that outside the agent itself.

farside-unsafe-chair

dgieselaar:
@trentm Hmm. Let me rephrase: if we have to pick between `await Promise.resolve()` and `agent._instrumentation.activeSpan = span`, what's the least worst option?

trentm:
> we'd rather not have usage like that outside the agent itself.

I don't know what I was thinking when I said that. This *is* a patch to the agent itself. 

>  if we have to pick between `await Promise.resolve()` and `agent._instrumentation.activeSpan = span`, what's the least worst option?

Promise.resolve - you can do without the agent needing to take this change or to wait for https://github.com/elastic/apm-agent-nodejs/pull/1890 to do similar. I don't have any idea the potential impact on Kibana perf with it doing that in (all?) its calls to ES.

I *do* want to get to these changes, but I probably won't get there quickly.

What if the agent had an option to elide the child HTTP span for ES queries altogether? (I'm assuming we can even do this relatively easily in the node agent.) Would that even be preferable for your case? It would effectively workaround this issue until parent-child spans in general are tackled.

dgieselaar:
> Promise.resolve - you can do without the agent needing to take this change or to wait for #1890 to do similar. I don't have any idea the potential impact on Kibana perf with it doing that in (all?) its calls to ES.

I'm currently using the `runInAsyncScope` workaround: https://github.com/elastic/kibana/pull/90403/files#diff-3aa1ae78694030a50794cbcc433f97b37cfd3084afaa451871d06e220c7366bfR42

awaiting `Promise.resolve()` works as well, but things get executed in a slightly different order then with `runInAsyncScope` (AFAICT), so I figured this is more predictable (I couldn't really see a difference in performance on a single request). 

> What if the agent had an option to elide the child HTTP span for ES queries altogether? (I'm assuming we can even do this relatively easily in the node agent.) Would that even be preferable for your case? It would effectively workaround this issue until parent-child spans in general are tackled.

I'm not sure what the consequences would be there. I do like the idea of less spans, but the spans might have different semantics which might be useful for a specific scenario (e.g., the HTTP span might record the network error, something like that?) I'm okay with postponing this for sure. 

What also might be an option is to have the agent add something like a `startSpanWithContext(spanOptions, runFn)`, where the active span for that context is the one that just started, and all spans created in that context are the direct child of that span. And the trigger context stays "unpolluted". Basically adding a more explicit API so there's no breaking change or unexpected consequences.

ddolcimascolo:
Hey guys, any progress on that one ? I'm started to get tired of all these double HTTP/Elasticsearch spans 😊

Thx,
David

dgieselaar:
@ddolcimascolo hello David, that's being tracked in this issue I think: https://github.com/elastic/apm-agent-nodejs/issues/2000

trentm:
@ddolcimascolo I'm currently looking into #2000 (and effectively this ticket as well). Nothing to report yet.

ddolcimascolo:
Thx guys

PR #1964, opened 2021-02-04
Traverse context to find active span/transaction

dgieselaar:
**(Keeping this in draft, as I expect this PR is mostly just suitable for discussion)**

I'm running into an issue where if the active span for the current async context ends, and a new span immediately starts, the ended span is the parent of the new span. 

This PR attempts to address this issue by:
- when a span ends, unset the active span for the current async context
- when activeSpan or activeTransaction is requested, walk back up the context hierarchy to return the first span or transaction it finds

This assumes that each transaction/span is executed in its own async context. If that's not the case, weird things might happen.

### DIFF
--- a/lib/instrumentation/async-hooks.js
+++ b/lib/instrumentation/async-hooks.js
@@ -11,16 +11,28 @@ module.exports = function (ins) {
   const asyncHook = asyncHooks.createHook({ init, before, destroy })
   const contexts = new WeakMap()
 
+  const resources = new Map()
+
   if (resettable) {
     if (_asyncHook) _asyncHook.disable()
     _asyncHook = asyncHook
   }
 
+  function traverseContext ( asyncId, map ) {
+    const event = map.get(asyncId)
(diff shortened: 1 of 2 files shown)


trentm:
@dgieselaar This looks pretty good! Thanks for this. I think we should pursue this. I also (eventually) want to get a sense of the possible performance impact here because the async hook functions are hot spots.

My understanding improved when (locally) renaming `resources` to `triggerAsyncIdFromAsyncId`. There is a fourth argument to the async hook's `function init` called "resource" that means something different.

dgieselaar:
@trentm I was thinking about the performance issue as well. I'll try to have a look at profiling next week, it's also going to be useful for me to learn more about the Kibana instrumentation performance overhead we discussed.

Other than the perf issue I think I also not fully understand the consequences of this change. But it does feel like a step in the right direction.

I hope to get back to this ~next week after I've had a look at the performance implications.

PR #1971, opened 2021-02-09
tav: docker-compose down with orphans and no fail

v1v:
### What

In the CI the `.ci/scripts/test.sh` runs in each specific node with a retry in case something bad happened (it seems). 

Ensure the docker-compose down manages the orphans and it does not fail if something bad happened, the reason, with ephemeral workers we don't need to tear-down the environment.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


trentm:
> with a retry in case something bad happened (it seems).

These retries were added in https://github.com/elastic/apm-agent-nodejs/pull/801 
There was some discussion of flakiness, but nothing specific on why those retries were added. It might be worth removing those retries (in a separate commit).

trentm:
> It might be worth removing those retries (in a separate commit).

I added https://github.com/elastic/apm-agent-nodejs/issues/1972 to look at that later.

.ci/scripts/test.sh
```diff
@@ -83,4 +83,4 @@ NODE_VERSION=${NODE_VERSION} docker-compose \
   --no-ansi \
   --log-level ERROR \
   -f ${DOCKER_FOLDER}/${DOCKER_COMPOSE_FILE} \
-  down -v
+  down -v --remove-orphans || true
```
trentm, line 86: 1. Adding `--remove-orphans` sounds reasonable to me.
2. I'm not sure I want to add the `|| true` yet because it could mask real issues, right?

Recall that I said it was crazy we were setting up so many docker containers for the "@hapi/hapi" TAV tests (e.g. https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/master/763/pipeline/1290)? I think I see the issue why. It is a fallback here:
https://github.com/elastic/apm-agent-nodejs/blob/c94c799502dc4df2b09e513275337d5c98ee0d4e/.ci/scripts/test.sh#L9-L38

At least for that particular Jenkins build, the `error while removing network: network docker_default id ... has active endpoints` failures were for the steps testing "@hapi/hapi" and "@elastic/elasticsearch" which are missing from https://github.com/elastic/apm-agent-nodejs/blob/c94c799502dc4df2b09e513275337d5c98ee0d4e/.ci/scripts/test.sh#L10

I think we should update that `case ...` to include those two.
I also think we should consider warning or *erroring* in the `*)` case: https://github.com/elastic/apm-agent-nodejs/blob/c94c799502dc4df2b09e513275337d5c98ee0d4e/.ci/scripts/test.sh#L37

It may be that `docker-compose down ...` here is eventually still an issue, but I don't want to `|| true` it out yet.

trentm (changes requested):

trentm:
Here is another case of that `error while removing network: network docker_default id cfb384f605476f50bdd4272a0a8b0e82463d64fabcca849e47f7e3e1ba24b326 has active endpoints
` failure: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/astorm%2Fundo-docs-change/1/pipeline

This docker issue comment: https://github.com/moby/moby/issues/17217#issuecomment-196607597
suggests:

> BTW, if the container is removed, but the endpoint is still seen, then one can force disconnect the endpoint using docker network disconnect -f {network} {endpoint-name} . You can get the endpoint-name from the docker network inspect {network} command.

I'm not sure how actionable that is for our case here. Would we want to do that `docker network disconnect -f ...` up front before `docker-compose down`? Or only if `docker-compose down` failed with that error case. Yuck.

If we just `docker-compose down ... || true` are we leaving behind docker networks that will cause issues for subsequent runs? I don't know if CI is running in a fresh VM/container.

v1v:
> If we just docker-compose down ... || true are we leaving behind docker networks that will cause issues for subsequent runs? I don't know if CI is running in a fresh VM/container

A fresh fungible VM per parallel stage, https://github.com/elastic/apm-agent-nodejs/blob/master/.ci/scripts/test.sh is called within the node closure -> https://github.com/elastic/apm-agent-nodejs/blob/6fb89e84d1a41155452206f1feb26a0aab1a5509/.ci/Jenkinsfile#L368-L388

trentm:
@v1v What is about the `node('...')` call (https://www.jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#node-allocate-node) that shows that this is a newly provisioned VM?

trentm:
Ah, I am learning about "gobld" in infra.git and how Jenkins workers are provided.

trentm:
FWIW, I'm playing a little bit with that `docker-compose down` in https://github.com/elastic/apm-agent-nodejs/pull/1974  but so far I can't get it to fail. :)

v1v:
I'll then close this PR in favour of https://github.com/elastic/apm-agent-nodejs/pull/1974

trentm:
I hope you don't mind, I'm going to re-open this. My playing in #1974 is likely just to inform this one. :)

trentm:
@v1v Would you be willing to update your patch here to drop the `|| true`. I'd like to get the `--remove-orphans` change in now. (I've closed my experimental #1974 PR.)

.ci/scripts/test.sh
trentm, line 86:
```suggestion
  down -v --remove-orphans
```

v1v:
@trentm , I missed your last comment, sorry,  I've just seen that you managed to do it, thanks 🙏

kuisathaverat (approved):

trentm:
@v1v No worries, thanks!

trentm (approved):

PR #1979, opened 2021-02-11
feat: transaction_sample_rate precision

trentm:
This implements the transaction_sample_rate precision spec (https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md#transaction_sample_rate-configuration), added in https://github.com/elastic/apm/pull/335.

Fixes: #1817

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -367,7 +367,11 @@ If the `errorOnAbortedRequests` property is `false`, this property is ignored.
 
 Specify the sampling rate to use when deciding whether to trace a request.
 
-The value should between `0.0` and `1.0` where `1.0` is 100% of all requests.
+The must be a value between `0.0` and `1.0`, where `1.0` means 100% of requests
+are traced. The value is rounded to four decimal places of precision (e.g.
+0.0001, 0.3333) to ensure consistency and reasonable size when propagating the
+sampling rate in the `tracestate` header for
+<<distributed-tracing,distributed tracing>>.
 
 [[hostname]]
 ==== `hostname`
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -367,7 +367,7 @@ If the `errorOnAbortedRequests` property is `false`, this property is ignored.
 
 Specify the sampling rate to use when deciding whether to trace a request.
 
-The must be a value between `0.0` and `1.0`, where `1.0` means 100% of requests
+This must be a value between `0.0` and `1.0`, where `1.0` means 100% of requests
 are traced. The value is rounded to four decimal places of precision (e.g.
 0.0001, 0.3333) to ensure consistency and reasonable size when propagating the
 sampling rate in the `tracestate` header for
--- a/lib/config.js
+++ b/lib/config.js
@@ -154,10 +154,6 @@ var CENTRAL_CONFIG = {
   sanitize_field_names: 'sanitizeFieldNames'
 }
 
-var VALIDATORS = {
-  transactionSampleRate: numberBetweenZeroAndOne
-}
-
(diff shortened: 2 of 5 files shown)


docs/configuration.asciidoc
```diff
@@ -367,7 +367,11 @@ If the `errorOnAbortedRequests` property is `false`, this property is ignored.
 
 Specify the sampling rate to use when deciding whether to trace a request.
 
-The value should between `0.0` and `1.0` where `1.0` is 100% of all requests.
+The must be a value between `0.0` and `1.0`, where `1.0` means 100% of requests
+are traced. The value is rounded to four decimal places of precision (e.g.
```
astorm, line 371: ~The must~  The value must? This must?

lib/config.js
```diff
@@ -320,21 +316,16 @@ class Config {
...
-                  // If a new log level value was provided, and we aren't
-                  // already using a custom-provided logger (_conf.logger),
-                  // then update the agent logger.
-                  agent.logger = agent._conf.logger = logging.createLogger(value)
-                  agent.logger.info(`Remote config success. Updated logger with new logLevel: ${value}`)
-                }
-                agent.logger.info(`Remote config success. Updated ${key}: ${value}`)
-              } else {
-                agent.logger.warn(`Remote config failure. Invalid value for ${key}: ${value}`)
+              const oldValue = agent._conf[key]
```
astorm, line 319: I see here we're dropping the use of the `numberBetweenZeroAndOne` validator from `VALIDATORS`.  I presume this is because this validation has been moved to `normalizeTransactionSampleRate` 👍

lib/config.js
```diff
@@ -422,9 +413,33 @@ function normalize (opts, logger) {
...
+// - with the special case that a value in the range (0, 0.0001] should be
+//   rounded to 0.0001 -- to avoid a small value being rounded to zero.
+//
+// https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md
+function normalizeTransactionSampleRate (opts, logger) {
+  if ('transactionSampleRate' in opts) {
+    // The value was already run through `Number(...)` in `normalizeNumbers`.
+    const rate = opts.transactionSampleRate
+    if (isNaN(rate) || rate < 0 || rate > 1) {
+      opts.transactionSampleRate = DEFAULTS.transactionSampleRate
```
astorm, line 432: I see this conditional performs a similar validation to the number that was done previously by `numberBetweenZeroAndOne`.  Also, since (per the comment) we've already nun the value through `Number`,  `isNaN` won't "false position" on numbers as strings.  👍

lib/config.js
```diff
@@ -422,9 +413,33 @@ function normalize (opts, logger) {
...
+      logger.warn('Invalid "transactionSampleRate" config value %s, falling back to default %s',
+        rate, opts.transactionSampleRate)
+    } else if (rate > 0 && rate < 0.0001) {
+      opts.transactionSampleRate = 0.0001
+    } else {
+      opts.transactionSampleRate = Math.round(rate * 10000) / 10000
+    }
```
astorm, line 439: Rounding rules checkout per the spec ✅

astorm (changes requested):
Does the thing -- just a small typo in the docs to fix and we should be good here.

docs/configuration.asciidoc
trentm: "This must". Fixed.

lib/config.js
trentm: >  I presume this is because this validation has been moved

Yup. The two separate stages: validation and normalization weren't necessary.

astorm (approved):

PR #1988, opened 2021-02-25
test: use the local apm-server schema files in tests

trentm:
With #1985 we have changed to have apm-server schema files commited to
repos.  This simplifies the tests to not have to download those schema
files at test-time.

Refs: #1985


test/integration/api-schema/_utils.js
```diff
@@ -48,30 +25,23 @@ exports.errorValidator = thunky(function (cb) {
   loadSchema('error.json', cb)
 })
 
-function loadSchema (relativePath, cb) {
-  schemaDir(function (err, dir) {
-    if (err) {
-      return cb(err)
+function loadSchema (schemaFileName, cb) {
+  const schemaPath = join(schemaDir, schemaFileName)
+  fs.readFile(schemaPath, { encoding: 'utf8' }, function (readErr, content) {
```
astorm, line 30: Just noting out loud that we've removed the `console.log('processing "%s"...', path)` debugging output.  I'm presuming this is intention 👍

astorm (approved):
👍 it does the thing, and tests run locally without issue. Approving.

test/integration/api-schema/_utils.js
trentm: Yup that was the intent: 1. I don't expect this to fail anymore, so isn't interesting. 2. That "processing ..." output isn't valid TAP output.

PR #1993, opened 2021-03-09
chore: drop unused 'nodejs.eventloop.delay.ns' metric

trentm:
The nodejs runtime metrics code define both
    nodejs.eventloop.delay.ns
    nodejs.eventloop.delay.avg.ms
but only the latter is every filled with a value. The former
is initialized to 0 and always reported as 0. This stems from the
original addition in #1021.

The apm-contrib Node.js dashboard only uses the latter.
https://github.com/elastic/apm-contrib/tree/master/apm-agent-nodejs/dashboards/

Refs: #1021

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,22 @@ Notes:
 [[release-notes-3.x]]
 === Node.js Agent version 3.x
 
+==== Unreleased
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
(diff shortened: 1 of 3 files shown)


astorm (approved):
seems safe to rename, and an extra 👍 for the due diligence in our contrib dashboards.

trentm:
Screen Shot 2021-03-10 at 11 45 24 AM

I'm adding a changelog entry.

PR #2005, opened 2021-03-16
chore: drop ancestors trace log (it adds nothing over 'startTrace' info) and the dep for it

trentm:


astorm (commented):
Question: My intuition is that the `require-ancestors` module is here to give anyone supporting the agent some insights as to whether the agent's been included at the start of a user's program (as required), or at some deeper level in their application (which might result in a module not being instrumented.  

With this gone is the there an equivalent way for a TSE to tell where/when the agent was included in a customer's application?

trentm:
@astorm

> With this gone is the there an equivalent way for a TSE to tell where/when the agent was included in a customer's application?

What is "TSE"?


I believe `stackTrace` provides the same and more info (line numbers). For example, here is part of the output from that logger.trace call for a simple express app:

```
% node myapp.js
agent configured correctly { pid: 62516,
  ppid: 62321,
  arch: 'x64',
  platform: 'darwin',
  node: 'v10.23.0',
  agent: '3.12.1',
  ancestors:
   [ '/Users/trentm/tmp/p3/node_modules/elastic-apm-node/index.js',
     '/Users/trentm/tmp/p3/myapp.js',
     [length]: 2 ],
  startTrace:
   [ 'at Agent.start (/Users/trentm/tmp/p3/node_modules/elastic-apm-node/lib/agent.js:179:11)',
     'at Object.<anonymous> (/Users/trentm/tmp/p3/myapp.js:1:41)',
     'at Module._compile (internal/modules/cjs/loader.js:778:30)',
     'at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)',
     'at Module.load (internal/modules/cjs/loader.js:653:32)',
     'at tryModuleLoad (internal/modules/cjs/loader.js:593:12)',
     'at Function.Module._load (internal/modules/cjs/loader.js:585:3)',
     'at Function.Module.runMain (internal/modules/cjs/loader.js:831:12)',
     'at startup (internal/bootstrap/node.js:283:19)',
     'at bootstrapNodeJSCore (internal/bootstrap/node.js:623:3)',
     [length]: 10 ],
  main: 'index.js',
  dependencies: { 'elastic-apm-node': '^3.12.1' },
```

Is there some value "ancestors" provides here that "stackTrace" doesn't?

astorm:
@trentm Ah, sorry for the previous job jargon -- TSE stands for Technical Support Engineer.  

The info in `startTrace` looks like it's an excellent substitute.  How would you feel about adding an item to `CHANGELOG.md` that mentions this change? With that in place this looks merge-able to me.

trentm:
> How would you feel about adding an item to `CHANGELOG.md`

Thanks. Sure, will do. Feels slightly overkill for dropping a field from a log.trace.

Issue #1956, opened 2021-01-29
[META 408] Instrumentation for AWS SQS


AlexanderWert:
Provide instrumentation for AWS SQS.

Meta issue: elastic/apm#408
SPEC: elastic/apm#409

astorm:
Some notes on the implementation details of the AWS SDK that will be helpful in instrumenting API methods for the SQS API as well as the other APIs we'll want to instrument. 

The AWS SDK provides an internal system for building constructor-functions that can inherit from one another and provide other features that an end-user-programmer might expect to find in a class based system.  This has been a popular pattern in javascript frameworks since the language's inception.

Here's an example of using this system to create a simple `ClassA extends BaseClass` hierarchy.

```javascript
    const AWS = require('aws-sdk')
    BaseClass = AWS.util.inherit({
      constructor: function BaseClass() {
        console.log("BaseClass constructor")
      },

      test: function(value) {
        console.log(`In BaseClass: ${value}`)
        this.science = 'science'
        return true
      }
    })

    ClassA = AWS.util.inherit(BaseClass, {
      constructor: function ClassA() {
        console.log("ClassA constructor")
        BaseClass.call(this)
        this.science = null
      },

      test: function(value) {
        console.log(`Before: this.science = ${this.science}`)
        console.log(`In ClassA: ${value}`)
        BaseClass.prototype.test.call(this,"goodbye")
        console.log(`After: this.science = ${this.science}`)
      }
    })
    const a = new ClassA()
    a.test("hello")
```

While `AWS.util.inherit` is [marked as a private API](https://github.com/aws/aws-sdk-js/blob/309d053672af4b0b6a480d54960a402501f5c823/lib/util.js#L613), this system has been in place since the 1.0 release of the framework and has remained remarkably stable.  If we need to extend any constructor-functions that aren't service-constructor-functions (see below) this seems like a reasonable API to use.

## Service Constructor Functions

In addition to these `AWS.util.inherit` objects, the `aws-sdk` also features special service-constructor-functions.  These service-constructor-functions allow end-user-programmer to create objects that allow them to use [the `aws-sdk`'s public APIs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/).  

For example, in this code

```javascript
    var sqs = new AWS.SQS({apiVersion: '2012-11-05'});
    var params = {/* ... */ }
    sqs.sendMessage(params, function(err, data) {
      if (err) {
        console.log("Error", err);
      } else {
[...]

PR #2008, opened 2021-03-17
feat: ignore_message_queues configuration

astorm:
Partially implements: https://github.com/elastic/apm-agent-nodejs/issues/1956

This PR adds [the `ignore_message_queues`](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-messaging.md#ignore_message_queues-configuration) configuration to the agent.  This is a prerequisite for supporting AWS SQS instrumentation.  This PR does not include a CHANGELOG entry (that will come later, once the first messaging system (SQS) is added)

- [x] Implement code
- [X] Add tests
- [X] Update TypeScript typings
- [X] Update documentation
- [X] Commit message follows [commit guidelines](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#commit-message-guidelines)

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -933,3 +933,33 @@ During startup the Node.js agent queries the local environment to determine whet
 - `none`: Do not query for any cloud provider information.
 
 If the value is not one of the five listed above, the agent will use the value of `auto`.
+
+[[ignore_message_queues]]
+==== `ignoreMessageQueues`
+* *Type:* Array
+* *Default:* `[]`
+* *Env:* `ELASTIC_IGNORE_MESSAGE_QUEUES`
+
+Array or comma-separated string of wildcard patterns that tell the agent to
+ignore certain queues/topics when instrumenting messaging systems.
+
+When an instrumented messaging system sends or receives a message, the agent
+will test the queue/topic name against each wildcard in this list. If
+the name matches the agent will skip instrumenting the operation.
+
+The `ignoreMessageQueues` property supports simple wildcard (`*`) patterns, and
+may not include commas.  Wildcard matches are case-insensitive by default. You
+may make wildcard searches case-sensitive by using the `(?-i)` prefix.
+
+Example usage:
+
+[source,js]
+----
+require('elastic-apm-node').start({
+  ignoreMessageQueues: [
+    'overnight_jobs',
+    'events_*',
+    '(?-i)/caseSensitiveSearch'
+  ]
+})
+----
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -934,18 +934,19 @@ During startup the Node.js agent queries the local environment to determine whet
 
 If the value is not one of the five listed above, the agent will use the value of `auto`.
 
-[[ignore_message_queues]]
+[[ignore-message-queues]]
 ==== `ignoreMessageQueues`
 * *Type:* Array
 * *Default:* `[]`
 * *Env:* `ELASTIC_IGNORE_MESSAGE_QUEUES`
+* <<dynamic-configuration, image:./images/dynamic-config.svg[] >> *Central config name:* `ignore_message_queues`
 
 Array or comma-separated string of wildcard patterns that tell the agent to
 ignore certain queues/topics when instrumenting messaging systems.
 
 When an instrumented messaging system sends or receives a message, the agent
 will test the queue/topic name against each wildcard in this list. If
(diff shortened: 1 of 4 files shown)


docs/configuration.asciidoc
```diff
@@ -933,3 +933,33 @@ During startup the Node.js agent queries the local environment to determine whet
 - `none`: Do not query for any cloud provider information.
 
 If the value is not one of the five listed above, the agent will use the value of `auto`.
+
+[[ignore_message_queues]]
```
trentm, line 937: Typically we use hyphens for the HTML anchors.

Aside: Do you have suggestion/ideas for where to include the actual string "ignore_message_queues" in changes like this? The Node.js Agent spells its config vars in the different ignoreMessageQueues pattern which breaks [the grep test](http://jamie-wong.com/2013/07/12/grep-test/). This need not block this change, however.

docs/configuration.asciidoc
```diff
@@ -933,3 +933,33 @@ During startup the Node.js agent queries the local environment to determine whet
...
+* *Type:* Array
+* *Default:* `[]`
+* *Env:* `ELASTIC_IGNORE_MESSAGE_QUEUES`
+
+Array or comma-separated string of wildcard patterns that tell the agent to
+ignore certain queues/topics when instrumenting messaging systems.
+
+When an instrumented messaging system sends or receives a message, the agent
+will test the queue/topic name against each wildcard in this list. If
+the name matches the agent will skip instrumenting the operation.
```
bmorelli25, line 948:
```suggestion
the name matches, the agent will skip instrumenting the operation.
```

docs/configuration.asciidoc
bmorelli25, line 937:
```suggestion
[[ignore-message-queues]]
```

How do you feel about dashes here to maintain consistency with the other anchors?

docs/configuration.asciidoc
```diff
@@ -933,3 +933,33 @@ During startup the Node.js agent queries the local environment to determine whet
...
+
+Example usage:
+
+[source,js]
+----
+require('elastic-apm-node').start({
+  ignoreMessageQueues: [
+    'overnight_jobs',
+    'events_*',
+    '(?-i)/caseSensitiveSearch'
```
trentm, line 962:
```suggestion
    '(?-i)caseSensitiveSearch'
```

Perhaps drop the slash... unless it is typical to have slashes in queue names. This example came from transactionIgnoreUrls where the slashes are less likely to be confused as being part of the syntax.

docs/configuration.asciidoc
```diff
@@ -933,3 +933,33 @@ During startup the Node.js agent queries the local environment to determine whet
 - `none`: Do not query for any cloud provider information.
 
 If the value is not one of the five listed above, the agent will use the value of `auto`.
+
+[[ignore_message_queues]]
+==== `ignoreMessageQueues`
+* *Type:* Array
+* *Default:* `[]`
+* *Env:* `ELASTIC_IGNORE_MESSAGE_QUEUES`
+
```
trentm, line 942: Are you adding support for central config later?

docs/configuration.asciidoc
astorm: 👍 -- making this change.  @bmorelli25 are there automated docs linting tools that would catch this sort of thing for us?

astorm: 👍 switching to hyphens.  I don't have any good suggestions/ideas for where the literal text `ignore_message_queues` should go.  The decision to stick with camel casing binds our hands there.  I think the best we can do is the more recent convention of prepending `ELASTIC_` to get the env variable name (which presumes a `grep -i` test).

docs/configuration.asciidoc
astorm: Ah -- good catch.  I'd missed/assumed that this wasn't a configuration intended for central config, but re-reading the spec confirms it is.   

I've now added it to the `CENTRAL_CONFIG` array in `lib/config.js` and updated these docs and tested the configuration with a mock APM server.   I'll also need to cargo cult this PR in Kibana to make this configuration a central config. https://github.com/elastic/kibana/pull/85655

docs/configuration.asciidoc
astorm: ✅

docs/configuration.asciidoc
astorm: ✅ making this change

trentm (approved):

Issue #2012, opened 2021-03-23
Amazon SQS: Transactions and Distributed Tracing Headers for `receiveMessage`


astorm:
The [current SQS implementation](https://github.com/elastic/apm-agent-nodejs/pull/2013/files) does _not_ automatically generate transactions or handle the distributed tracing `traceparent`/`tracecontext` headers for the `receiveMessage` API method.  We currently recommend users use the `startTransaction` API to decide when/how they want transactions started for this use case.  

We should consider implementing automatic handling of both the transaction creation and distributed tracing headers for SQS queues.  

### Where to set the `traceparent` header.

- SQS has no concept of a "header".  The only place we have available to store a header is with [the message attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html).  This it not ideal, as these are the same attributes a user might use to set data on their messages and Amazon limits messages to 10 headers.  In addition to taking up 10% of the available headers, it's also unclear what name we should use for this message.  Since this is a space users can set their own headers an un-namespaced  `traceparent` might not be a "safe" value (this is unlike HTTP where `traceparent` _is_ considered safe since it's a W3C standard that's designed to be shared by multiple vendors.

### When to Create a Transaction

- There's no clear/ideal place to start/stop a transaction.  The `receiveMessage` method is a polling style API with a callback.  Users call `receiveMessage`, passing in a callback.  The SQS SDK will call this callback once it's finished fetching data and can return between 1 and 10 messages
    - A transaction could be automatically started for the duration of the callback.  This approach might interfere if users want to have a transaction per message in order to use distributed tracing.  
    - Another approach is to attempt to start a transaction per message -- however the response object isn't typed -- it's an object that includes a `Messages` property that contains [an array of Message objects](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_Message.html) -- but the message objects are just native javascript objects.  Wrapping either the iteration or access of these message is a non-trivial task. 

- In addition to the challenges of where to set a distributing tracing header, there's also the question of how transaction parenting should work with this polling-via-callback style API.  A single call to `receiveMessage` can return messages that may have started in different transactions.  We don't have an implementation for anything like span links (which would allow a spans to have multiple parent-like spans), and attempting to link via a transaction for each message runs into the previously mention problem of not being able to wrap either the iteration or access to native message array.

Issue #2000, opened 2021-03-15
exclude the HTTP span under an Elasticsearch span


trentm:
From discussion on today's apm-agent-nodejs call and from https://github.com/elastic/apm/pull/420 we think the HTTP span "under" an Elasticsearch (ES) span from instrumentation of the "elasticsearch" and "@elastic/elasticsearch" modules should be excluded. The ES span is an "exit" span, and arguably the HTTP request it makes is an implementation detail. As well, eliding that span would make breakdown metrics much more useful/realistic. (FWIW, it would also side-step the current issue with the HTTP span being a *sibling* of, rather than a child of, the ES span for the instrumentation of the newer "@elastic/elasticsearch" client.)

From memory (though might be nice to confirm), the Python agent excludes the HTTP span (via `leaf=True` in the code). I recall one other language agent optionally including the HTTP span, but I think it was off by default.

dgieselaar:
How do we capture retries if the HTTP span is dropped?

joshdover:
+1 on this for Kibana's usage. We'd like to start enabling our support team to use APM instrumentation to debug performance problems and I think this duplication creates confusing noise.

> How do we capture retries if the HTTP span is dropped?

Could we add a `context.http.attempts` field?

trentm:
> How do we capture retries if the HTTP span is dropped?

I'll ask the APM group. I don't see any current apm-server intake fields, no ECS http fields that could easily hold this value.

trentm:
Note to self: because a span synchronously created under a current span (e.g. the HTTP span under an ES span) currently isn't set as a *child* might complicate implementation of this. If we had the child relationship (as we eventually want), then the implementation could just be something like the Python agent's `span.leaf=true` and use that to avoid creating new span children.

trentm:
Note that I haven't pursued the "how to track retries". We discussed it at an APM Agents call (2022-02-01). There isn't a currently suitable ECS field for this. I considered adding a span label for this, but it would be first case of any Node.js APM agent instrumentation adding span labels automatically -- I'm not sure if span labels should remain the domain of the user.

User code (e.g. Kibana code) should be able to add an [ES Client "response" handler](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/observability.html#_events) to add a label to the `apm.currentSpan` with the number of attempts.  Something like:

```js
client.on('response', (_err, result) => {
  // Obviously missing defensive guards here.
[...]

Issue #1977, opened 2021-02-10
MetricsRegistry and High Cardinality 


astorm:
We've received reports that [the current implementation of our metrics registry](https://github.com/elastic/kibana/pull/90955) can create situations where a service or application with a large number of distinct URLs will create a large number of distinct transaction names which, in turn, will create a large number of unique "metric names + labels/dimensions" instances which, over time, can overwhelm APM Server with traffic.  This is particularly concerning when a low traffic endpoint creates a metric once that is rarely incremented, but continues to be reported because it was seen once. 

In other words, the High Cardinality problem.

The [`captureBreakdown` method](https://github.com/elastic/apm-agent-nodejs/blob/fb9923fbd143064bd548e7b0adad1fe9a9ca27ee/lib/instrumentation/transaction.js#L268) is where these problematic metrics originate. 

Our current workaround is to recommend that users use [the `setTransactionName` API method](https://www.elastic.co/guide/en/apm/agent/nodejs/master/agent-api.html#apm-set-transaction-name) to reduce the unique transaction names being generated.

This issue is track whether other users are seeing similiar issues with this and discuss possible work around (ejecting stale metrics, allowing configuration of what breakdown metrics are reported, etc,)

trentm:
Notes to myself on some details of the issue:

Here is the base metricset that the Node.js APM agent will send every metricsInterval:

```
    {
        "metricset": {
            "samples": {
                "system.cpu.total.norm.pct": { "value": 0 },
                "system.memory.total": { "value": 68719476736 },
                "system.memory.actual.free": { "value": 36078088192 },
                "system.process.cpu.total.norm.pct": { "value": 0.06212579838674182 },
                "system.process.cpu.system.norm.pct": { "value": 0.007457950524313062 },
                "system.process.cpu.user.norm.pct": { "value": 0.05466784786242876 },
                "system.process.memory.rss.bytes": { "value": 41197568 },
                "nodejs.handles.active": { "value": 3 },
                "nodejs.requests.active": { "value": 0 },
                "nodejs.eventloop.delay.avg.ms": { "value": 0 },
                "nodejs.memory.heap.allocated.bytes": { "value": 17809408 },
                "nodejs.memory.heap.used.bytes": { "value": 11838984 },
                "nodejs.memory.external.bytes": { "value": 1730938 },
                "nodejs.memory.arrayBuffers.bytes": { "value": 234834 }
            },
            "timestamp": 1626884724383000,
[...]

PR #2013, opened 2021-03-23
feature: sqs queue instrumentation 

astorm:
Part of #1956

This PR provides instrumentation for Amazon's SQS queues and partially implements the messaging spec.

This PR _does not_ automatically generate transactions during message processing -- see #2012 for more information.  To work around this limitation we provide documentation showing a few different ways users could use the API to generate transactions when processing messages.  

We instrument the SQS method calls by wrapping the underlying AWS SDK `AWS.Request.send` method.  The main instrumentation entry point is built in such a way that should allow us to instrument future aws-sdk services in a similar fashion.

We also generate a queue latency metric using the existing [measured based](https://www.npmjs.com/package/measured-core?activeTab=readme) metric system.  These new metrics are query-able via the Kibana `apm` index pattern

dimensions


docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
+[[message-quques]]
+== Message Queues
+
+The Node.js Agent will automatically create spans for any activity to/from your Amazon SQS message queues.  In order to have those spans recorded, your message queue activity must  occur during a transaction.
```
bmorelli25, line 4:
```suggestion
The Node.js Agent will automatically create spans for any activity to/from your Amazon SQS message queues.  To record these spans, your message queue activity must occur during a transaction.
```

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+
+If you're performing queue operations during an HTTP request from a <<compatibility-frameworks,supported framework>>     the agent will start a transaction automatically.  However, if you're performing queue operations in a stand alone program (such as a message processor) you'll need to use the Node.js Agent's <<apm-start-transaction,`startTransaction()`>> method to create transactions for your messages.
```
bmorelli25, line 6:
```suggestion
If you're performing queue operations during an HTTP request from a <<compatibility-frameworks,supported framework>>, the agent will start a transaction automatically.  However, if you're performing queue operations in a stand-alone program (such as a message processor), you'll need to use the Node.js Agent's <<apm-start-transaction,`startTransaction()`>> method to manually create transactions for your messages.
```

bmorelli25: I'm not sure that this paragraph should be separate from the preceding one.

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+  } else {
+    console.log(`Length: ${data.Messages.length}`)
+    /* process messages */
+  }
+  // end the transaction
+  transaction.end()
+})
+----
+
+In the above code you can see we've started a transaction before calling the `sqs.receiveMessage` method, and ended it _after_ the queue's processing callback finishes executing.  The will ensure a transaction is active while processing your queue messages.
```
bmorelli25, line 35: I'm a sucker for code callouts, but nbd either way.

```suggestion
const transaction = apm.startTransaction("Process Messages", 'cli') <1>
sqs.receiveMessage(params, function(err, data) {
  if(err) {
    console.log("Receive Error", err);
  } else {
    console.log(`Length: ${data.Messages.length}`)
    /* process messages */
  }
  // end the transaction
  transaction.end() <2>
})
----
<1> Prior to calling the `sqs.receiveMessage` method, start a new transaction.
<2> Only end the transaction _after_ the queue's processing callback finishes executing. The will ensure a transaction is active while processing your queue messages.
```

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
+[[message-quques]]
+== Message Queues
```
bmorelli25, line 2: We use [sentence case capitalization](https://apastyle.apa.org/style-grammar-guidelines/capitalization/sentence-case) for headings.

```suggestion
== Message queues
```

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+  // end the transaction
+  transaction.end()
+})
+----
+
+In the above code you can see we've started a transaction before calling the `sqs.receiveMessage` method, and ended it _after_ the queue's processing callback finishes executing.  The will ensure a transaction is active while processing your queue messages.
+
+[float]
+[[message-queues-distributed-tracing]]
+=== Distributed Tracing and Messaging Queues
```
bmorelli25, line 39: Sentence case again

```suggestion
=== Distributed tracing and messaging queues
```

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+
+Similarly, if you want to connect your queue scheduling and queue processing via distributed tracing, you'll need to use the Node.js Agent's API to _store_ a `traceparent` header with your queue message and then provide that `traceparent` header when starting the new transaction.
```
bmorelli25, line 41: It might not be immediately clear what we're referring to when we say, "similarly". I would simplify to something like:

```suggestion
To enable queue scheduling and queue processing with distributed tracing, use the Node.js Agent's API to _store_ a `traceparent` header with your queue message; then, provide that `traceparent` header when starting a new transaction.
```

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+
+Here's an example that uses the Node.js Agent API to store the `traceparent` as a message attribute and then uses that attribute to link your new transaction with the original.
+
+First, when sending the message, we add the trace as one of the `MessageAttributes`
```
bmorelli25, line 45:
```suggestion
First, when sending the message, we add the trace as one of the `MessageAttributes`.
```

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+}
+sqs.sendMessage(params, function(err, data) {
+  /* ... */
+});
+----
+
+Then, when receiving the messages, we start and end our transaction while looping over the messages.
```
bmorelli25, line 69: I'm confused here. The previous text only refers to a singular message, but now we're talking about multiple messages:

> First, when sending the message, we add the trace as one of the `MessageAttributes`

> Then, when receiving the messages, we start and end our transaction while looping over the messages.

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+      childOf:traceparent
+    })
+    /* ... process message ... */
+    transactionMessage.end()
+  }
+})
+
+----
+
+Unlike the first code sample, this approach will create a single transaction for each processed queue message.  This allows you to link each individual message with the transaction it started in. The <<apm-start-transaction,`startTransaction()`>> API allows _you_ to control how your queue data is reported to Elastic APM.
```
bmorelli25, line 98: I'm confused here too.

> Unlike the first code sample, ...

But the first example is about sending the message?

docs/message-queues.asciidoc
```diff
@@ -0,0 +1,98 @@
+[[message-quques]]
```
bmorelli25, line 1: typo

```suggestion
[[message-queues]]
```

bmorelli25 (commented):
Neato! I have some suggestions and questions.

CHANGELOG.asciidoc
```diff
@@ -36,6 +36,7 @@ Notes:
 [float]
 ===== Features
 
+* Adds support for Amazon SQS queues via `aws-sdk` instrumentation.
```
trentm, line 39: A link to the message-queues doc here would be very helpful, especially given there is some nuance to SQS instrumentation.

I haven't looked at the queue delay metrics yet, but I'm guessing it would be good to mention their being added in the changelog as well.

.tav.yml
```diff
@@ -353,3 +353,8 @@ body-parser:
   versions: '>=1.19.0'
   commands:
     - node test/sanitize-field-names/express.js
+
+aws-sdk:
+  versions: '>=2.858 <3'
```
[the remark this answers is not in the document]
astorm: Good question -- I picked the version based the the current release and then went back 15 versions.  Per the above, this was in order to limit the number of previous versions the TAV runner would need to churn through.  It looks like in the time between then and now there's been four releases.  

I've opened an issue to address this more broadly: https://github.com/elastic/apm-agent-nodejs/issues/2020

docs/message-queues.asciidoc
astorm: appologies -- looks like a copy/pasta error with the code sample.  I will correct

CHANGELOG.asciidoc
astorm: _nod_ that makes sense.  I've added a link to messaging spec and called out the queue metrics by name.

lib/instrumentation/modules/aws-sdk.js
```diff
@@ -0,0 +1,33 @@
...
+  if (request.service.serviceIdentifier === 'sqs') {
+    return instrumentationSqs(orig, origArguments, request, AWS, agent, { version, enabled })
+  }
+
+  // if we're still here, then we still need to call the original method
+  return orig.apply(request, origArguments)
+}
+
+// main entry point for aws-sdk instrumentation
+module.exports = function (AWS, agent, { version, enabled }) {
```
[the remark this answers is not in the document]
astorm: _nod_ -- good call.  I'd done a spot check and seen a direct guard used in some place and not used in others, but doing a more comprehensive checks show a guard here the rule, and not the exception.  Added a guard.

lib/instrumentation/modules/aws-sdk.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!semver.satisfies(version, '>1 <3')) {
+    agent.logger.debug('aws-sdk version %s not supported - aborting...', version)
+    return AWS
+  }
+
+  shimmer.wrap(AWS.Request.prototype, 'send', function (orig) {
+    return function _wrappedAWSRequestSend () {
+      return instrumentOperation(orig, arguments, this, AWS, agent, { version, enabled })
```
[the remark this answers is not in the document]
astorm: It's mainly uncertainty about what we'll use in the future. Both wanting to make sure the SQS instrumentation file has everything it _might_ need in the future, as well as make sure the additional services (SNS, S3) we'll add into `aws-sdk.js::instrumentOperation` will have everything they might need. 

I'm going to leave as is for now, but if your opinion here is strong let me know and I'm happy to change it.

lib/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -0,0 +1,202 @@
...
+}
+
+module.exports = {
+  instrumentationSqs,
+  getToFromFromOperation,
+  getActionFromRequest,
+  getQueueNameFromRequest,
+  getRegionFromRequest,
+  getMessagingDestinationContextFromRequest,
+  shouldIgnoreRequest
```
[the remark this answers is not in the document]
astorm: ✅ comment added

docs/message-queues.asciidoc
astorm: @bmorelli25 I think I see the confusion here. It was not copy/pasta as I originally thought.  

The SQS API _schedules_ a single message at a time.  However, when you're consuming messages, the API always returns an array of messages.  My original text assumed the reader had some experience with SQS queues already and would have internalized this. 

I've made some small adjustments to the section to address this, and the other place I introduced confusion.  If this is still a head scratcher lets sync up via slack or zoom to talk it out.

docs/message-queues.asciidoc
astorm: @bmorelli25 OK, I see the confusion here.  It was not copy/pasta as I originally thought.  I wasn't referring to the immediately previous code sample, I was referring to the code sample at the _top_ of this page.  

I've pulled this last paragraph in favor of some callouts.  I also decided that trying to communicate the difference in where a transaction is started might impact how spans and linked might be too much information for this doc. 

If there's still confusion around this lets sync up via slack or zoom to talk it out.

bmorelli25: Doh! That makes sense. Let me take another read-through.

bmorelli25 (commented):
Great clarifications in the docs. It's a 👍 from me.

test/agent.js
```diff
@@ -649,7 +649,7 @@ test('filters', function (t) {
         setTimeout(function () {
           t.end()
           server.close()
-        }, 50)
+        }, 200)
```
astorm, line 652: This test was failing in a race condition way without this change.  It's unclear if this change actually fixed the test, or if the internet weather on the GHA virtual machines changed.  Failures included below for posterity. 

```
not ok 241 plan != count
  ---
    operator: fail
    expected: 1
    actual:   0
    at: Timeout._onTimeout (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/test/agent.js:650:13)
    stack: |-
      Error: plan != count
          at Test.assert [as _assert] (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:304:54)
          at Test.bound [as _assert] (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:91:32)
          at Test.fail (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:398:10)
          at Test.bound [as fail] (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:91:32)
          at completeEnd (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:243:18)
          at next (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:217:13)
          at Test._end (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:236:5)
          at Test.bound [as _end] (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:91:32)
          at Test.end (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:191:10)
          at Test.bound [as end] (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/node_modules/tape/lib/test.js:91:32)
          at Timeout._onTimeout (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/test/agent.js:650:13)
          at ontimeout (timers.js:427:11)
          at tryOnTimeout (timers.js:289:5)
          at listOnTimeout (timers.js:252:5)
          at Timer.processTimers (timers.js:212:10)
  ...


```

docs/message-queues.asciidoc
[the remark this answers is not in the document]
astorm: ✅ 👍

astorm:
@trentm Re:

> 2. Decide whether to drop the HTTP span underneath an instrumented SQS action. This could be handled in a separate ticket... where perhaps it would be nicer to have a clearer discussion of the issue. This should link to, or perhaps could piggyback on https://github.com/elastic/apm-agent-nodejs/issues/2000 (the equivalent issue for `@elastic/elasticsearch` instrumentation).
> 3. Start a separate ticket for whether to add a "traceparent" SQS message attribute when **sending** a message.  I don't know if this is implied by `For distributed tracing, the SQS API has "message attributes" that can be used in lieu of headers.` from the spec. The spec should probably be clearer.

I've opened new issues for these items

https://github.com/elastic/apm-agent-nodejs/issues/2021

https://github.com/elastic/apm-agent-nodejs/issues/2022

Re: 
> 1. Fix the sync=true span.start()/end() creation in the send and receive instrumentation.

This was more of a journey than we (or at least I) originally thought when we spoke on the phone, but I believe it has been corrected. It looks like out standard pattern here is to use the `bindFunction` method of the Instrumentation object.  This allows us to tell the agent that a specific callback function should be associated with the span that's active when `bindFunction` gets called.

In addition to refactoring the instrumentation module to use `bindFunction` I've also got a first draft going that documents what `bindFunction` is for and how it can/should be used.  Current span relationships will look like this in the trace viewer.

sqs-receive-with-delete
sqs-send

lib/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -0,0 +1,193 @@
...
+  // our complete listener MUST be registered after
+  // calling the original Request.send.  Request.send
+  // invokes its callback function by registering its
+  // own complete callback. We need our complete callback
+  // to fire _after_ the one registered by Request.send
+  // see: https://github.com/aws/aws-sdk-js/blob/e5c12fad7ebc87b4de447d777628030be4792945/lib/request.js#L368
+  request.on('complete', function () {
+    span.end()
+  })
+  return origResult
```
trentm, line 180: From our discussion, here is the alternative take that just uses the 'complete' event handler:

- Some "XXX"s to handle with the captureError.
- This (arguably) fixes the span duration to *not* include the user's callback code.

```js
function instrumentationSqs2 (orig, origArguments, request, AWS, agent, { version, enabled }) {
  if (shouldIgnoreRequest(request, agent)) {
    return orig.apply(request, origArguments)
  }

  const type = TYPE
  const subtype = SUBTYPE
  const action = getActionFromRequest(request)
  const name = getSpanNameFromRequest(request)
  const span = agent.startSpan(name, type, subtype, action)
  span.setDestinationContext(getMessagingDestinationContextFromRequest(request))

  request.on('complete', (response) => {
    if (response && response.error) {
      // XXX Have to do the same skipOutcome + // span._setOutcomeFromErrorCapture()
      //     dance as in elasticsearch.js before the "current span" has moved
      //     to the HTTP child span.
      //     (Consider an option to pass the span to `agent.captureError()`)
      // XXX Consider also the same `captureAttributes: false` + explicit custom
      //     fields dance.
      agent.captureError(response.error)
    }

    // We know, from knowledge of the aws-sdk, that this is async.
    span.sync = false

    span.end()

    if (request.operation === 'receiveMessage' && response && response.data) {
      recordMetrics(getQueueNameFromRequest(request), response.data, agent)
    }
  })

  return orig.apply(request, origArguments)
}
```

lib/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -0,0 +1,193 @@
...
+  if (shouldIgnoreRequest(request, agent)) {
+    return orig.apply(request, origArguments)
+  }
+
+  const type = TYPE
+  const subtype = SUBTYPE
+  const action = getActionFromRequest(request)
+  const name = getSpanNameFromRequest(request)
+  const span = agent.startSpan(name, type, subtype, action)
+  span.setDestinationContext(getMessagingDestinationContextFromRequest(request))
```
trentm, line 154: Should we also be added `span.context.message.*` here, per https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-messaging.md#context-fields
`message.age.ms` and `message.queue.name`.

docs/index.asciidoc
```diff
@@ -39,6 +39,8 @@ include::./source-maps.asciidoc[]
 
 include::./distributed-tracing.asciidoc[]
 
+include::./message-queues.asciidoc[]
+
 include::./performance-tuning.asciidoc[]
```
trentm, line 44: There should be an addition to supported-technologies.asciidoc for SQS now.

test/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -0,0 +1,409 @@
...
+
+    request.params.QueueUrl = null
+    t.equals(getQueueNameFromRequest(request), 'unknown')
+    request.params.QueueUrl = 5
+    t.equals(getQueueNameFromRequest(request), 'unknown')
+    request.params.QueueUrl = 'foo/baz/bar'
+    t.equals(getQueueNameFromRequest(request), 'unknown')
+
+    request.params.QueueUrl = 'http://foo/baz/bar'
+    t.equals(getQueueNameFromRequest(request), 'bar')
```
trentm, line 101: Given the "URL" comment above, I would expect this one to fail on node 8.6.

test/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -0,0 +1,409 @@
...
+      })
+      const sqs = new AWS.SQS({ apiVersion: '2012-11-05' })
+      const params = getParams('sendMessage', listener.address().port)
+      sqs.sendMessage(params, function (err, data) {
+        t.error(err)
+        listener.close()
+      })
+    })
+  })
+
```
trentm, line 372: Would it be possible (I'm not sure with the mocking here) to test the promise-y AWS SDK usage (https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/using-promises.html)?

trentm (changes requested):

test/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -140,8 +140,15 @@ tape.test('AWS SQS: Unit Test Functions', function (test) {
     const agent = {
       _conf: {
         ignoreMessageQueuesRegExp: []
+      },
+      logger: {
+        trace: function () {
+        }
```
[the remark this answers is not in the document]
astorm: Good instinct, but no.  This is a mocked logger object.  Without this these tests crash when they hit this conditional branch. 

```
  if (!agent.currentTransaction) {
    agent.logger.trace('no active transaction found, skipping sqs instrumentation')
    return true
  }
```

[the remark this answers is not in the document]
trentm: Ah, could perhaps use `logger: logging.createLogger('off')` to get the full logging API.

lib/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// Parses queue/topic name from AWS queue URL
+function getQueueNameFromRequest (request) {
+  const unknown = 'unknown'
+  if (!request || !request.params || !request.params.QueueUrl) {
+    return unknown
+  }
+  try {
+    const url = new URL(request.params.QueueUrl)
```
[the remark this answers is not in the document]
astorm: > Does this mean that this is accidentally falling back to "unknown" for 8.6 tests unnecessarily?

Worse -- the tests weren't running in the CI due to the suite not being added in `test/test.js`.  Adding the `aws-sdk` folder resulted in the expected test failures, which will be addressed.

Also dinging the _maybe long term we should auto-scan the directories for tests_ bell. 🛎️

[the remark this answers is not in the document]
trentm: I have a side-project (timeboxing it to short bursts every day or so) going to switch to tap where discovering the test files will be `tap test/**/*.test.js`.

lib/instrumentation/modules/aws-sdk/sqs.js
astorm: Good catch.  We can support queue.name for sure -- but the message age will be less straight forward.I've opened an issue for future consideration here: https://github.com/elastic/apm-agent-nodejs/issues/2026

test/agent.js
astorm: Capturing this context in this issue: https://github.com/elastic/apm-agent-nodejs/issues/2027

docs/index.asciidoc
astorm: _nod_ added a 

```
|https://www.npmjs.com/package/aws-sdk[aws-sdk] |>1 <3 |Will instrument SQS send/receive/delete messages
```

to `supported-technologies.asciidoc`.

test/instrumentation/modules/aws-sdk/sqs.js
astorm: Per previous comments -- the `test/instrumentation/modules/aws-sdk/sqs.js` test file wasn't running in CI.  This has been corrected.  We saw Node 8 failures with the above test.  We corrected this by importing the `URL` symbol.

test/instrumentation/modules/aws-sdk/sqs.js
astorm: Good call -- using the promise API was generating HTTP spans, but not SQS spans.  The promise API bypasses the `send` method and instead [adds directly to that state-machine-like-thing](https://github.com/aws/aws-sdk-js/blob/3691e826f4e785d7ead9763f10d58fa098a4078d/lib/request.js#L797) that's under the hood.  I've added a wrapper for `promise` as well to catch this case.

test/instrumentation/span.js
```diff
@@ -240,7 +240,7 @@ test('#_encode() - with meta data', function myTest2 (t) {
     t.strictEqual(payload.type, 'bar')
     t.strictEqual(payload.timestamp, span._timer.start)
     t.ok(payload.duration > 0)
-    t.deepEqual(payload.context, { db: { statement: 'foo', type: 'bar' }, http: undefined, tags: { baz: '1' }, destination: undefined })
+    t.deepEqual(payload.context, { db: { statement: 'foo', type: 'bar' }, http: undefined, tags: { baz: '1' }, destination: undefined, message: undefined })
     assert.stacktrace(t, 'myTest2', __filename, payload.stacktrace, agent)
```
astorm, line 244: A small adjustment to this test was necessary to reflect the new structure of a span.

test/instrumentation/modules/aws-sdk/sqs.js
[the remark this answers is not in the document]
astorm: ✅ story checks out -- switched to the `logging.createLogger('off')` logger.

lib/instrumentation/modules/aws-sdk/sqs.js
astorm: I made all these change _except_ the captureAttributes one.  It feels like a premature optimization and it's uncertain which explicit custom fields would be the right ones. I intend to hold off on that for now

trentm: +1

I think the logic is backwards and that the agent should always be explicit about what fields to capture on the error. But that's a general and separate issue. Not for here.

astorm:
jenkins run the tests please

astorm:
@trentm the span strategy, `promise` tests and other nits have been addressed.  The CI failures are the integration tests [listed here](https://github.com/elastic/apm-integration-testing/issues/1095) and seemingly unrelated to these changes.

astorm:
jenkins run the tests please

lib/instrumentation/span.js
```diff
@@ -96,6 +97,10 @@ Span.prototype.setDestinationContext = function (context) {
   this._destination = Object.assign(this._destination || {}, context)
 }
 
+Span.prototype.setMessagingContext = function (context) {
```
trentm, line 100: In the function and var names you used "messaging" rather than the spec'd field name "message". Is there a reason for not having those be a match, to possibly assist with grepping?

The others (setDbContext, setHttpContext, setDestinationContext) all match the payload field name.

CHANGELOG.asciidoc
```diff
@@ -36,6 +36,10 @@ Notes:
 [float]
 ===== Features
 
+* Adds support for Amazon SQS queues via `aws-sdk` instrumentation that
+  partially implements the https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-messaging.md[APM messaging spec],
+  and adds `queue.latency.min.ms`, `queue.latency.max.ms`, and `queue.latency.avg.ms` 
+  metrics for SQS queues.
```
trentm, line 42: Q: How does a user view these? I'm showing my ignorance with metrics here.
Perhaps you could add a comment to the PR showing a Kibana screenshot looking at those metrics?

lib/instrumentation/modules/aws-sdk.js
```diff
@@ -0,0 +1,40 @@
+'use strict'
+const semver = require('semver')
+const shimmer = require('../shimmer')
+const { instrumentationSqs } = require('./aws-sdk/sqs')
+
+// Called in place of AWS.Request.send
```
trentm, line 6: nit: and .promise now as well.

lib/metrics/queue.js
```diff
@@ -0,0 +1,85 @@
...
+  const collector = new QueueMetricsCollector()
+  registry.registerCollector(collector)
+  for (const metric of Object.keys(collector.stats)) {
+    registry.getOrCreateGauge(
+      metric,
+      function returnCurrentValue () {
+        return collector.stats[metric]
+      },
+      { queue_name: queueOrTopicName },
+      10
```
trentm, line 76: This hardcoded 10 is causing queue metrics to get reported every 10s regardless of the value of `metricsInterval`. E.g. I set my conf to metricsInterval:30 and I get the queue metrics every 10s, and the rest of the metrics every 30s.

trentm (changes requested):

lib/metrics/queue.js
```diff
@@ -0,0 +1,85 @@
...
+    this.total += time
+  }
+
+  // standard `collect` method for stats collectors
+  //
+  // called by the MetricReporter object prior to its sending data
+  //
+  // @param {function} cb callback function
+  collect (cb) {
+    // update average based on count and total
```
trentm, line 40: Given https://github.com/elastic/apm-agent-nodejs/issues/1977 (and its root https://github.com/elastic/kibana/pull/90955) should we be:

1. Avoiding reporting metrics for this period if count==0? Currently, once a queue metric is reported for a particular queue, it will be reported every interval, even if there is no further activity on that queue.
2. Consider changing `const queueMetics = new Map()` in sqs.js to be a data structure that evicts entries if they haven't been used in a long time. Currently if a long running app cycles through lots of queue names over time, then we'll add (small) mem usage for those over time.  This data structure could be accomplished via https://github.com/isaacs/node-lru-cache (using max=Infinity, and maxAge=...some timeout value...).

The former (1) is a bigger issue.

For example, I setup my app to use queue "trentm-play-sqs3" for a while, then stop using and use queue "trentm-play-sqs2". The reported metrics every interval show empty metrics for the now unused "trentm-play-sqs3":

```
    {"metricset":{"samples":{"queue.latency.min.ms":{"value":825300},"queue.latency.max.ms":{"value":3382475},"queue.latency.avg.ms":{"value":1163652.6875}},"timestamp":1617651611908000,"tags":{"queue_name":"trentm-play-sqs2","hostname":"pink.local","env":"development"}}}
    {"metricset":{"samples":{"queue.latency.min.ms":{"value":0},"queue.latency.max.ms":{"value":0},"queue.latency.avg.ms":{"value":0}},"timestamp":1617651611908000,"tags":{"queue_name":"trentm-play-sqs3","hostname":"pink.local","env":"development"}}}
```

lib/instrumentation/span.js
astorm: Hmm, that's an interesting point.  This wasn't a conscious choice but (guessing) I'd say it fell our of the span's type being specced as `messaging` and not `message`.  This feels like a six of one situation (there's grep considerations on both side with any [gerund](https://www.dictionary.com/browse/gerund)), but it's a quick change.  Done: https://github.com/elastic/apm-agent-nodejs/commit/6c88f7fa64babaa69abb3ca9898f734f9cb6e6e5

lib/instrumentation/modules/aws-sdk.js
astorm: ✅ https://github.com/elastic/apm-agent-nodejs/commit/d99a68f64fc4b61a7948966da03107448bd6255d

CHANGELOG.asciidoc
astorm: Good question -- it looks like metrics come to rest in  the `apm-[version]-metric-...` elasticsearch indices, which should be query-able via the `apm*` or `apm-*` kibana index pattern. Screenshot added the the PR description above.

lib/metrics/queue.js
astorm: _nod_ fair.  I'll remove the hard coded value and have opened an issue for us to consider whether having the ability to configure individual reporting periods for metrics is worth while. https://github.com/elastic/apm-agent-nodejs/issues/2031

lib/metrics/queue.js
astorm: Good questions -- I see three.  Working backwards: 

Re: `node-lru-cache` or something similar -- that feels like a premature optimization.  A user would need to have a significant number of queues running and instrumented for this to be noticeable.  If node had a built-in data structure that expired the least recently used items or if the agent already has a system for doing this I'd say we go for it now. I don't think (at this time) it's worth the cost of a new third party dependency, or building out or own "expire the old stuff" memory cache. If we see evidence SQS instrumentations create high memory problems we can/should revisit.

astorm: Re: skipping a metric if the count is zero -- that's another good question without a clear win tradeoff wise.  To take the other side of the argument a measurement of 0 is still a measurement that might provide insight into how a system is running, and the absence of a metric isn't necessarily a value of 0.  If we _are_ going to change the behavior of reporting 0-value metrics, it seems like the right place to do that is when metrics [are serialized and reported](https://github.com/elastic/apm-agent-nodejs/blob/a2a6152c8009887efec4b10e29f4ac47b7d7dab2/lib/metrics/reporter.js#L18), and I'd prefer to do it separate from this PR.

astorm: Re: the high cardinality problem, we're less susceptible that that here.  The problem Kibana ran into is they had [a slightly pathological case](https://elastic.slack.com/archives/C5M02D1GE/p1612974533191700?thread_ts=1612965746.189300&cid=C5M02D1GE) 

> e.g. there’s an endpoint that’s basically `/:buildHash/bundles/:plugin/:filename`, which is called for every asset loaded from the browser, and per file/build hash/plugin it generates a unique transaction name

_Every_ frontend asset served (in dev mode)

    ex. https://kibana.example.com:9243/39309/bundles/plugin/mapsLegacy/mapsLegacy.chunk.1.js
    
was creating a unique transaction name when the file was first served.  I presume that's because, in dev mode, the static files are actually served/bundled by node.js on the fly as requests for them come in.  On the home page this can be _literally_ hundreds of different transaction names.  

This bad problem was made worse by [the transaction breakdown metrics](https://github.com/elastic/apm-agent-nodejs/blob/a2a6152c8009887efec4b10e29f4ac47b7d7dab2/lib/instrumentation/transaction.js#L276).  These metrics include transaction name as a metric dimension (tag/label).  So hundreds, possibly thousands, of unique metric sets were created and then stuck around.

The only extra metric dimension/label/tag we use here is queue name.  It's important to capture this metric as it allows filtering.  A single application/service would need to be a very large number of queues for this to become a perf. issue.  So, again, I think the right trade-off here is to launch for the 80% case, and if the 20% cases start rolling in we can address it then.  Regardless of when we address it, it feels like a separate PR that targets metric reporting generally.

astorm:
jenkins run the tests please

trentm (approved):

Issue #2023, opened 2021-03-30
Set Sample Rate on Transactions and Spans


astorm:
The transaction sample rate must be set on both transactions and spans via a `sample_rate` property, per the [`sample_rate`](https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md).

In addition to being spec compliant, we believe this will solve https://github.com/elastic/kibana/issues/95693

This work is being done to ensure that dependencies show up on the 

    Observability -> APM -> [Service] -> [Overview Tab]

page.

PR #2038, opened 2021-04-08
feat: add sample_rate to spans and transactions

astorm:
Closes: #2023

This PR finishes [the incomplete tracing-sampling](https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md) spec by ensuring the sample rate value stored in the tracestate is also set on the final serialized spans and transactions. 

Preexisting Functionality: 

- A "sampled" transaction (also called "recorded") is one where spans have been generated
- A root level transaction is the first transaction in a distributed trace
- The agent makes a sampling decision for root level transactions, and the resulting boolean is stored in the `traceparent` header
- Non-root transactions use the `traceparent` header to make their sampling decision
- The sampling rate the agent used to determine if a root transaction was sampled or not is stored on the `tracestate` header
- The sample rate from `tracestate` is never used to make a sampling decision. We're only propagating it so its value can be sent to APM Server and used in metrics calculations

New Functionality:

- We've added a `sample_rate` property to transactions and spans that duplicates the value in the `tracestate` header
- There are certain scenarios, outlined [in the spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md), where `sample_rate` is not set, or `sample_rate` is set to zero


lib/instrumentation/generic-span.js
```diff
@@ -67,6 +67,17 @@ Object.defineProperty(GenericSpan.prototype, 'sampled', {
   }
 })
 
+Object.defineProperty(GenericSpan.prototype, 'sample_rate', {
+  enumerable: true,
+  get () {
+    const rate = parseFloat(this._context.tracestate.getValue('s'))
```
astorm, line 73: If it's a legitimate number, return the number.  Otherwise return 0.

lib/instrumentation/transaction.js
```diff
@@ -161,6 +161,23 @@ Transaction.prototype.toJSON = function () {
     }
   }
 
+  // add sample_rate to transaction
```
astorm, line 164: This is the edge case handling for places where we're not just passing on the sample rate. (unsampled MUST have a sample_rate of 0, invalid/empty trace state, etc.).  Tests for these edge cases start here: https://github.com/elastic/apm-agent-nodejs/pull/2038/files#diff-800eab1f0e0cb3be24e7db4feea6cd572dc4ccbce5caf260923b1828ea3ff290R47

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
+const agent = require('../..').start({
```
trentm, line 1: 'use strict' at the top?

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+// @return {Transaction}
+function startUnSampledTransaction () {
+  let transaction = { sampled: true }
+  let guardCount = 0
+  while (transaction.sampled && guardCount < 10000) {
+    transaction = agent.startTransaction('foo', 'bar', 'baz', 'bing')
+    guardCount++
+  }
+  if (guardCount >= 10000) {
+    throw new Error('startUnSampledTransaction could not start a sampled transaction')
```
trentm, line 215:
```suggestion
    throw new Error('startUnSampledTransaction could not start an unsampled transaction')
```

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+const suite = require('tape')
+
+suite('Sample Rate Propagation', function (test) {
+  const TRACEPARENT_RECORDED = '00-2b9e333caa56c38e6e4d682f3184da23-a6bea2be0b62fb54-01'
+  const TRACEPARENT_NOTRECORDED = '00-c054b05d8262aaca165ed62e32898f55-4fcce7bf3ec36951-00'
+
+  test.test('sample rate is set', function (t) {
+    agent._conf.transactionSampleRate = 0.499
+    // const transaction = agent.startTransaction('foo', 'bar', 'baz', 'bing', {
+    // })
```
trentm, line 19: nit: Debugging code to remove?

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+    })
+    const span = transaction.startSpan('foo')
+    t.equals(transaction.sample_rate, 0, 'sample rate set')
+    t.ok(!span, 'span not started')
+    t.end()
+  })
+
+  // test that a sampled/recorded root transaction produces samples and
+  // transactions with a sample_rate equal to the sample rate
+  test.test('recorded root transactions have a sample rate of 0', function (t) {
```
trentm, line 63: copypasta: Update this test name to match what is tested.

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+    const transaction = agent.startTransaction('foo', 'bar', 'baz', 'bing')
+    const span = transaction.startSpan('foo')
+    t.equals(transaction.sample_rate, 0, 'sample rate set')
+    t.ok(!span, 'no span for unsampled transactions')
+    t.end()
+  })
+
+  // test that an unsampled transaction from a non-zero sample rate
+  // still produces a final serialized span with a sample rate of 0
+  test.test('recorded root transactions have a sample rate of 0', function (t) {
```
trentm, line 85: copypasta: Update this test name to match what is tested.

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+    const span = transaction.startSpan('foo')
+    const serialized = transaction.toJSON()
+    t.equals(serialized.sample_rate, 0, 'serialized sample rate is zero')
+    t.ok(!span, 'no span for unsampled transactions')
+    t.end()
+  })
+
+  // Test that an invalid tracestate in a recorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
```
trentm, line 100: copypasta: Update this test name to match what is tested.

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+      t.equals(transactionSerialized.sample_rate, undefined, 'serialized transaction should have no sample_rate')
+      t.equals(spanSerialized.sample_rate, undefined, 'serialized span should have no sample_rate')
+
+      t.end()
+    })
+  })
+
+  // Test that an invalid tracestate in an unrecorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
```
trentm, line 144: copypasta: Update this test name to match what is tested.

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+    t.equals(serialized.sample_rate, 0, 'serialized sample rate is zero')
+    t.ok(!span, 'no span for unsampled transactions')
+    t.end()
+  })
+
+  // Test that an invalid tracestate in a recorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
+    // sort of gross while shenanagins to let us get an unsampled
+    // transaction with a non-zero sample rate
```
trentm, line 102: Is this comment copypasta as well? It says "unsampled transaction", but this is a sampled one.

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+
+      t.end()
+    })
+  })
+
+  // Test that a blank tracestate in a recorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
+    // sort of gross while shenanagins to let us get an unsampled
+    // transaction with a non-zero sample rate
```
trentm, line 124: copypasta?

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+
+      t.end()
+    })
+  })
+
+  // Test that an invalid tracestate in an unrecorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
+    // sort of gross while shenanagins to let us get an unsampled
+    // transaction with a non-zero sample rate
```
trentm, line 146: Copypasta?

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+    t.equals(transactionSerialized.sample_rate, undefined, 'serialized transaction should have no sample_rate')
+    t.ok(!span, 'no span for unsampled transaction')
+    t.end()
+  })
+
+  // Test that a blank tracestate in an unrecorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
+    // sort of gross while shenanagins to let us get an unsampled
+    // transaction with a non-zero sample rate
```
trentm, line 162: Copypasta?

test/tracecontext/sample_rate.js
```diff
@@ -0,0 +1,218 @@
...
+    const span = transaction.startSpan('foo')
+    const transactionSerialized = transaction.toJSON()
+    t.equals(transactionSerialized.sample_rate, undefined, 'serialized transaction should have no sample_rate')
+    t.ok(!span, 'no span for unsampled transaction')
+    t.end()
+  })
+
+  // Test that a blank tracestate in an unrecorded transaction
+  // results in a serialized span without a sample_rate, per the spec.
+  test.test('blank tracestate, recorded', function (t) {
```
trentm, line 160: copypasta: Update this test name to match what is tested.

test/tracecontext/sample_rate.js
astorm: ✅

test/tracecontext/sample_rate.js
astorm: ✅

test/tracecontext/sample_rate.js
astorm: ✅

test/tracecontext/sample_rate.js
astorm: ✅

test/tracecontext/sample_rate.js
astorm: ✅

lib/instrumentation/span.js
```diff
@@ -219,6 +219,11 @@ Span.prototype._encode = function (cb) {
       outcome: self.outcome
     }
 
+    // if a valid sample rate is set (truthy or zero), set the property
+    if (self.sample_rate || self.sample_rate === 0) {
+      payload.sample_rate = self.sample_rate
+    }
```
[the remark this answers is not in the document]
astorm: replaced with a `getTraceContextSampleRate` method, pending discussion on Tuesday.

lib/instrumentation/transaction.js
```diff
@@ -161,6 +161,18 @@ Transaction.prototype.toJSON = function () {
...
+  // https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md
+  // Only set sample_rate on transaction payload if a valid trace state
+  // variable is set.
+  //
+  // "If there is no tracestate or no valid es entry with an s attribute,
+  //  then the agent must omit sample_rate from non-root transactions and
+  //  their spans."
+  if (this.sample_rate || this.sample_rate === 0) {
+    payload.sample_rate = parseFloat(this.sample_rate)
+  }
```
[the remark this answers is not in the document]
astorm: replaced with a `getTraceContextSampleRate` method, pending discussion on Tuesday.  Also, re: the `parseFloat` -- good catch.  That was left over from a more complex version of this.

test/tracecontext/sample_rate.js
astorm: 🤦 when the comment is right but the test name is 🍝 , fixed

test/tracecontext/sample_rate.js
astorm: 🤦 when the comment is right but the test name is 🍝 , fixed

test/tracecontext/sample_rate.js
astorm: it was plain pasta, but a stale comment from a more complicated version.  Removing.

test/tracecontext/sample_rate.js
astorm: 🤦 when the comment is right but the test name is 🍝 , fixed

test/tracecontext/sample_rate.js
astorm: 🤦 when the comment is right but the test name is 🍝 , fixed

test/tracecontext/sample_rate.js
astorm: decrufted

astorm:
@trent per offline discussion, we're avoiding calling the property/getter multiple times in the same function and have renamed it to `sampleRate`, and have de-carbed all that test name copy/pasta. Test are passing, but back up for review.

test/tracecontext/sample_rate.js
trentm: I meant this (and a few other) "sort of gross while shenanagins  ..." comment. That feels like it was a copypasta comment from before you factored out `start(Un)SampledTransaction()`

lib/instrumentation/generic-span.js
```diff
@@ -67,6 +67,17 @@ Object.defineProperty(GenericSpan.prototype, 'sampled', {
   }
 })
 
+Object.defineProperty(GenericSpan.prototype, 'sampleRate', {
+  enumerable: true,
+  get () {
+    const rate = parseFloat(this._context.tracestate.getValue('s'))
+    if (rate || rate === 0) {
```
trentm, line 74: 1. Should this discard values outside of `[0, 1]` as invalid?  The *comments* in apm-server (https://github.com/elastic/apm-server/blob/b92191369f58c939ac894b478f1bb6385c07dd63/model/modeldecoder/v2/model.go#L768-L771) and the spec suggest it should be in that range, though it doesn't look to me that apm-server validates that.

2. nit: Also perhaps use `if (!isNaN(rate)) {`  to guard the result from `parseFloat()`.

lib/instrumentation/span.js
```diff
@@ -219,6 +219,12 @@ Span.prototype._encode = function (cb) {
       outcome: self.outcome
     }
 
+    // if a valid sample rate is set (truthy or zero), set the property
+    const sampleRate = self.sampleRate
+    if (sampleRate || sampleRate === 0) {
```
trentm, line 224: nit: Perhaps use `if (sampleRate !== null)` given that the sampleRate getting returns `null` for an invalid value.

lib/instrumentation/transaction.js
```diff
@@ -161,6 +161,19 @@ Transaction.prototype.toJSON = function () {
...
+  // add sample_rate to transaction
+  // https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md
+  // Only set sample_rate on transaction payload if a valid trace state
+  // variable is set.
+  //
+  // "If there is no tracestate or no valid es entry with an s attribute,
+  //  then the agent must omit sample_rate from non-root transactions and
+  //  their spans."
+  const sampleRate = this.sampleRate
+  if (sampleRate || sampleRate === 0) {
```
trentm, line 173: nit: Perhaps use `if (sampleRate !== null)` given that the sampleRate getting returns `null` for an invalid value.

lib/instrumentation/generic-span.js
trentm: FWIW, I'm fine punting the range check to a separate issue .

astorm: Longer term it's probably worth thinking about where/when values that could be validated should be validated. 

For now, validating this value here seems fine, as it also ends up protecting against `NaN`.

```
    if(rate >=0 || rate <= 1) {
      return rate
    }
```

lib/instrumentation/transaction.js
astorm: ✅

lib/instrumentation/generic-span.js
```diff
@@ -71,7 +71,7 @@ Object.defineProperty(GenericSpan.prototype, 'sampleRate', {
   enumerable: true,
   get () {
     const rate = parseFloat(this._context.tracestate.getValue('s'))
-    if (rate || rate === 0) {
+    if (rate >= 0 || rate <= 1) {
```
trentm, line 74:
```suggestion
    if (rate >= 0 && rate <= 1) {
```

You want `&&` here, I think.

lib/instrumentation/span.js
trentm, line 224:
```suggestion
    if (sampleRate !== null) {
```

Do the !== null check here as well?

trentm (changes requested):

trentm (approved):

PR #2039, opened 2021-04-10
fix: span url hostname not ip

### LINKED ISSUE #2035: Hostname to IP Address Resolution in http-request-to-url can Overwhelm Service Maps
We've received reports that the "Hostname to final IP Address Resolution" for outgoing span URLs  behavior of the agent can overwhelm service maps when a service url has a dynamic IP address.  

In one specific case, requests made to the Amazon S3 API can result in over 100 different IP addresses being reported, resulting in a service map that looks something like this

service-map

However, this is not a problem restricted to the AWS SDK.  This will happen for any service URL that doesn't have a fixed IP address.  

We should reconsider whether resolving the hostname to its final IP destination is the right default behavior for instrumentation on outgoing HTTP spans. 

## Technical Details

This behavior stems from how we're setting the URL portion of the span's HTTP Context.  

https://github.com/elastic/apm-agent-nodejs/blob/8d4f10da32ccc16704c6a80ad526613c6cce98b8/lib/instrumentation/http-shared.js#L212-L216

The `url` variable is set by the `http-request-to-url` module

https://github.com/elastic/apm-agent-nodejs/blob/8d4f10da32ccc16704c6a80ad526613c6cce98b8/lib/instrumentation/http-shared.js#L6

https://github.com/elastic/apm-agent-nodejs/blob/8d4f10da32ccc16704c6a80ad526613c6cce98b8/lib/instrumentation/http-shared.js#L195

The `http-request-to-url` module waits for the socket's `connect` event, and extract the `remoteAddress` and `remotePort` from the socket object, and uses these to create a URL.

astorm:
Fixes #2035

This PR add a `getUrlFromRequestAndOptions` function to the `lib/instrumentation/http-shared.js` module.  The `http-shared` module also uses this new function in place of the old `http-request-to-url` module to set the `url` property on a span.

The `getUrlFromRequestAndOptions` function accepts a `ClientRequest` object and a generic options object and uses the information from those objects to return the same final URL that the ClientRequest will use.  The options object is either 

1. The options object passed to the `request` method
   ```
    http.request(options)
   ```
       
2. Or, in the case of `http.request`'s second form    
   ```
    http.request(url, options)
   ```
   
   the `options` object [is merged with the URL information from `url`](https://github.com/elastic/apm-agent-nodejs/blob/842be710cbd8736bfff0867e2413ebb7cb0d229a/lib/instrumentation/http-shared.js#L146) to create a new object.
   
In addition to the usual test suite, this was tested across all supported node versions with [the following sludge-hammer of a bash script](https://gist.github.com/astorm/35ac16d0e2fb71b465f20381197e446f).


astorm:
Tests failing in `test/instrumentation/modules/http/outgoing.js` for nodes 12.0 and 14.0, but passing in v14.14 or v12.22.1.  Likely due to some subtly of localhost vs. 127.0.0.1.  Will investigate.

lib/instrumentation/http-shared.js
```diff
@@ -260,3 +257,40 @@ function isAWSSigned (opts) {
...
+  if (!req) {
+    return false
+  }
+  options = options || {}
+  req = req || {}
+  req.agent = req.agent || {}
+
+  const port = options.port ? `:${options.port}` : ''
+  const host = req.host || options.hostname || options.host || 'localhost'
+  const protocol = req.protocol || req.agent.protocol
```
trentm, line 284: Would it be worth a comment (or overkill?) to note that req.host and req.protocol are only available in v14.5.0, v12.19.0 and later?

Also, `req.agent` isn't actually a documented attributes, but I agree this is the way to get to the protocol as a fallback. ... *or* we pass in the protocol given that we know if we are instrumenting `http.request` or `https.request`.

lib/instrumentation/http-shared.js
```diff
@@ -260,3 +257,40 @@ function isAWSSigned (opts) {
...
+// to http.request if the caller uses the the http.request(url,options,...)
+// method signature. The agent normalizes the url and options into a single
+// options array. This function expects those pre-normalized options.
+//
+// @param {ClientRequest} req
+// @param {object} options
+// @return string
+function getUrlFromRequestAndOptions (req, options) {
+  if (!req) {
+    return false
```
trentm, line 276: Returning false here results in `"url":false` in the span. E.g.:

```
{"span":{"name":"GET httpbin.org/Get","type":"external",..."duration":190.626,"context":{"http":{"method":"GET","status_code":404,"url":false}}...
```

lib/instrumentation/http-shared.js
```diff
@@ -260,3 +257,40 @@ function isAWSSigned (opts) {
...
+  req.agent = req.agent || {}
+
+  const port = options.port ? `:${options.port}` : ''
+  const host = req.host || options.hostname || options.host || 'localhost'
+  const protocol = req.protocol || req.agent.protocol
+  let url
+  try {
+    url = new URL(`${protocol}//${host}${port}${req.path}`)
+  } catch (e) {
+    // console.log(e)
```
trentm, line 289: Debug log to remove?

lib/instrumentation/http-shared.js
```diff
@@ -260,3 +257,40 @@ function isAWSSigned (opts) {
...
+    return false
+  }
+
+  return `${url.protocol}//${url.host}${url.pathname}${url.search}`
```
trentm, line 293: Why the step to `new URL`-parse the values we have, and then reconstruct it to a string again?

trentm (changes requested):

lib/instrumentation/http-shared.js
trentm: BTW, forget the "... or we pass in the protocol" part. That's just me talking out loud. I'd only bother with this *if* we later go the route of trying to reconstruct the url from *just the given arguments*.

trentm: Okay, I *do* have a baroque example of breaking this:
- if using a version of node before `req.protocol` was added
- and if passing in a custom `options.agent` that implements `function addRequest` (as required), but does *not* provide `agent.protocol`.

What happens currently is the `try { ... }` block below attempts to parse, e.g.: `undefined//localhost/get?foo=bar`, which throws and we return the sentinel (currently `false`).  I think that's reasonable for now. If we hear a complain, we can talk about passing in the protocol. 

Or we *could* pass in `moduleName` which we have already as the protocol.

lib/instrumentation/http-shared.js
astorm: ✅ removing

lib/instrumentation/http-shared.js
astorm: The comments are a good idea, I'll add them.

Also, since we already have the moduleName, passing it into `getUrlFromRequestAndOptions` and using it as a fallback value is a quick win.  I'll make these changes.

lib/instrumentation/http-shared.js
astorm: Per discussion elsewhere, this is to remove the username:password portion of a hostname.  We rely on the URL object's parsing, and then reconstruct the URL to be just the parts we want.

lib/instrumentation/http-shared.js
astorm: switched this up from false to undefined to ensure its dropped from the serialized span.

astorm:
Per discussions, we've removed the "pass everything through a `URL` object to sanitize" in order to simplify the function (and adjusted the final return value to use local variables instead of properties from that URL object).  We've also added a  parameter to the function that lets us pass in the `moduleName` to use as a fallback protocol.  In tests we've moved the `username:password`  to the `auth` section of the options and added some test assertions that check for the presence of either string in the final URL. We've also added a test to ensure that port 80, if supplied, is represented in the URL.

lib/instrumentation/http-shared.js
```diff
@@ -260,3 +256,47 @@ function isAWSSigned (opts) {
...
+  const port = options.port ? `:${options.port}` : ''
+  // req.host and req.protocol are node versions v14.5.0/v12.19.0 and later
+  const host = req.host || options.hostname || options.host || 'localhost'
+  const protocol = req.protocol || req.agent.protocol || fallbackProtocol
+  // let url
+  // try {
+  //   url = new URL(`${protocol}//${host}${port}${req.path}`)
+  // } catch (e) {
+  //   return undefined
+  // }
```
trentm, line 297: Drop these.

lib/instrumentation/http-shared.js
```diff
@@ -191,13 +189,11 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
         return emit.apply(req, arguments)
       }
 
-      let url, statusCode
-      httpRequestToUrl(req).then(_url => {
-        url = _url
-      }).catch(() => {
+      const url = getUrlFromRequestAndOptions(req, options, moduleName)
```
trentm, line 192:
```suggestion
      const url = getUrlFromRequestAndOptions(req, options, moduleName + ':')
```

Need the colon, otherwise you get something like: `http//localhost/get` in the *very* special circumstances where `fallbackProtocol` is used. E.g. with node 10 or less, using this:

```
http.get(
  {
    hostname: 'localhost',
    path: '/get',
    // A custom agent that implements the minimum to pass muster, but does
    // *not* define `agent.protocol`.
    agent: {
      addRequest() {}
    }
  }
)
```

trentm (changes requested):

lib/instrumentation/http-shared.js
astorm: got it -- I've made this change, and added the test I should have to test this particular case.

trentm (approved):

Issue #1916, opened 2020-12-15
Add filtering support for metadata payload


joshdover:
In Kibana's APM instrumentation, we are now capturing stats from developer's local environments. One issue is that we cannot filter some PII such as the developer's username from the path that is contained in the `process.args` array in the metadata that is sent along with APM traces:

![image](https://user-images.githubusercontent.com/1813008/102268563-ba09e200-3ed8-11eb-9935-59baf65fdf23.png)

Currently, we're including a notice that some personal data may be captured with an option for disabling this. Ideally, we'd be able filter this out similar to how we can filter out transaction data.

trentm:
Some quick notes and a first hack:

1. patch to apm-nodejs-http-client.git

```diff
diff --git a/index.js b/index.js
index c27fb71..2bb3a0d 100644
--- a/index.js
+++ b/index.js
@@ -495,7 +495,9 @@ function onStream (client, onerror) {

     // All requests to the APM Server must start with a metadata object
     if (!client._encodedMetadata) {
-      client._encodedMetadata = client._encode({ metadata: client._conf.metadata }, Client.encoding.METADATA)
+      // XXX HACK: This should be done *once*, but as late as possible.
+      var filteredMetadata = client._conf.metadataFilters.process(client._conf.metadata)
+      client._encodedMetadata = client._encode({ metadata: filteredMetadata }, Client.encoding.METADATA)
     }
     stream.write(client._encodedMetadata)
   }
```

2. patch to apm-agent-nodejs.git

```
diff --git a/lib/agent.js b/lib/agent.js
index cd75056..56038c1 100644
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -42,6 +42,7 @@ function Agent () {
   this._errorFilters = new Filters()
   this._transactionFilters = new Filters()
   this._spanFilters = new Filters()
+  this._metadataFilters = new Filters()
   this._transport = null

   this.lambda = lambda(this)
@@ -221,6 +224,12 @@ Agent.prototype.addFilter = function (fn) {
   this.addErrorFilter(fn)
   this.addTransactionFilter(fn)
   this.addSpanFilter(fn)
+  // XXX Decide if this would be a breaking change. For example the *default
+  // filter example in the docs* will break because it assumes
+  // `payload.context`. However that should *already* break for filtering
+  // error payloads.
+  // https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-add-filter
[...]

PR #2045, opened 2021-04-15
feat: add apm.addMetadataFilter(fn)

trentm:
Fixes: #1916

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

Note: still draft until the http-client PR (https://github.com/elastic/apm-nodejs-http-client/pull/154) and package.json is updated.


lib/config.js
```diff
@@ -279,6 +279,7 @@ class Config {
           globalLabels: maybePairsToObject(conf.globalLabels),
           hostname: conf.hostname,
           environment: conf.environment,
+          metadataFilters: agent._metadataFilters,
```
astorm, line 282: I don't see where/how this is used.  There's no `agent._metadataFilters` property assigned anywhere else -- `addMetadataFilter` adds its functions directly to the transport/client object.  Old code, or something else?

astorm (commented):
Overall looks good.  The new API method does what it says it does and has reasonable guards in place.  The new docs are clear to me, and the caveats added to the other filter functions make sense.  

One question about a lingering object property before approving, but otherwise 👍

lib/config.js
trentm: Good catch. Cruft left over from an earlier impl. Dropped in commit 26b897a.

astorm (approved):
👍

PR #2047, opened 2021-04-15
chore: silence a warning from 'docker-compose --no-ansi ...' usage

trentm:
The message is:
    --no-ansi option is deprecated and will be removed in future versions. Use `--ansi never` instead.

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -23,7 +23,7 @@ fi
 
 if [ -x "$(command -v docker-compose)" ]; then
   docker-compose \
-    --no-ansi \
+    --ansi never \
     --log-level ERROR \
     -f .ci/docker/docker-compose-all.yml \
     pull --quiet --ignore-pull-failures
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -51,7 +51,7 @@ NODE_VERSION=${NODE_VERSION} \
 TAV_VERSIONS=${TAV_VERSIONS} \
 USER_ID="$(id -u):$(id -g)" \
 docker-compose \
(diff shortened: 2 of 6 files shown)


trentm:
I get this warning from the docker-compose I have on my dev machine:

```
% docker-compose --version
docker-compose version 1.29.0, build 07737305
```

But CI has docker-compose version:

```
[2021-04-15T23:19:54.810Z] + docker-compose --version
[2021-04-15T23:19:55.387Z] docker-compose version 1.25.5, build 8a1c60f6
```

which doesn't support the newer `--ansi never` option, so until CI's docker-compose is updated, I'll just have to live with the warning. :)

* * * 

@elastic/observablt-robots A question for you: What defines what version of docker-compose is used in the CI containers? I'm guessing, but is this? https://github.com/elastic/infra/blob/master/ansible/roles/docker/defaults/main.yml#L6

I think the new "--ansi" flag was added in docker-compose v1.28.0 (https://docs.docker.com/compose/release-notes/#1280), but I don't think it is worth upgrading docker-compose just for this warning. However, let me know if you think it is.

kuisathaverat:
>@elastic/observablt-robots A question for you: What defines what version of docker-compose is used in the CI containers? I'm guessing, but is this?

the workers have a version that is updated when it is requested, these changes affect all projects so we have to be sure we will not break anything. However, docker-compose is something we can install in a Python virtual environment so we can use whatever version we need on any project easily. If you want it you only have to request it and we will manage to modify the pipeline to give you the docker-compose version you need.

trentm:
Understood, thanks. I don't want to bother updating docker-compose just for this case (avoiding a warning).

PR #1025, opened 2019-04-19
ci(jenkins): add tav tests

kuisathaverat:
* Move YAML files to .ci folder
* Add tav test
* Add .ch to .npmignore file
* Script to convert tap tests output to junit - .ci/convert_tap_to_junit.sh
* Script to lint commit messages - .ci/lint-commits.sh
* Script to launch test and tav tests - .ci/test.sh
* Script to perfom basic checks - .ci/test_basic.sh

PR #2048, opened 2021-04-15
tests: drop lingering support for junit output

trentm:
Usage of the "JUNIT" envvar by CI to get the test suite to product
junit-format output was dropped in #1025. These days tap->junit
conversion is done by the tap2Junit function from apm-pipeline-library.

@kuisathaverat Can you sanity check that my understanding is correct here?


kuisathaverat:
Does this generate TAP output files `*-output.tap`? if so the test result will be in Jenkins UI

```
  if [ -z "$COVERAGE" ]
  then
    node test/test.js
  else
    nyc node test/test.js
  fi
```

kuisathaverat (approved):

PR #2059, opened 2021-04-22
chore: drop unused createFilter function

trentm:
This was added in commit 4cbe2f9e85f3d23808ddbec4e2833205c4550ae8,
likely in early revs of its PR, but never used.


astorm (approved):
👍

Issue #2056, opened 2021-04-21
Perf Work: Investigate TraceContext/TraceState


astorm:
We have a few indicators that the performance of our TraceState implementation could be improved.  Flame graphs show an abnormal amount of CPU usage, and the overhead benchmarks showed a 20 - 40 microsecond bump after landing this work.

<img width="1874" alt="Screen Shot 2021-04-21 at 10 40 00 AM" src="https://user-images.githubusercontent.com/184372/115597999-c5a65480-a28e-11eb-9d3e-b33ba0d44663.png">

This Issue represents the work to 

1. Compare a No-Op tracestate implementation with our Kibana demo journey to determine if these perf issue effect our load time statistics

2. Investigate alternate implementations of TraceState, including the [Open Telemetry implementation](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-core/src/trace/TraceState.ts)

astorm:
Results of benchmarking with runs that include

- A kibana with the agent disabled
- A kibana with the agent enabled
- A kibana with the agent enabled, but a "no op/low op" `TraceState` class that does no work

Over the response times numbers (from the Gatling global information section of the report) looked like this 
```
╔════════════════════════╤══════╤═══════╤═══════╤═══════╗
║                        │ 50th │ 75th  │ 95th  │ 99th  ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
║ Agent Disabled         │ 41ms │ 112ms │ 345ms │ 747ms ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
║ Agent Enabled Run 1    │ 63ms │ 154ms │ 422ms │ 812ms ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
║ Agent Enabled Run 2    │ 65ms │ 154ms │ 430ms │ 812ms ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
║ Agent Enabled Run 3    │ 65ms │ 158ms │ 387ms │ 838ms ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
║ Agent Enabled Run 4    │ 61ms │ 154ms │ 379ms │ 810ms ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
║ No Op TraceState Run 1 │ 66ms │ 148ms │ 377ms │ 716ms ║
╟────────────────────────┼──────┼───────┼───────┼───────╢
[...]

PR #2063, opened 2021-04-27
fix: remove passthrough props in tracecontext

astorm:
This PR removes the dynamic property accessors from the TraceContext class in order to simplify this class and make some small performance improvements. The `definePassthroughProp` function was showing up as a top-of-stack in flame graphs, and microbenchmarking suggests we can reclaim ~5 microseconds of the 20 - 25 microsecond bump seen in https://github.com/elastic/apm-agent-nodejs/issues/2056

This PR also removes the warning about `_context` being considered public API due to its use in the OpenTracing bridge.  After investigating [the open tracing bridge](https://github.com/elastic/apm-agent-nodejs-opentracing/), it does not appear that a context object is ever _directly_ accessed.  Instead, the OT Bridge accesses [the `traceparent` property accessor](https://github.com/elastic/apm-agent-nodejs-opentracing/blob/master/lib/span_context.js#L16) that's part of the agent's `GenericSpan` class.  The naming here is somewhat confusing -- `this._context` in the OT Bridge refers to a `Transaction` or `Span` object, whereas `this._context` in `GenericSpan` refers to a `TraceContext` object. 

The intended scope of this PR is strictly to address removing the property accessors. Additional perf. improvements may be made to these classes in separate PRs.


trentm (approved):
Looks good. Any idea how much of an improvement in the benchmarks you saw with this?

astorm:
@trentm What I saw in the microbenchmarks suggests we'll reclaim around 5μs of the 20μs - 25μs microsecond bump we saw after the introduction of tracestate -- we'll know better once this lands and we can follow the trentline on the bare meta benchmarks.   Flame graphs for this change are here: https://drive.google.com/drive/folders/1IG-VTCM6L9U5GZEBa6wVKNySEishzwJN?usp=sharing

PR #2072, opened 2021-05-04
test: ensure sanitizeFieldNames is applied to captured errors

trentm:
Specifically that captured request and response headers, and request
body (if captureBody config var allows it) are sanitized.

Refs: https://github.com/elastic/apm-dev/issues/623


astorm (approved):
👍 Does the thing it says it does.  Approving.

PR #2073, opened 2021-05-04
refactor: clarify func name

trentm:
There is a local 'http-headers' module (in lib/filters) that does
something different.

Possibly a neurotic refactor. When I was walking through code for https://github.com/elastic/apm-agent-nodejs/pull/2072 there were two "things" matching "http-headers" in the agent code that do different things.


astorm (approved):
_comments about several eras of "best practices" on what a module should expose, de-structuring, etc_

neurotic or not if it helps prevent similar symbol conflation in the future the 👍 -- approving.

PR #2074, opened 2021-05-05
fix: drop obsolete .context.user.is_authenticated field

trentm:
This field is dropped by apm-server on injest. This is suspected to be
cruft from opbeat days.

AFAICT apm-server has never accepted this field. Currently at least it is dropped by apm-server.

* * *

FYI @elastic/apm-agent-python, the Python agent is the only other APM agent to have `user.is_authenticated` refs:

```
apm-agent-python/elasticapm/contrib/django/client.py
93:            if hasattr(user, "is_authenticated"):
94:                if callable(user.is_authenticated):
95:                    user_info["is_authenticated"] = user.is_authenticated()
97:                    user_info["is_authenticated"] = bool(user.is_authenticated)

apm-agent-python/tests/contrib/django/django_tests.py
186:    assert "is_authenticated" in user_info
187:    assert not user_info["is_authenticated"]
200:    assert "is_authenticated" in user_info
201:    assert user_info["is_authenticated"]
216:    with mock.patch("django.contrib.auth.models.User.is_authenticated") as is_authenticated:
217:        is_authenticated.side_effect = DatabaseError("Test Exception")
245:        assert "is_authenticated" in user_info
246:        assert user_info["is_authenticated"]
```

I haven't verified if that is also about the APM event `context.user.is_authenticated` field as well. If so, it might be something you want to drop as well.

### DIFF
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -137,4 +137,4 @@ NODE_VERSION=${NODE_VERSION} docker-compose \
   --no-ansi \
   --log-level ERROR \
   -f ${DOCKER_FOLDER}/${DOCKER_COMPOSE_FILE} \
-  down -v
+  down -v --remove-orphans
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,6 +42,9 @@ Notes:
 [float]
 ===== Bug fixes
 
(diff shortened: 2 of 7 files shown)


astorm (approved):
👍 looks like the due diligence was done here and this field can be safely removed.

PR #2079, opened 2021-05-19
feat: prefer 'traceparent' header over 'elastic-apm-traceparent' header

trentm:
The w3c traceparent header is standardized and widely used enough now in Elastic tools that we should prefer it over the proprietary 'elastic-apm-traceparent' header.

See similar for [Java](https://github.com/elastic/apm-agent-java/pull/1821), [.NET](https://github.com/elastic/apm-agent-dotnet/pull/1302).

### Checklist

- [x] Implement code

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,6 +28,21 @@ Notes:
 [[release-notes-3.x]]
 === Node.js Agent version 3.x
 
+==== Unreleased
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
+- Prefer W3C "traceparent" header over "elastic-apm-traceparent" for incoming
+  requests. {pull}2079[#2079]
+
+[float]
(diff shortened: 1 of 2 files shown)


astorm (dismissed):
Approving pending CI passing. 

Just talking through the consequences of this out loud, there's a _theoretical_ change in behavior if a user is setting both of these headers and we start preferring the W3C trace header over the old proprietary header.  

However, this doesn't seem like anything that should cause a breaking change -- the behavior change will be better interoperability with the standards based header.  

The only users who will see a degradation in their experience are folks who might have _some_ services that add both headers, and other services that only add the old proprietary header.  That seems like an edge case on an edge case (and they can always remain on an older agent), so I don't think we'll need a major version change for this change.

astorm:
jenkins run the tests

astorm:
jenkins run the tests please

PR #2081, opened 2021-05-25
perf: don't collect stack traces on short spans 

astorm:
This PR implements the `ELASTIC_APM_SPAN_FRAMES_MIN_DURATION` configuration variable, allowing users to set a minimum duration/threshold that spans must reach before the agent will collect a stack trace.  

This PR also sets a default value of 10ms for this new configuration variable.  

[Spec](https://github.com/elastic/apm/blob/94600df247e00baa3a7a154b28173d4205c85d9b/specs/agents/tracing-spans.md#span-stack-traces)

Future work not covered in this PR: 
- Agents may [move away from the "0 is none, -1 is all"](https://github.com/elastic/apm/issues/94) behavior
- This PR keeps the `captureSpanStackTraces` boolean -- when set to false NO stack traces will be sent.  When set to true the agent will reference the new `spanFramesMinDuration` value.  We'll consider dropping this in the next major release. 
 
### Checklist

- [x] Write Description
- [x] Implement code
- [x] Add tests
- [x] Update documentation


astorm:
jenkins run the tests please

astorm:
jenkins run the tests please

astorm:
jenkins run the tests please

astorm:
jenkins run the tests please

lib/instrumentation/span.js
```diff
@@ -79,6 +80,29 @@ Span.prototype.end = function (endTime) {
 
   this.ended = true
   this._agent.logger.debug('ended span %o', { span: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type, subtype: this.subtype, action: this.action })
+
+  // TODO: This is expensive! Consider if there's a way to cache some of this
+  if (this._capturedStackTrace !== null &&
+     (this.duration() / 1000 > this._agent._conf.spanFramesMinDuration)) {
```
astorm, line 86: The new check that says "don't do this work is the span is fast".

docs/configuration.asciidoc
```diff
@@ -551,7 +551,18 @@ A capture location stack trace is never generated for uncaught exceptions.
...
+Set this option to `false` to disable capture of stack traces.
+
+When set to `true` the agent will capture stack traces for spans whose duration falls within the <<span-frames-min-duration, `spanFramesMinDuration`>> threshold.
+
+[[span-frames-min-duration]]
+==== `spanFramesMinDuration`
+
+* *Type:* String
+* *Default:* `10ms`
+* *Env:* `ELASTIC_APM_CAPTURE_SPAN_STACK_TRACES_THRESHOLD`
```
trentm, line 563: Should this instead be `ELASTIC_APM_SPAN_FRAMES_MIN_DURATION` ?

docs/configuration.asciidoc
```diff
@@ -551,7 +551,18 @@ A capture location stack trace is never generated for uncaught exceptions.
...
+
+A time value to control whether the agent will report stack traces for measured spans. If a span's duration reaches this threshold, the agent will report stack traces.
```
trentm, line 565: Would be helpful to mention interaction with `captureSpanStackTraces` setting... for those that link directly to this anchor and don't notice that `captureSpanStackTraces` is defined just above.

index.d.ts
```diff
@@ -202,6 +202,7 @@ interface AgentConfigOptions {
   captureExceptions?: boolean;
   captureHeaders?: boolean;
   captureSpanStackTraces?: boolean;
+  captureSpanStackTracesThreshold?: string;
```
trentm, line 205: old name?

CHANGELOG.asciidoc
```diff
@@ -28,6 +28,21 @@ Notes:
...
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
+* Added <<span-frames-min-duration, `spanFramesMinDuration`>>
+  configuration field, allowing users to set a time threshold value that spans
+  must reach before the agent will add a stack trace to the span.
```
trentm, line 41: I think this should be expanded to point out that *by default* now, fast spans won't have stack traces.

CHANGELOG.asciidoc
```diff
@@ -28,6 +28,21 @@ Notes:
...
+
+[float]
```
trentm, line 43: Separately, the "Span Stack Traces" section of performance.asciidoc should be updated to mention this option.  IIUC the state will be:
- captureSpanStackTraces off: Least overhead.
- captureSpanStackTraces=true, spanFramesMinDuration=$somePositiveValue: Still some overhead for each span as `Error.captureStackTrace` is called on span creation.

I think we should get some sanity perf numbers to validate whether the overhead is low enough. That could be on a separate ticket to decide if we'd *still* want `captureSpanStackTraces` *off* by default.

lib/instrumentation/span.js
trentm: Looks good.

nit: A comment on the `.duration()` method would eventually be nice to state that it is in milliseconds. One currently has to wade through to lib/instrumentation/timers.js and its `elapsed()` method. :)

IIUC this moves the start of `stackman.callsites` processing from span *start* to span *end*... which looks good to me.

test/capture-stacktrace-duration.js
```diff
@@ -0,0 +1,67 @@
+'use strict'
```
trentm, line 1: nit: would be nice if the test file name was based on "spanFramesMinDuration" and/or a comment at the top said this is about testing that config var.

test/capture-stacktrace-duration.js
```diff
@@ -0,0 +1,67 @@
...
+      // t.ok(false,'we are parsing the callsites after we are encoded the span')
+      span._encode(function (err, data) {
+        t.error(err)
+        t.ok(!data.stacktrace, 'stacktrace not set')
+        t.end()
+      })
+    }, 50)
+  })
+
+  suite.test('over configured value has has stack trace', function (t) {
```
trentm, line 50:
```suggestion
  suite.test('over configured value has stack trace', function (t) {
```

lib/instrumentation/span.js
```diff
@@ -35,6 +35,7 @@ function Span (transaction, name, ...args) {
   this._destination = null
   this._message = null
   this._stackObj = null
+  this._capturedStackTrace = null
```
trentm, line 38: Below at:

```js
  if (this._agent._conf.captureSpanStackTraces) {
          this._recordStackTrace()
```

should we consider also skipping if `_conf.spanFramesMinDuration <= 0`?  This would allow the same behaviour as the Java agent's https://www.elastic.co/guide/en/apm/agent/java/current/config-stacktrace.html#config-span-frames-min-duration where a negative value (or `0` as you are using in tests below) will turn off the `Error.captureStackTrace()` calls and save some more cycles.  This might be relevant if/when `spanFramesMinDuration` is added to central config.

test/instrumentation/async-hooks.js
```diff
@@ -136,13 +136,28 @@ test('sync/async tracking', function (t) {
...
   setImmediate(() => {
     t.strictEqual(trans.sync, false)
-    t.strictEqual(span1.sync, false)
     t.strictEqual(span2.sync, true,
       'span2.sync=true later after having ended sync')
     t.end()
   })
 })
 
+test('sync/async tracking: span ended across async boundry is not sync', function (t) {
```
trentm, line 145:
```suggestion
test('sync/async tracking: span ended across async boundary is not sync', function (t) {
```

astorm:
jenkins run the tests please

test/instrumentation/async-hooks.js
astorm: ✅

lib/instrumentation/span.js
astorm: > from span start to span end

You understand that correctly.  In order to decide whether or not to do this work based on span duration, we need to move this work to the end.  and 👍 on the comment for `GenericSpan.prototype.duration`

docs/configuration.asciidoc
astorm: yup, good eye.  ✅

docs/configuration.asciidoc
astorm: Sounds like an improvement to me.  Done. ✅

index.d.ts
astorm: _nod_ -- good eye.  Replaced with `spanFramesMinDuration?: string;`

CHANGELOG.asciidoc
astorm: _nod_ that makes sense.  I've added some initial words to `performance.asciidoc` about this new configuration value, and have also created an issue to capture the intent around turning this off by default: https://github.com/elastic/apm-agent-nodejs/issues/2092

test/capture-stacktrace-duration.js
astorm: ah, good call.  Also, dings my "I really ought to check the specs for env names _before_ picking one" bell 🛎️

test/capture-stacktrace-duration.js
astorm: done done ✅

lib/instrumentation/span.js
astorm: Good question! I did some thinking and testing on this one today.

First -- I think we'll be ok going either way on this.  The behavior [isn't speced](https://github.com/elastic/apm/blob/master/specs/agents/tracing-spans.md#span-stack-traces) so we have some flexibility.

Second -- if [I'm reading those docs correctly](https://www.elastic.co/guide/en/apm/agent/java/current/config-stacktrace.html#config-span-frames-min-duration) -- then zero and negative values actually have different meanings.  Zero is appears to disable traces

> To disable stack trace collection for spans completely, set the value to 0ms.

While negative values indicate every span should have a stack trace

> When setting this option to a negative value, like -1ms, stack traces will be collected for all spans.

It sounds like [this has come up before](https://github.com/elastic/apm/issues/94) and that both Andrew and Felix find the current state of affairs in the java agent confusing.  I tend to agree with them -- a `0` to me means collect stack traces for all the spans.

So, for now my preference would be to skate to where the java agent is headed and just honor `0` as the "always collect stack traces" value since (almost?) all spans should have some positive duration, and allow the existing `captureSpanStackTraces` to be the global toggle instead of fiddling around with a negative values.

CHANGELOG.asciidoc
astorm: Fair -- added a _As a result, many short spans that previously included stack traces by default no longer will._

lib/instrumentation/span.js
trentm: I agree with "skate to where the java agent is headed" in general.
Are we going move to:

1. Having a negative value meaningful -- to mean "never collect span stack traces" -- and then eventually deprecate `captureSpanStackTraces` in favour of just the single numeric config var; or
2. Continue with two separate vars: (a) a boolean to turn it on or off, plus (b) a numeric to just specify the duration if the boolean is "on"; and suggest other agents do the same.

lib/instrumentation/generic-span.js
```diff
@@ -84,6 +84,7 @@ Object.defineProperty(GenericSpan.prototype, 'traceparent', {
   }
 })
 
+// duration is returned as an int representing the time in miliseconds
```
trentm, line 87:
```suggestion
// duration is returned as an int representing the time in milliseconds
```

nit

trentm (commented):
All looks good. ... just the big Q about what config var name and its semantics.

lib/instrumentation/span.js
astorm: Capturing a decision made elsewhere

1. Implement the behavior as the other agents have (-1 means collect all, and a 0 means collect none)
2. Leave our captureSpanStackTraces boolean as is for backward compatibility reasons.  If captureSpanStackTraces=false then nothing is collected.  If it's true then the value in spanFramesMinDuration is also consulted.
3. Restart the discussion in https://github.com/elastic/apm-agent-nodejs/pull/2081 in order to decide what everyone should do moving forward.

docs/configuration.asciidoc
```diff
@@ -564,6 +564,10 @@ When set to `true` the agent will capture stack traces for spans whose duration
 
 A time value to control whether the agent will report stack traces for measured spans. If a span's duration reaches this threshold, the agent will report stack traces. If <<capture-span-stack-traces, `captureSpanStackTraces`>> is set to `false` the agent will ignore the `spanFramesMinDuration` value.
 
+When set to `0` the agent will not record stack traces for any spans.
+
+When set to a negative value (like `-1`) the agent will always record stack traces for spans.
```
trentm, line 569:
```suggestion
When set to a negative value (like `-1ms`) the agent will always record stack traces for spans.
```

Should include units so it isn't ambiguous. (Though, granted that in this case for a negative value compared against duration that the units don't matter.)

lib/instrumentation/span.js
```diff
@@ -125,6 +124,18 @@ Span.prototype.setMessageContext = function (context) {
   this._message = Object.assign(this._message || {}, context)
 }
 
+Span.prototype.shouldSetSpanStackTrace = function () {
+  if (this._agent._conf.spanFramesMinDuration === -1) {
```
trentm, line 128: The [Java agent docs](https://www.elastic.co/guide/en/apm/agent/java/current/config-stacktrace.html#config-span-frames-min-duration) (and its [impl](https://github.com/elastic/apm-agent-java/blob/a837d43aa3fa189c19ae5dadaa4e0e28ae52d97a/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/ElasticApmTracer.java#L397-L399)) say:

> When setting this option to a negative value, like -1ms, stack traces will be collected for all spans. 

In other words, it isn't `-1` specifically that is a special case; it is *any* negative value.  And actually it doesn't need to be special-cased because any `duration` will be greater than a negative value.  So the implementation above could be simplified to:

```
if (this._capturedStackTrace !== null && this._agent._conf.spanFramesMinDuration !== 0 && this.duration() / 1000 > this._agent._conf.spanFramesMinDuration) { ... }
```

and potentially not have the separate `.shouldSetSpanStackTrace()` method -- though that is fine too.

test/span-frames-min-duration.js
```diff
@@ -65,5 +65,37 @@ tape.test(function (suite) {
...
+      // t.ok(false,'we are parsing the callsites after we are encoded the span')
+      span._encode(function (err, data) {
+        t.error(err)
+        t.ok(!data.stacktrace, 'stacktrace is not set')
+        t.end()
+      })
+    }, 1)
+  })
+
+  suite.test('spanFrameMinDuration=-1 always sets', function (t) {
```
trentm, line 84: nit
```suggestion
  suite.test('spanFrameMinDuration=<negative> always sets', function (t) {
```

trentm (changes requested):

lib/instrumentation/span.js
astorm: fair -- I might argue the separate method makes the intent of the long conditional clearer, but we don't have any linting rules against those long conditionals to I'll let that lie ;)   Dropped the above code in place and tests (plus a new one with a spanFramesMinDuration: '-2' are passing.

trentm (approved):
Thanks.

PR #2087, opened 2021-05-27
test: limit testing of 'edge' node builds (node.js nightly and RC builds) to useful versions

trentm:
"Edge" builds or "Edge Test" is this repo's name for running the Agent test suite against pre-release (RC or nightly) builds of Node.js.

1. node.js "nightly" builds: It is only useful testing those if there are no current releases for that major branch. For example, a nightly build of "v16.0.0-nightly20210420a0261d231c" (the current latest v16 nightly build) is pointless because there are v16.x node.js releases after that that we already test.  At time of writing only v17 has nightly builds and no releases. Once there is a v17.0.0 release my interpretation of https://nodejs.org/download/nightly/ is that there will be no more v17.x nightly builds -- or at least no v17 nightly builds of a commit past the latest v17 release commit sha. (I'd welcome an authoritative source for that. I couldn't find it. E.g. https://github.com/nodejs/node/blob/master/doc/guides/releases.md#8-produce-a-nightly-build-optional isn't really it.)

2. node.js "RC" builds: Only useful to check for active majors (per https://nodejs.org/en/about/releases/). Similar to the above, if the latest v16 RC build is "v16.0.0-rc.5" (as is the case currently), then there is no point in testing against it because there is a v16.0.0 actual release of node.js.

This PR:
- changes the Node.js versions of "nightly" and "rc" pre-release builds that we consider testing against; and
- makes ".ci/scripts/test.sh" smarter to **skip out** of running the agent test suite against a nightly or RC pre-release build that isn't useful to test against (according to the descriptions above). 

The result is that the execution of "Edge Tests" (currently run daily by Jenkins on a timer) will, at time of writing, only run a test against the latest node v17 nightly build. There aren't currently any RCs for upcoming node.js releases. If one lands in https://nodejs.org/download/nightly/ or https://nodejs.org/download/rc/   then the next "Edge Tests" run will test it.

### DIFF
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -11,45 +11,143 @@
 #     /bin/bash -c ".ci/scripts/docker-test.sh"
 #
 # Usage:
-#     .ci/scripts/tests.sh NODE_VERSION [TAV_MODULE] [IS_EDGE]
+#     .ci/scripts/tests.sh [-b release|rc|nightly] [-f] [-t TAV_MODULE] NODE_VERSION
 #
 # - NODE_VERSION is a version of Node.js, e.g. "14", usable both to select
 #   a image from Docker Hub:
 #       docker pull docker.io/library/node:$NODE_VERSION
 #   and to install Node with nvm:
 #       nvm install $NODE_VERSION
-# - TAV_MODULE, if specified and not the empty string, will result in
-#   "test-all-versions" (TAV) tests being run. It identifies the name of a
-#   module in ".tav.yml", e.g. "redis" or "pg". If this argument is not given
-#   then the regular agent tests (i.e. the same as 'npm test') will be run.
-# - IS_EDGE is "true" or "false" (the default). If true, then the node version
-#   used for testing will be installed via:
-#       NVM_NODEJS_ORG_MIRROR=${NVM_NODEJS_ORG_MIRROR} nvm install ${NODE_VERSION}
-#   where typically the mirror URL is also set to one of:
-#       https://nodejs.org/download/nightly/
-#       https://nodejs.org/download/rc/
-#   to test against a recent pre-releases of node.
+# - TAV_MODULE, if specified, will result in "test-all-versions" (TAV) tests
+#   being run. It identifies the name of a module in ".tav.yml", e.g. "redis" or
+#   "pg". If this argument is not given then the regular agent tests (i.e. the
+#   same as 'npm test') will be run.
+# - The "-b" option (BUILD_TYPE) can be used to test with pre-release "rc" or
+#   "nightly" builds of Node.js. This will skip out if there is already a
+#   release build for the given version.
+# - The "-f" option (FORCE) can be used to force a "nightly" or "rc" test,
+#   even if there is already a release build for the same version.
 #
 # Examples:
-#     .ci/scripts/test.sh 14                # regular tests against latest docker:14 image
-#     .ci/scripts/test.sh 14 "" false       # same
-#     .ci/scripts/test.sh 8.6               # regular tests against latest docker:8.6 image
-#     .ci/scripts/test.sh 14 "redis" false  # redis TAV tests against latest docker:14 image
-#     NVM_NODEJS_ORG_MIRROR=https://nodejs.org/download/nightly/ \
-#       .ci/scripts/test.sh 14 "" true      # regular tests against latest node 14 nightly release
+#     .ci/scripts/test.sh 14                # run agent tests against latest docker:14 image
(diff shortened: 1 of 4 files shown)


trentm:
This (jenkins build 6 of this PR: https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2087/6/) worked as expected:
- All the RC builds skipped out early with a message like: "There is already a release build (v12.22.0) of the latest v12 rc (v12.22.0-rc.0). Skipping tests."
- The v17 nightly build ran and passed.

However:

Screen Shot 2021-05-27 at 1 25 43 PM

Jenkins marked the RC build steps (and the whole build) "UNSTABLE"... because I don't know why. 

1. Is it because those stages ran way faster than before?
2. Is it because those stages didn't produce TAP output leading to this sad:

Screen Shot 2021-05-27 at 1 16 49 PM

vs. this happy (with the same output):

Screen Shot 2021-05-27 at 1 16 30 PM


The full console output (https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2087/6/consoleFull) mentions "UNSTABLE" only once at the end:

```
Finished: UNSTABLE
```

with no justification. Thanks Jenkins.

trentm:
I'm not sure if this "UNSTABLE" designation will continue if/when this is merged to master.  The impact is that an "UNSTABLE" build will result in GitHub PR and commit checks *not* being a pass.

trentm:
Another theory on Jenkins marking these UNSTABLE is because it considers there to be failed tests... where "tests" here are the Junit format tests that this Jenkinsfile step produces:

```
            if (!tav?.trim()) {
              tap2Junit(pattern: '*-output.tap', suffix: 'junit.xml', package: 'Node.js', failNever: true)
            }
```


Screen Shot 2021-05-27 at 3 12 02 PM


(Aside: I don't see where in the Jenkinsfile the mapping from "there is a SOMETHING-junit.xml file created" to "you should process that junit.xml file and show it as tests in the Jenkins UI" exists. I don't get any value from it.)

(Grumble: That `failNever: true` I would expect to mean we don't get marked "failed tests" ... but that's not quite what it means.)

A potential hack is to **create a minimal "test-suite-output.tap" output file that this `tap2Junit` function expects.**

trentm:
That works, and nicely shows "skipped" counts in the Jenkins "Test Result" summary. For example: 
Screen Shot 2021-05-27 at 3 48 43 PM

.ci/scripts/test.sh
```diff
@@ -11,54 +11,166 @@
...
+#     .ci/scripts/test.sh -b release 14     # same
+#     .ci/scripts/test.sh 8.6               # run agent tests against latest docker:8.6 image
+#     .ci/scripts/test.sh -t redis 14       # run redis TAV tests against latest docker:14 image
+#     .ci/scripts/test.sh -b nightly 17     # run agent tests against the latest v17 node.js nightly build
+#     .ci/scripts/test.sh -f -b nightly 17  # ... same, but force doing tests even if there
+#                                           #     is already a node.js v17 release
 #
 
-set -exo pipefail
+set -eo pipefail
```
astorm, line 41: Presuming -x has been removed to make this all a bit less chatty.  This is a six of one, half dozen of the other decision for me -- just noting it out loud.

astorm (approved):
Reasoning about the nightlies and RCs lines up with my (brief) review of what's available for download and their last modified dates :thumbsup:

The changes to the script from "is or is not edge" to "build type" also make sense, and I see we've updated our one invocation of `.ci/scripts/test.sh` (in `.ci/Jenkinsfile`) to take this into account.  I don't know of anything else that's using/invoking this script, so :thumbsup:

With the _Unstable_ problem taken care of this seems ready to go -- approving.

.ci/scripts/test.sh
trentm: It was added again below after arg processing:

```
# Turn on xtrace output only after processing args.
set -x
```

because otherwise for a simple `.ci/scripts/test.sh -h` to get help output, it resulted in lots of noise.


If we were over-engineering this, I personally typically have a starter Bash script block like this:

```bash
if [ "$TRACE" != "" ]; then
    export PS4='${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
    set -o xtrace
fi
set -o errexit
set -o pipefail
```

which allows full xtrace with the "TRACE=1" envvar. (The PS4 also gives detail like bash function and line number for each logged line.) We could do that at some point, if helpful.

Issue #2028, opened 2021-04-01
tracking bug for agent performance impact on Kibana


trentm:
This is a tracking bug for a timeboxed effort to look at improving the Node.js APM agent's perf impact in Kibana.
See https://github.com/elastic/kibana/issues/78792 for most details. This ticket will help for tracking agent-side work.

trentm:
Yesterday my thoughts on a plan for this issue were:

1. Investigate each of the top items over ~2% top-of-stack from flamegraphs
   of the loaded demo "pathosapp.js"
   (truncation/encoding, tracecontext, Transaction._captureBreakdown)
2. Implement the obvious "captureSpanStackTraces=false by default", and
   a new setting to only capture span stack traces for slow spans.
3. Profile and possibly improve stack trace capturing (used for
   captured errors and ^^^ slow spans).
4. Run Kibana DemoJourney with APM agent stripped to barebones core
   async-hook that is the foundation of tracing async context as a
   guide for "how low we can go" on overhead.
5. Learn (and possibly integrate) agent ability to get profiles of
   part of an app run using pprof-nodejs (building on Dario's PoC
   https://github.com/dgieselaar/apm-agent-nodejs/tree/pprof-poc/).
   This could be used for debugging customer perf issues (get a
   profile from prod to analyze offline) and is the likely basis
   for a profiling feature in the Node.js APM agent.
6. Add one or more benchmarks to the agent's current benchmark system
   that covers encoding of spans/transactions/errors for ongoing
   tracking.

Timebox to another 2 weeks max, then time to move on.

Future work:
- Investigate cachegrind technique for stable benchmarking
  (https://github.com/pythonspeed/cachegrind-benchmarking).
- WG Perf (https://github.com/elastic/observability-dev/issues/1248)

trentm:
>     1. Investigate each of the top items over ~2% top-of-stack from flamegraphs
>        of the loaded demo "pathosapp.js"
>        (truncation/encoding, tracecontext, Transaction._captureBreakdown)

That flamegraph is here: https://gist.github.com/trentm/c70154cc6674a000f6b4c0827e3f0183
Screenshot:
![screenshot](https://user-images.githubusercontent.com/46866/115898297-3de95300-a412-11eb-9cc3-5d19ed48a07b.png)
Top "top-of-stack" items there are:

[...]

PR #1920, opened 2020-12-17
test: add tests for parsers._moduleNameFromFrames

trentm:
Also refactor lib/parsers.js slightly. No functional change.

Fixes: #1918

PR #2094, opened 2021-05-31
perf: refactor error capture and stacktrace collection

trentm:
Refactor agent.captureError(...) and stacktrace collection for better
performance. This moves the stackman dep into lib/stacktraces.js and
lib/errors.js to allow breaking its compatibility for perf improvements.

Refs: #2028
Fixes: #2095

### Checklist

- [x] Implement code
- [x] Tests


trentm:
### perf test setup

- The script being tested in `node pathosapp.js err` -- a simple express app whose `/` route does the synthetic:

```
function err(req, res, next) {
    next(new Error('oof'))
}
```

(I'll propose adding this script to the repo later. I find it useful for perf testing.)

- load that up with `hey -c 50 -q 10 -z 60s http://localhost:3000/` (a 60s run of 50 concurrent callers at 10 req/s, for a total load of 500 req/s)

- Profiling with `node --prof ...` to get a flamegraph (using the 0x tool).

Ultimately the command being run here is:
```
0x -o -D "$DATADIR" -P 'hey -c 50 -q 10 -z 60s http://localhost:3000/' bin/pathosapp.js -m "$NAME" err
```

### without these changes

Testing against the current release elastic-apm-node@3.15.0

```
$ ./scripts/flame-it err-pr2 err            # the command run (my wrapper script)
CPU: ~89%                 # observed process CPU usage during the run via `htop`

# the CPU usage for the run of pathosapp.js using `process.cpuUsage(...)`
CPU usage (μs since process start):  { user: 49386192, system: 6246167 }

# the latency distribution from `hey` output
Latency distribution:
10% in 0.0015 secs
25% in 0.0028 secs
50% in 0.0047 secs
75% in 0.0077 secs
90% in 0.0110 secs
95% in 0.0124 secs
99% in 0.0168 secs
```

A screenshot of the flamegraph for what it is worth:

Screen Shot 2021-06-04 at 4 48 14 PM

### with the current PR changes

```
$ ./scripts/flame-it err-pr10-rmgetcwd err -s

CPU: ~70%
CPU usage (μs since process start):  { user: 39872179, system: 4153932 }

Latency distribution:
  10% in 0.0011 secs
  25% in 0.0024 secs
  50% in 0.0042 secs
  75% in 0.0069 secs
  90% in 0.0092 secs
  95% in 0.0102 secs
  99% in 0.0147 secs
```

Screen Shot 2021-06-04 at 4 51 44 PM


### observations

- For the same load CPU usage is down from 90% to 70%.
- For this (albeit extreme and synthetic workload: app erroring at full tilt) the latency distribution numbers are a 10-20% improvement.

(These flamegraph screenshots don't really provide much info after the fact. They are useful for focusing perf work during dev.)

trentm:
## Rough summary of changes

Perf issues:
- `{error,span}.stacktrace` logic was largely in the "stackman" module.
  It had a few perf issues:
    1. It monkey-patched (or wrapped?) v8 native CallSite objects with
       a bunch of properties for convenience/sugar. Using these in
       stacktrace generation was very slow.
    2. It had a bug in its caching of sourcemap file loading for each
       stackframe: when a file had no separate sourcemap file (the common
       case), this wasn't cached. That means we would be stat'ing for
       a sourcemap file for almost every frame.
    3. A stack frame's `filename` is a relative path to cwd. The code ended
       up calling `process.cwd()` for *every* frame, which is expensive.
  Dealing with the first issue can't be done without either completely breaking
  "stackman" backwards compat, or just no longer using "stackman" and vendoring
  some of its code. I've done the latter. Error and stacktrace code has moved to
  lib/errors.js and lib/stacktraces.js respectively.
- This adds caching to the conversion of a CallSite object to an APM error
  stacktrace frame (frameFromCallSite), which had a significant impact on
  CPU usage in the perf test mentioned above (it was about 6-7% of the diff).

Other issues:
- Fix sourcemap "sourcesContent" usage (#2094).
  Test case: test/stacktraces/stacktraces.test.js

lib/agent.js
```diff
@@ -313,153 +314,161 @@ Agent.prototype.addMetadataFilter = function (fn) {
...
-    : trans && trans.res
-  var captureAttributes = opts && opts.captureAttributes
-  const skipOutcome = opts && opts.skipOutcome
-  var _isError = isError(err)
-
-  if ((!opts || opts.handled !== false) &&
+  const agent = this
+  let callSiteLoc = null
+  const errIsError = isError(err)
+  const handled = opts.handled !== false // default true
```
trentm, line 364: REVIEW-NOTE: This changes `handled` to be set `true` in cases where it was sometimes left `undefined` by accident/bug. Previously it was set like this: `handled = !opts || opts.handled`. The result of the latter is that adding unrelated opts, e.g. `captureError(err, { request: res}, cb)` changes `handled` from true to undefined.

lib/errors.js
```diff
@@ -0,0 +1,267 @@
...
+//                 ^^^^^^^^^^^^^^^
+// or on Windows:
+//    node_modules\@myorg\mymodule\lib\subpath\index.js
+//                 ^^^^^^^^^^^^^^^
+let SEP = path.sep
+if (SEP === '\\') {
+  SEP = '\\' + SEP // Escape this for use in a regex.
+}
+const MODULE_NAME_REGEX = new RegExp(`node_modules${SEP}([^${SEP}]*)(${SEP}([^${SEP}]*))?`)
+function _moduleNameFromFrames (frames) {
```
trentm, line 50: REVIEW-NOTE: This just moved from parsers.js unchanged.

lib/errors.js
```diff
@@ -0,0 +1,267 @@
...
+  if (moduleName && moduleName[0] === '@' && match[3]) {
+    // Normalize the module name separator to '/', even on Windows.
+    moduleName += '/' + match[3]
+  }
+  return moduleName
+}
+
+// Gather properties from `err` to be used for `error.exception.attributes`.
+// If there are no properties to include, then it returns undefined.
+function attributesFromErr (err) {
```
trentm, line 66: REVIEW-NOTE: This is based off stackman.properties() that was used before (identical except for the handling of `key === 'code'` moved here from elsewhere, and formatting).

lib/errors.js
```diff
@@ -0,0 +1,267 @@
...
+      var match = errMsg.match(MYSQL_ERROR_MSG_RE)
+      if (match) {
+        error.exception.code = match[1]
+      }
+    }
+
+    // Optional add an alternative error message as well as the exception message.
+    if (args.message && typeof args.message === 'string') {
+      error.log = { message: args.message }
+    }
```
trentm, line 176: REVIEW-NOTE: This is a slight change in guards on `opts.message` from before that I think more closely matches the API docs.  This changes the behaviour so that `error.log.message` is set whenever a message option is passed to `agent.captureError` like this:

```
agent.captureError(new Error('foo'), { message: 'foo' })
```

The implementation before was to have a special case where `error.log.message` was not added in that case *because err.message and opts.message were the same*. 

Here are the API docs on `captureError`:

> ** `message` - If the `error` argument is an +{type-error}+ object,
> it's possible to use this option to supply an additional message string that will be stored along with the error message under `log.message`

I think it is slightly surprising that a given `opts.message` might not get used if the err.message happens to be the same.

If you care strongly, I'm happy to revert to the old behaviour and tweak the API doc to be clearer.

lib/errors.js
```diff
@@ -0,0 +1,267 @@
...
+      null, // filterCallSite
+      function (_err, stacktrace) {
+        // _err from gatherStackTrace is always null.
+
+        const culprit = culpritFromStacktrace(stacktrace)
+        if (culprit) {
+          error.culprit = culprit
+        }
+        const moduleName = _moduleNameFromFrames(stacktrace)
+        if (moduleName) {
```
trentm, line 200: REVIEW-NOTE: This fixes a bug added in #1920 where the variable is `const moduleName`, but the if-block is `if (module) {`  -- write var name.

test/stacktraces/stacktraces.test.js
```diff
@@ -0,0 +1,343 @@
...
+      library_frame: false,
+      abs_path: __filename
+    },
+    'stacktrace top frame is as expected'
+  )
+
+  t.end()
+})
+
+tape.test('gatherStackTrace()', function (suite) {
```
trentm, line 247: REVIEW-NOTE This test case replaces '#parseCallsite()' test from test/parsers.js

test/errors.test.js
```diff
@@ -0,0 +1,340 @@
...
+    if (!Object.prototype.hasOwnProperty.call(o, field)) {
+      return undefined
+    }
+    o = o[field]
+  }
+  return o
+}
+
+// Test processing of Error instances by `createAPMError`.
+tape.test('#createAPMError({ exception: ... })', function (suite) {
```
trentm, line 30: REVIEW-NOTE: This reproduces the old test/parsers.js '#parseError()' tests (except the 'should be able to choose number of source context line per frame type' tests that I think are more test code than helpful to ensure an int is passed through correctly)

test/errors.test.js
```diff
@@ -0,0 +1,340 @@
...
+        }
+      )
+    })
+  })
+
+  suite.end()
+})
+
+// Test the various forms of the `logMessage` arg to `createAPMError`.
+tape.test('#createAPMError({ logMessage: ... })', function (suite) {
```
trentm, line 208: REVIEW-NOTE: This reproduces the old test/parsers.js '#parseMessage()' tests.

trentm:
On the test failure: It failed in a single one of the test runs and failed right at the end, during Docker cleanup:

```
[2021-06-05T00:36:37.820Z] error while removing network: network docker_default id 8f9957080cae6d9f1130514145c7a06cedf877e3b26bfdb2afb04ada4c9ff3a9 has active endpoints
```

trentm:
> On the test failure:

And then a re-run of tests (including the TAV and edge tests)... *worked*, but the runner lost the plot in finishing the "edge tests":

https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2094/21/
Blue ocean view showing the "RC Test" step that just hung and never completed: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-2094/21/pipeline/5295

astorm:
jenkins run the tests please

astorm:
jenkins run the tests please

astorm (commented):
I took a first pass at reviewing this and did some limited/basic testing. In my tests the new code appeared to be creating identical stack traces to the old :+1:

I've left some questions/comments -- there's still a few things I'd like to review but I wanted to get you some feedback before the day was out.

lib/stacktraces.js
```diff
@@ -0,0 +1,416 @@
...
+    loadSourceMap(filename, function onSourceMap (err, sourcemap) {
+      // Translate sourcemap===undefined to null, because 'async-cache'
+      // treats `undefined` as a cache miss. Without this there is no
+      // caching for files that have no sourcemap (the common case).
+      cb(err, sourcemap || null)
+    })
+  }
+})
+
+const frameCache = new LRU({ max: 1000 })
```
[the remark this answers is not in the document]
trentm: > Do we have any back of the napkin numbers on how much this will could, potentially, increase overall memory use for the agent?

Good Q. Given an example frame is ~500 bytes when JSON serialized... I'd have hoped it would be as simple as saying "it'll take up ~500kB of added memory". However playing around a bit with the implementation and applying the load (500 req/s) to pathosapp.js, it isn't so simple. The numbers below are "RES" (resident size) from watching `htop` during a load run. Note that RES oscillated plus/minus 5-10MB during runs.

- using the current agent release (v3.15.0): RES=~122M
- using the current PR adjusted to comment out `frameCache` usage: RES=~144M
- using the current PR adjusted to bust the cacheKey so we hit 1000 entries *and* always have a cache miss: RES=~155M
- using the current PR adjusted to have a `frameCache = new LRU({ max: 1 })`: RES=~155M
- using the current PR adjusted with the patch just below to hit 1000 entries, but still cache effectively: RES=~123M
- ditto, but using node 16: RES=~110MB (10-12MB less than node v14!)

```diff
diff --git a/lib/stacktraces.js b/lib/stacktraces.js
index 2809080f..aa7cad9b 100644
--- a/lib/stacktraces.js
+++ b/lib/stacktraces.js
@@ -241,7 +241,8 @@ function frameFromCallSite (log, callsite, cwd, sourceLinesAppFrames, sourceLine
   const colno = callsite.getColumnNumber()

   // Caching
-  const cacheKey = [filename, lineno, colno, sourceLinesAppFrames, sourceLinesLibraryFrames].join(':')
+  let cacheKey = [filename, lineno, colno, sourceLinesAppFrames, sourceLinesLibraryFrames].join(':')
+  if (frameCache.itemCount < 1000) cacheKey += Math.random().toString()
   const cachedFrame = frameCache.get(cacheKey)
   if (cachedFrame !== undefined) {
     frameCacheStats.hits++
@@ -251,6 +252,7 @@ function frameFromCallSite (log, callsite, cwd, sourceLinesAppFrames, sourceLine
   function cacheAndCb (frame) {
     frameCacheStats.misses++
     frameCache.set(cacheKey, frame)
+    frameCacheStats.itemCount = frameCache.itemCount
     cb(null, frame)
   }

```

My conclusions:
- It's complicated.
- A heap profile might help given more concrete data, but I'm not proposing to do that right now.
- I *believe* the memory usage of that cache is mostly lost in the noise of the changes in behaviour (handling concurrent operations differently under load) with the effective caching. Recall that when getting cache hits the CPU usage dropped 7-10%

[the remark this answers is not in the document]
trentm: > Also, more generally, any thoughts on whether the `lru-cache` module is what we want to use here long term, or if this seems like something to target for simplification in a future PR?

For what it's worth, I *did* do a load run with a simple `const frameCache = new Map()` (which didn't do any `max: 1000` handling) vs. using `new LRU({ max: 1000 })` and for CPU usage or flamegraph I saw no change. So CPU performance-wise, I think lru-cache is fine.

Also, for what it's worth, I've been using lru-cache for almost a decade and I'm pretty happy with it. For a while I had https://github.com/trentm/node-expiring-lru-cache which added expiry support to `lru-cache`. Then isaacs added `maxAge` support to `lru-cache` making my module unnecessary.  Happy to re-evaluate later at some point, but I think it'll be low prio. `lru-cache` is pretty battle tested.

lib/agent.js
```diff
@@ -313,153 +314,161 @@ Agent.prototype.addMetadataFilter = function (fn) {
...
         })
-      })
-    } else if (cb) {
-      // TODO: Swallow this error just as it's done in agent.flush()?
-      process.nextTick(cb.bind(null, new Error('cannot capture error before agent is started'), id))
-    }
-  }
+      } else if (cb) {
+        // TODO: Swallow this error just as it's done in agent.flush()?
+        cb(new Error('cannot capture error before agent is started'), id)
```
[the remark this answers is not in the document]
trentm: Yes. I don't see the point given that we are already in the explicitly async `setImmediate` further up.

lib/agent.js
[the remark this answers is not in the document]
trentm: Yup, fair. I'll add a changelog entry.

Q: do you even know what the visible diff is in APM UI for a handled vs unhandled exception?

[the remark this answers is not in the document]
trentm: commit 51f8e995

trentm:
@astorm FYI: I found a bug. Because (a) frame Objects cache be retrieved from a cache now and (b) because we use a JSON serializer for APM events that replaces an object it has already seen during serialization with `"[Circular]"`, if a given stacktrace has two frames at the same location, then the serializer will produce a serialized `error.exception.stacktrace` with some elements `"[Circular]"`. This happens with the Express-based pathosapp, e.g.:

```
...
                    {
                        "filename": "node_modules/express/lib/router/index.js",
                        "lineno": 284,
                        "function": "<anonymous>",
                        "library_frame": true,
                        "abs_path": "/Users/trentm/el/apm-nodejs-dev/perf/node_modules/express/lib/router/index.js",
                        "pre_context": [
                            "      }",
                            ""
                        ],
                        "context_line": "      trim_prefix(layer, layerError, layerPath, path);",
                        "post_context": [
                            "    });",
                            "  }"
                        ]
                    },
                    "[Circular]",
                    "[Circular]",
                    {
                        "filename": "node_modules/express/lib/middleware/query.js",
                        "lineno": 45,
                        "function": "query",
                        "library_frame": true,
                        "abs_path": "/Users/trentm/el/apm-nodejs-dev/perf/node_modules/express/lib/middleware/query.js",
                        "pre_context": [
                            "    }",
                            ""
                        ],
                        "context_line": "    next();",
                        "post_context": [
                            "  };",
                            "};"
                        ]
                    },
                    "[Circular]",
                    "[Circular]",
                    "[Circular]",
                    "[Circular]",
                    "[Circular]",
                    "[Circular]",
                    {
                        "filename": "node_modules/express/lib/router/index.js",
                        "lineno": 174,
                        "function": "handle",
                        "library_frame": true,
...
```

I'm working on fix.

lib/stacktraces.js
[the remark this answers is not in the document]
astorm: 👍

trentm:
For the coming fix, this shows the newly added test failing without the fix:

```
% node stacktraces.test.js
TAP version 13
# error.exception.stacktrace
ok 1 throw-an-error.js errored out
ok 2 err.message includes /Error: boom/: "Command failed: /Users/trentm/.nvm/versions/node/v14.16.1/bin/node fixtures/throw-an-error.js\nError: boom\n at main (/Users/trentm/el/apm-agent-nodejs2/test/stacktraces/fixtures/throw-an-error.js:15:9)\n at Object.<anonymous> (/Users/trentm/el/apm-agent-nodejs2/test/stacktraces/fixtures/throw-an-error.js:18:1)\n at Module._compile (internal/modules/cjs/loader.js:1063:30)\n at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)\n at Module.load (internal/modules/cjs/loader.js:928:32)\n at Function.Module._load (internal/modules/cjs/loader.js:769:14)\n at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)\n at internal/main/run_main_module.js:17:47\n"
ok 3 APM server got event metadata object
ok 4 APM server got error event
ok 5 stacktrace top frame is as expected
# error.exception.stacktrace does not have "[Circular]" frames
ok 6 circular-stack.js exited non-zero
ok 7 APM server got first metadata object
ok 8 APM server got first error event
ok 9 APM server got second metadata object
ok 10 APM server got second error event
ok 11 "boom on zero" stacktrace frame is as expected
not ok 12 found 9 "boom on zero" stacktrace frames
  ---
    operator: equal
    expected: 9
    actual:   1
    at: done (/Users/trentm/el/apm-agent-nodejs2/test/stacktraces/stacktraces.test.js:106:11)
    stack: |-
      Error: found 9 "boom on zero" stacktrace frames
          at Test.assert [as _assert] (/Users/trentm/el/apm-agent-nodejs2/node_modules/tape/lib/test.js:304:54)
          at Test.bound [as _assert] (/Users/trentm/el/apm-agent-nodejs2/node_modules/tape/lib/test.js:91:32)
          at Test.strictEqual (/Users/trentm/el/apm-agent-nodejs2/node_modules/tape/lib/test.js:468:10)
          at Test.bound [as equal] (/Users/trentm/el/apm-agent-nodejs2/node_modules/tape/lib/test.js:91:32)
          at done (/Users/trentm/el/apm-agent-nodejs2/test/stacktraces/stacktraces.test.js:106:11)
          at ChildProcess.exithandler (child_process.js:299:7)
          at ChildProcess.emit (events.js:315:20)
          at maybeClose (internal/child_process.js:1048:16)
          at Socket.<anonymous> (internal/child_process.js:439:11)
          at Socket.emit (events.js:315:20)
  ...
```

lib/errors.js
astorm: no strong opinions here, 👍 to the change.

astorm (approved):
I did my second pass on this and I think we're good/adequate for merging. Approving.

Just to say it out loud, given the size of the PR and the late discovery of the `[Circular]` problem I have vague concerns that our new `stacktraces.js` and `errors.js` might behave slightly differently than the old stackman module -- (ex. it doesn't look like we pulled [the stackman fixtures/tests over 1:1](https://github.com/watson/stackman/tree/master/test/fixtures/lib)) -- but those differences don't seem to be critical, and if they're significant to users/customers we can address them over time.

trentm:
Thanks for the review!

> (ex. it doesn't look like we pulled [the stackman fixtures/tests over 1:1](https://github.com/watson/stackman/tree/master/test/fixtures/lib))

FWIW, the last change to those files was 2017-03-30, and the last change to *these* files was that same date:
https://github.com/elastic/apm-agent-nodejs/tree/master/test/sourcemaps/fixtures/lib
I believe these are the same fixtures.

> I have vague concerns that our new stacktraces.js and errors.js might behave slightly differently than the old stackman module

I understand the general sentiment. It *is* a pretty big change, moving around a lot of code.
For what little it is worth, I just eyeballed all the commits to stackman.git's test/ dir going back to 4y ago and I don't see anything that we are missing in existing tests in the agent. That was from a fairly quick scan, however.

PR #2097, opened 2021-06-01
fix: missing typescript defs

astorm:
After fixing up a missing type in another PR, I noticed there were a number of configuration values used by `lib/config.js` and documented by `docs/configuration.asciidoc` that were missing from our `index.d.ts` file.   This PR adds them in.

### Checklist

- [x] Implement code
- [x] Update documentation


index.d.ts
```diff
@@ -243,6 +248,7 @@ interface AgentConfigOptions {
   transactionIgnoreUrls?: Array<string>;
   transactionMaxSpans?: number;
   transactionSampleRate?: number;
+  useElasticTraceparentHeader?: boolean;
```
trentm, line 251: This one isn't in configuration.asciidoc, FWIW.

Issue #2075, opened 2021-05-06
Align type, subtype and action for HTTP spans with SPEC


AlexanderWert:
The Node.js agent uses different subtypes for http spans (e.g. http, https, http2).
As defined in the [corresponding SPEC](https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-http.md#http-client-spans) http spans should use `external` for type and `http` for subtype.

trentm:
Per discussion on the apm-agents call earlier this week, Felix suggested this need not be considered a breaking change. Should given Dale a heads up if/when near merging this to Kibana.

trentm:
@AlexanderWert @felixbarny Should we look to add guidance to the spec on the "action" field? Currently, for outgoing http spans, Node.js is setting `action="http"`, Ruby is setting `action="GET|POST|..."`, and others are leaving it null (this per Alex's inconsistencies doc). I don't know what `action` is used for in the UI. I can ask around (in chat or one of the weeklies) if you think that is best.

trentm:
Second question: How does a user know if a span is HTTP 1, 2 (or 3 when we have QUIC integration)? An example span from the Node.js agent tests:

```
{
  id: 'a5b4d2f6d911e11c',
  transaction_id: '7d9e35adc8deba0c',
  parent_id: '7d9e35adc8deba0c',
  trace_id: '85108ecdae4d8e6e8e5011f8d5b48b10',
  name: 'GET https://localhost:55588/sub',
  type: 'external',
  subtype: 'http2',
  action: null,
  timestamp: 1623281643740481,
  duration: 2.466,
  context: {
    db: undefined,
    http: {
      method: 'GET',
      status_code: 200,
      url: 'https://localhost:55588/sub'
    },
    tags: undefined,
    destination: { service: [Object], address: 'localhost', port: 55588 },
    message: undefined
  },
  stacktrace: undefined,
  sync: false,
  outcome: 'success',
  sample_rate: 1
}
```

This is likely a question for a separate issue. :)

AlexanderWert:
> @AlexanderWert @felixbarny Should we look to add guidance to the spec on the "action" field? Currently, for outgoing http spans, Node.js is setting action="http" (actually action=null for http2), Ruby is setting action="GET|POST|...", and others are leaving it null (this per Alex's inconsistencies doc). I don't know what action is used for in the UI. I can ask around (in chat or one of the weeklies) if you think that is best.

[...]

PR #2105, opened 2021-06-09
fix: use subtype=http and action=$method for all HTTP spans

trentm:
Use subtype 'http', even for https:// and http2 requests.

Fixes: #2075

### Checklist

- [x] Implement code
- [x] Add tests


trentm:
I will add this CHANGELOG entry once the changelog updates from https://github.com/elastic/apm-agent-nodejs/pull/2108 are merged later today (hopefully):

```
* Fix inconsistencies in HTTP spans from other APM agents.
  <<span-subtype, `span.subtype`>> will now be "http" for HTTP, HTTPS, and
  HTTP/2 outgoing spans -- previously it was "http", "https", "http2",
  respectively. As well, <<span-action, `span.action`>> will now be the HTTP
  method (e.g. "GET", "PUT", "POST"), rather than "http". ({pull}2075[#2075])
```

lib/instrumentation/http-shared.js
```diff
@@ -130,7 +130,7 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
     return function (...args) {
       // TODO: See if we can delay the creation of span until the `response`
       // event is fired, while still having it have the correct stack trace
-      var span = agent.startSpan(null, 'external', moduleName, 'http')
+      var span = agent.startSpan(null, 'external', 'http')
```
astorm, line 133: nit: It looks like `span.action = req.method` could be removed by passing in req.method as the fourth argument here.

astorm (approved):
Approving -- looks like it does the thing.

lib/instrumentation/http-shared.js
trentm: We don't have `req` until we do `      var req = orig.apply(this, newArgs)` a little while *after* the span is created.

Issue #2110, opened 2021-06-15
regression on 3.16.0


sibelius:
**Describe the bug**

```jsx
node_modules/elastic-apm-node/lib/logging.js:142
...ecsFormat({ apmIntegration: false })
^

TypeError: ecsFormat is not a function
at Object.createLogger (app/node_modules/elastic-apm-node/lib/logging.js:142:8)
at new Config (app/node_modules/elastic-apm-node/lib/config.js:260:27)
at config (app/node_modules/elastic-apm-node/lib/config.js:222:10)
at Agent._config (app/node_modules/elastic-apm-node/lib/agent.js:145:16)
at new Agent (app/node_modules/elastic-apm-node/lib/agent.js:38:8)
at Object.<anonymous> (app/node_modules/elastic-apm-node/index.js:5:18)
```

**To Reproduce**

bumped from 3.15.0 to 3.16.0

**Expected behavior**

it should not break

**Environment (please complete the following information)**

- OS: [e.g. Linux] linux docker, mac os x
- Node.js version: v14.16.1
- APM Server version: 
- Agent version:

**How are you starting the agent? (please tick one of the boxes)**

- [ ] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [x] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**

- Agent config options 
  <details>
    <summary>Click to expand</summary>

    ```
    replace this line with your agent config options
    ```
  </details>
- `package.json` dependencies:
  <details>
    <summary>Click to expand</summary>

    ```
    replace this line with your dependencies section from package.json
    ```
  </details>

trentm:
@sibelius Thanks for the report. I can't yet repro. Are you able to show me the output of this command in your ".../app" dir please?

```
npm ls @elastic/ecs-pino-format
```

trentm:
In my repro attempt I have this:

```
% npm ls @elastic/ecs-pino-format
p@1.0.0 /Users/trentm/tmp/p
└─┬ elastic-apm-node@3.16.0
  └── @elastic/ecs-pino-format@1.1.2
```

trentm:
More details on an attempted repro (I can't repro yet):

```
% docker run -ti --rm node:14 /bin/bash
...
root@072aa6477f80:/# mkdir app
root@072aa6477f80:/# cd app
root@072aa6477f80:/app# npm init -y
...
[...]

PR #2111, opened 2021-06-15
fix: avoid possible circular require bug between this and ecs-pino-format

trentm:
https://github.com/elastic/ecs-logging-nodejs/issues/79 was a circular
require issue that could happen between 'elastic-apm-node' and
'@elastic/ecs-pino-format' (versions before 1.1.2) -- if user code
directly used both modules. This bumps the version of the latter to help
avoid that situation in user apps.

Refs: #2110

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   },
   "homepage": "https://github.com/elastic/apm-agent-nodejs",
   "dependencies": {
-    "@elastic/ecs-pino-format": "^1.1.0",
+    "@elastic/ecs-pino-format": "^1.1.2",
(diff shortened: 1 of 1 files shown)


astorm (approved):

trentm:
My guess is that `"@elastic/ecs-pino-format": "^1.1.0"` allowed a situation where a user had:

```
myapp
└── @elastic/ecs-pino-format@1.1.2
├─┬ elastic-apm-node
│ └── (use the top-level @elastic/ecs-pino-format because it is good enough)
...
```

PR #2114, opened 2021-06-16
fix: SQS crashing error with no active transaction

### LINKED ISSUE #2113: SQS instrumentation can cause Errors
**Describe the bug**

When a transaction is not set and and `sqs.sendMessage` is called an error is thrown:
`Cannot read property 'setDestinationContext' of null`

This was introduced in the SQS instrumentation added in version 3.14.0.

The problem is that `span` can be null here but there is no check for that case - https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/aws-sdk/sqs.js#L154-L156

image

**Expected behavior**

No error, should be handled gracefully.

- OS: Linux
- Node.js version: 16.3.0
- Agent version: 3.15.0 (present in 3.14.0 upwards)

**Additional context**

- Agent config options 
  <details>

    ```
  serviceName: 'servicename',
  secretToken: 'mysecret',
  serverUrl: 'https://myserver',
  active: true
    ```
  </details>

astorm:
Fixes #2113

This PR modifies adds an additional guard clause to the SQS instrumentation that checks to see if a span was actually started or not.  It also removed the old `currentTransaction` guard clause from `shouldIgnoreRequest` and modifies the unit test accordingly. 

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -46,7 +46,7 @@ Notes:
   method (e.g. "GET", "PUT", "POST"), rather than "http". ({pull}2075[#2075])
 
 * Fixed error where SQS messages sent without an active transactions could
-  crash the agent.
+  crash the agent. ({issues}2113[#2113])
 
 [[release-notes-x.x.x]]
 ==== 3.16.0 - 2021/06/14
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -45,6 +45,8 @@ Notes:
   respectively. As well, <<span-action, `span.action`>> will now be the HTTP
   method (e.g. "GET", "PUT", "POST"), rather than "http". ({pull}2075[#2075])
 
+* Fixed error where SQS messages sent without an active transactions could
+  crash the agent.
 
 [[release-notes-x.x.x]]
 ==== 3.16.0 - 2021/06/14
--- a/lib/instrumentation/modules/aws-sdk/sqs.js
+++ b/lib/instrumentation/modules/aws-sdk/sqs.js
@@ -130,7 +130,7 @@ function shouldIgnoreRequest (request, agent) {
     }
   }
 
-  if (!agent.currentTransaction) {
+  if (!agent.currentTransaction || agent.currentTransaction.ended) {
     agent.logger.trace('no active transaction found, skipping sqs instrumentation')
     return true
   }
--- a/lib/instrumentation/modules/aws-sdk/sqs.js
+++ b/lib/instrumentation/modules/aws-sdk/sqs.js
@@ -147,7 +147,7 @@ function instrumentationSqs (orig, origArguments, request, AWS, agent, { version
   const action = getActionFromRequest(request)
   const name = getSpanNameFromRequest(request)
   const span = agent.startSpan(name, type, subtype, action)
-  if(span) {
+  if (span) {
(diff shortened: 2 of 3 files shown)


lib/instrumentation/modules/aws-sdk/sqs.js
```diff
@@ -130,7 +130,7 @@ function shouldIgnoreRequest (request, agent) {
     }
   }
 
-  if (!agent.currentTransaction) {
+  if (!agent.currentTransaction || agent.currentTransaction.ended) {
```
trentm, line 133: @astorm What about something like this instead:

```
--- a/lib/instrumentation/modules/aws-sdk/sqs.js
+++ b/lib/instrumentation/modules/aws-sdk/sqs.js
@@ -152,6 +152,10 @@ function instrumentationSqs (orig, origArguments, request, AWS, agent, { version
   const action = getActionFromRequest(request)
   const name = getSpanNameFromRequest(request)
   const span = agent.startSpan(name, type, subtype, action)
+  if (!span) {
+    return orig.apply(request, origArguments)
+  }
+
   span.setDestinationContext(getMessageDestinationContextFromRequest(request))
   span.setMessageContext(getMessageContextFromRequest(request))
```

This is the typical technique I see in the other instrumentations, e.g.:

```
tedious.js
72:        const span = agent.startSpan(null, 'db', 'mssql', 'query')
73-        if (!span) {
74-          return super.makeRequest(...arguments)
75-        }

cassandra-driver.js
45:      const span = agent.startSpan('Cassandra: Connect', 'db', 'cassandra', 'connect')
46-      if (!span) {
47-        return original.apply(this, arguments)
48-      }

mongodb.js
44:    const span = agent.startSpan(name, 'db', 'mongodb', 'query')
45-    if (span) {
46-      activeSpans.set(event.requestId, span)
```

That allows keeping the logic for whether a span could be created inside `agent.startSpan()` instead of duplicating that logic.

trentm (changes requested):

CHANGELOG.asciidoc
```diff
@@ -45,6 +45,8 @@ Notes:
   respectively. As well, <<span-action, `span.action`>> will now be the HTTP
   method (e.g. "GET", "PUT", "POST"), rather than "http". ({pull}2075[#2075])
 
+* Fixed error where SQS messages sent without an active transactions could
+  crash the agent.
```
trentm, line 49: Nit: can we link to the issue number?

trentm (dismissed):
One nit.

Assuming tests pass, LGTM.

astorm:
jenkins run the tests please

CHANGELOG.asciidoc
astorm: 👍

lib/instrumentation/modules/aws-sdk/sqs.js
astorm: _nod_ -- change made.  We're now checking to see if a span is successfully created after calling startSpan, and if not bailing out by calling the original method.  also, dings the "something to consider for a custom linting rule at a future date" bell 🛎️ .

trentm: >  also, dings the "something to consider for a custom linting rule at a future date" bell 🛎️ .

We could re-implement the agent in Rust which would require us to handle the `Option<Span>`. ;)

trentm (approved):

Issue #2118, opened 2021-06-18
export AgentConfigOptions, Transaction, etc. from index.d.ts


trentm:
@mshustov asked:

> Is there a specific reason why this interface [`AgentConfigOptions`] is not exported? https://github.com/elastic/apm-agent-nodejs/blob/master/index.d.ts#L192
> Kibana has to work it around https://github.com/elastic/kibana/blob/4584a8b570402aa07832cf3e5b520e5d2cfa7166/packages/kbn-apm-config-loader/src/types.ts
> ...
> I think the agent could export all the public types: AgentConfigOptions, TransactionOptions, Transaction, Span, SpanOptions. that user interact with.

I don't believe there is a design reason to not export these types.

There might be a *practical* reason in the difference between [`export =`](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) and [`export default ...`](https://www.typescriptlang.org/docs/handbook/modules.html#default-exports). The agent current uses the former: https://github.com/elastic/apm-agent-nodejs/blob/a07eb3a6616d25bf91cb0cf4ce982d7c1af18be4/index.d.ts#L5
which, IIUC, is models the actual CommonJS export style: https://github.com/elastic/apm-agent-nodejs/blob/a07eb3a6616d25bf91cb0cf4ce982d7c1af18be4/index.js#L5

I don't have a solid enough experience with TypeScript to know if changing to this will have some unexpected issue:

```
export default agent
export interface AgentConfigOptions {
...
```

Experience from others is welcome.

trentm:
Some guidance here on "DTS files": https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html

mshustov:
@trentm APM-RUM agent `d.ts` file https://github.com/elastic/apm-agent-rum-js/blob/master/packages/rum/src/index.d.ts

trentm:
Notes (somewhat to self).

### option B: define everything in index.d.ts like we have now.

This seems to work, but I don't see a way to export "Agent" itself.

```typescript
declare interface Agent {
  start (options?: agent.AgentConfigOptions): Agent;
  isStarted (): boolean;
  // ...
}

declare const agent: Agent;
export = agent

// Re-export some public API types.
declare namespace agent {
  export interface AgentConfigOptions {
    abortedErrorThreshold?: string;
    active?: boolean;
    captureExceptions?: boolean;
    serviceName?: string;
    // ...
  }
[...]

PR #2119, opened 2021-06-18
feat: export more TypeScript types for Public APIs

trentm:
Fixes: #2118

### Notes for reviewers

- The diff on index.d.ts is *much* smaller if whitespace is ignored: https://github.com/elastic/apm-agent-nodejs/pull/2119/files?w=1

### Checklist

- [x] Add tests


trentm:
https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html#default-exports describes a way to still use `export =` style without requiring consumers to use `esModuleInterop`:

> If you can’t have esModuleInterop: true in your project, such as when you’re submitting a PR to Definitely Typed, you’ll have to use the export= syntax instead. This older syntax is harder to use but works everywhere. Here’s how the above example would have to be written using export=: ...

I wonder if that would be preferred here.

index.d.ts
```diff
@@ -1,426 +1,21 @@
-/// <reference types="node" />
```
trentm, line 1: REVIEW NOTE: This diff is huge because I've moved most of the types to lib/agent.d.ts and this file now just creates the `agent` singleton and re-exports some types from agent.d.ts.  A better representation of the diff is the change from old index.d.ts to new agent.d.ts:  https://gist.github.com/trentm/67b13e5c99abf2ca8001a98e7dd7416e

trentm:
@mshustov I am on PTO next week, but if you have the time, you could try out the current state of this PR. `AgentConfigOptions` (and other interfaces requested in https://github.com/elastic/apm-agent-nodejs/issues/2118) are now exported in the types file.

mshustov:
@trentm answered in the root issue https://github.com/elastic/apm-agent-nodejs/issues/2118#issuecomment-878338293

index.d.ts
```diff
@@ -1,426 +1,339 @@
...
+    ignoreUserAgents?: Array<string | RegExp>;
+    instrument?: boolean;
+    instrumentIncomingHTTPRequests?: boolean;
+    kubernetesNamespace?: string;
+    kubernetesNodeName?: string;
+    kubernetesPodName?: string;
+    kubernetesPodUID?: string;
+    logLevel?: LogLevel;
+    logUncaughtExceptions?: boolean;
+    logger?: PinoLogger | Logger;
```
trentm, line 237: REVIEWER NOTE: This is an independent change that could be done in a separate commit. A while back we added support for the given logger being a pino logger instance.

index.d.ts
```diff
@@ -1,426 +1,339 @@
...
+    stackTraceLimit?: number;
+    transactionIgnoreUrls?: Array<string>;
+    transactionMaxSpans?: number;
+    transactionSampleRate?: number;
+    useElasticTraceparentHeader?: boolean;
+    usePathAsTransactionName?: boolean;
+    verifyServerCert?: boolean;
+  }
 
-interface Taggable {
```
trentm, line 330: REVIEWER NOTE: interfaces `Taggable` and `StartSpanFn` served no purpose. They are not meaningful to external users. They are not constructs in the actual code, and the type definitions here weren't actually using the inheritance: the `setLabel` et al functions were still defined on the interfaces above.

index.d.ts
```diff
@@ -1,426 +1,339 @@
...
-    action: string | null,
-    options?: SpanOptions
-  ): Span | null;
-}
+  interface Labels {
+    [key: string]: LabelValue;
+  }
 
-// Inlined from @types/aws-lambda - start
-declare namespace AwsLambda {
```
trentm, line 361: REVIEWER NOTE: I don't see the value in inlining these rather than include "@types/aws-lambda" in devDependencies (as we do for node and pino types).

trentm:
@mshustov This is ready to try again, if you are able to give it a go. The issue with re-exporting that you [identified in your last comment](https://github.com/elastic/apm-agent-nodejs/issues/2118#issuecomment-878338293) is dealt with now.

astorm (approved):
👍 no objections, and not in-lining the AWS Lambda types makes sense to me (it does make me curious why they were in-lined in the first place).  Approving.

All that said, I'm not a working TypeScript programmer so feedback from @mshustov (or otherwise) would be a wise thing.

trentm:
(I've added a changelog entry and merged from master since Alan's review. I'll assume this still passes review.

mshustov:
@trentm I will update the agent version in the Kibana repo and give it a shot. Which version of the agent will include these changes?

trentm:
@mshustov There isn't an agent release with this yet, because I just merged this yesterday. I'll ping you here once we have a release with this change in it.  (Does it matter to you if that isn't for a week or so, yet? I can do a release sooner if that helps.)

mshustov:
> Does it matter to you if that isn't for a week or so

No, it can wait. We are several versions behind anyway. I just sent a PR to update Kibana deps to the latest version of agents https://github.com/elastic/kibana/pull/109877

trentm:
> Which version of the agent will include these changes?

Version 3.21.0 of the agent now has it.

PR #2123, opened 2021-06-22
ci: drop testing with node v15

trentm:
It is no longer a supported release:
https://nodejs.org/en/about/releases/

@astorm What do you think? The only real value I see in testing odd node majors is the next one *before* there are releases of the next even release. For example, once there is a v17 release then we should start testing v17 releases up until https://nodejs.org/en/about/releases/ says it is no longer supported.

### Checklist

- [x] Implement code


astorm (approved):
Approving. Keeping the old odds around doesn't make sense to me.  As you said, the main value of testing vs. them is when they're under active development.

PR #1974, opened 2021-02-09
investigating 'docker-compose down ...' failures in CI

trentm:
This PR is me playing with some issues with our CI being discussed in https://github.com/elastic/apm-agent-nodejs/pull/1971

**Update (20210430):** The main symptom being looked at here was tests running in docker failing at the the `docker-compose down ...` step with `error while removing network: network docker_default id cfb384f605476f50bdd4272a0a8b0e82463d64fabcca849e47f7e3e1ba24b326 has active endpoints`

PR #2142, opened 2021-07-05
ci: tear down docker network to retry explicitly

v1v:
### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### What

The docker compose environment failed when tearing down the network. Since there is a retry in this particular stage in the CI, it's needed to happen otherwise the next iteration will fail.

Let's run an explicit `docker network prune --force` before running the `docker-compose down` since the later does not support a `force` feature.

### Why

Reduce any environmental flakiness.

Log errors can be seen in https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/master/963/pipeline/#step-600-log-1841

```
[2021-07-05T06:08:58.530Z] + docker-compose --no-ansi --log-level ERROR -f .ci/docker/docker-compose-all.yml down -v --remove-orphans
[2021-07-05T06:08:59.471Z] Stopping docker_mssql_1         ... 
...
[2021-07-05T06:09:10.904Z] Removing docker_mongodb_1       ... done
[2021-07-05T06:09:10.904Z] error while removing network: network docker_default id 8731047bca1c48562f34b7a5016ebcbd46d36fe4b1d1354ddba24ad812af506d has active endpoints

```

### Tests

image

### DIFF
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -250,6 +250,9 @@ docker-compose \
   --abort-on-container-exit \
   node_tests
 
+echo "Workaround issues with docker-compose down while removing network"
+docker network prune --force || true
+
 NODE_VERSION=${NODE_VERSION} docker-compose \
   --no-ansi \
   --log-level ERROR \
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -250,9 +250,6 @@ docker-compose \
   --abort-on-container-exit \
   node_tests
 
-echo "Workaround issues with docker-compose down while removing network"
-docker network prune --force || true
-
 NODE_VERSION=${NODE_VERSION} docker-compose \
   --no-ansi \
   --log-level ERROR \
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -365,8 +365,6 @@ def generateStep(Map params = [:]){
             retryWithSleep(retries: 2, seconds: 5, backoff: true) {
               sh(label: "Run Tests", script: """.ci/scripts/test.sh -b "${buildType}" -t "${tav}" "${version}" """)
             }
-          } catch(e){
-            error(e.toString())
           } finally {
             // TAV tests run 'npm run test:tav' which does *not* produce TAP
             // output.
--- a/.ci/docker/docker-compose-all.yml
+++ b/.ci/docker/docker-compose-all.yml
@@ -76,3 +76,7 @@ volumes:
     driver: local
   nodelocalstackdata:
     driver: local
+
+networks:
+  apm-agent-nodejs:
+    driver: bridge
--- a/.ci/docker/docker-compose-cassandra.yml
+++ b/.ci/docker/docker-compose-cassandra.yml
@@ -16,3 +16,7 @@ services:
 volumes:
   nodecassandradata:
     driver: local
+
+networks:
+  apm-agent-nodejs:
+    driver: bridge
--- a/.ci/docker/docker-compose-edge.yml
+++ b/.ci/docker/docker-compose-edge.yml
@@ -76,3 +76,7 @@ volumes:
     driver: local
   nodecassandradata:
     driver: local
+
+networks:
+  apm-agent-nodejs:
+    driver: bridge
--- a/.ci/docker/docker-compose-elasticsearch.yml
+++ b/.ci/docker/docker-compose-elasticsearch.yml
@@ -16,3 +16,7 @@ services:
 volumes:
   nodeesdata:
     driver: local
+
+networks:
+  apm-agent-nodejs:
+    driver: bridge
--- a/.ci/docker/docker-compose-localstack.yml
+++ b/.ci/docker/docker-compose-localstack.yml
@@ -16,3 +16,7 @@ services:
 volumes:
   nodelocalstackdata:
     driver: local
+
+networks:
+  apm-agent-nodejs:
+    driver: bridge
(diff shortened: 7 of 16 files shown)


.ci/scripts/test.sh
```diff
@@ -250,6 +250,9 @@ docker-compose \
   --abort-on-container-exit \
   node_tests
 
+echo "Workaround issues with docker-compose down while removing network"
+docker network prune --force || true
```
trentm, line 254: Does this actually force removal of the `docker_default` network? 

```
% docker network prune --help

Usage:  docker network prune [OPTIONS]

Remove all unused networks

Options:
      --filter filter   Provide filter values (e.g. 'until=<timestamp>')
  -f, --force           Do not prompt for confirmation
```

That command says it removes *unused* networks. Before `docker-compose down` is called, if I understand correctly, the `docker_default` network will still be in used.

trentm: A possible alternative here would be something based on this comment: https://github.com/elastic/apm-agent-nodejs/pull/1974#issuecomment-830275247

For example, something like this:

```diff
diff --git a/.ci/scripts/test.sh b/.ci/scripts/test.sh
index 26af1d4e..9d39867e 100755
--- a/.ci/scripts/test.sh
+++ b/.ci/scripts/test.sh
@@ -250,8 +250,15 @@ docker-compose \
   --abort-on-container-exit \
   node_tests

-NODE_VERSION=${NODE_VERSION} docker-compose \
-  --no-ansi \
-  --log-level ERROR \
-  -f .ci/docker/${DOCKER_COMPOSE_FILE} \
-  down -v --remove-orphans
+if ! NODE_VERSION=${NODE_VERSION} docker-compose \
+    --no-ansi \
+    --log-level ERROR \
+    -f ${DOCKER_FOLDER}/${DOCKER_COMPOSE_FILE} \
+    down -v; then
+  # Workaround for this commonly seen error:
+  #   error while removing network: network docker_default id $id has active endpoints
+  echo "Unexpected error in 'docker-compose down ...'. Forcing removal of unused networks."
+  docker network inspect docker_default || true
+  docker network inspect -f '{{range .Containers}}{{ .Name }} {{end}}' docker_default || true
+  docker network prune --force || true
+fi
```

Thoughts?

kuisathaverat:
why not `docker-compose -f ci/docker/docker-compose-all.yml down`? this should remove all provisioned resources

v1v:
> why not docker-compose -f ci/docker/docker-compose-all.yml down? this should remove all provisioned resources

That's happening already but the `docker_default` network sometimes gets stalled :( 

> Does this actually force removal of the docker_default network?

I actually got mislead with the command `docker network prune --force`, and assumed it did remove everything when using the flag `--force` :/ 

Maybe to force a network entry in the docker-compose files could help in this case?

kuisathaverat:
>Maybe to force a network entry in the docker-compose files could help in this case?

sure, the resource will be only used by the resources you are deprovisioning.

kuisathaverat (approved):

trentm:
Do any of you have prior experience suggesting explicitly listing a network like this helps? I haven't used docker-compose enough to have good experience with this.

It is worth a try if we think it will help.

v1v:
There is a known issue with the provisioner that allocates existing workers in some cases rather than allocating a new ephemeral worker. Maybe that's the reason we see this issue from time to time.

Good question, the change should work without affecting. Worst case scenario we face the same issue. Though, if you don't feel this is the way you'd like to move forward, I'm happy to close this pull request

kuisathaverat:
> Do any of you have prior experience suggesting explicitly listing a network like this helps? I haven't used docker-compose enough to have good experience with this.
> 
> It is worth a try if we think it will help.

On the APM ITs we have a similar process that starts and stop docker-compose we do not have issues destroying the resources. On the APM ITs we use a named network, this is the pattern we have in others docker-compose, I think this is the only one that does not have a named network.

trentm:
> I think this is the only one that does not have a named network.

I don't see a network listed in https://github.com/elastic/apm-agent-python/blob/master/tests/docker-compose.yml
or https://github.com/elastic/apm-agent-go/blob/master/scripts/docker-compose-testing.yml or https://github.com/elastic/apm-agent-js-core/blob/master/test/docker-compose.yml but I don't trust my reading of docker compose config files. :)

> Worst case scenario we face the same issue.

I think it is worth a try.

---

I had also proposed some extra code around `docker-compose down`[ in a comment above](https://github.com/elastic/apm-agent-nodejs/pull/2142#discussion_r664056450) to attempt to get some info on what is using that default network. Did you have thoughts on trying that?

trentm:
Note that in https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/PR-2143/8/pipeline we hit 

```
[2021-07-08T16:52:58.231Z] error while removing network: network docker_default id d78cf221351a77d301c6593e917e6b7b80a16226eff5d7f2fdcb6c26d28f42b6 has active endpoints
script returned exit code 1
```

again. This PR branch included the https://github.com/elastic/apm-agent-nodejs/pull/2145 change.

v1v:
I'll close this PR now, feel free to reopen it if needed

.ci/scripts/test.sh
cachedout: This seems like it is likely to work and there's certainly no harm in giving it a try. 👍

trentm: Attempting this here: https://github.com/elastic/apm-agent-nodejs/pull/2184

PR #2163, opened 2021-07-20
fix: remove zero count metrics from registry 

### LINKED ISSUE #1977: MetricsRegistry and High Cardinality

astorm:
Closes #1977

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/metrics/reporter.js
+++ b/lib/metrics/reporter.js
@@ -31,10 +31,10 @@ class MetricsReporter extends Reporter {
         // due to `metricsLimit` leave empty slots in the list.
         if (!metric) continue
 
-        // if (this.isStaleMetric(metric)) {
-        //   this.removeMetricFromRegistry(metric, this._registry)
-        //   continue
-        // }
+        if (this.isStaleMetric(metric)) {
+          this.removeMetricFromRegistry(metric, this._registry)
+          continue
+        }
 
         const data = seen.ensure(metric.dimensions, () => {
           const metricData = unflattenBreakdown(metric.dimensions)
--- a/lib/metrics/reporter.js
+++ b/lib/metrics/reporter.js
@@ -6,6 +6,7 @@ const ObjectIdentityMap = require('object-identity-map')
 
 class MetricsReporter extends Reporter {
   constructor (agent, options = {}) {
+    options.defaultReportingIntervalInSeconds = 5
     super(options)
     this.enabled = options.enabled
     this._agent = agent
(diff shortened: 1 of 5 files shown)


test/metrics/breakdown.test.js
```diff
@@ -144,7 +159,7 @@ test('does not include breakdown when not sampling', t => {
 
     const { metricsets } = data
 
-    assertMetricSet(t, 'transaction', metricsets, {
+    assertMetricSet(t, 'transaction not sampling', metricsets, {
```
astorm, line 162: Previously this test asserted that `transaction.breakdown.count` was reported.  However, with the new behavior the `transaction.breakdown.count` metric is no longer reported _in this specific scenario_ [because its count is zero](https://github.com/elastic/apm-agent-nodejs/blob/28e77afa0ea542094e4b655dea713c6769c63880/lib/instrumentation/transaction.js#L322).

astorm:
jenkins run the tests please

lib/metrics/reporter.js
```diff
@@ -55,6 +61,23 @@ class MetricsReporter extends Reporter {
...
+  isStaleMetric (metric) {
+    // if a metric is a counting metric and that count is
+    // zero, then the metric is considered stale
+    if (metric.metricImpl && metric.metricImpl._count === 0) {
+      return true
+    }
+    return false
+  }
+
+  removeMetricFromRegistry (metric) {
```
trentm, line 74: Unfortunate the node-measured libs are not more usable here (e.g. no `deleteMetric` on https://github.com/yaorg/node-measured/blob/master/packages/measured-reporting/lib/registries/DimensionAwareMetricsRegistry.js and no `getValue()` or something accessor for Counter._count), but that's not at all a job for this PR.

trentm (commented):
Just the changelog nit. Otherwise looks perfect.

lib/metrics/reporter.js
astorm: yeah, I didn't like dipping into the private-ish APIs to accomplish this -- but the library left me with no choice.

trentm (approved):

Issue #2170, opened 2021-07-27
mongodb 4.0.x causes startup to hang/fail


stevenhair:
**Describe the bug**

With mongodb 4.0.1 installed, application startup will either hang or fail.

**To Reproduce**

Steps to reproduce the behavior: 
1. Add elastic-apm-node to a project that uses mongodb 4 (or upgrade an existing one)
2. Start the application

**Expected behavior**

The application should start

**Environment (please complete the following information)**

- OS: CentOS 7/macOS 11.4
- Node.js version: 14.7.3
- APM Server version: N/A
- Agent version: 3.18.0

**How are you starting the agent? (please tick one of the boxes)**

- [ ] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [X] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**

- Agent config options 
  <details>
    <summary>Click to expand</summary>

    ```
    module.exports = {
      serverUrl: 'https://elastic-apm.mydomain.com/',
      secretToken: '...',
      environment: 'dev',
      captureBody: 'all',
      centralConfig: false,
      serviceName: 'MyService',
      active: true,
      logLevel: 'debug',
    };
    ```
  </details>
- `package.json` dependencies:
  <details>
    <summary>Click to expand</summary>

    ```
    "@nestjs/axios": "0.0.1",
    "@nestjs/common": "^8.0.4",
    "@nestjs/core": "^8.0.4",
    "@nestjs/platform-express": "^8.0.4",
    "@nestjs/schedule": "^1.0.0",
    "amqplib": "^0.8.0",
    "axios": "^0.21.1",
    "body-parser": "^1.18.2",
    "cookie-parser": "^1.4.5",
    "csv-parse": "^4.15.0",
    "date-fns": "^2.21.0",
    "elastic-apm-node": "^3.18.0",
    "express": "^4.16.2",
    "express-openapi-validator": "^4.12.15",
    "hs-fuzzy-dates": "1.0.1",
    "is-base64": "^1.1.0",
    "minimist": "^1.2.5",
    "moize": "^6.0.3",
    "mongodb": "^4.0.1",
    "pako": "^2.0.3",
    "queue-promise": "^2.2.1",
    "rascal": "^13.1.0",
    "reflect-metadata": "^0.1.13",
    "rxjs": "^7.2.0",
[...]

PR #2171, opened 2021-07-28
fix: adds supports for MongoDB 4.0+

astorm:
Fixes: https://github.com/elastic/apm-agent-nodejs/issues/2170

This PR adds back support for the recently release MongoDB 4.0+, and adds a `<5` constraint to the version check on the MongoDB instrumentation module. TAV test constraint remains at `>3.3` in order to catch problems with a theoretical 5.0 in the future. 

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/modules/mongodb.js
+++ b/lib/instrumentation/modules/mongodb.js
@@ -8,34 +8,39 @@ const HOSTNAME_PORT_RE = /^([^:]+):(\d+)$/
 
 module.exports = (mongodb, agent, { version, enabled }) => {
   if (!enabled) return mongodb
-  if (!semver.satisfies(version, '>=3.3')) {
+  if (!semver.satisfies(version, '>=3.3 < 5.0')) {
     agent.logger.debug('mongodb version %s not instrumented (mongodb <3.3 is instrumented via mongodb-core)', version)
     return mongodb
   }
 
   const activeSpans = new Map()
-  if(mongodb.instrument) {
+  if (mongodb.instrument) {
     const listener = mongodb.instrument()
 
     listener.on('started', onStart)
     listener.on('succeeded', onEnd)
     listener.on('failed', onEnd)
-  } else {
+  } else if (mongodb.MongoClient) {
+    // mongo 4.0+ removed the instrument() method in favor of
+    // listeners on the instantiated client objects.
     class MongoClient extends mongodb.MongoClient {
-      constructor(url, options) {
-        super(url, options)
+      constructor () {
+        super(...arguments)
+        this.on('commandStarted', onStart)
+        this.on('commandSucceeded', onEnd)
+        this.on('commandFailed', onEnd)
       }
     }
     Object.defineProperty(
       mongodb,
-      "MongoClient",
+      'MongoClient',
       {
         enumerable: true,
         get: function () {
           return MongoClient
         }
       }
-    );
+    )
   }
   return mongodb
 
--- a/lib/instrumentation/modules/mongodb.js
+++ b/lib/instrumentation/modules/mongodb.js
@@ -8,7 +8,7 @@ const HOSTNAME_PORT_RE = /^([^:]+):(\d+)$/
 
 module.exports = (mongodb, agent, { version, enabled }) => {
   if (!enabled) return mongodb
-  if (!semver.satisfies(version, '>=3.3 < 5.0')) {
+  if (!semver.satisfies(version, '>=3.3 <5.0')) {
     agent.logger.debug('mongodb version %s not instrumented (mongodb <3.3 is instrumented via mongodb-core)', version)
     return mongodb
   }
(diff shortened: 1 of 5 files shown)


lib/instrumentation/modules/mongodb.js
```diff
@@ -8,18 +8,40 @@ const HOSTNAME_PORT_RE = /^([^:]+):(\d+)$/
 
 module.exports = (mongodb, agent, { version, enabled }) => {
   if (!enabled) return mongodb
-  if (!semver.satisfies(version, '>=3.3')) {
+  if (!semver.satisfies(version, '>=3.3 < 5.0')) {
```
trentm, line 11: nit:

```suggestion
  if (!semver.satisfies(version, '>=3.3 <5.0')) {
```

If you bother to take this suggestion, don't bother waiting for a test re-run to pass on my account.

lib/instrumentation/modules/mongodb.js
```diff
@@ -8,18 +8,40 @@ const HOSTNAME_PORT_RE = /^([^:]+):(\d+)$/
...
+      mongodb,
+      'MongoClient',
+      {
+        enumerable: true,
+        get: function () {
+          return MongoClient
+        }
+      }
+    )
+  }
```
trentm, line 44: Was there a reason to add this via defineProperty rather than something like:

```js
mongodb.MongoClient = MongoClient
// or
return Object.assign(mongodb, { MongoClient: MongoClient })
```

trentm (approved):
All nits and Qs.

- Thoughts on bumping ` "mongodb": "^3.5.3",` in package.json devDependencies up to "^4.x" or something so that local dev testing is using the latest supported major?


Kind of painful that the [changelog](https://github.com/mongodb/node-mongodb-native/blob/4.0/HISTORY.md) doesn't give a hint (that I could see) of this API removal. 3.x docs had an APM page (https://mongodb.github.io/node-mongodb-native/3.6/reference/management/apm/). Ah, this is the current equivalent: https://docs.mongodb.com/drivers/node/current/fundamentals/monitoring/command-monitoring/

lib/instrumentation/modules/mongodb.js
astorm: Good question.  That was my first instinct as well.  However, this was the result

> TypeError: Cannot set property MongoClient of `#<Object>` which has only a getter

The mongo library/driver uses `defineProperty` to set those 

```
    #File: node_modules/mongodb/lib/index.js
    /* ... */
    Object.defineProperty(exports, "MongoClient", { enumerable: true, get: function () { return mongo_client_1.MongoClient; } });
    /* ... */
```

which means (or appears to mean?) that we need to redefine the property.

(Also, it's worth noting that it looks like MongoDB has moved to a pure typescript implementation, and the commonjs libraries are the results of compilation.)

trentm: Ah, fair. This sounds part of the path I went through with `@aws-sdk/client-s3` described at https://github.com/elastic/apm-agent-nodejs/blob/35a5a56b9b58e079740d15adedf2f7fad80ea1f6/lib/instrumentation/modules/%40aws-sdk/client-s3.js#L178-L221



> (Also, it's worth noting that it looks like MongoDB has moved to a pure typescript implementation, and the commonjs libraries are the results of compilation.)

Ah. Hence the `Object.defineProperty(exports, "MongoClient",  ...` for properties. I'm slightly curious why defineProperty worked for you on `mongodb` but did not for me for client-s3. Hrm.

astorm: > I'm slightly curious why defineProperty worked for you on mongodb but did not for me for client-s3. Hrm.

An excellent nerd snipe.  It looks like it's this line that's the difference.

> Gist of compiled lib/index.js from mongodb package
> https://gist.github.com/astorm/bfe2ed7ac34445046867000176693c8e#file-mongodb-lib-index-js-L3

It seems like most/every exported property is first assigned a value of `void 0`.  This appears to have the side effect of making the property [configurable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty).  Without this line I also see the `TypeError: Cannot redefine property:` errors. 

So Re _worked for you on mongodb but did not for me for client-s3_ -- luck and compiler flags.

trentm (approved):

lib/instrumentation/modules/mongodb.js
trentm: WAT. Wow, excellent snipe response.

curtdept:
@astorm you going to tag a new release for this fairly soon?

astorm:
@curtdept tentative plans for a release next week.  Out of curiosity, did you run into this issue as well?

curtdept:
> @curtdept tentative plans for a release next week. Out of curiosity, did you run into this issue as well?

We have, yes. We used a third party package for a job scheduling UI called agendash which pulls in the mongo 4 client.

trentm:
@curtdept elastic-apm-node@3.19.0 has now been published with this.

Issue #2172, opened 2021-08-02
Master branch tests broken


cachedout:
Hello!

It seems [that the master branch tests are broken](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/master/1000/pipeline) when testing with Apollo Server Express.

The relevant log lines appear to be as follows:

```
[2021-08-02T06:52:37.424Z] node_tests_1  | -- running test "node test/instrumentation/modules/apollo-server-express.test.js" with apollo-server-express

[2021-08-02T06:52:37.424Z] node_tests_1  | /app/test/instrumentation/modules/apollo-server-express.test.js:50

[2021-08-02T06:52:37.424Z] node_tests_1  |   apollo.start().then(function () {

[2021-08-02T06:52:37.424Z] node_tests_1  |          ^

[2021-08-02T06:52:37.424Z] node_tests_1  | 

[2021-08-02T06:52:37.424Z] node_tests_1  | TypeError: apollo.start is not a function

[2021-08-02T06:52:37.424Z] node_tests_1  |     at Test.<anonymous> (/app/test/instrumentation/modules/apollo-server-express.test.js:50:10)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at Test.bound [as _cb] (/app/node_modules/tape/lib/test.js:96:32)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at Test.run (/app/node_modules/tape/lib/test.js:115:31)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at Test.bound [as run] (/app/node_modules/tape/lib/test.js:96:32)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at Immediate.next (/app/node_modules/tape/lib/results.js:88:19)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at runCallback (timers.js:705:18)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at tryOnImmediate (timers.js:676:5)

[2021-08-02T06:52:37.424Z] node_tests_1  |     at processImmediate (timers.js:658:5)

[2021-08-02T06:52:37.424Z] node_tests_1  | -- detected failing command, flushing stdout...

```

astorm:
Local Run:

% TAV=apollo-server-express ./node_modules/.bin/tav --compat
Testing compatibility with apollo-server-express:
✔ 3.1.1
[...]

PR #2173, opened 2021-08-02
fix: remove older tav versions

astorm:
Fixes: #2172 

This PR removes older versions of Apollo Server from the TAV test matrix in order to make the test cases compatible for both apollo-server-express@3.x and apollo-server-express@2.x

The apollo-server-express@3.x tests require us to call `apollo.start` to start the apollo server. 

However, the some older version of apollo-server-express@2.x  (< 2.8.2) don't have a `start` method.

This PR removed these over two years old versions of apollo server from the test matrix in order to get a clean run.  Other options include

1. Breaking off separate TAV suites for apollo < 2.8.2
2.Rewriting the apollo tests to detect versions and use `listen` or `start` accordingly to start the server. 

### Checklist

- [x] Implement code


astorm:
@trentm Re:

>  If not, then I'd bias to still testing 2.x (perhaps not back to 2.0).

Agreed.  The version we're cutting off at (2.8.2) was released almost two years ago (23 months and some change)

```
% npm view apollo-server-express time | grep 2.8.2
  '2.8.2': '2019-08-20T18:03:00.440Z',
```

.tav.yml
```diff
@@ -221,7 +221,7 @@ apollo-server-express-2_12:
   # but only the latest MAJOR.MINOR.x to reduce the test matrix.
```
trentm, line 221: nit: can we update the "We want this version range: ..." comment to say why we are clipping at >2.8?

trentm (approved):

.tav.yml
trentm: meh, nevermind. Encoded in the PR discussion here.

trentm:
Some context for:

> If not, then I'd bias to still testing 2.x (perhaps not back to 2.0).

This was about proposing to allow dropping TAV tests of versions older than 18months, somewhat in line with https://www.elastic.co/support/eol policy.

.tav.yml
astorm: Update the versions in the comments, thanks for catching that.

package.json
```diff
@@ -137,7 +137,6 @@
     "dependency-check": "^4.1.0",
     "elasticsearch": "^16.5.0",
     "express": "^4.17.1",
-    "express-graphql": "^0.9.0",
```
trentm, line 140: Was this unintentional? Won't that break:

```
test/config.test.js
715:  var expressGraphqlVersion = require('express-graphql/package.json').version
727:    modules.delete('express-graphql')

test/instrumentation/modules/express-graphql.test.js
17:var graphqlHTTP = require('express-graphql')
```

?

trentm (approved):

Issue #2177, opened 2021-08-04
hang in mongodb tests with mongodb@4.1.0


trentm:
There is a frequent hang in the mongodb tests. The recent #2171 that added MongoDB v4 support is a likely suspect. Reproducible (at least for me on my first attempt) via `.ci/scripts/test.sh -b "release" -t "" "10"` on my Mac. 

Note that tests *did* pass for #2171 initially in Jenkins. Also the same test run in GitHub Actions on today's latest master passed: https://github.com/elastic/apm-agent-nodejs/runs/3235248141

Jenkins builds 1002 and 1003 at https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/master/ have hit it. For example the "Node.js-10" test stage (https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp/detail/master/1003/pipeline/164) has this at the end of its log for the `.ci/scripts/test.sh -b "release" -t "" "10" ` build step:

```
[2021-08-04T05:56:21.405Z] node_tests_1     | running (cwd: ./): node --require /app/test/_promise_rejection.js test/instrumentation/modules/mongodb.test.js
[2021-08-04T05:56:23.939Z] node_tests_1     | TAP version 13
[2021-08-04T05:56:23.940Z] node_tests_1     | # instrument simple command
[2021-08-04T05:56:23.940Z] node_tests_1     | ok 1 no connect error
[2021-08-04T05:56:23.940Z] node_tests_1     | ok 2 got a valid server connection
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 3 no insert error
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 4 inserted three records
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 5 no update error
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 6 updated one record
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 7 no delete error
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 8 deleted one record
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 9 no cursor next error
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 10 found record #2
[2021-08-04T05:56:24.198Z] node_tests_1     | ok 11 no cursor next error
[...]

PR #2178, opened 2021-08-04
fix: ensure instrumentation of mongodb >=4.1.0 works

trentm:
In mongodb@4.1.0 (https://jira.mongodb.org/browse/NODE-3513) the
monitorCommands option was reverted to defaulting to false. To
instrument we ensure it is true.

Fixes: #2177

### DIFF
--- a/lib/instrumentation/modules/mongodb.js
+++ b/lib/instrumentation/modules/mongodb.js
@@ -21,10 +21,16 @@ module.exports = (mongodb, agent, { version, enabled }) => {
     listener.on('succeeded', onEnd)
     listener.on('failed', onEnd)
   } else if (mongodb.MongoClient) {
-    // mongo 4.0+ removed the instrument() method in favor of
+    // mongodb 4.0+ removed the instrument() method in favor of
     // listeners on the instantiated client objects.
     class MongoClient extends mongodb.MongoClient {
       constructor () {
+        // The `command*` events are only sent if `monitorCommands: true`.
+        if (!arguments[1]) {
+          arguments[1] = { monitorCommands: true }
+        } else if (arguments[1].monitorCommands !== true) {
+          arguments[1] = Object.assign({}, arguments[1], { monitorCommands: true })
+        }
         super(...arguments)
         this.on('commandStarted', onStart)
         this.on('commandSucceeded', onEnd)
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
     "memcached": "^2.2.2",
     "mimic-response": "^2.1.0",
     "mkdirp": "^0.5.1",
-    "mongodb": "^4.0.0",
+    "mongodb": "^4.1.0",
     "mongodb-core": "^3.2.7",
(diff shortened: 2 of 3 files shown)


lib/instrumentation/modules/mongodb.js
```diff
@@ -21,10 +21,16 @@ module.exports = (mongodb, agent, { version, enabled }) => {
...
+    // mongodb 4.0+ removed the instrument() method in favor of
     // listeners on the instantiated client objects.
     class MongoClient extends mongodb.MongoClient {
       constructor () {
+        // The `command*` events are only sent if `monitorCommands: true`.
+        if (!arguments[1]) {
+          arguments[1] = { monitorCommands: true }
+        } else if (arguments[1].monitorCommands !== true) {
+          arguments[1] = Object.assign({}, arguments[1], { monitorCommands: true })
+        }
```
astorm, line 33: If I'm reading this right, it looks like we're removing the ability for an end user to disable the `monitorCommands` events entirely.  This means if a user wanted to disable the mongodb events for reasons (perf., etc.) but still take advantage of other agent features they couldn't.  

Do you think it's worth/possible to handle the above use case by keeping the `if (!arguments[1]) {` conditional, but dropping the `else if (arguments[1].monitorCommands !== true) {` and honoring explict requests to turn off the command events?  Or would that complicate things?

astorm (changes requested):

lib/instrumentation/modules/mongodb.js
trentm: > This means if a user wanted to disable the mongodb events for reasons (perf., etc.) but still take advantage of other agent features they couldn't.

They could still set `disableInstrumentations=mongodb` config to the APM agent to disable.

* * *

It would certainly be possible to honour an explicit `arguments[1].monitorCommands === false`. That would allow a (savvy) user to be able to have APM instrumentation on one `MongoClient()` instance and not on another, if they had a use case for that. 

The *fabricated and unlikely* "use case" that will be broken with this change would be the ops person deploying an app for which they want mongodb APM tracing, but the app code has a lingering `{monitorCommands: false}` for whatever reason.

The mongodb code (nothing mentioned in the docs): https://github.com/mongodb/node-mongodb-native/blob/4ecaa37f72040ed8ace6eebc861b43ee9cb32a99/src/mongo_client.ts#L369-L378 says "Intended for APM use only" for the accessors of this config option. That slightly implies it is unexpected that an end user would be fiddling with this config option.

I'm slightly biased against adding the subtlety here, but not strongly. I'm happy to make the change (to honour `arguments[1].monitorCommands === false`) if you prefer. Thoughts?

astorm (approved):
👍 Looks good from here. This will fix the tests and ensures that the instrumentation is automatically on regardless of what the mongodb core code sets `monitorCommands` to.  Per offline discussion with @trentm if users _want_ to disable `monitorCommands` they can set `disableInstrumentations=mongodb` in the agent configuration.

PR #2179, opened 2021-08-05
3.19.0

trentm:
@astorm Thoughts on the change here clearing out old 1.x and 2.x entries from the EOL table?


astorm (approved):
> Thoughts on the change here clearing out old 1.x and 2.x entries from the EOL table?

So long as we have the final versions and EOL dates to point to (as we do) clearing these out seems fine. 

```
|2.17.x |2021-03-19 |(now EOL)
|1.14.x |2020-05-06 |(now EOL)
```

approved pending ci passing

trentm:
v3.19.0 tagged
elastic-apm-node@3.19.0 published to npm

Issue #147, opened 2018-01-02
Add `transaction.id` to all errors


watson:
Associate all logged errors with the transaction that was active at the time the error occurred.

This is done by adding the following property and object to the root of the error JSON document sent to the intake API:

```json
"transaction": {
    "id": "<uuid>"
}
```

Related APM Server PR: elastic/apm-server/pull/437

PR #2181, opened 2021-08-06
feat: new "run context" management to fix numerous span hierarchy issues

trentm:
### Status

This is still very much draft work.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -128,10 +128,19 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
 
   return function (orig) {
     return function (...args) {
+      // XXX See if can avoid instrumenting request to APM server... would
+      //     be nice if could be more explicit from the APM server client code
+      //     itself. Special header? Use uninstrumented endpoint (would need
+      //     shimmer to expose access to the original via a method)? Invoke
+      //     in a separate no-op RunContext???
+      console.warn('XXX traceOutgoingRequest: start')
+      // console.warn('XXX curr span:', agent._instrumentation.currentSpan?.name) // considers 'bindingSpan'
+
       // TODO: See if we can delay the creation of span until the `response`
       // event is fired, while still having it have the correct stack trace
       var span = agent.startSpan(null, 'external', 'http')
       var id = span && span.transaction.id
+      // console.warn('XXX curr span (after startSpan):', agent._instrumentation.currentSpan?.name) // considers 'bindingSpan'
 
       agent.logger.debug('intercepted call to %s.%s %o', moduleName, method, { id: id })
 
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -9,6 +9,11 @@ var { Ids } = require('./ids')
 var NamedArray = require('./named-array')
 var shimmer = require('./shimmer')
 var Transaction = require('./transaction')
+const {
+  RunContext,
+  BasicRunContextManager,
+  AsyncHooksRunContextManager
+} = require('../run-context')
 
 var MODULES = [
   '@elastic/elasticsearch',
@@ -54,6 +59,10 @@ function Instrumentation (agent) {
   this._agent = agent
   this._hook = null // this._hook is only exposed for testing purposes
   this._started = false
+
+  this._log = agent.logger.child({ 'event.module': 'instrumentation' })
+
+  // XXX TODO: handle all these curr tx/span properties
   this.currentTransaction = null
 
   // Span for binding callbacks
@@ -85,6 +94,22 @@ function Instrumentation (agent) {
     })
   }
 }
+
+Instrumentation.prototype.currTx = function () {
+  return this._runCtxMgr.active().tx || null
+}
+Instrumentation.prototype.currSpan = function () {
+  const spanStack = this._runCtxMgr.active().spans
+  if (spanStack.length === 0) {
+    return null
+  } else {
+    return spanStack[spanStack.length - 1]
+  }
+}
+// XXX unneeded?
+// Instrumentation.prototype.currParent = function () {
+//   return this._runCtxMgr.active().topSpanOrTx()
+// }
 
 Object.defineProperty(Instrumentation.prototype, 'ids', {
   get () {
@@ -136,8 +161,11 @@ Instrumentation.prototype.start = function () {
   this._started = true
 
   if (this._agent._conf.asyncHooks) {
-    require('./async-hooks')(this)
+    // XXX
+    // require('./async-hooks')(this)
+    this._runCtxMgr = new AsyncHooksRunContextManager(this._log)
   } else {
+    this._runCtxMgr = new BasicRunContextManager(this._log)
     require('./patch-async')(this)
   }
 
@@ -148,6 +176,7 @@ Instrumentation.prototype.start = function () {
     }
   }
 
+  this._runCtxMgr.enable()
   this._startHook()
 }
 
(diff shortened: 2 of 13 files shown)


test/instrumentation/index.test.js
```diff
@@ -617,7 +525,7 @@ test('nested spans', function (t) {
     t.strictEqual(s0.transaction_id, trans.id, 's0 transaction_id matches transaction id')
 
     const s1 = findObjInArray(data.spans, 'name', 's1')
-    t.strictEqual(s1.parent_id, trans.id, 's1 should directly descend from the transaction')
+    t.strictEqual(s1.parent_id, s0.id, 's1 should descend from s0')
```
trentm, line 528: REVIEW NOTE: This is the intentional change in behaviour that sync span creation results in parent/child relationships, rather than siblings.

```
Expected hierarchy before:
  transaction "foo"
  `- span "s0"
    `- span "s01"
  `- span "s1"
    `- span "s11"
    `- span "s12"

After:
  transaction "foo"
  `- span "s0"
    `- span "s1"
      `- span "s11"
      `- span "s12"
    `- span "s01"
```

The change is that "s1" is a child of "s0". See discussion at https://github.com/elastic/apm-agent-nodejs/issues/1889

lib/instrumentation/modules/pg.js
```diff
@@ -81,74 +82,43 @@ function patchClient (Client, klass, agent, enabled) {
...
+            queryOrPromise.on('end', onQueryEnd)
+            queryOrPromise.on('error', onQueryEnd)
+            if (queryOrPromise instanceof EventEmitter) {
+              agent._instrumentation.bindEmitter(queryOrPromise)
+            }
+          } else if (typeof queryOrPromise.then === 'function') {
+            queryOrPromise.then(
+              () => { onQueryEnd() },
+              onQueryEnd
+            )
```
trentm, line 112: REVIEW NOTE: No need to pass back our modified promise because context tracking automatically ensures the `queryOrPromise` chain of handlers will run with the appropriate context.  Also no need to `throw err` in our reject/catch because we aren't returning this promise now.

lib/instrumentation/modules/pg.js
```diff
@@ -81,74 +82,43 @@ function patchClient (Client, klass, agent, enabled) {
...
-        span.end()
-        if (cb) return cb.apply(this, arguments)
-      }
-    }
-  }
-
-  // The client maintains an internal callback queue for all the queries. In
-  // 7.0.0, the queries are true promises (as opposed to faking the Promise API
-  // in ^6.3.0). To properly get the right context when the Promise API is
-  // used, we need to patch all callbacks in the callback queue.
```
trentm, line 136: REVIEW NOTE: The bindFunction and bindEmitter above ensure `onQueryEnd` runs in the context of the span (and its transaction). This obviates the need for this wrapPulseQueryQueue. A full TAV=pg test run passes.

test/instrumentation/modules/memcached.test.js
```diff
@@ -64,11 +67,16 @@ test(function (t) {
         port: 11211
       })
     })
+    spans.forEach(span => {
+      t.equal(span.parent_id, data.transactions[0].id,
+        'span is a child of the transaction')
+    })
```
trentm, line 73: REVIEW NOTE: The `ins.bindFunction(callback)` added in "lib/instrumentation/memcached.js" fixed an odd parentage issue. For this test case:

- Before this PR, the parent child relationship of the memcached commands was:

    ```
    transaction "myTrans"
    `- span "memcached.set"
      `- span "memcached.replace"
      `- span "memcached.get"
      `- span "memcached.touch"
      `- span "memcached.delete"
      `- span "memcached.get"
    `- span "memcached.get"
    ```

    I.e. weird. The first `cache.get` under `cache.set` is parented to the
    transaction, and thereafter every other `cache.$command` is parented to
    the initial `cache.set`.

- After this PR:

    ```
    transaction "myTrans"
    `- span "memcached.set"
    `- span "memcached.get"
    `- span "memcached.replace"
    `- span "memcached.get"
    `- span "memcached.touch"
    `- span "memcached.delete"
    `- span "memcached.get"
    ```

test/instrumentation/index.test.js
```diff
@@ -304,20 +210,6 @@ test('errors should have a transaction id - non-ended transaction', function (t)
...
-    const trans = data.transactions[0]
-    t.strictEqual(data.errors[0].transaction_id, trans.id)
-    t.strictEqual(typeof data.errors[0].transaction_id, 'string')
-    t.end()
-  })
-  agent.captureError = origCaptureError
-  agent.startTransaction('foo').end()
-  agent.captureError(new Error('bar'))
-})
-
```
trentm, line 320: REVIEW NOTE: This is an intentional behaviour change. Before this PR an ended transaction would linger as `agent.currentTransaction`. Not any longer.  This test was added in https://github.com/elastic/apm-agent-nodejs/issues/147. My read of that is that there is no need to associate an error with a transaction if it is captured *after* the transaction has ended.

lib/instrumentation/http-shared.js
```diff
@@ -228,17 +228,11 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       }
 
       function onresponse (res) {
-        // Work around async_hooks bug in Node.js 12.0 - 12.2 (https://github.com/nodejs/node/pull/27477)
-        ins._recoverTransaction(span.transaction)
```
trentm, line 232: REVIEW NOTE: See separate, long review note about `_recoverTransaction` being dropped.

lib/instrumentation/span.js
```diff
@@ -80,8 +80,6 @@ Span.prototype.end = function (endTime) {
 
   this._setOutcomeFromSpanEnd()
 
-  this._agent._instrumentation._recoverTransaction(this.transaction)
```
trentm, line 83: REVIEW NOTE: See separate, long review note about _recoverTransaction being dropped.

lib/instrumentation/index.js
```diff
@@ -303,115 +391,107 @@ Instrumentation.prototype.setTransactionName = function (name) {
...
+// This can be used to ensure `fn` does *not* run in the context of the current
+// transaction or span.
+Instrumentation.prototype.bindFunctionToEmptyRunContext = function (fn) {
+  if (!this._started) {
+    return fn
+  }
+  return this._runCtxMgr.bindFn(new RunContext(), fn)
 }
 
-Instrumentation.prototype._recoverTransaction = function (trans) {
```
trentm, line 407: REVIEW NOTE: See separate, long review note about _recoverTransaction being dropped.

test/instrumentation/index.test.js
```diff
@@ -187,101 +188,6 @@ test('stack branching - no parents', function (t) {
...
-          t2.end()
-          setImmediate(function () {
-            trans.end()
-          })
-        })
-      })
-    })
-  })
-})
-
```
trentm, line 284: REVIEW NOTE: This PR drops `Instrumentation#_recoverTransaction(trans)`. It was a function that would reset the currentTransaction to the given one (and log a warning) if they differed. This was called in two places:

1. Originally `_recoverTransaction` and the "recoverable"-named test cases in "test/instrumentation/index.test.js" (that I've now removed) were added in [this commit from 5 years ago](https://github.com/elastic/apm-agent-nodejs/commit/f9d15b55fc469edccdf91878327f8b75a49ff3d1) in relation to dealing with internal user-land callback queues (particularly in db modules). See https://www.youtube.com/watch?v=omOtwqffhck&t=892s which describes the problem. That commit added `Instrumentation#_recoverTransaction(trans)` and called it from what was to become `Span#end()` (after renamings).
    
    A better answer for ensuring these delayed and possibly user-land-queued callbacks are executed in the appropriate run context is to wrap them with `ins.bindFunction(callback)`. This has already been done for any instrumentations that were failing the test suite because of this:
    - memcached: `query.callback = agent._instrumentation.bindFunction(...)`
    - mysql: `return ins.bindFunction(function wrappedCallback ...`
    - tedious: `request.userCallback = ins.bindFunction(...`
    - pg: `args[index] = agent._instrumentation.bindFunction(...`
    - s3: `request.on('complete', ins.bindFunction(...`
    
    The "recoverable" tests here are not meaningfully translatable to the new RunContext management. What is more meaningful are tests on the particular instrumented modules, e.g. see the changes to "test/instrumentation/modules/memcached.test.js". Also each of the memcached, mysql, tedious, and pg tests failed when `_recoverTransaction` was made a no-op and before the added `bindFunction` calls. In other words there were meaningful tests for these cases.

2. Much later `_recoverTransaction` was added in http-shared.js in the "response" handler for outgoing HTTP requests, e.g. calls to `http.request(url, onResponse)`.

    ```js
        function onresponse (res) {
          // Work around async_hooks bug in Node.js 12.0 - 12.2 (https://github.com/nodejs/node/pull/27477)
          ins._recoverTransaction(span.transaction)
          ...
    ```

    With the old/current "async-hooks.js"-based context management and with Node.js v12.0 - 12.2, when this onresponse ran, the `currentTransaction` was null, because of [this core Node async_hooks bug](https://github.com/nodejs/node/pull/27477) where an "init" async hook would not be called for a re-used HTTPParser.

    This was worked around in the agent in #1339 by adding `ins._recoverTransaction(span.transaction)` which just sets `ins.currentTransaction = span.transaction`.  Before the workaround, `node test/instrumentation/modules/http/outgoing.test.js` would fail:

    ```
    not ok 19 test exited without ending: http.request(options, callback)
    ---
      operator: fail
      at: process.<anonymous> (.../node_modules/tape/index.js:94:23)
      stack: |-
        Error: test exited without ending: http.request(options, callback)
    ```

    A possibly better workaround would have been to use ins.bindFunction to bind this `onresponse` to the current span and transaction.

    With the new AsyncHooksRunContextManager, neither of these are necessary. The `init` async hook *is* still missed in node v12.0-12.2, but this `onresponse()` runs within the context of this trans and span in all outgoing.test.js and in manual tests. I.e. adding `assert(ins.currSpan() === span)` here, still passes the test suite (with all versions of node).

PR #2184, opened 2021-08-09
ci: workaround for common 'network docker_default id $id has active endpoints' jenkins error

trentm:
A relatively frequent error in Jenkins CI is this at the end of an
otherwise successful test run:
    error while removing network: network docker_default id 873ad... has active endpoints

This is an attempted workaround to force removal and to print some
possibly helpful debug info.


cachedout (approved):
I like it. Let's see how it goes. :)

Issue #2156, opened 2021-07-12
[META 352] AWS Lambda support


AlexanderWert:
Add support for AWS lambda as described in the meta issue: https://github.com/elastic/apm/issues/352

astorm:
An initial To Do list for the Node.js Agent's Lambda support

- Determine the fate of instrumentation for context.success, context.fail, and context.done (no longer documented – should we keep these?)
- Handling async handler functions
- Include processing of tracestate header along with traceparent
- Collect data per the draft spec
- Create “metadata from env” fetcher
- Cold Start detector
- Solving the “Thawed execution environment but no APM Server Connection” errors

trentm:
This depends on:

- [ ] https://github.com/elastic/apm-nodejs-http-client/issues/165
- [ ] https://github.com/elastic/apm-agent-nodejs/issues/2294

PR #2185, opened 2021-08-09
feat: add tracestate header when starting lambda transaction

astorm:
### Checklist

Part of https://github.com/elastic/apm-agent-nodejs/issues/2156

This PR ensures that if the Lambda function is responding to an API Gateway request (or other request that includes a `tracestate` header) that its transaction will start using that tracestate.  This is needed because our existing lambda instrumentation was written before `tracestate` was a thing.

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -87,7 +87,7 @@ module.exports = function elasticApmAwsLambda (agent) {
              * elastic-apm-traceparent if available.
              */
             parentId = value
-          } else if(lowerCaseKey === 'tracestate') {
+          } else if (lowerCaseKey === 'tracestate') {
             tracestate = value
           }
         }
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -74,7 +74,7 @@ module.exports = function elasticApmAwsLambda (agent) {
 
     return function wrappedLambda (payload, context, callback) {
       let parentId
-      let traceState
+      let tracestate
       if (payload.headers !== undefined) {
         for (const [key, value] of Object.entries(payload.headers)) {
           const lowerCaseKey = key.toLowerCase()
@@ -88,7 +88,7 @@ module.exports = function elasticApmAwsLambda (agent) {
              */
             parentId = value
           } else if(lowerCaseKey === 'tracestate') {
-            traceState = value
+            tracestate = value
           }
         }
       }
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -74,6 +74,7 @@ module.exports = function elasticApmAwsLambda (agent) {
 
     return function wrappedLambda (payload, context, callback) {
       let parentId
+      let traceState
       if (payload.headers !== undefined) {
         for (const [key, value] of Object.entries(payload.headers)) {
           const lowerCaseKey = key.toLowerCase()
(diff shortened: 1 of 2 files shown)


lib/lambda.js
```diff
@@ -86,11 +87,14 @@ module.exports = function elasticApmAwsLambda (agent) {
              * elastic-apm-traceparent if available.
              */
             parentId = value
```
trentm, line 89: 1. The handling will have to change to avoid that `break` preferring 'elastic-apm-traceparent' to 'traceparent'. IIUC, with the current logic it will miss tracestate if it comes after 'elastic-apm-traceparent' in `payload.headers`.
2. Also, per #2079 the logic should change (could be a separate PR) to prefer 'traceparent' to 'elastic-apm-traceparent'.

trentm (changes requested):

lib/lambda.js
astorm: @trentm Ugh -- that's a weird construct.  Thanks for catching it. Any strong preference between 

- more code and fewer loop cycles (matching most closely with current intent/pattern)

```
    if (payload.headers !== undefined) {
      for (const [key, value] of Object.entries(payload.headers)) {
        const lowerCaseKey = key.toLowerCase()
        if (lowerCaseKey === 'traceparent') {
          parentId = value
        } else if (!parentId && lowerCaseKey === 'elastic-apm-traceparent') {
          parentId = value
        } else if(lowerCaseKey === 'tracestate') {
          tracestate = value
        }
        if(parentId && tracestate) {
          break
        }
      }
    }
```

- or less code but slightly more work normalizing all the headers

```
      if (payload.headers !== undefined) {
        const normalizedHeaders = {}
        for (const [key, value] of Object.entries(payload.headers)) {
          const lowerCaseKey = key.toLowerCase()
          normalizedHeaders[lowerCaseKey] = value
        }
        parentId = normalizedHeaders.traceparent ? normalizedHeaders.traceparent : normalizedHeaders['elastic-apm-traceparent']
        tracestate = normalizedHeaders.tracestate
      }
```

trentm: The former will prefer 'elastic-apm-traceparent' over 'traceparent' if it and 'tracestate' come first, so doesn't work as written.

No strong preference between those two until perf profiles show that the section is a hot spot.

Slightly related: from old https://elastic.slack.com/archives/C5M02D1GE/p1619802623086700?thread_ts=1619742610.083400&cid=C5M02D1GE `for ([key, value] of Object.entries(obj))` is dog slow. Perhaps premature optimization, but I'd generally bias against using it.

astorm:
I've replaced the `Object.entries()` loop with an `Object.keys()` loop. Always up for quashing these then we see them and remember.

I've opted for the clearer (to me) code of normalizing _all_ the headers rather than the micro-optimization of finishing once we have the specific headers we're interested in. 

I've added test cases with both `elastic-apm-traceparent` by itself and `elastic-apm-traceparent` both before and after the `traceparent` header to ensure that we prefer `traceparent` when it's there, but still pickup `elastic-apm-traceparent` when it's by itself.

trentm (approved):
Thanks.

Issue #2187, opened 2021-08-11
Test Suite Errors with `@elastic/elasticsearch:7.14.0`


astorm:
Tests are failing after the release of @elastic/elasticsearch:7.14.0.

**To Reproduce**

Steps to reproduce the behavior: 
1. Remove your `node_modules` folder and any package lock files
2. Run `npm install`
3. Run `npx tape test/instrumentation/modules/@elastic/elasticsearch.test.js`

**Expected behavior**

Tests pass.

**Actual behavior**

Tests fail (see below)

**Additional context**

<details>
<summary>Click for test output</summary>

```
    % npx tape test/instrumentation/modules/@elastic/elasticsearch.test.js
    TAP version 13
    # client.ping with promise
    ok 1 should have 1 transaction
    not ok 2 should have 2 spans
      ---
        operator: equal
        expected: 2
        actual:   4
        at: <anonymous> (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/test/instrumentation/modules/@elastic/elasticsearch.test.js:496:7)
        stack: |-
          Error: should have 2 spans
              at Test.assert [as _assert] (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/node_modules/tape/lib/test.js:311:54)
              at Test.bound [as _assert] (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/node_modules/tape/lib/test.js:96:32)
              at Test.strictEqual (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/node_modules/tape/lib/test.js:475:10)
              at Test.bound [as equal] (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/node_modules/tape/lib/test.js:96:32)
              at /Users/astorm/Documents/github/elastic/apm-agent-nodejs/test/instrumentation/modules/@elastic/elasticsearch.test.js:496:7
              at Timeout._onTimeout (/Users/astorm/Documents/github/elastic/apm-agent-nodejs/test/_mock_http_client.js:65:7)
              at listOnTimeout (internal/timers.js:554:17)
              at processTimers (internal/timers.js:497:7)
      ...
    ok 3 should have expected transaction name
    ok 4 should have expected transaction type
    ok 5 have an elasticsearch span
    ok 6 should be strictly equal
    ok 7 should be strictly equal
    ok 8 should be strictly equal
    ok 9 have an http span
    ok 10 should be strictly equal
    ok 11 should be strictly equal
[...]

PR #2188, opened 2021-08-11
fix: test failures and possible crash in @elastic/elasticsearch@7.14.0 instrumentation

trentm:
This will also workaround
https://github.com/elastic/elasticsearch-js/issues/1516  It is up for
debate whether it is an elasticsearch-js bug that the Client "request"
event can be missing '.meta.request.params'.

Fixes: #2187

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/elasticsearch-shared.js
+++ b/lib/instrumentation/elasticsearch-shared.js
@@ -26,8 +26,10 @@ exports.pathIsAQuery = pathIsAQuery
 //
 //    {}
 //    {"query":{"query_string":{"query":"pants"}}}
+//
+// `path`, `query`, and `body` can all be null or undefined.
 exports.setElasticsearchDbContext = function (span, path, query, body) {
-  if (pathIsAQuery.test(path)) {
+  if (path && pathIsAQuery.test(path)) {
     // From @elastic/elasticsearch: A read of Transport.js suggests query and
     // body will always be serialized strings, however the documented
     // `TransportRequestParams` allows for non-strings, so we will be defensive.
--- a/lib/instrumentation/modules/@elastic/elasticsearch.js
+++ b/lib/instrumentation/modules/@elastic/elasticsearch.js
@@ -19,16 +19,6 @@ module.exports = function (elasticsearch, agent, { version, enabled }) {
     return elasticsearch
   }
 
-  function generateSpan (params) {
-    const span = agent.startSpan(null, 'db', 'elasticsearch', 'request')
-    if (span === null) {
-      return null
-    }
-
-    span.name = `Elasticsearch: ${params.method} ${params.path}`
-    return span
-  }
-
   class ApmClient extends elasticsearch.Client {
     constructor (...args) {
       super(...args)
@@ -43,13 +33,23 @@ module.exports = function (elasticsearch, agent, { version, enabled }) {
 
       this.on('request', (err, result) => {
         const request = result.meta.request
+        let paramMethod = null
+        let paramPath = null
+        let paramQueryString = null
+        let paramBody = null
+        if (request.params) {
+          // request.params can be null with ProductNotSupportedError and
+          // DeserializationError.
+          paramMethod = request.params.method
+          paramPath = request.params.path
+          paramQueryString = request.params.querystring
+          paramBody = request.params.body
+        }
         agent.logger.debug('intercepted call to @elastic/elasticsearch "request" event %o',
-          { id: request.id, method: request.params.method, path: request.params.path })
+          { id: request.id, method: paramMethod, path: paramPath })
 
         let span = spanFromEsResult.get(result)
 
-        // As of @elastic/elasticsearch@7.10.0-rc.1 this event's `err` will
-        // always be null, but we'll be future-proof.
         if (err) {
           agent.captureError(err)
           if (span !== undefined) {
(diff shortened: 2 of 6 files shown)


trentm:
Here is a TAV test run for the package (with node v12 on my mac):

```
% TAV=@elastic/elasticsearch ./node_modules/.bin/tav --compat
Testing compatibility with @elastic/elasticsearch:
✔ 7.14.0
✔ 7.13.0
✔ 7.12.0

✔ 7.11.0
✔ 7.10.0
✔ 7.9.1
✔ 7.9.0
✔ 7.8.0
✔ 7.7.1
✔ 7.6.1
✔ 7.6.0
✔ 7.5.1
✔ 7.5.0
✔ 7.4.0
✔ 7.3.0
✔ 7.2.0
✔ 7.1.0
✔ 7.0.1
✔ 7.0.0
```

lib/instrumentation/elasticsearch-shared.js
```diff
@@ -26,8 +26,10 @@ exports.pathIsAQuery = pathIsAQuery
 //
 //    {}
 //    {"query":{"query_string":{"query":"pants"}}}
+//
+// `path`, `query`, and `body` can all be null or undefined.
 exports.setElasticsearchDbContext = function (span, path, query, body) {
-  if (pathIsAQuery.test(path)) {
+  if (path && pathIsAQuery.test(path)) {
```
astorm, line 32: nit: Is the `path &&` needed here?  My naive reading is `pathIsAQuery.test(path)` should be false for all the false-ish values.

```
const pathIsAQuery = /\/(_search|_msearch|_count|_async_search|_sql|_eql)(\/|$)/

> pathIsAQuery.test(null)
false
> pathIsAQuery.test(undefined)
false
> pathIsAQuery.test('')
false
> pathIsAQuery.test(false)
false

etc.
```

lib/instrumentation/modules/@elastic/elasticsearch.js
```diff
@@ -73,12 +74,13 @@ module.exports = function (elasticsearch, agent, { version, enabled }) {
...
         }
 
-        setElasticsearchDbContext(span, request.params.path,
-          request.params.querystring, request.params.body)
+        setElasticsearchDbContext(span, paramPath, paramQueryString, paramBody)
 
-        const { hostname, port } = result.meta.connection.url
-        span.setDestinationContext(
-          getDBDestination(span, hostname, port))
+        if (result.meta.connection && result.meta.connection.url) {
```
astorm, line 79: nit: here, and elsewhere, we're still assuming that `result.meta` is a set value, which means we're still open to future client change causing another crashing error. (A moot point if we get our friends to incorporate the agent in their integration tests, ergo `nit`)

test/instrumentation/modules/@elastic/elasticsearch.test.js
```diff
@@ -348,11 +392,16 @@ if (semver.gte(pkgVersion, '7.7.0')) {
 
         const err = data.errors
           .filter((e) => e.exception.type === 'RequestAbortedError')[0]
-        t.ok(err, 'sent an error to APM server')
-        t.ok(err.id, 'err.id')
-        t.equal(err.exception.message, 'Request aborted', 'err.exception.message')
-        t.equal(err.exception.type, 'RequestAbortedError',
-          'err.exception.type is RequestAbortedError')
+        if (semver.eq(esVersion, '7.14.0')) {
```
astorm, line 395: Presuming this is intentional, but just calling out that this is an `eq` test, which means if this behavior remains in 7.14.1, or 7.15.0 we'll have failures. No changes needed/expected.

astorm (approved):
👍 I see we're explicitly checking for `request.params` now, fixing the deep access problem of `request.[null|undefined].method` -- a few nits but this looks like a reasonable fix.  Approving.

lib/instrumentation/elasticsearch-shared.js
trentm: Yah, true, but passing other than a string into `RegExp.prototype.test` strikes me as messy. It is doing implicit type conversion to string and doing a regex match on that. For example:

```
> pathIsAOops = /foo|bar|null|baz$/
/foo|bar|null|baz$/
> pathIsAOops.test(null)
true
```

test/instrumentation/modules/@elastic/elasticsearch.test.js
trentm: Yes, that was intentional... but perhaps just making more work for us. I'm slightly banking on the clients team getting a 7.14.1 or 7.15.0 that fixes https://github.com/elastic/elasticsearch-js/issues/1517

lib/instrumentation/modules/@elastic/elasticsearch.js
trentm: Yah, fair point. I'll give it a try. I guess I'll have to [wait for node v14 for `foo?.bar?.baz`](https://node.green/#ES2020-features-optional-chaining-operator------optional-property-access).

trentm: Done in commit f8d3f516

trentm:
I'll assume astorm's re-review on this :) so I can get it in and get a release PR going.

Issue #1921, opened 2020-12-17
Expose Max Db Statement Size as a Config Option


gingerwizard:
Currently the `span.context.db.statement` is limited to 10000 characters after which truncation occurs. This limit is easily exceeded on REST API based dbs that use JSON e.g. Elasticsearch. Proposal is to make this configurable similar tothe  truncation limit for error messages.

gingerwizard:
Note is particularly important when debugging requests to Elasticsearch from Kibana.

trentm:
A temporary work around is to apply this patch to one's node_modules/elastic-apm-node/lib/config.js:

```diff
diff --git a/lib/config.js b/lib/config.js
index 38d6408..a5520b0 100644
--- a/lib/config.js
+++ b/lib/config.js
@@ -257,6 +257,8 @@ class Config {
           // Sanitize conf
           truncateKeywordsAt: config.INTAKE_STRING_MAX_SIZE,
           truncateErrorMessagesAt: conf.errorMessageMaxLength,
+          // XXX
+          truncateQueriesAt: 20000,
           // HTTP conf
           secretToken: conf.secretToken,
```

trentm:
^^ proposing this for 7.13. Currently Dale is having "to maintain a fork for tests and any automation" for his es rally work.

trentm:
Notes to self:
- perhaps do this together with changes to db.statement format for ES: https://github.com/elastic/apm-agent-nodejs/issues/2019
- tracing-instrumentation-db.md spec says "We store up to 10000 Unicode characters per database statement." FWIW
- TODO: verify if >10k gets stored in ES docs or if apm-server clips or rejects. I don't (with a quick naive look) see clipping/length limiting in apm-server for this field.

trentm:
@gingerwizard Once we publish an agent release with this change you'll be able to set the truncation limie for `span.context.db.statement` via the `longFieldMaxLength: <number>` config var.  I try to post here again once we have that release.

trentm:
@gingerwizard If it helps for https://github.com/elastic/rally-internal-tracks/issues/119  `elastic-apm-node@3.21.0` has now been released that includes `longFieldMaxLength` to allow you to increase the max size for `span.context.db.statement`.

PR #2193, opened 2021-08-13
feat: Add "longFieldMaxLength" config var

trentm:
https://github.com/elastic/apm/pull/493 specifies a `long_field_max_length` config option we can use for `span.context.db.statement`. Currently this is hardcoded to 10000... bytes. 

- We'll also change to be 10000 unicode characters (rather than bytes). This is in line with the json-schema spec for "maxLength" and is now [the apm spec for `long_field_max_length`](https://github.com/elastic/apm/blob/master/specs/agents/field-limits.md#long_field_max_length-configuration). See [the http-client PR](https://github.com/elastic/apm-nodejs-http-client/pull/162) for this part of the change.
- This also deprecates the existing node.js-agent-only `errorMessageMaxLength` config option. It now defaults to the `longFieldMaxLength` value if not specified. Otherwise `errorMessageMaxLength` is still supported. (We should drop it in the "next major".)
- This change also applies the truncation to the other fields mentioned in [the `long_field_max_length` spec](https://github.com/elastic/apm/blob/master/specs/agents/field-limits.md#long_field_max_length-configuration). This *does* mean that some existing usage out there might get truncation of some of these fields where they didn't before. That's debatable whether that constitutes a "breaking change". So far this PR takes the point of view that this is *not* a breaking change that would require either a *separate* config var (which means not conforming with the APM spec link above) or only applying truncation to these fields in a new major version later. Objections?

Fixes: #1921

### Checklist

- [x] Implement code
- [x] update package.json dep once https://github.com/elastic/apm-nodejs-http-client/pull/162 is merged and v10.0.0 is published
- [x] Add tests
- [x] Update documentation


trentm:
https://github.com/elastic/apm/issues/488 is a discussion for a cross-agent config option name for this var. Let's hold until that discussion plays out.

trentm:
I've started https://github.com/elastic/apm/pull/493 for the spec addition for this field name: `long_field_max_length`.

trentm:
^^ Integration Tests failure is the current https://github.com/elastic/apm-integration-testing/issues/1188 flake

trentm:
Here is a small test script to demonstrate `longFieldMaxLength` working on `span.context.db.statement`:

```
  longFieldMaxLength: 100000
})

const { Client } = require('@elastic/elasticsearch')
var client = new Client({
  node: 'http://localhost:9200',
  auth: { username: 'admin', password: 'changeme' }
})

const t1 = apm.startTransaction('t1')
// Build up a search body >10k in size.
aggs = {}
for (var i = 0; i < 1000; i++) {
  aggs[`aggs${i}`] = { "terms": { "field": "author" } }
}
client.search(
    {
      body: {
        size: 1,
        aggregations: aggs
      }
    },
    function (err, result) {
      t1.end()
    }
)
```

docs/configuration.asciidoc
```diff
@@ -671,16 +671,20 @@ The default value `0` means that no source code will be collected for span libra
...
+This option is **deprecated** -- use <<long-field-max-length,`longFieldMaxLength`>> instead.
+
 The maximum length allowed for error messages.
 It is expressed in bytes or includes a size suffix such as `2kb`.
 Size suffixes are case-insensitive and include `b`,
 `kb`,
 `mb`,
 and `gb`.
 Messages above this length will be truncated before being sent to the APM Server.
+Note that while the configuration option accepts a number of *bytes*, truncation
```
astorm, line 686: nit: As written, I'm a little uncertain what this extra message is trying to convey about the behavior of `errorMessageMaxLength`, and how the bytes configured will be converted into characters.

package.json
```diff
@@ -86,7 +87,7 @@
     "basic-auth": "^2.0.1",
     "cookie": "^0.4.0",
     "core-util-is": "^1.0.2",
-    "elastic-apm-http-client": "^9.9.0",
+    "elastic-apm-http-client": "elastic/apm-nodejs-http-client#trentm/truncate-in-chars-not-bytes",
```
astorm, line 90: I presume the intention is to land this PR after the client changes land and that the github repo and `XXX` will be swapped out.

astorm (commented):
Overall looks it does the thing the way the agent wants the thing done. Will approve once the `package.json` is updated to reflect what we want to ship.

docs/configuration.asciidoc
trentm: Given the field is now deprecated, should I just drop that comment to avoid adding confusion?

What is happening is that we accept a config value that is "some number of bytes" (because the suffixes are "b", "kb", etc.). However that is just converted to an integer: `"2kb"` becomes `2048`. Then that `2048` is used as a count of *characters* (not bytes) for translation. It is a little subtle, but almost certainly won't affect anyone -- only if they were having error messages near the limit and the error messages had lots of emoji or chars outside the BMP in them.

package.json
trentm: Yup, done.

docs/configuration.asciidoc
astorm: Ah -- I see, so the string is converted into a byte value, and then that byte value is used as a character length.  I'm not sure whether it's better to say nothing or keep what we have to indicate that's it's not a true limiting by byte size.  Nit withdrawn.

astorm (approved):

PR #2194, opened 2021-08-13
feat: lambda faas transaction properties and cold start detection

astorm:
Part of: https://github.com/elastic/apm-agent-nodejs/issues/2156

The intent of this PR is to 

1. Set up the scaffolding for reporting transaction data
2. Set up the new `faas` data on the transaction for reporting
3. Serialize the new `faas` data to the transaction payload
4. Implement a cold start detection feature
5. Implements non-metadata context fields

### Checklist

- [x] Implement code
- [x] Add tests
- [x] API Gateway Events (V1/V2)
- [x] SQS Gateway Event
- [x] SNS Gateway Event
- [x] Finish S3 Lambda event
- [x] Finish Generic Lambda event
- [x] Serialization of new context objects
- [x] set transaction.result


lib/instrumentation/transaction.js
```diff
@@ -92,6 +93,11 @@ Transaction.prototype.setUserContext = function (context) {
   this._user = Object.assign(this._user || {}, context)
 }
 
+Transaction.prototype.setFaas = function (context) {
```
trentm, line 96: nit: Perhaps change this "context" argument to "faas" or "faasFields" or something.

"context" is a very overloaded word in the agent code:

- the "context" object on spans on transactions
- async "context" in "async-hooks.js"
- trace-"context" for the W3C trace-context headers/data

lib/instrumentation/transaction.js
```diff
@@ -133,7 +139,8 @@ Transaction.prototype.toJSON = function () {
     span_count: {
       started: this._builtSpans
     },
-    outcome: this.outcome
+    outcome: this.outcome,
+    faas: this._faas
```
trentm, line 143: It would be nice to either have the `this._faas` initial value be `undefined` or only add `faas` to `payload` behind a `if (this._faas)` guard. Otherwise, in the common non-faas case we will be adding `"faas": null` to serialized transaction APM events.

lib/lambda.js
```diff
@@ -2,6 +2,28 @@
 
 const shimmer = require('./instrumentation/shimmer')
 
+// used to detect cold starts
+let isFirstRun = false
```
trentm, line 6: THis should default to `true`, shouldn't it?

lib/lambda.js
```diff
@@ -2,6 +2,28 @@
...
+
+function getTriggerTypeFromEvent (event) {
+  if (event && event.requestContext && event.requestContext.requestId) {
+    return 'http'
+  }
```
trentm, line 11: Do you intend to leave the SNS/SQS/S3-trigger handling to later?

nit: Given that the spec says `faas.trigger.request_id` is dependent on `faas.trigger.type`, I would consider having a single function that handles those two together rather than the pattern currently here which has the `event && event.requestContext && event.requestContext.requestId`-logic in two places.

lib/lambda.js
astorm: _nod_, it should.  I've changed the default value and added a test to catch similar problems in the future.

lib/instrumentation/transaction.js
astorm: ✅ in 60b6f4a166422407f3b9c78e88ad211cc3c4b23c

lib/instrumentation/transaction.js
astorm: ✅ in 61ec4029b86c843b9b2831991525cdd5405ac800

lib/lambda.js
astorm: resolving as obsolete -- approach has changed significantly.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  transaction.setCloudContext(cloudContext)
+
+  let age
+  if (record.attributes && record.attributes.SentTimestamp) {
+    age = Date.now() - record.attributes.SentTimestamp
+  }
+  const messageContext = {
+    queue: record.eventSourceARN,
+    age: age,
```
trentm, line 133: Is the https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-aws-lambda.md#sqs--sns wrong here? The APM Server transaction.json spec says these should be `context.message.queue.name` and `context.message.age.ms`.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  }
+  transaction.setCloudContext(cloudContext)
+
+  let age
+  if (message.Timestamp) {
+    age = Date.now() - (new Date(message.Timestamp).getTime())
+  }
+  const messageContext = {
+    queue: topicArn,
+    age: age,
```
trentm, line 197: Same as for SQS above: Is the apm.git lambda spec wrong? The APM server transaction.json spec says `.queue.name` and `.age.ms`.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+    body: undefined,
+    headers: Object.assign(message.MessageAttributes)
+  }
+
+  if (agent._conf.captureBody) {
```
trentm, line 202: Same as for SQS above: check for `'all'` or `'transactions'` values of `captureBody`.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+
+  let age
+  if (message.Timestamp) {
+    age = Date.now() - (new Date(message.Timestamp).getTime())
+  }
+  const messageContext = {
+    queue: topicArn,
+    age: age,
+    body: undefined,
+    headers: Object.assign(message.MessageAttributes)
```
trentm, line 199: Similar to above for SNS:
1. `context.message.headers` should only be added if `_conf.captureHeaders` is true.
2. This Object.assign() isn't doing anything. Is the intent to get a copy of the attributes so that possible later sanitization and truncation doesn't change the original? If so then, `Object.assign({}, ...)`.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  }
+  transaction.setMessageContext(messageContext)
+}
+
+function isSnsEvent (event) {
+  const records = event.Records ? event.Records : []
+  return records.length > 0 && records[0].EventSource === 'aws:sns'
+}
+
+function setS3SingleData (transaction, event, context, isColdStart) {
```
trentm, line 213: Given this is only called if there is a single `event.Records[0]`, can all the `record &&` guards be dropped to reduce the code in this function?

test/instrumentation/transaction.test.js
```diff
@@ -315,7 +315,7 @@ test('#_encode() - ended', function (t) {
   t.ok(payload.duration > 0)
   t.strictEqual(payload.timestamp, trans._timer.start)
   t.strictEqual(payload.result, 'success')
-  t.deepEqual(payload.context, { user: {}, tags: {}, custom: {} })
+  t.deepEqual(payload.context, { user: {}, tags: {}, custom: {}, service: {}, cloud: {}, message: {} })
```
trentm, line 318: Will all our transactions now always have these extra fields? I think it would be nice to minimize that. It could be a separate PR, if you prefer.

Oh actually, if you take the suggestion in transaction.js to *not* have the `|| {}`, then I think this will clear up.

trentm (changes requested):

lib/instrumentation/transaction.js
```diff
@@ -152,7 +175,10 @@ Transaction.prototype.toJSON = function () {
         this._user
       ),
       tags: this._labels || {},
-      custom: this._custom || {}
+      custom: this._custom || {},
+      service: this._service || {},
+      cloud: this._cloud || {},
+      message: this._message || {}
```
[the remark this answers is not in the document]
trentm: Actually I think the `|| {}` here is why our encoded transactions have empty objects for unused `context.$subkey`s. I think it would be nicer if those were not included when empty.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  transaction.setFaas(faas)
+
+  transaction.type = 'http'
+  let httpMethod = event && event.requestContext && event.requestContext.httpMethod
+  if (!httpMethod) {
+    httpMethod = event && event.requestContext && event.requestContext.http && event.requestContext.http.method
+  }
+  let path = event && event.requestContext && event.requestContext.resourcePath
+  if (!path) {
+    path = event && event.requestContext && event.requestContext.http && event.requestContext.http.path
```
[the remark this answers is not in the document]
trentm: Ah, I see that https://github.com/elastic/apm/pull/537 clarifies this somewhat.
Note that for v2 usage of `event.requestContext.http.path` it suggests stripping off the `stage` part of the path component.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  }
+
```
trentm, line 57: Also the spec says:

> The agent should use the information in the request and response objects to fill the HTTP context (context.request and context.response) fields in the same way it is done for HTTP transactions.

Perhaps a separate issue to add `transaction.context.{request,response}` fields? We won't be able to re-use the `parsers.getContextFrom{Request,Response}()` functions that take core http objects.

lib/lambda.js
```diff
@@ -18,17 +282,20 @@ function elasticApmAwsLambda (agent) {
...
     })
   }
 
-  function wrapContext (trans, payload, context) {
+  function wrapContext (trans, event, context) {
     shimmer.wrap(context, 'succeed', (succeed) => {
       return function wrappedSucceed (result) {
+        if (result && result.statusCode && result.statusCode >= 500 && isApiGatewayEvent(event)) {
+          trans.result = 'HTTP ' + result.statusCode.toString()[0] + 'xx'
+        }
```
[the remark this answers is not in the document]
basepi: Yes, we look for a case-insensitive `statuscode` key in the response. If present, we convert it to an `HTTP Xxx` and set that in `transaction.result`. So, if the function decides it's a `5xx` then we'll set `5xx` even without an exception. [code](https://github.com/elastic/apm-agent-python/blob/e989439a2702e1cf8256687e6ba1caf8dc20b44d/elasticapm/contrib/serverless/aws.py#L166-L176)

lib/instrumentation/transaction.js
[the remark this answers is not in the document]
astorm: I had a similar thoughts when adding this but was unsure what effect, if any, not serializing these would have, so I erred on the side of cargo cult-ing the existing behavior.  I've opened a separate issue to investigate what effect (if any) _not_ including the properties when empty would have. https://github.com/elastic/apm-agent-nodejs/issues/2416

test/instrumentation/transaction.test.js
astorm: Added a separate issue to address this: https://github.com/elastic/apm-agent-nodejs/issues/2416

lib/lambda.js
```diff
@@ -94,34 +363,40 @@ function elasticApmAwsLambda (agent) {
...
-          const value = payload.headers[key]
+        for (const key of Object.keys(event.headers)) {
+          const value = event.headers[key]
           const lowerCaseKey = key.toLowerCase()
           normalizedHeaders[lowerCaseKey] = value
         }
         parentId = normalizedHeaders.traceparent ? normalizedHeaders.traceparent : normalizedHeaders['elastic-apm-traceparent']
         tracestate = normalizedHeaders.tracestate
       }
+      const isColdStart = !!isFirstRun
```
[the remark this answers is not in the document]
astorm: Can't remember why this was here -- and removing it doesn't seem to effect the tests so sure, it's gone.

lib/lambda.js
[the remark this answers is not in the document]
astorm: Thanks for the note on the case insensitivity @basepi, I'd missed that.  Code updated in 0e250479e5bd31e7cecec11075e696afea5c6fe5

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  } else if (isS3SingleEvent(event)) {
+    setS3SingleData(transaction, event, context, isColdStart)
+  }
+}
 
 function elasticApmAwsLambda (agent) {
   const log = agent.logger
 
-  function captureContext (trans, payload, context, result) {
+  function captureContext (trans, event, context, result) {
```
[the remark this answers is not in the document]
astorm: That's a good point but outside the scope of what I wanted to accomplish with this PR.  I've opened a separate issue to address this: https://github.com/elastic/apm-agent-nodejs/issues/2417

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+    coldstart: isColdStart,
+    execution: context.awsRequestId,
+    trigger: {
+      type: 'http'
+    }
+  }
+  return faas
+}
+
+function setApiGatewayData (transaction, event, context, isColdStart) {
```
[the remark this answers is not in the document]
astorm: I don't have a strong opinion here and the tests should catch any "undefined nested objects" issues so I'm happy to make the change (done in 777458fa1bc8a40b377b0e9acf0d66aa83181652) -- but my default operating mode is that any one function shouldn't be dependent on the behavior of another function, so you'll probably see me do similar things in the future :)

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  const faas = getSharedFaasData(context, isColdStart)
+  faas.trigger.request_id = event && event.requestContext && event.requestContext.requestId
+  faas.trigger.type = 'http'
+  transaction.setFaas(faas)
+
+  transaction.type = 'http'
```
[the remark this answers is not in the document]
astorm: Done in 38051317c91ffdf75d016d9930d31fc774a5e28c

lib/lambda.js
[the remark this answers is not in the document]
astorm: Good call on stripping stage -- made that change in 6545652a2902b4230e2fdc354df86fde05241c6a, including the requested comment on this being about V1/V2.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+      },
+      account: {
+        id: event && event.requestContext.accountId
+      }
+    }
+  }
+  transaction.setCloudContext(cloudContext)
+}
+
+function isApiGatewayEvent (event) {
```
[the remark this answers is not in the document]
astorm: Seems outside the scope of this PR -- I've opened an issue over here for discussion: https://github.com/elastic/apm-agent-nodejs/issues/2418

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+
+  if (agent._conf.captureBody) {
+    messageContext.body = message.Message
+  }
+  transaction.setMessageContext(messageContext)
+}
+
+function isSnsEvent (event) {
+  const records = event.Records ? event.Records : []
+  return records.length > 0 && records[0].EventSource === 'aws:sns'
```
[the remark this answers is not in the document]
astorm: Nice catch -- I saw that for the S3 events but not the SNS/SQS.  Updated in 4683f030c56bd87edb04e330f8503f0061d50883

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+  transaction.setCloudContext(cloudContext)
+}
+
+function isApiGatewayEvent (event) {
+  return !!(event && event.requestContext && event.requestContext.requestId)
+}
+
+function setSqsData (agent, transaction, event, context, isColdStart) {
+  const record = event && event.Records && event.Records[0]
+  record.eventSourceARN = record.eventSourceARN ? record.eventSourceARN : ''
```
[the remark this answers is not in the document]
astorm: Done in adc39b8f4a68d9f5255d5ced0d05563c1aa247c7

lib/lambda.js
astorm: Opened an issue to look into this. https://github.com/elastic/apm-agent-nodejs/issues/2419

lib/lambda.js
astorm: Done in 326912eceae698d4453f27ebb1057b5fa9174f5a

lib/lambda.js
```diff
@@ -94,34 +363,40 @@ function elasticApmAwsLambda (agent) {
...
         parentId = normalizedHeaders.traceparent ? normalizedHeaders.traceparent : normalizedHeaders['elastic-apm-traceparent']
         tracestate = normalizedHeaders.tracestate
       }
+      const isColdStart = !!isFirstRun
+      if (isFirstRun) {
+        isFirstRun = false
+      }
 
-      const trans = agent.startTransaction(context.functionName, type, {
+      const transaction = agent.startTransaction(context.functionName, type, {
```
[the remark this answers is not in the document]
astorm: Can't remember if there was an specific reason for doing this other than I needed to work in this function-space and I always prefer an full `transaction` to `trans`.  I've moved this back to `trans` in 53fee17a32180731cf9d9dee306cafa3623f1039

lib/lambda.js
astorm: 90d4d0d00508b007856e67641085ca138149fc72

lib/lambda.js
astorm: I'm going to honor the APM Server spec for the time being since that's the code that's deployed to the world at the moment. changed in dcc73aae2961b6a9ba1a6b48ff682ab7bb566bb1

lib/lambda.js
astorm: Honoring APM Server spec for now since it's deployed the the world.  Changed in 61ef4f8d0025763b8e8429e29746b1c385bd2d13

lib/lambda.js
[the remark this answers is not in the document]
trentm: @astorm I think this `trans.result = ...` handling will also need to be in `wrapLambdaCallback`. IIUC, when a callback is used the Lambda runtime does *not* call our wrapped `context.succeed()`.

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+
+  let age
+  if (record.attributes && record.attributes.SentTimestamp) {
+    age = Date.now() - record.attributes.SentTimestamp
+  }
+  const messageContext = {
+    queue: record.eventSourceARN,
+    age: age,
+    body: undefined,
+    headers: Object.assign(record.attributes ? record.attributes : {})
```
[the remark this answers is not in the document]
astorm: Done in 832efb9f208865facbd4d5e06c87a530e17455cb

lib/lambda.js
```diff
@@ -2,11 +2,275 @@
...
+
+  if (agent._conf.captureBody) {
+    messageContext.body = record.body
+  }
+  transaction.setMessageContext(messageContext)
+}
+
+function isSqsEvent (event) {
+  const records = event.Records ? event.Records : []
+  return records.length > 0 && records[0].eventSource === 'aws:sqs'
```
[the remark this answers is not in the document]
astorm: done in 692916c2ecc3da447f0e744aabfecb90c850bf03

lib/lambda.js
```diff
@@ -77,9 +345,10 @@ function elasticApmAwsLambda (agent) {
         // sub-span.
         agent.captureError(err, { skipOutcome: true })
         trans.setOutcome(constants.OUTCOME_FAILURE)
+        setTransactionFailure(trans, event)
```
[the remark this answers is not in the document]
astorm: no objection to this being done, but I'm going to keep it out of this specific PR.

lib/lambda.js
```diff
@@ -2,11 +2,301 @@
...
+    },
+    age: {
+      ms: age,
+    },
+    body: undefined,
+    headers: undefined
+  }
+
+  if(agent._conf.captureHeaders) {
+    messageContext.headers = Object.assign({}, record.messageAttributes ? record.messageAttributes : {})
```
trentm, line 154: nit:

```suggestion
    messageContext.headers = Object.assign({}, record.messageAttributes)
```

A falsy value for an argument to Object.assign is skipped, so no need for the extra check.

lib/lambda.js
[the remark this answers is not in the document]
astorm: Sure thing trent, done in 236f6193504ef6da0cbb7d3bb4acb440dadbb41a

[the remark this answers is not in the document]
trentm: @basepi I'm curious why the case-insensitive check. My read of https://docs.aws.amazon.com/apigateway/latest/developerguide/handle-errors-in-lambda-integration.html is that a Lambda "proxy" integration with API Gateway (which, IIUC, is what we are talking about; vs. "custom" integrations) must have a result object with this format:

```
{
  "isBase64Encoded" : "boolean",
  "statusCode": "number",
  "headers": { ... },
  "body": "JSON string"
}
```

Actually, this looks like a better reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format

astorm:
Fixes, changes, and nits addressed, including the spec changes around the diff between the Lambda spec and APM Server's intake spec.

lib/lambda.js
```diff
@@ -2,11 +2,311 @@
...
+  const records = event.Records ? event.Records : []
+  return records.length === 1 && records[0].eventSource === 'aws:s3'
+}
+
+function setTransactionResultFromApiGatewayResponse (trans, result, event) {
+  if (!result) {
+    return
+  }
+  const statusCode = result.statusCode ? result.statusCode : result.statuscode
+  if (statusCode >= 500 && isApiGatewayEvent(event)) {
```
trentm, line 280: I asked in slack if the Python impl isn't guarding on `statusCode >= 500` here. However, that can come afterwards if different -- and with clarification of the spec.

test/lambda/mock/transaction.js
```diff
@@ -14,6 +15,23 @@ module.exports = class TransactionMock {
...
+    Object.assign(this._message = message)
+  }
+
+  setServiceContext (serviceContext) {
+    if (!serviceContext) return
+    this._service = Object.assign(this._service || {}, serviceContext)
+  }
+
+  setFaas (context) {
+    Object.assign(this._faas, context)
```
trentm, line 32: The impl of setCloudContext, setMessageContext, setFaas are wrong here -- though are working by luck.   You perhaps (tho not required) want `this._cloud = undefined` et al in the constructor; then you want the implementations to be like `setServiceContext`.

trentm (changes requested):
Just a thing in  test/lambda/mock/transaction.js

test/lambda/mock/transaction.js
astorm: ab9a381164daa60f93783a8935387c841f872bab

lib/lambda.js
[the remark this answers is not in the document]
basepi: @trentm I don't actually have a good answer. I guess I was trying to be flexible, because `statusCode` seemed like a convention more than a requirement. That said, I just tested with `statuscode` and it threw an internal server error. So I'm going to tear out my case-insensitivity. @astorm I recommend you do the same, sorry for misguiding you with my weird assumptions. :flushed:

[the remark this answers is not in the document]
trentm: Okay, that makes sense. Thanks very much for checking.

trentm (approved):
I played a bit with API Gateway and S3 triggers using this branch. LGTM.
You can drop the `result.statuscode` lowercase thing here, or in a follow-up PR.  Likewise with setting `trans.result = 'HTTP Nxx'` for non-error responses for API Gateway triggers.

Issue #1799, opened 2020-07-26
Calling registerMetric() throws when active: false


MartinKolarik:
**Describe the bug**

The exception is the same as in #1257.

**Environment (please complete the following information)**
- OS: Windows 10
- Node.js version: 14
- APM Server version: N/A
- Agent version: 3.6.1

**How are you starting the agent? (please tick one of the boxes)**
- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

MartinKolarik:
Just an update - this bug is still present in the current version.

astorm:
Thanks for re-bringing this to our attention @MartinKolarik -- we're better staffed these days :)

We've got a PR going to get this issue fixed: https://github.com/elastic/apm-agent-nodejs/pull/2290  -- If you have a moment and could test out the PR branch and confirm that it solves your issue we'd appreciate it (no worries if not though).

MartinKolarik:
Yes, #2290 fixes it, thanks!

trentm:
elastic-apm-node@3.21.0 is now published with this fix.

PR #2290, opened 2021-08-18
fix: crashing error when active:false and agent.registerMetrics is called

astorm:
Closes: #1799

This PR adds guard clauses to the `Metrics` class to prevent crashing errors when `this[registrySymbol]` is not set.  

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


astorm:
jenkins run the tests please

CHANGELOG.asciidoc
```diff
@@ -43,6 +43,7 @@ Notes:
 * Ensure the internal HTTP(S) client requests made by the APM agent to APM
   server are not themselves traced. ({issues}1168[#1168], {issues}1136[#1136])
 
+* Fix crashing error with `agent.registerMetric` and `active:false` configuration. ({issues}[#1799], {pull}[#2290])
```
trentm, line 46:
```suggestion
* Fix crashing error with `agent.registerMetric` and `active:false` configuration. ({issues}1799[#1799], {pull}2290[#2290])
```

test/metrics/index.test.js
```diff
@@ -211,6 +211,18 @@ test('noop counter when not active', function (t) {
   t.end()
 })
 
+test('Metrics objects do not throw/crash when not started', function (t) {
```
trentm, line 214: nit: Could perhaps add a test between these two lines to test `agent.registerMetric(...)` before `agent.start() is called: https://github.com/elastic/apm-agent-nodejs/blob/22423a5b409b8af0a9923624e9af32e591a8159e/test/agent.test.js#L1423-L1424

astorm: Hmmm -- good suggestion but after looking into it my current thinking is to hold off on that -- I'm not sure that additional call brings much value to the test suite.  The `_metrics` object is mocked in that test so we're not actually hitting any of the code paths that were a problem.   Also, it's not clear what we could assert on there given that `registerMetrics` returns nothing regardless of whether the metric is created successfully or not. 

Unless you think changing `registerMetric` to return the value of the call to `this._metrics.getOrCreateGauge(name, callback, labels)` is worthwhile and not a breaking change? @trentm

trentm: > The _metrics object is mocked in that test so we're not actually hitting any of the code paths that were a problem. 

You had me at "this is a mock".

astorm:
CHANGLOG links are fixed (thank you) and additional test was investigated, but ultimately discarded due to the fact we'd be testing a mock.

trentm (approved):

PR #2300, opened 2021-08-24
refactor: clean up types slightly (drop GenericSpan, s/agent./apm./ namespace)

trentm:
- Drop the "GenericSpan" interface in our types. It isn't an external
  concept and there wasn't a lot of utility in using inheritance for it
  in the types file.
- Change from 'agent.' to 'apm.' usage in the types test file because
  'apm' is the prefix we encourage in docs (for good reason I think, it
  is less ambiguous).

### Checklist

- [x] Implement code

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -127,31 +127,27 @@ declare namespace apm {
     setSpanOutcome(outcome: Outcome): void;
   }
 
-  type Outcome = 'unknown' | 'success' | 'failure'
+  type Outcome = 'unknown' | 'success' | 'failure';
 
-  interface GenericSpan {
+  // Transaction API
+  // https://www.elastic.co/guide/en/apm/agent/nodejs/current/transaction-api.html
+  export interface Transaction {
     // The following properties and methods are currently not documented as their API isn't considered official:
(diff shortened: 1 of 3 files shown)


trentm:
I'm considering the checks green. The failure of Integration Tests is the unrelated, ongoing https://github.com/elastic/apm-integration-testing/issues/1188

index.d.ts
```diff
@@ -127,31 +127,27 @@ declare namespace apm {
     setSpanOutcome(outcome: Outcome): void;
   }
 
-  type Outcome = 'unknown' | 'success' | 'failure'
+  type Outcome = 'unknown' | 'success' | 'failure';
```
astorm, line 130: Not necessary for this PR -- but seeing the added `;` makes me wonder if we want to get some sort of auto-fix linter looking at these typescript files as well.

astorm (approved):
Approving, with the caveat that all the use cases for these types aren't clear to me and I'll be curious to see if removing with type causes any weird downstream issues.

index.d.ts
trentm: Yah, agreed it would be nice, but I didn't want to go there in this PR.

PR #2305, opened 2021-08-26
test: refactor to move knowledge for in-process Agent re-use into Agent methods

trentm:
Move away from using the test/_agent.js wrapper in tests, in favour of using the real Agent class directly and `agent.destroy()` at the end of a test to clean up. `Agent#destroy()` has been improved and `Instrumentation#stop()` added to properly handle agent clean-up (within reason) such that tests can serially create multiple instances of the Agent.

`Agent#destroy()` and `Instrumentation#stop()` are mostly what the "test/_agent.js" cleanup was doing with some improvements:
- restore Error.stackTraceLimit
- disable the running asyncHook

Not all usage of test/_agent.js has been removed -- just the major users agents.test.js and config.test.js.
I'll look at doing the others in subsequent changes. This PR is big enough already.

The motivation for this work is to help with test suite updates for the [async context work](https://github.com/elastic/apm-agent-nodejs/pull/2181). The internals of `Instrumentation` are changing, so it has been a pain that the test suite assumes so many internals.


trentm (commented):
Add some reviewer notes.

astorm (approved):
The changes here look reasonable and seem like appropriate steps towards being able to better test changes we're making to our context handling. 

Approving, pending one discretionary question.

I noticed that a run of `test/agent.test.js` went from 853 assertions  to 572 assertions with these changes.  

    $ npx tape test/agent.test.js
    //...
    1..853
    # tests 853
    # pass  853

    # ok
    
    $ git checkout trentm/sans-mock-agent
    $ npx tape test/agent.test.js
    //...
    1..572
    # tests 572
    # pass  572

    # ok

These seemed to be from test suites that used the now removed `APMServerWithDefaultAsserts` function.  Here's one example

Before: 
```
# capture location stack trace - off (param msg)
ok 539 should be a POST request
ok 540 should be sent to the intake endpoint
ok 541 metadata should always be sent first
ok 542 should be strictly equal
ok 543 should be deeply equivalent
ok 544 should be deeply equivalent
ok 545 should be deeply equivalent
ok 546 should be strictly equal
ok 547 should be strictly equal
ok 548 should be strictly equal
ok 549 should be truthy
ok 550 should be strictly equal
ok 551 should have a pid greater than 0
ok 552 should have a process title
ok 553 should be strictly equal
ok 554 should be deeply equivalent
ok 555 should have at least two process arguments
ok 556 should be strictly equal
ok 557 should not have a log.stacktrace
ok 558 should not have an exception
```

After:
```
# capture location stack trace - off (param msg)
ok 384 APM server got 2 events
ok 385 should be strictly equal
ok 386 should not have a log.stacktrace
ok 387 should not have an exception
```

Is there any concern about this lost coverage? Or were the tests that came along with the old `APMServerWithDefaultAsserts` now redundant?

trentm:
> Is there any concern about this lost coverage? Or were the tests that came along with the old `APMServerWithDefaultAsserts` now redundant?

There is no concern in my humble opinion. Here is what the old test code would do for each of the 33 usages of APMServerWithDefaultAsserts:

```js
function APMServerWithDefaultAsserts (t, agentOpts, mockOpts) {
  var server = APMServer(agentOpts, mockOpts)
    .on('request', validateRequest(t))
    .on('data-metadata', validateMetadata(t))
 // ...
}

function validateRequest (t) {
  return function (req) {
    t.strictEqual(req.method, 'POST', 'should be a POST request')
    t.strictEqual(req.url, '/intake/v2/events', 'should be sent to the intake endpoint')
  }
}

function validateMetadata (t) {
  return function (data, index) {
    t.strictEqual(index, 0, 'metadata should always be sent first')
    assertMetadata(t, data)
  }
}
```

- The `validateRequest` asserts are redudant because the `MockAPMServer` being used now only captures events for tests if the request is a `POST /intake/v2/events`:

```js
      } else if (req.method === 'POST' && parsedUrl.pathname === '/intake/v2/events') {
        body
          .split(/\n/g) // parse each line
          .filter(line => line.trim()) // ... if it is non-empty
          .forEach(line => {
            this.events.push(JSON.parse(line)) // ... append to this.events
          })
```

- The `assertMetadata` is still being done, but just not for every test -- in about 12 of them instead of 33.

astorm:
@trentm sounds good!  Approved x2

trentm:
Thanks for the review!

PR #2313, opened 2021-08-31
tests: refactor handling of config var type in config.test.js

trentm:
No functional change.

### background on why this change

This is work that came out of earlier work on https://github.com/elastic/apm-agent-nodejs/pull/2193

The `longFieldMaxLength` config var for that PR was *originally* a "strict bytes" type. "Strict bytes" here is a config var type similar to the existing "bytes" type, but per https://github.com/elastic/apm/blob/master/specs/agents/configuration.md#durationsize-config-legacy-considerations all new "bytes" config vars must be *strictly* a string with the units. Our current "bytes" vars are loose (don't require a suffix like "b" or "kb", and allow an integer instead of a string) here, so a separate type on "config.js" will be necessary if/when we add one.

If/when we add a "strict" bytes/size type, then we'll need this refactoring of config.test.js -- or something better than this, to not have this test code get lost in the special overlapping meanings of these vars:

```js
    var bool = typeof fixture[2] === 'boolean'
    var url = fixture[0] === 'serverUrl' // special case for url's so they can be parsed using url.parse()
    var file = fixture[0] === 'serverCaCertFile' // special case for files, so a temp file can be written
    var number = typeof fixture[2] === 'number'
    var array = Array.isArray(fixture[2])
```

A single "type" for the type of the config var, and `switch`es below to handle them is cleaner.

### DIFF
--- a/test/config.test.js
+++ b/test/config.test.js
@@ -100,11 +100,23 @@ var truthyValues = [true, 'true']
 
 optionFixtures.forEach(function (fixture) {
   if (fixture[1]) {
-    var bool = typeof fixture[2] === 'boolean'
-    var url = fixture[0] === 'serverUrl' // special case for url's so they can be parsed using url.parse()
-    var file = fixture[0] === 'serverCaCertFile' // special case for files, so a temp file can be written
-    var number = typeof fixture[2] === 'number'
-    var array = Array.isArray(fixture[2])
+    var type
+    if (typeof fixture[2] === 'boolean') {
+      type = 'bool'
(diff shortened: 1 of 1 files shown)


trentm:
For the record, here is what adding a new strict bytes var to the Node.js agent could look like. This is code I had for earlier #2193 work, but don't need right now: https://gist.github.com/trentm/ae3490fa9e2bc800e75a58bd13d6b96f

astorm (approved):
✅ 👍

Issue #2294, opened 2021-08-19
agent.flush() doesn't flush an ended span that is currently in `span._encode()`


trentm:
**Describe the bug**

Currently the agent's API does not provide a way to know when an ended span has had a chance to be sent to agent._transport. On `span.end()`, the encoding of the span (`span._encode()`) is done **asynchronously** before the span object is sent to the transport (`agent._transport.sendSpan(encodedSpanObject)`).  Currently a call to `agent.flush([cb])` calls `agent._transport.flush(cb)`, which only flushes objects already on the transport's queue.

For one, this is a pain for testing. The following does not suffice:

```js
// ...
mySpan.end()
agent.flush(function () {
    // test that APM server got 'mySpan' here
})
```

The typical (poor) handling in tests is:

```js
mySpan.end()
setTimeout(function () {
    agent.flush(function () {
        // test that APM server got 'mySpan' here
    })
}, SOME_DELAY)
```
where that $SOME_DELAY ms delay is a race.

Another use case is AWS lambda support. I suspect that a ending a span and then immediately completing the lambda handler function (using `captureSpanStackTraces:true, spanFramesMinDuration:-1` to ensure longer `span._encode()` time) may result in that span not being flushed before the lambda container is frozen.

**To Reproduce**

TODO: A simple script repro. Also show a lambda repro.

**Expected behavior**

It would be helpful for testing and perhaps (to be demonstrated) required for lambda to have some programmatic way to wait for ended-and-currently-encoding spans to be flushed.

**Additional context**

A potential candidate here would be to extend the current `agent.flush([cb])` to wait for these inflight encoding spans. Thinking out loud:

- Add an `Instrumentation.flush` that handles this; and get `Agent.flush` to call that.
- There would be some subtlety there to not starve waiting for span._encode()s that come in after agent.flush() was called.

trentm:
Likewise, `agent.flush()` does not ensure that a just-called `agent.captureError()` will get flushed to APM server before calling back. `agent.captureError(..., cb)` *does* support its own callback, so this isn't as much of any issue for testing. However, if expanding `Agent.prototype.flush()` to mean "flush all in-flight APM event processing", then it is worth considering this as well.

trentm:
Somewhat similarly, the node.js agent doesn't have any concept of flushing metrics at a given time -- only periodically doing so at `metricsInterval`.

[...]

PR #2316, opened 2021-09-03
tests: refactor breakdown metrics tests to not override Agent internals

trentm:
This refactors breakdown.test.js to not need a `resetAgent(...)` that knows so much about Agent internals. Now that we have a usable `agent.destroy()`, use that and create a new Agent for each test case. 

This also moves away from the test/_mock_http_client.js testing style of waiting for an expected number of sent events to decide when to end a test case. This style is brittle when there is a test failure: either hanging waiting for an event that isn't coming, or crashing with `throw new Error('too many writes')`.

### DIFF
--- a/lib/metrics/reporter.js
+++ b/lib/metrics/reporter.js
@@ -52,8 +52,10 @@ class MetricsReporter extends Reporter {
         }
       }
 
-      for (const metric of seen.values()) {
-        this._agent._transport.sendMetricSet(metric)
+      if (this._agent._transport) {
+        for (const metric of seen.values()) {
+          this._agent._transport.sendMetricSet(metric)
+        }
       }
     })
 
--- a/test/_capturing_transport.js
+++ b/test/_capturing_transport.js
@@ -1,0 +1,77 @@
+'use strict'
+
+// An Agent transport -- i.e. the APM server client API provided by
+// elastic-apm-http-client -- that just captures all sent events.
+//
+// Usage:
+//    const testAgentOpts = {
+//      // ...
+//      transport () { return new CapturingTransport() }
+//    }
+//
+//    test('something', function (t) {
+//      const agent = new Agent().start(testAgentOpts)
+//      // Use `agent`, then assert that
+//      // `agent._transport.{spans,transactions,errors,metricsets}` are as
+//      // expected.
+//      agent.destroy()
+//      t.end()
+//    })
+//
+// Note: This is similar to _mock_http_client.js, but avoids the testing model
+// of using an expected number of sent APM events to decide when a test should
+// end.
+
+class CapturingTransport {
+  constructor () {
+    this.spans = []
+    this.transactions = []
+    this.errors = []
+    this.metricsets = []
+  }
+
+  config (opts) {}
+
+  addMetadataFilter (fn) {}
+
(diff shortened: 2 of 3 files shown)


lib/metrics/reporter.js
```diff
@@ -52,8 +52,10 @@ class MetricsReporter extends Reporter {
         }
       }
 
-      for (const metric of seen.values()) {
-        this._agent._transport.sendMetricSet(metric)
+      if (this._agent._transport) {
```
trentm, line 55: Reviewer note: This was added to not crash in the metrics reporting interval if, in the interim, `agent.destroy()` has been called -- which sets `this._agent._transport` to null.  Similar is already done for `sendSpan` here: https://github.com/elastic/apm-agent-nodejs/blob/ec719c29384ef721d615339e9a6ea3541d15154d/lib/instrumentation/index.js#L268

astorm (approved):
On objection to this -- it does what it says on the tin and it's hard to argue with better isolated tests. Approving.

One thing I did notice -- in this new branch the `breakdown.test.js` tests take significantly longer to run. On my local dev machine the `test/metrics/breakdown.test.js` tests took `14.345` total seconds where the old suites took `0.369` seconds.  This was using `npx` to invoke the tests via `tape`


> With the new branch

    $ time npx tape test/metrics/breakdown.test.js
    // ...
    ok 101 does not have un-ended spans

    1..101
    # tests 101
    # pass  101

    # ok
    $ npx tape test/metrics/breakdown.test.js  0.42s user 0.08s system 3% cpu 14.345 total

> Old/current master

    $ time npx tape test/metrics/breakdown.test.js 
    // ...
    ok 101 does not have un-ended spans

    1..101
    # tests 101
    # pass  101

    # ok

    $ npx tape test/metrics/breakdown.test.js  0.33s user 0.06s system 104% cpu 0.369 total

Not a deal breaker (it's stills seconds, not minutes), just an observation.

trentm:
> One thing I did notice -- in this new branch the `breakdown.test.js` tests take significantly longer to run.

Yes, that is unfortunate. The earlier test model of "do test assertions and end **as soon as I get the expected number of items on the transport**" is faster, because the items we are waiting for (the encoded span and the breakdown metrics) are sent (via `_transport.sendSpan()` and `_transport.sendMetricSet()`) very soon later -- but asynchronously.

The test model I switched to here is:
1. Do the thing (create some transactions and spans)
2. Use whatever API is available on the public API to wait until we are sure the agent will have sent all items (transactions, spans, metric sets).
3. Only *then*, do test assertions.

This model is, IMHO, less error prone, as mentioned in the PR description. However, that step 2 is in this case way slower because the only "API available on the public API" to know that we've gotten the breakdown metrics for all ended transactions and spans is "wait for the next metricsInterval". I think eventually we'll want our agent.flush() or similar API to improve on this (https://github.com/elastic/apm-agent-nodejs/issues/2294).


So yes, this adds ~14s to our test run of ~15 minutes (in CI: https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/master/), which I think should be okay. Eventually if/when we can move to `tap` and parallel running of test files, then this will not be a slowdown of the overall test run time at all.

PR #2319, opened 2021-09-07
tests: refactor away need for test/instrumentation-specific mocks

trentm:
Drop the need for the mocks _agent.js and _instrumentation.js under
test/instrumentation that reach into Agent internal implementation
details. This will help coming re-work of some Instrumentation
internal implementation details.

This depends on #2316 going in first (for the added
"test/_capturing_transport.js").


astorm (approved):
Overall this look good -- as in it does the thing and moves us away from the mock agent.  Also, no noticeable perf. regressions in the time these particular new tests take to run.   Approving.

One thing to note -- the `npx tape test/instrumentation/transaction.test.js` has a different number of assertions on this branch when compared to master (`140` on master, `136` on this branch)  I did not dig deeply into it but my presumption is these are assertions lost due to how `mockInstrumentation` worked.  Not a deal break, just an observation in case this wasn't expected.

trentm:
> the `npx tape test/instrumentation/transaction.test.js` has a different number of assertions on this branch when compared to master (`140` on master, `136` on this branch)

We *add* two assertions near the top, e.g. here is one added assertion: https://github.com/elastic/apm-agent-nodejs/pull/2319/files#diff-115ec49fff221477451b17e1fca2a5208448489ccbb84c4eb7b53c6c4016f3caR189-R203

Then for six replacements of mockInstrumentation, we drop a `    t.pass('should end the transaction')` assertion, e.g. here: https://github.com/elastic/apm-agent-nodejs/pull/2319/files#diff-115ec49fff221477451b17e1fca2a5208448489ccbb84c4eb7b53c6c4016f3caR356-R357
Those are safe to drop because the subsequent assertions on `agent._transport.transactions[0]` ensures the transaction *was* ended and sent.

Thanks for the review.

PR #2321, opened 2021-09-10
refactor: remove the unhelpful test suite namespacing

trentm:
No functional change.
Namespacing these test cases under an "acceptance" group doesn't
add anything, and it gets in the way of using `test.only` when
debugging this test file.


trentm:
Most of the diff is whitespace. This diff is nicer: https://github.com/elastic/apm-agent-nodejs/pull/2321/files?w=1

PR #2323, opened 2021-09-13
tests: make breakdown.test.js less racy

trentm:
This improvements these tests to be less susceptible to a metricsInterval
and sendMetricSet race. In the context mgmt feature branch, CI would
fairly reliably fail on this race. The agent doesn't provide a mechanism
to "wait for breakdown metrics for all currently ended transactions and
spans".

### DIFF
--- a/test/metrics/breakdown.test.js
+++ b/test/metrics/breakdown.test.js
@@ -119,6 +119,52 @@ const testAgentOpts = {
   metricsInterval: testMetricsInterval
 }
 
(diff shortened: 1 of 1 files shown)


astorm (approved):
👍 approving.  Also, per semi-standard practice

Before this change we have 101 passing tests with `time npx tape test/metrics/breakdown.test.js` reporting `14.356 total` (seconds)

After this change we have 101 passing tests with `time npx tape test/metrics/breakdown.test.js` reporting `1.780 total` (seconds)

PR #2325, opened 2021-09-15
docs: start an examples/ dir with short demos of tracing

trentm:
@astorm Objections to starting an examples/ dir with small working demo scripts?

I find scripts like these a very useful working starting point for
understanding tracing concepts or tracing of a particular library.

/cc @bmorelli25 FYI only :)


bmorelli25:
Love this idea

astorm (dismissed):
No objection, these can be nice examples to have around.  Approving.

Question: Are you concerned that folks might see these examples and believe they _need_ to start their own transactions with Memcached, mysql, or postgres?

trentm:
> Question: Are you concerned that folks might see these examples and believe they _need_ to start their own transactions with Memcached, mysql, or postgres?

Yah, fair. I'll add a comment about that.

astorm (approved):

Issue #2292, opened 2021-08-19
stop tracking and sending `transaction.sync`


trentm:
The APM server intake model does not include it: https://github.com/elastic/apm-server/blob/2c65c70cda74aa3eb4633c022417548ef52d89c7/model/modeldecoder/v2/model.go#L797-L846
It is only defined for spans: https://github.com/elastic/apm-server/blob/2c65c70cda74aa3eb4633c022417548ef52d89c7/model/modeldecoder/v2/model.go#L647-L648

The current Node.js agent is:
- sending it here: https://github.com/elastic/apm-agent-nodejs/blob/1207b49e8b08f9c1d7153a21dd369aa13faa1661/lib/instrumentation/transaction.js#L132
- updating it here: https://github.com/elastic/apm-agent-nodejs/blob/1207b49e8b08f9c1d7153a21dd369aa13faa1661/lib/instrumentation/index.js#L363
- and here: https://github.com/elastic/apm-agent-nodejs/blob/1207b49e8b08f9c1d7153a21dd369aa13faa1661/lib/instrumentation/async-hooks.js#L97-L99

trentm:
@astorm Do you recall why you split this test case in test/instrumentation/async-hooks.js into two separate tests cases?
https://github.com/elastic/apm-agent-nodejs/pull/2081/files#diff-7452104585e564b9dc20f0b64aa156c0701046c833513e01aaf2c55a1f172260

Issue #1996, opened 2021-03-12
Async spans are being reported as sync


dgieselaar:
Some clearly async spans are being reported as sync, and show up in the UI as blocking. I'm not sure when this was first introduced. Here's an example:

![image](https://user-images.githubusercontent.com/352732/110910811-2054a400-8312-11eb-8deb-e71787dca179.png)

I don't think the problem is exclusive to elasticsearch spans, I've seen it happen with custom spans as well. It seems like it was introduced in a fairly recent version of the nodejs agent:

![image](https://user-images.githubusercontent.com/352732/110910912-4417ea00-8312-11eb-8783-a804b12a4844.png)

astorm:
@dgieselaar Thanks for reporting this behavior.

First -- looking through the changelog, [3.9 includes a fix](https://github.com/elastic/apm-agent-nodejs/blob/master/CHANGELOG.asciidoc#390---20201130) where we were incorrectly setting the `sync` property on a span.  Could this be the version where you've seen this change in behavior?

Second -- what about the spans involved makes you believe they should be marked as `sync=false`?  Related, can you provide some isolated code/a repro that demonstrates the behavior?  

With the above information we'll be able to have a better information conversation about all this :) 

Thanks again for reporting!

dgieselaar:
@astorm:
> First -- looking through the changelog, 3.9 includes a fix where we were incorrectly setting the sync property on a span. Could this be the version where you've seen this change in behavior?

Could be, yeah. I haven't tested all versions.

> Second -- what about the spans involved makes you believe they should be marked as sync=false?

It's an elasticsearch span, and it was previously reported as async. Maybe I'm missing something, but it seems like it (obviously?) should not be reported as blocking?

> Related, can you provide some isolated code/a repro that demonstrates the behavior?

The issue seems fairly obvious to me and I'm not sure why a repro here is necessary. I probably don't have time for that, but let me know if I can help out otherwise, happy to chat on Slack about it next week if needed.

astorm:
That's fair @dgieselaar -- here's my take on things.  

1. We fixed a bug in 3.9 where spans that began and ended in the same Node.js async context were incorrectly reported as `async` when they should have been reported as `sync`.  (an `async` span is a span that starts execution in one async context, but ends in another)

2. You've reported seeing this change in behavior

[...]

PR #2326, opened 2021-09-15
fix: more reliable span.sync determination, drop unused transaction.sync

trentm:
Current `span.sync` is tracked via:
- the "before" async hook setting it false on the "active" span, and
- Instrumentation#bindFunction's wrapper setting it false, on the
  fair assumption that all usages of bindFunction are for async
  callbacks.
The former has issues when there are multiple active spans within a
single async task -- as is the case with Elasticsearch instrumentation
(issue #1996) and the aws-sdk instrumentations (which have manual
workarounds).

This changes to set sync=false if the executionAsyncId() at end-time
is different than at start-time. This works for whatever `asyncHooks`
config var value.

One change in behaviour is that the value of span.sync is only
guaranteed to be accurate after span.end() is called. Given that
span.sync isn't a public field, I believe this is fine.

This change also drops the 'transaction.sync' fields. It was never
used by APM server.

Fixes: #1996
Fixes: #2292

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Add CHANGELOG.asciidoc entry (waiting on 3.21.0 release first, #2329)

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -353,11 +353,7 @@ var wrapped = Symbol('elastic-apm-wrapped-function')
 // the callback function gets instrument on "the right" transaction and span.
 //
 // The instrumentation programmer is still responsible for starting a span,
-// and ending a span.  Additionally, this function will set a span's sync
-// property to `false` -- it's up to the instrumentation programmer to ensure
-// that the callback they're binding is really async.  If bindFunction is
-// passed a callback that the wrapped function executes synchronously, it will
-// still mark the span's `async` property as `false`.
+// and ending a span.
 //
 // @param {function} original
 Instrumentation.prototype.bindFunction = function (original) {
@@ -379,8 +375,6 @@ Instrumentation.prototype.bindFunction = function (original) {
     ins.currentTransaction = trans
     ins.bindingSpan = null
     ins.activeSpan = span
-    if (trans) trans.sync = false
-    if (span) span.sync = false
     var result = original.apply(this, arguments)
     ins.currentTransaction = prevTrans
     return result
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -35,6 +35,7 @@ function Span (transaction, name, ...args) {
   this._message = null
   this._stackObj = null
   this._capturedStackTrace = null
+  this.sync = true
   this._startXid = executionAsyncId()
 
   this.transaction = transaction
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { executionAsyncId } = require('async_hooks')
 var util = require('util')
 
 var Value = require('async-value-promise')
@@ -34,6 +35,7 @@ function Span (transaction, name, ...args) {
   this._message = null
   this._stackObj = null
   this._capturedStackTrace = null
+  this._startXid = executionAsyncId()
 
   this.transaction = transaction
   this.name = name || 'unnamed'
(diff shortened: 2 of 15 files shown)


lib/instrumentation/span.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+const { executionAsyncId } = require('async_hooks')
```
astorm, line 3: Question: Are there any perf. concerns with the `executionAsyncId()` function calls?  With two calls per span, a span heavy service/application is going to be calling the function a moderate amount.  I'm not familiar enough the async_hooks implementation to know how heavy the call to `executionAsyncId` is.

lib/instrumentation/index.js
```diff
@@ -379,8 +375,6 @@ Instrumentation.prototype.bindFunction = function (original) {
     ins.currentTransaction = trans
     ins.bindingSpan = null
     ins.activeSpan = span
-    if (trans) trans.sync = false
-    if (span) span.sync = false
```
astorm, line 383: Question: Am I correct that a consequences of removing "`if (span) span.sync = false`" is that folks using the patch-async nodejs-context tracker (i.e. [the `asyncHooks` configuration ](https://www.elastic.co/guide/en/apm/agent/nodejs/master/configuration.html#async-hooks) is set to false) will no longer have any tracking of `span.sync` value?  If so, and if we're comfortable with this trade off (I am)  this seems like it's worth mentioning in the configuration docs for the `asyncHooks` configuration and/or calling out with a CHANGELOG entry.

astorm (commented):
Neat! (and smart!) -- overall I like this change and this seems like a better/more-accurate way of tracking whether a span has cross an asynchronous boundary (I'm increasingly skeptical of how _useful_ this is to track for the general Node.js user -- but that's a separate issue :)) 

Happy to get on the approval train once the perf. and patch-async questions are worked through.

lib/instrumentation/index.js
trentm: Actually, even with `asyncHooks=false` this still works. `span.sync` determination is now totally separate from what context manager is being used (patch-async, or async-hooks ... and eventually AsyncLocalStorage-based, etc.). The only requirement is that `async_hooks.executionAsyncId()` is available -- which it is since Node v8.2.0.

lib/instrumentation/span.js
trentm:
```js
const Benchmark = require('benchmark')
const { executionAsyncId } = require('async_hooks')

console.log(`# Benchmark executionAsyncId()`);

var suite = new Benchmark.Suite({
  onStart: function () {
  }
})

suite
  .add('executionAsyncId', function() {
    executionAsyncId()
  })

  .on('cycle', function(event) {
    console.log(String(event.target));
  })
  .on('complete', function() {
    console.log('Fastest is: ' + this.filter('fastest').map('name'));
  })
  .run({ async: false });
```

```
$ node bench-executionAsyncId.js
# Benchmark executionAsyncId()
executionAsyncId x 930,504,788 ops/sec ±1.42% (85 runs sampled)
```

Comparing to other benchmarks, that is the same speed order as just doing a field lookup in a mapping. Checking the source, that appears to be exactly what it is: https://github.com/nodejs/node/blob/v16.9.1/lib/internal/async_hooks.js#L561-L563

Note that we call `executionAsyncId()` currently for every `agent.currentTransaction` or `agent.currentSpan` (when asyncHooks=true).

lib/instrumentation/index.js
astorm: Ah -- it's not reliant our hook specifically 👍

lib/instrumentation/span.js
astorm: 👍

astorm (approved):
👍

PR #2330, opened 2021-09-15
fix: ensure metrics collection is actually disabled with metricsInterval=0s

trentm:
Also guard against:
- `metricsInterval:null` resulting in metrics being enabled without any
  log message, which is perhaps slightly surprising.
- `metricsInterval:-1s` (any negative value) resulting in metrics being
  collected *as fast as possible* via
  `setInterval(collectAllTheMetrics, -1000)`.

An invalid value for this config var, and the other "POSTIVE_TIME_OPTS",
will result in (a) a log.warning and (b) falling back to the default
value.

A slight semantic change on the internal Metrics object means that the
`getOrCreate...()` metric functions now return undefined when metrics
are disabled via `metricsInterval`. This doesn't affect any public API.

### Repro

```js
// no-metrics-please.js

require('./').start({
  serviceName: 'foo',
  metricsInterval: '0s'
})

const http = require('http')

const hostname = '127.0.0.1'
const port = 3000
const server = http.createServer((req, res) => {
  res.end('pong')
})
server.listen(port, hostname, () => {
  console.log(`Server running at http://${hostname}:${port}/`)
})
```

Run the above script, then do at least one request via `curl -i localhost:3000/`.

The agent's internal `Metrics` object will have `enabled === false` and calls:

```
      this[registrySymbol] = new MetricsRegistry(this[agentSymbol], {
        reporterOptions: {
          defaultReportingIntervalInSeconds: 0,
          enabled: false,
          unrefTimers: !refTimers,
          logger: new NoopLogger()
        }
      })
```

The `enabled: false` there is just for our `MetricsRegistry` class.
Internally the `Reporter` from measured-reporting gets the `defaultReportingIntervalInSeconds: 0` and does this:

```js
    this._defaultReportingIntervalInSeconds =
      options.defaultReportingIntervalInSeconds || DEFAULT_REPORTING_INTERVAL_IN_SECONDS;
```

hence defaulting to its internal 10s interval. Then the first time `<reporter>.reportMetricOnInterval(metricKey)` is called (which happens for breakdown metrics on that `curl ...` call above), that `defaultReportingIntervalInSeconds: 10` is used in a `setInterval` to call metrics collectors.

We don't *report* any metrics to APM server because of this in MetricsRegistry:

```js
  _reportMetrics (metrics) {
    if (!this.enabled) return
...
```

However, collection can still be happening, which is work that shouldn't be done.

### Repro metricsInterval=-1s

With a negative value you get approaching 100% CPU usage as system and runtime metrics are collected in a setInterval with a negative delay -- which means run as fast as possible (subject to a JS engine-defined minimum delay).

### Checklist

- [x] Implement code
- [x] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,24 @@ Notes:
 === Node.js Agent version 3.x
 
 
+==== Unreleased
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
+[float]
+===== Bug fixes
+
+* Guard against a negative value of `metricsInterval`, which can lead to
+  high CPU usage as metrics are collected as fast as possible. Also ensure
+  no metrics collection can happen if `metricsInterval="0s"` as intended.
+  Before this change it was possible for some metric collection to still
+  happen, even though none would be reported. ({pull}2330[#2330])
+
+
 [[release-notes-3.21.0]]
 ==== 3.21.0 2021/09/15
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -36,8 +36,6 @@ Notes:
 
 [float]
 ===== Features
-[[release-notes-3.21.1]]
-==== 3.21.1 2021/09/16
 
 [float]
 ===== Bug fixes
@@ -52,6 +50,14 @@ This change also guards against negative and invalid values in the following
 configuration options: `abortedErrorThreshold`, `apiRequestTime`, and
 `serverTimeout`. If an invalid value is given, then will fallback to their
 default value.
+
+
+[[release-notes-3.21.1]]
+==== 3.21.1 2021/09/16
+
+[float]
+===== Bug fixes
+
 * Update types to avoid imports of `@types/...` modules (other than
   `@types/node`), so that TypeScript users of elastic-apm-node need not
   manually `npm install @types/connect @types/pino @types/aws-lambda` to
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -45,6 +45,11 @@ Notes:
   no metrics collection can happen if `metricsInterval="0s"` as intended.
   Before this change it was possible for some metric collection to still
   happen, even though none would be reported. ({pull}2330[#2330])
++
+This change also guards against negative and invalid values in the following
+configuration options: `abortedErrorThreshold`, `apiRequestTime`, and
+`serverTimeout`. If an invalid value is given, then will fallback to their
+default value.
 
 
 [[release-notes-3.21.0]]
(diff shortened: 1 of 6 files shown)


lib/config.js
```diff
@@ -400,6 +403,7 @@ function normalize (opts, logger) {
   normalizeNumbers(opts)
   normalizeBytes(opts)
   normalizeArrays(opts)
+  normalizePositiveTime(opts, logger)
```
astorm, line 406: 👍 I ran some tests and this function appears to behave as expected -- non-positive and null/undefined values use the values from DEFAULTS.

CHANGELOG.asciidoc
```diff
@@ -29,6 +29,24 @@ Notes:
...
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
+[float]
+===== Bug fixes
+
+* Guard against a negative value of `metricsInterval`, which can lead to
```
astorm, line 43: Is it worth calling out in the CHANGELOG that the configuration handling for 'abortedErrorThreshold', 'apiRequestTime' are also changed/fixed such that negative or null/undefined values now result in the default values being used?

lib/config.js
```diff
@@ -512,9 +516,24 @@ function normalizeBytes (opts) {
   }
 }
 
+function normalizePositiveTime (opts, logger) {
```
astorm, line 519: nit: Would it be worth getting some unit tests around both `normalizePositiveTime` and `toSeconds`?  The case for this: They both seem like classic candidates for these sorts of tests (clear inputs, clear outputs) and having the edge-cases/invalid configurations in tests makes the intended behavior of these functions clearer in the future.

astorm (approved):
👍  Overall looks like a positive incremental improvement of the agent.  A few questions/suggestions below but I'm comfortable with a merge as is.  Approving.

lib/config.js
trentm: I pale a little bit at the warty test/config.test.js ... :)  but I'll give it a go.

trentm: Added in https://github.com/elastic/apm-agent-nodejs/pull/2330/commits/2160e98ac0a1d4eca8d7191180804eed143b7e5a

astorm (dismissed):

CHANGELOG.asciidoc
trentm: Done in commit f9e992a842607f931d708e75331aa76dfe738c86

trentm:
Failure in GH Action CI:

```
not ok 16 captureError is fast (<3ms): 3.688309ms
```

is due to the inherent race in disable-send.test.js with possibly-slow CI:

```js
      // Test captureError speed as a proxy for testing that it avoids
      // stacktrace collection when disableSend=true. It isn't a perfect way
      // to test that.
      const durationMs = duration[0] / 1e3 + duration[1] / 1e6
      const THRESHOLD_MS = 3 // Is this long enough for slow CI?
      t.ok(durationMs < THRESHOLD_MS, `captureError is fast (<${THRESHOLD_MS}ms): ${durationMs}ms`)
```

Just retrying for now.

trentm:
> but I'm comfortable with a merge as is. Approving.

I'm taking that as a +1.

Issue #2331, opened 2021-09-16
I cannot longer compile with latest version (3.21.0)


blaart:
node_modules/elastic-apm-node/index.d.ts:6:33 - error TS2307: Cannot find module 'aws-lambda' or its corresponding type declarations.

6 import type * as AwsLambda from 'aws-lambda';

AckerApple:
Confirming having same issue

trentm:
Thanks. This will be from #2118 (PR: #2119) that (among other things) moved from inlining AwsLambda types to referring to `@types/aws-lambda`. 

Note that that `@types/aws-lambda` was added to *devDependencies* and not *dependencies*. However the same is true of other deps. These are all in devDependencies. 

```
    "@types/aws-lambda": "^8.10.77",
    "@types/connect": "^3.4.34",
    "@types/node": "^13.7.4",
    "@types/pino": "^6.3.8",
```

`@types/node` had been in devDependencies already in previous releases and I gather that was working for folks. I don't know if `@types/node` is a special case or if it is the existing `/// <reference` https://github.com/elastic/apm-agent-nodejs/blob/6547c1e5d75479f20cc83e476f480e9084d10c75/index.d.ts#L1 that makes the difference.

trentm:
@blaart @AckerApple The

```js
import type * as Connect from 'connect';
```

line appears before the AwsLambda import. I'm slightly curious why your error wasn't first from that line. Do you perhaps already have `@types/connect` installed?

```
npm ls @types/connect
```

trentm:
I'd based some of the recent work on fastify's handling of types files. I had not previous seen this: https://github.com/fastify/fastify/blob/e77bfa26809e08e82a5f3180572ccd4d4c4d771a/types/logger.d.ts#L2-L18

It is their current justification for avoiding directly importing `@types/pino` in their types file (and then requiring it to be included in their "dependencies").

I am not leaning to doing the same. I regret having switched to *not* inlining the types from deps that only have their types published as a separate 3rd-party `@types/...`. I will try to get to reverting that soon.

AckerApple:
I have run npm list to give you a dependency tree to study.

I have actually moved to NOT directly importing dependencies because a few times lately its had me stuck and the best solution was to not have direct dependencies. I caution you, that if an additional dependency is needed do NOT make it a direct dependency. Instead put in the README.md install process what's needed such as `npm install aws-lambda  apm-agent-nodejs`

```sh
npm list @types/connect

[...]

PR #2335, opened 2021-09-16
fix: avoid import of any '@types/...' packages in types file

trentm:
This ensures that TypeScript users do not need to install separate
'@types/...' packages to build -- with the lone exception of
'@types/node'.

We avoid imports by instead inlining the required types. This is a
trade-off. We don't want '@types/...' in "dependencies" because they
can be big and aren't really "official" types for those packages.

Fixes: #2331

### Checklist

- [x] Implement code
- [ ] ~Add tests~  I'm skipping this because it is relatively difficult to test. The only way to simulate the failure is to `npm prune --production` or to install a fully separate `node_modules/...` tree with just the "dependencies" of elastic-apm-node. I don't think it is worth the churn during tests.
- [x] Update documentation


astorm (approved):
👍

PR #2342, opened 2021-09-22
[CI] use obltGitHubComments

v1v:
## What does this PR do?

Use https://github.com/elastic/apm-pipeline-library/blob/master/vars/obltGitHubComments.txt

## Why is it important?

Avoid hardcoded strings and easy to normalise the GitHub comments in all the oblt projects

You can also see in the GitHub comment that there is a specific section for **some** of those GitHub comments called `🤖 GitHub comments`

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -25,7 +25,7 @@ pipeline {
     quietPeriod(10)
   }
   triggers {
-    issueCommentTrigger('(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?(?:(module|benchmark)\\W+)?tests(?:\\W+please)?.*')
+    issueCommentTrigger("(${obltGitHubComments()}|^run (module|benchmark) tests)")
   }
   parameters {
     booleanParam(name: 'Run_As_Master_Branch', defaultValue: false, description: 'Allow to run any steps on a PR, some steps normally only run on master branch.')


kuisathaverat (approved):

cachedout (approved):

trentm:
This changes the GH comment pattern for running the "module" (aka TAV) tests and the "benchmark" tests I believe. Before "jenkins run the module tests please" would work. Not it must be "^run the module tests".  Also perhaps https://github.com/elastic/apm-agent-nodejs/blob/6f67c8e2fac1d364acefc3efa9141315f05b454f/.ci/Jenkinsfile#L401 should be updated now as some of those regex match groups or now obsolete.

I am totally fine with changing the syntax for triggering builds/tests with GH comments. I'd appreciate the syntax being simplified so that *fewer* ways to request it are available. The newer "/test ..." style seems good because it is more obvious that it is syntax specifically for a bot/tool.

There are a few mentions of the old forms in dev docs (some of which will now be broken):
- https://github.com/elastic/apm-agent-nodejs/blob/master/DEVELOPMENT.md#how-to-trigger-a-benchmark-run-for-a-pr
- https://github.com/elastic/apm-agent-nodejs/blob/master/TESTING.md#jenkins

v1v:
Thanks @trentm for the heads up.

`Jenkins run the` is now deprecated, so it will be removed from the docs but people can still use it.

`run the module tests` matches the regex `(regex: '(?i).*(?:jenkins\\W+)?run\\W+(?:the\\W+)?module\\W+tests\\W+for\\W+(.+)') `, so it can be replaced to be more accurate. That's something Ill do it in addition to the `.md` files too.

PR #2347, opened 2021-09-23
types: add Span.ids and Transaction.ids to types

trentm:
Back in #1335 the "ids" property was added to Span and Transaction, but
it wasn't added to the TypeScript types.

/cc @mshustov 

### Checklist

- [x] Add tests

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -40,6 +40,8 @@ Notes:
 [float]
 ===== Bug fixes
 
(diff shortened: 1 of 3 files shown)


trentm:
BTW, this came up in discussion of https://github.com/elastic/kibana/issues/102699 partly because `apm.currentTraceIds` doesn't include "transaction.id" when there is a current span. That's a separate issue I'm following up on.

astorm (approved):

PR #2354, opened 2021-09-29
docs: add a hapi tracing example

trentm:
Also change GH Action tests to skip PRs that only touch examples/.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,7 @@ on:
     - '*.md'
     - '*.asciidoc'
     - 'docs/**'
+    - 'examples/**'
   pull_request:
     branches:
     - master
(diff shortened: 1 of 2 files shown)


trentm:
> Also change GH Action tests to skip PRs that only touch examples/.

It would be nice to do the same for Jenkinsfile tests:

```groovy
            // Skip all the stages except docs for PR's with asciidoc or md changes only
            env.ONLY_DOCS = isGitRegionMatch(patterns: [ '.*\\.(asciidoc|md)' ], shouldMatchAll: true)
```

but my naive/quick read of https://github.com/elastic/apm-pipeline-library/blob/master/vars/isGitRegionMatch.groovy makes me think adding multiple regexes to `patterns` will *not* do as intended. I don't want to get into that right now.

Issue #1944, opened 2021-01-14
[META 382] Zero-configuration support


AlexanderWert:
Implement zero-configuration support for the agent as described in: https://github.com/elastic/apm/issues/382

SPEC issue: https://github.com/elastic/apm/issues/401

trentm:
- [ ] handle the rare case for "service.name" discovery to not error out if the $CWD of the process is not inside the directory tree holding the package.json file. Instead we'd change it to fallback to some default "unknown nodejs service" string.

PR #2410, opened 2021-11-03
feat: zero-configuration support by always successfully inferring a serviceName

trentm:
If a bogus serviceName is explicitly configured, the agent will still
log.error and de-activate. However, if the agent is inferring a
serviceName, then with this change it will always succeed. This change
improves inference from a package.json by dealing with some cases where
a package.json "name" was an invalid serviceName. And then ultimately it
falls back to "nodejs_service" if a name cannot otherwise be inferred.

This changes Agent configuration somewhat. Before this change the agent
would `this._config()` twice: once at Agent creation (resulting in a
agent._conf that is incorrect if options are later passed to
agent.start()), and once at agent.start(). This commit changes to only
configure at agent.start(). A separate `config.configLogger()` handles
setting `agent.logger` to a valid early logger when the Agent is
created.  All Agent methods have been updated as necessary to guard
against usage before the agent is started.  One visible semantic change
is that `agent.getServiceName()` will now always return `undefined`
before the agent is started. Arguably the earlier behaviour was a
subtle bug, but the returned serviceName could be *wrong* if the agent
was later started with a given serviceName.

Closes: #1944

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


lib/agent.js
```diff
@@ -113,7 +109,7 @@ Agent.prototype.destroy = function () {
   // for tests that may use multiple Agent instances in a single test process.
   global[symbols.agentInitialized] = null
 
-  if (Error.stackTraceLimit === this._conf.stackTraceLimit) {
+  if (this._origStackTraceLimit && Error.stackTraceLimit !== this._origStackTraceLimit) {
```
trentm, line 112: REVIEW NOTE: This is a guard against agent.destroy() being called before agent.start()

lib/agent.js
```diff
@@ -127,7 +123,7 @@ Agent.prototype._getStats = function () {
   const stats = {
     frameCache: frameCacheStats
   }
-  if (typeof this._transport._getStats === 'function') {
+  if (this._transport && typeof this._transport._getStats === 'function') {
```
trentm, line 126: REVIEW NOTE: This is a guard against agent._getStats() being called before agent.start()

lib/agent.js
```diff
@@ -556,9 +558,9 @@ Agent.prototype.handleUncaughtExceptions = function (cb) {
   this._uncaughtExceptionListener = function (err) {
     agent.logger.debug({ err }, 'Elastic APM caught unhandled exception')
     // The stack trace of uncaught exceptions are normally written to STDERR.
-    // The `uncaughtException` listener inhibits this behavor, and it's
+    // The `uncaughtException` listener inhibits this behavior, and it's
     // therefore necessary to manually do this to not break expectations.
-    if (agent._conf.logUncaughtExceptions === true) {
+    if (agent._conf && agent._conf.logUncaughtExceptions === true) {
```
trentm, line 563: REVIEW NOTE: This is a guard against calling agent.handleUncaughtExceptions() and then capturing an uncaught exception before agent.start()

lib/config.js
```diff
@@ -18,13 +18,6 @@ const { isLambdaExecutionEnviornment } = require('./lambda')
 
 let confFile = loadConfigFile()
 
-let serviceName, serviceVersion
-try {
-  const { name, version } = readPkgUp.sync().packageJson
-  serviceName = name
-  serviceVersion = version
-} catch (err) {}
```
trentm, line 26: REVIEW NOTE: The equivalent of this was moved to `getPkg()` below.

lib/instrumentation/index.js
```diff
@@ -59,8 +59,7 @@ function Instrumentation (agent) {
   this._hook = null // this._hook is only exposed for testing purposes
   this._started = false
   this._runCtxMgr = null
-
-  this._log = agent.logger.child({ 'event.module': 'instrumentation' })
+  this._log = agent.logger
```
trentm, line 62: REVIEW NOTE: There was no value in the `event.module` field in the log records, so I dropped it rather than having to reproduce it in the `Instrumentation.prototype.start` change below.

lib/logging.js
```diff
@@ -163,8 +163,9 @@ function setLogLevel (logger, levelName) {
   const pinoLevel = PINO_LEVEL_FROM_LEVEL_NAME[levelName]
   if (!pinoLevel) {
     logger.warn('unknown log levelName "%s": cannot setLogLevel', levelName)
+  } else {
+    logger.level = pinoLevel
   }
-  logger.level = pinoLevel
```
trentm, line 167: REVIEW NOTE: This was an existing small bug that was exposed by testing the `configLogger()` addition.

test/agent.test.js
```diff
@@ -91,14 +91,10 @@ function deep (depth, n) {
 test('#getServiceName()', function (t) {
   const agent = new Agent()
 
-  // Before agent.start(), config will have already been loaded once, which
-  // typically means a `serviceName` determined from package.json.
+  // Before agent.start() the agent hasn't configured yet.
   t.ok(!agent.isStarted(), 'agent should not have been started yet')
-  t.strictEqual(agent.getServiceName(), packageJson.name)
-  t.strictEqual(agent.getServiceName(), agent._conf.serviceName)
+  t.strictEqual(agent.getServiceName(), undefined)
```
trentm, line 96: REVIEW NOTE: This is the semantic change that was mentioned in the PR description and CHANGELOG message.

test/config.test.js
```diff
@@ -538,14 +537,6 @@ test('should overwrite option property active by ELASTIC_APM_ACTIVE', function (
   t.end()
 })
 
-test('should default serviceName to package name', function (t) {
-  var agent = new Agent()
-  agent.start()
-  t.strictEqual(agent._conf.serviceName, 'elastic-apm-node')
-  agent.destroy()
-  t.end()
-})
```
trentm, line 547: REVIEW NOTE: This test is a duplicate of "serviceName/serviceVersion zero-conf: valid" below.

test/config.test.js
```diff
@@ -620,144 +618,105 @@ test('valid serviceName => active', function (t) {
   t.end()
 })
 
-test('serviceName defaults to package name', function (t) {
```
trentm, line 623: REVIEW NOTE: This suite of test cases was replaced by the 'serviceName/serviceVersion zero-conf: *' test cases below. Instead of having all this goop to create directory trees and packages in which to test, I just created the various test scenarios in "test/fixtures/pkg-zero-conf-*/" dirs.

test/filters.test.js
```diff
@@ -1,5 +1,11 @@
 'use strict'
 
+// `filterHttpHeaders` below indirectly (*very* indirectly) depends on there
+// being a configured agent, because "is-secret.js" reaches into `agent._conf`.
+require('..').start({
+  disableSend: true
+})
+
```
trentm, line 8: REVIEW NOTE: This file and a few test files below implicitly depended on a configured agent because this code reaches into the agent singleton to get its config: https://github.com/elastic/apm-agent-nodejs/blob/ba570ba2af5243fa99cc1a36421cc2eb5ea930c3/lib/redact-secrets/is-secret.js#L26   A "configured agent" now means a "started agent", hence my starting an agent for these test files.

trentm (commented):
Add review notes.

lib/agent.js
```diff
@@ -26,17 +26,12 @@ var version = require('../package').version
...
+  this.logger = config.configLogger()
 
-  this.logger = null
   this._conf = null
   this._httpClient = null
   this._uncaughtExceptionListener = null
-
-  // Early configuration to ensure `agent.logger` works before `.start()`
-  // is called.
-  this._config()
```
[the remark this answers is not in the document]
trentm: *I* am. ;)

My first justification is that this is an internal/undocumented property. We cannot, IMHO, *in general* have internal code changes be disallowed by possible/unknown external usage.

The main justification is that `agent._conf` between Agent creation and `agent.start(...)` is incomplete -- a stronger word is that this _conf is *wrong* and misleading. External code should not be making behaviour decisions based on a config value that will be different after `agent.start(...)`. It could lead to a more subtle bug. (The obvious exception here is the logLevel, which we need for internal logging before `agent.start()`. If a good use case for another exception comes up, I think having the specific use case would much better inform how we support it.)

The last justification is that I would be surprised if there is any meaningful usage of the Agent instance in customer code *before* `agent.start(...)` is called.

lib/config.js
```diff
@@ -409,6 +456,78 @@ function readEnv () {
...
 
+function validateServiceName (s) {
+  if (typeof s !== 'string') {
+    throw new Error('not a string')
+  } else if (!/^[a-zA-Z0-9 _-]+$/.test(s)) {
+    throw new Error('contains invalid characters (allowed: a-z, A-Z, 0-9, _, -, <space>)')
+  }
+}
+
+// getPkg() attempts to find a "package.json" file (looking up from the current
```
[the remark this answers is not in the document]
trentm: I think that would be a good thing to add. Or even prefer the main JS file's dir over process.cwd().

```
% cd /

% node ~/el/apm-agent-nodejs/whats-my-dir.js
/Users/trentm/el/apm-agent-nodejs/whats-my-dir.js

% cat ~/el/apm-agent-nodejs/whats-my-dir.js
console.log(require.main.filename)
```

I'll open a separate ticket for that.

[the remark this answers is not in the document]
trentm: https://github.com/elastic/apm-agent-nodejs/issues/2420

Issue #2417, opened 2021-11-04
Is old Lambda captureContext Necesary?


astorm:
The pre-extension lambda handling in the Node.js agent included a `captureContext` method.

https://github.com/elastic/apm-agent-nodejs/blob/ba570ba2af5243fa99cc1a36421cc2eb5ea930c3/lib/lambda.js#L9

These fields are no longer part of the Lambda spec for captured data: https://github.com/elastic/apm/blob/master/specs/agents/tracing-instrumentation-aws-lambda.md

If we can safely no longer capture this data, we should consider removing this block.

trentm:
> If we can safely no longer capture this data

What argument is there for keeping them?

trentm:
Here is an example lambda transaction showing all the `context.custom.lambda.*` that is added:

```
{
    "id": "cee87407f4b5004b",
    "trace_id": "7539b41a881c0b83fe6c1fbe38bc71cc",
    "name": "ObjectCreated:Put trentm-play-bukkit3",
    "type": "request",
    "subtype": null,
    "action": null,
    "duration": 539.998,
    "timestamp": 1636133780186082,
    "result": "success",
    "sampled": true,
    "context": {
        "user": {},
        "tags": {},
        "custom": {
            "lambda": {
                "functionName": "trentm-play-fn1",
                "functionVersion": "$LATEST",
                "invokedFunctionArn": "arn:aws:lambda:us-west-2:627286350134:function:trentm-play-fn1",
                "memoryLimitInMB": "128",
                "awsRequestId": "01d37a73-7464-4d75-a239-a4c6566c9d75",
                "logGroupName": "/aws/lambda/trentm-play-fn1",
                "logStreamName": "2021/11/05/[$LATEST]69e03e801dce4ee491d7c8a9751dc67a",
                "executionEnv": "AWS_Lambda_nodejs14.x",
                "region": "us-west-2",
                "input": {
                    "Records": [
                        {
                            "eventVersion": "2.1",
                            "eventSource": "aws:s3",
                            "awsRegion": "us-west-2",
                            "eventTime": "2021-11-05T17:36:17.726Z",
                            "eventName": "ObjectCreated:Put",
                            "userIdentity": {
                                "principalId": "AWS:AIDAZEDJODE3PMCHOJU7W"
                            },
                            "requestParameters": {
                                "sourceIPAddress": "50.68.122.122"
                            },
                            "responseElements": {
[...]

PR #2422, opened 2021-11-05
fix: remove unspec'd custom context for lambda transactions

trentm:
Closes: #2417

Skipping changelog entry, because if this is in the same release as the recent Lambda transaction data work, then it falls under the same changelog entry.

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -347,24 +347,6 @@ function setLambdaTransactionData (agent, trans, event, context, isColdStart) {
 function elasticApmAwsLambda (agent) {
   const log = agent.logger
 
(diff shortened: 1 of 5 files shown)


astorm (approved):
Look safe and sane to me. We have a spec for capturing specific data now so grabbing everything like this doesn't make sense anymore. The `captureContext` function was private to this module, so no explicit to implicit APIs were changed. If users _do_ want this data then [`setCustomContext`](https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-set-custom-context) is there for them.

PR #2426, opened 2021-11-08
docs: new lambda docs for Node.js agent

astorm:
Updates Lambda docs to undeprecate Lambda and provide information on installing with [the Lambda extension](https://github.com/elastic/apm-aws-lambda/)

### Checklist

- [x] Update documentation

### DIFF
--- a/docs/lambda.asciidoc
+++ b/docs/lambda.asciidoc
@@ -7,9 +7,7 @@ NOTE: For the best reading experience,
 please view this documentation at https://www.elastic.co/guide/en/apm/agent/nodejs/current/lambda.html[elastic.co]
 endif::[]
 
-=== Get started with Lambda (Deprecated)
-
-WARNING: *Lambda functions are no longer supported by the Elastic APM Node.js Agent.*
+=== Get started with Lambda (Beta)
 
 Getting Elastic APM set up for your lambda functions is easy,
 and there are various ways you can tweak it to fit your needs.
@@ -19,6 +17,20 @@ check out the <<api,API Reference>>.
 [float]
 [[lambda-installation]]
 ==== Installation
+Setting up your Lambda function is a two step process. First, you’ll
+need to install https://github.com/elastic/apm-aws-lambda/[the Elastic
+APM Lambda extension] into your Lambda environment. Elastic uses a
+Lambda extension to forward data on to APM Server in a way that does not
+interfere with the execution of your Lambda function. You can
+https://github.com/elastic/apm-aws-lambda/blob/main/apm-lambda-extension/docs/aws-lambda-extension.asciidoc[read
+more about installing and using this extension here].
+
+Once you have the extension installed in your environment, you’ll need
+to configure your Lambda handler to instrument your function.
+
+[float]
+[[lambda-instrumenting]]
+==== Instrumenting your Lambda Handler
 
 Add the `elastic-apm-node` module as a dependency to your application:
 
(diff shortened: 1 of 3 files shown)


docs/lambda.asciidoc
```diff
@@ -19,6 +17,20 @@ check out the <<api,API Reference>>.
 [float]
 [[lambda-installation]]
 ==== Installation
+Setting up your Lambda function is a two step process. First, you’ll
+need to install https://github.com/elastic/apm-aws-lambda/[the Elastic
```
trentm, line 21: Thoughts on pointing that directly to the extension subdir: https://github.com/elastic/apm-aws-lambda/tree/main/apm-lambda-extension ?

docs/lambda.asciidoc
```diff
@@ -60,6 +59,8 @@ exports.handler = apm.lambda(function handler (payload, context, callback) {
 
 The agent will now monitor the performance of your lambda function.
 
+*IMPORTANT*: During installation of the Lambda extension you'll have set a number of environmental variables for configuring the agent.  We recommend relying on these environmental variables -- while it's possible to set configuration values via the configuration object passed to `start`, doing so may interfere with the configuration required by the Lambda extension.
```
trentm, line 62: nit: I think "environment variable" is the more common terminology than "environmental variable" and is the term used in the current docs.  One data point: https://trends.google.com/trends/explore?geo=CA&q=%22environment%20variable%22,%22environmental%20variable%22

trentm, line 62: > doing so may interfere with the configuration required by the Lambda extension.

That is just for the `ELASTIC_APM_SERVER_URL` variable, correct? Or is there other "interfere"nce I'm not aware of. I think it would be helpful to be specific.  I agree to encourage configuring via envvars, but mainly because I think it leads to a nice developer experience if/when we use a layer for the agent and automatically instrument the handler rather than needing the explicit `apm.lambda()` call.

trentm (approved):
LGTM. I have a few nits/questions.

docs/lambda.asciidoc
bmorelli25: ++

docs/lambda.asciidoc
```diff
@@ -19,6 +17,20 @@ check out the <<api,API Reference>>.
 [float]
 [[lambda-installation]]
 ==== Installation
+Setting up your Lambda function is a two step process. First, you’ll
+need to install https://github.com/elastic/apm-aws-lambda/[the Elastic
+APM Lambda extension] into your Lambda environment. Elastic uses a
+Lambda extension to forward data on to APM Server in a way that does not
```
bmorelli25, line 23:
```suggestion
Lambda extension to forward data to APM Server in a way that does not
```

docs/lambda.asciidoc
bmorelli25: As-is, the link to the [repo](https://github.com/elastic/apm-aws-lambda/) doesn't provide much context. You also can't immediately see a docs directory or readme, so it doesn't seem very user-friendly. I think it would be a better user experience if we link to the documentation before linking to the repo itself. I also wonder if we should break installation of the AWS Lambda extension into its own step like Instrumenting your Lambda Handler is. For example, something like this:


```
[float]
[[lambda-installation]]
==== Installation

Setting up your Lambda function is a two-step process.

1. <<lambda-extension>>
2. <<lambda-instrumenting>>

First, https://github.com/elastic/apm-aws-lambda/blob/main/apm-lambda-extension/docs/aws-lambda-extension.asciidoc[install the Elastic
APM Lambda extension] into your Lambda environment. Elastic uses a Lambda
extension to forward data to APM Server in a way that does not interfere with the
execution of your Lambda function. 

[float]
[[lambda-extension]]
==== Install the Elastic APM Lambda extension

Elastic uses a Lambda extension to forward data to APM Server in a way that does not interfere with the
execution of your Lambda function.

See the https://github.com/elastic/apm-aws-lambda/blob/main/apm-lambda-extension/docs/aws-lambda-extension.asciidoc[installation documentation] to get started.

Once you have the extension installed in your environment, you can
configure your Lambda handler to instrument your function.

[float]
[[lambda-instrumenting]]
==== Instrumenting your Lambda Handler
```

astorm:
Committing on previous approval from @trentm 

Additionally -- I've made the environmental/environment switch, and restructured the intro to have an explicit "install the extension" section per suggestions from @bmorelli25 here 70f6791db5b65c56a6910c972c6fe5fa240e5a5f

Re: linking to the blank-looking repo, that a very valid point but also something that seems like a problem with `elastic/apm-aws-lambda` repo. I'd rather (for ease of maintenance) keep the number of deep links to a minimum, so I've opened an issue to address the landing page experience for `elastic/apm-aws-lambda` here: https://github.com/elastic/apm-aws-lambda/issues/74

Re: being more specific with _doing so may interfere with the configuration required by the Lambda extension._ -- I'd rather not get too specific here and risk overwhelming the reader _and_ having these specific-detail-that-might change in multiple places.  If they've worked their way through the installation doc they'll be exposed to those details.

Issue #2437, opened 2021-11-10
`apm.middleware.connect()` and 'finalhandler' instrumentation result in double-reporting errors


trentm:
Way back in the early days of this APM agent (about 10y ago) `.middleware.connect()` was added to support `apm.captureError()`'ing with Connect middleware:

```
var app = connect()
// app.use(...)
app.use(apm.middleware.connect())  // installs error middleware to `captureError()`
```

About three years ago in #629 support for [finalhandler](https://www.npmjs.com/package/finalhandler) was added to better handle error reporting for Express usage. Express uses finalhandler.  At the time an `errorReportedSymbol` symbol was added to the reported errors to coordinate between the 'express' and 'finalhandler' instrumentation to avoid double-reporting the errors.

What was missed is that Connect also uses finalhandler. The `errorReportedSymbol`-based coordination is not being done in "lib/middleware/connect.js". So if one uses `apm.middleware.connect()`, then server response errors will be double-reported.

I see no current value in `apm.middleware.connect()`. I think we should:

1. Deprecate it and drop it in the next major
2. Document that it is automatically handled by 'finalhandler' middleware.
3. Fix 'finalhandler' middleware to be disablable via `disableInstrumentations`. Currently it is not.

Low priority: This has been the state for 3y and there are no issues about it. Connect and finalhandler have not had releases for 2y+ I think.

trentm:
Here is a "play-connect.js" to demonstrate:

```js
'use strict'

const apm = require('./').start({
  serviceName: 'play-connect',
  centralConfig: false,
  captureExceptions: false,
  metricsInterval: '0s',
  cloudProvider: 'none',
  stackTraceLimit: 2,
  // This doesn't work:
  // disableInstrumentations: 'finalhandler'
})

const connect = require('connect')
const http = require('http')

const app = connect()
app.use(function handleReq (req, res, next) {
  switch (req.url) {
    case '/error':
      res.writeHead(500)
      res.end()
      next(new Error('responding with an Error'))
      break
    case '/throw':
      throw new Error('throwing an Error')
    default:
      res.end()
  }
})
app.use(apm.middleware.connect())

const server = http.createServer(app)
server.listen(function () {
  var port = server.address().port
  var baseUrl = 'http://localhost:' + port

  http.get(baseUrl + '/error', function () {
    http.get(baseUrl + '/throw', function () {
      apm.flush(function () {
        server.close()
      })
    })
  })
})
```

PR #2438, opened 2021-11-10
refactor: drop unused 'npm run test:cli'; add test for apm.middleware.connect()

trentm:
Let's drop test/script/cli.js. It is a little CLI, exposed as `npm run test:cli`,
that prompts interactively for which of these test types to run:

```js
    case 'standard': return standardTest()
    case 'http': return httpTest()
    case 'restify': return restifyTest()
    case 'connect': return connectTest()
    case 'express': return expressTest()
    case 'transaction': return transactionTest()
```

The select test is run, which generates some APM data (sent to the default
`serverUrl`).

- standard: calls `apm.captureError()` a few times
- http: exercises `http.createServer()` and `http.get()` instrumentation
- restify: exercises `restify.createServer()` and restify-clients' `client.get()`
- connect: exercises `connect` instrumentation
- express: exercises `express` instrumentation
- transaction: spends 65s generating transactions with some child spans, e.g.:
    ```
    transaction dc3e29 "foo"
    `- span fa63c8 "sig1"
      `- span 8da1e6 "sig2"
        `- span 46866e "sig5"
          `- span e4d3b6 "sig6"
        `- span 36309d "sig3"
          `- span b5f043 "sig4"
    ```

This appears to be the only place that `connect` and `restify-clients` are
exercised under "test/...". This exercising useless because no one runs this
(correct me if I'm wrong there) and the output isn't checked.

- connect: I wonder if this module is pretty much EOL already. The last release
  was `"3.7.0": "2019-05-18T00:58:45.162Z"`. The repo is getting what looks like
  automated commits. **We can add a small test/connect.test.js for this.** Adding
  TAV tests is overkill.
  NOTE: Adding this test turned up an issue: https://github.com/elastic/apm-agent-nodejs/issues/2437
- restify-clients: **I see no value in adding more test code for restify-clients.**
  We don't document support for restify-clients, we have never otherwise
  tested it. The restify-clients just use the Node.js core `http.request`
  under the hood.

### Checklist

- [x] Implement code
- [x] Add tests


trentm:
FWIW, here is the output of a run of each of the test/script/cli.js test types:

  <details>

    ```
% npm run test:cli

> elastic-apm-node@3.24.0 test:cli /Users/trentm/el/apm-agent-nodejs5
> node test/script/cli.js

? Service name foo
? Secret token
? APM Server URL http://localhost:8200
? Test suite standard
? Save answers? Yes
Saved config: /Users/trentm/.config/elastic-apm-node.json
Capturing error...
The error have been logged
Capturing message...
Something went wrong: write after end
The message have been logged
Capturing parameterized message...
Something went wrong: write after end
The parameterized message have been logged
Throwing exception...
The uncaught exception have been logged

    ```
  </details>

astorm (dismissed):

PR #2446, opened 2021-11-16
fix: offline-queued redis commands would be double-instrumented

trentm:
Ensure that we do not create a second span for offline-queued redis
commands (those commands that redis queues before the client is
"ready").

### Checklist

- [x] Implement code
- [x] Add tests


lib/instrumentation/modules/redis.js
```diff
@@ -33,16 +35,24 @@ module.exports = function (redis, agent, { version, enabled }) {
...
     })
+    wrappedCallback[isWrappedRedisCbSym] = true
+    return wrappedCallback
   }
 
   function wrapInternalSendCommand (original) {
     return function wrappedInternalSendCommand (commandObj) {
+      // Avoid re-wrapping internal_send_command called *again* in
+      // send_offline_queue for commands queued before "ready".
+      if (commandObj && commandObj.callback && commandObj.callback[isWrappedRedisCbSym]) {
```
astorm, line 52: Question: I see we've altered the behavior of the wrapped `internal_send_command` but not the older wrapped `send_command`.  If offline queuing not a thing in the older version of the client with `send_command`, or is 2.5.3 too old to worry about? Or some third thing.

astorm (approved):
Looks like a reasonable change 👍 approving with a question below. Also, just to restate what I see it looks like we're altering the behavior of the redis instrumentation so that when we're wrapping the callback to either the `send_command` method (used in versions up to 2.5.3) or `internal_send_command` method (used in versions greater than 2.5.3)  that we add a property to that callback that indicates it has been wrapped. 

Then, in our **wrapped** `internal_send_command` method we check for this flag, and if present, we skip out of generating the usual span.  We're basically using the "is wrapped" property of the callback as a signal that indicates "`internal_send_command` was already called once for this callback, no need to to generate the span again".

trentm:
> to either the send_command method (used in versions up to 2.5.3)

@astorm Good point. I hadn't checked the older version's behaviour. Perhaps I should? This *is* a version of the redis client that is from 2016, however.  The current behaviour doesn't cause any harm, but also won't fix the equivalent issue, if it occurs in the older version.

trentm:
```
% npm install redis@2.5.3
...
% node example-redis-offline-queue-spans.js
```

results in this trace:

```
    transaction 74b2dc "transBeforeRedis"
    `- span 0ffffe "INFO"
    `- span 15e8af "SET"
      `- span 658225 "GET"
    transaction 6c1f81 "transForClientCalls"
    `- span e16c1b "SET"
      `- span bc02bf "GET"
```

So, yah, the same issue happens in the old one.  I'll try for a fix ... though it seems wrong to spend too much time on something this old.

trentm:
And down the rabbit hole...

The handling for older redis client versions breaks calling the callback for calling forms like this: `client.set(['k', 'v', myCb])`.

astorm (approved):
> I'll try for a fix ... though it seems wrong to spend too much time on something this old.

_nod_ -- glad it was a relatively quick fix.  Approving. 

I'm going to add an item to our weekly agenda to discuss deprecation of older versions of of these libraries.

PR #2452, opened 2021-11-17
fix: no need to escape chars in serviceName for User-Agent

trentm:
When creating the User-Agent, there is no need to escape chars in
serviceName, because it is already limited to legal User-Agent chars.

(From some private chat on escaping for User-Agent: https://elastic.slack.com/archives/C9A6ANR5Y/p1637163729166800)

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -845,15 +845,19 @@ function loadServerCaCertFile (opts) {
 // The format of User-Agent is governed by https://datatracker.ietf.org/doc/html/rfc7231.
 //    User-Agent = product *( RWS ( product / comment ) )
 // We do not expect `$repoName` and `$version` to have surprise/invalid values.
-// However, `$serviceName` and `$serviceVersion` are provided by the user
-// and could have invalid characters. `comment` is defined by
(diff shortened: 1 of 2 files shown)


astorm (approved):
> because it [ed. `serviceName`] is already limited to legal User-Agent chars.

Confirmed, 👍 approving.   

For completeness -- the `serviceName` is [validated here](https://github.com/elastic/apm-agent-nodejs/blob/8390df45d32c95f47f02a8263401a1a37514d285/lib/config.js#L312), and an agent with an invalid service name will log an error and not start.  This means it never gets to the point where it sets a user agent.  

Also, the [validation regular expression](https://github.com/elastic/apm-agent-nodejs/blob/8390df45d32c95f47f02a8263401a1a37514d285/lib/config.js#L482) in use (`/^[a-zA-Z0-9 _-]+$/`) for `serviceName` is essentially equivalent.  It's my understanding and experience that the character ranges in javascript generally and node specifically _do not_ include characters like ä, å, etc. (much to the minor annoyance of the non-english speaking world).

Issue #2430, opened 2021-11-08
run-context for spans created by instrumentations should not bleed out to user code


trentm:
## the issues

Typically a package that makes some external call will have an API like this:

```js
var foo = require("foo")
var client = foo.createClient(/* ... */)

// Callback API
client.doAThing(/* opts */, function afterTheThingIsDone (err, result) {
  // ... do whatever next ...
})
// ... synchronous code after client.doAThing() ...

// And/or a Promise API
client.doAThing(/* opts */)
  .then(function afterTheThingIsDone (result) {
    // ... do whatever next ...
  })
// ... synchronous code after client.doAThing() ...

// Using the Promise API via await.
async function main() {
  const result = await client.doAThing(/* opts */)
  // ... do whatever next ...
}
main()
```

Typically, the APM agent's instrumentation of these types of packages will create a span (an *exit* span) for the `doAThing` external call. 

1. When the user's `... do whatever next ...` code is called, it is important that the "run context" does *not* include that automatically created span -- it should be the same run context when `client.doAThing()` was called (the "parent" context). If `... do whatever next ...` is bound to the create span context, even if that span has now ended, it *can* result in subsequently created spans being a child of that automatically created span.
2. Likewise, when the user's `... synchronous code after client.doAThing() ...` code is called, it is important that the run context does *not* include the automatically created span.

In the current redis instrumentation, for example, this code:

```
var client = redis.createClient('6379', process.env.REDIS_HOST)
agent.startTransaction('t0')
client.set('key1', 'val1', function (err, reply) {
  // ...
})
client.set('key2', 'val2')
```

results in the span for `SET key2` being a *child* of the span for `SET key1`. I.e. the redis instrumentation suffers from issue 2. IIRC the `@elastic/elasticsearch` instrumentation suffers from issues 1 and 2.

This is a problem for handling exit spans. https://github.com/elastic/apm/blob/master/specs/agents/tracing-spans.md#child-spans-of-exit-spans says:

> Exit spans MUST not have child spans that have a different type or subtype.

And exit span handling is a prerequisite for [span compression](https://github.com/elastic/apm-agent-nodejs/issues/2100) and [work for `span.context.destination.service.resource`](https://github.com/elastic/apm-agent-nodejs/issues/2103).

## checklist

[...]

PR #2444, opened 2021-11-16
fix: drop no-longer-necessary half-disabling of instrumentations

trentm:
When disableInstrumentations support was added in #353 it only
half-disabled these instrumentations, briefly mentioning "continuation
patches". I believe this was about user-land callback queue handling
that should no longer be necessary after run-context and #2430 work.

In #2498 it was discussed and decided that if listed in 
`disableInstrumentations` the agent should not touch the module at all.

### Checklist

- [x] Implement code: update from main, this is currently a bit out of date.
- [x] Test with TAV run.
- [x] Add CHANGELOG.asciidoc entry (an entry was added in the [3.26.0 release](https://github.com/elastic/apm-agent-nodejs/pull/2505/files)).

PR #2460, opened 2021-11-19
fix: rewrite ioredis instrumentation: run-context, add destination context, captureError

### LINKED ISSUE #2459: ioredis: offline-queued commands are double-instrumented
The ioredis client supports queuing commands to send to the redis server after the connection is "ready". This agent's current ioredis instrumentation will create a span *both* for the initial call, and the delayed internal call after "ready". If there is a currentTransaction when the ioredis.Redis client is created then that transaction will be in scope for those delayed internal calls and duplicate spans will be created.

(This is the equivalent of https://github.com/elastic/apm-agent-nodejs/pull/2446 but for the ioredis instrumentation.)

# repro

Running this:

```js
// example-ioredis-double-instrument.js
  serviceName: 'example-ioredis-double-instrument'
})
const Redis = require('ioredis')

const t0 = apm.startTransaction('t0')
const redis = new Redis()
redis.on('ready', function () {
  redis.set('foo', 'bar')
  redis.get('foo')
  redis.quit(function () {
    t0.end()
  })
})
```

Results in this trace:
(Ignore the parent/child relationships. That is being handled in a separate change.)

```
    transaction 2df0dd "t0"
    `- span 64d988 "INFO"
    `- span ee46c5 "SET"
      `- span 916945 "GET"
        `- span d6aea4 "QUIT"
```

If we comment out waiting for "ready":

```js
// example-ioredis-double-instrument.js
  serviceName: 'example-ioredis-double-instrument'
})
const Redis = require('ioredis')

const t0 = apm.startTransaction('t0')
const redis = new Redis()
// redis.on('ready', function () {
  redis.set('foo', 'bar')
  redis.get('foo')
  redis.quit(function () {
    t0.end()
  })
// })
```

Then:

```
    transaction c5b255 "t0"
    `- span 9e8a34 "SET"
      `- span b0425e "GET"
        `- span 08773d "QUIT"
    `- span ced6cb "INFO"
    `- span e771fa "SET"
      `- span ce91c2 "GET"
        `- span 870c5c "QUIT"
```

trentm:
- This fixes run-context handling for proper parent/child relationships.
  The run-context in code in the same tick and the optional callback are
  no longer changed.
- Add destination context, so ioredis usage shows redis on the Service Map.
- Capture an error and set the span outcome to "failure" if the command
  errors.
- No longer double-instrument queued ioredis commands.

Fixes: #2459
Refs: #2430

### Checklist

- [x] perhaps fix #2459
- [x] Implement code
- [x] Add tests


trentm:
This passed a full `TAV=ioredis ./node_modules/.bin/tav` on my machine.

lib/instrumentation/modules/ioredis.js
```diff
@@ -1,81 +1,69 @@
...
+  }
   if (!semver.satisfies(version, '>=2.0.0 <5.0.0')) {
     agent.logger.debug('ioredis version %s not supported - aborting...', version)
     return ioredis
   }
 
-  agent.logger.debug('shimming ioredis.Command.prototype.initPromise')
-  shimmer.wrap(ioredis.Command && ioredis.Command.prototype, 'initPromise', wrapInitPromise)
-
-  if (!enabled) return ioredis
```
trentm, line 18: REVIEW NOTE: This update of the ioredis instrumentation no longer wraps `Command.prototype.initPromise`; it still wraps `Redis.prototype.sendCommand`. The only utility of having wrapped `initPromise` before was something about "continuation patches" that may have been necessary in the past. This is why the previous instrumentation still wrapped `initPromise` even if `enabled=false` (from `disableInstrumentations` use). I believe this is no longer necessary after the run-context work. See https://github.com/elastic/apm-agent-nodejs/pull/2444 (a draft PR where I intend to remove this "half-disabling" of instrumentations) for some more discussion.

lib/instrumentation/modules/ioredis.js
```diff
@@ -1,81 +1,69 @@
...
-
   return ioredis
 
-  // wrap initPromise to allow us to get notified when the callback to a
-  // command is called. If we don't do this we will still get notified because
-  // we register a callback with command.promise.finally the
-  // wrappedSendCommand, but the finally call will not get fired until the tick
-  // after the command.callback have fired, so if the transaction is ended in
-  // the same tick as the call to command.callback, we'll lose the last span
-  // as it hasn't yet ended.
```
trentm, line 31: REVIEW NOTE: In ioredis, `sendCommand(cmd)` is called with a `Command` object that always has a `.promise` and can optionally have a user-supplied `.callback` function. When the command is complete: (a) the callback is called, if any, and (b) the promise is resolved or rejected. Because this is a Promise that resolve/reject is in the next tick.

A change this update makes is to only instrument the promise. The code before this would instrument both, and `span.end()` on the first one. This means that a slight behaviour change with this update is that when a user-supplied callback is called, then span is not yet ended. It will be ended in the next tick with the promise. Not that the run-context fixes in this change mean that the span is intentionally *not* the `currentSpan` in the user-callback, so the slight change on end time doesn't have any impact that I can see.

The comment above about "we'll lose the last span" is wrong: the Elastic APM system can handle a trace with a span ending after the transaction ends just fine. It looks like this:

Screen Shot 2021-11-19 at 12 38 51 PM

I like that the code is less complex: only needing to `command.promise.then()` to handle `span.end()`.

test/instrumentation/modules/ioredis.test.js
```diff
@@ -110,13 +109,16 @@ test('rejections_handled', function (t) {
...
+    t.equal(data.errors.length, 1, 'captured 1 error')
+    t.equal(data.errors[0].exception.type, 'ReplyError', 'exception.type')
+    t.equal(data.errors[0].transaction_id, data.transactions[0].id, 'error.transaction_id')
+
     setTimeout(function () {
       t.notOk(unhandledRejection)
       t.end()
     }, 0)
   })
-  agent.captureError = function (err) { throw err }
```
trentm, line 119: REVIEW NOTE: This and the `agent.captureError =` override below were removed because they conflict with testing now that the ioredis instrumentation is actually using `agent.captureError()`.

lib/instrumentation/modules/ioredis.js
astorm: > I believe this is no longer necessary after the run-context work.

Does remove the `initPromise` wrapping have any effect on the spans/span-tree if someone's using the `patch-async` /`asyncHooks:false`?  If so are you OK with changing that functionality and/or should we mention it somewhere?

astorm (commented):
Generally looks ok -- does what it says on the tin.  One question below before signing off.

lib/instrumentation/modules/ioredis.js
trentm: > Does remove the initPromise wrapping have any effect on the spans/span-tree if someone's using the patch-async /asyncHooks:false?

No, I don't believe it has any effect. The removed wrappedInitPromise didn't do any promise-y context propagation or binding. My quoting of "continuation patches" was from this single comment on the change that added the "enabled" flag to instrumentations: https://github.com/elastic/apm-agent-nodejs/pull/353#issuecomment-392212569   It provides no more details.

The test suite passes with `asyncHooks: false`. Also the parent/child output from `ELASTIC_APM_ASYNC_HOOKS=false node examples/trace-ioredis.js` is as expected:

```
    transaction 035eaa "t1"
    `- span 46aff5 "SET"
    `- span 4824ad "GET"
    `- span 77b874 "GET"
    `- span 0d4192 "MULTI"
    `- span 537be3 "SET"
    `- span 35b9c7 "GET"
    `- span 3a4fa9 "EXEC"
    `- span 6610bb "HSET"
    `- span 21c383 "GET"
      `- error 579426 "WRONGTYPE Operation against a key holding the wrong kind of value"
```

astorm: 👍

astorm (approved):
👍 approving.

PR #2462, opened 2021-11-23
fix: the error captured for an erroring redis span should be the span

trentm:
Before this the error.parent_id was set to the transaction.id, rather
than span.id.

### Checklist

- [x] Implement code
- [x] Add tests


astorm (approved):
approving 👍

PR #2463, opened 2021-11-23
fix: ensure run-context for mysql span does not spill into user code

trentm:
This fixes the 'mysql' instrumentation to not have the mysql span
context be active in user code. This ensures that user code cannot
create a child span of the mysql span, which would (a) be misleading
and (b) cause problems for coming exit span and compressed span work.

Refs: #2430

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/modules/mysql.js
+++ b/lib/instrumentation/modules/mysql.js
@@ -98,59 +98,61 @@ function wrapQueryable (connection, objType, agent) {
 
   function wrapQuery (original) {
     return function wrappedQuery (sql, values, cb) {
-      var span = agent.startSpan(null, 'db', 'mysql', 'query')
-      var id = span && span.transaction.id
+      agent.logger.debug('intercepted call to mysql %s.query', objType)
+
+      var span = ins.createSpan(null, 'db', 'mysql', 'query')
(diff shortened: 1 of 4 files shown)


trentm:
This passed a full `TAV=mysql ./node_modules/.bin/tav` run using node v12.22.6 on my machine.

lib/instrumentation/modules/mysql.js
```diff
@@ -98,59 +98,61 @@ function wrapQueryable (connection, objType, agent) {
 
   function wrapQuery (original) {
     return function wrappedQuery (sql, values, cb) {
-      var span = agent.startSpan(null, 'db', 'mysql', 'query')
-      var id = span && span.transaction.id
+      agent.logger.debug('intercepted call to mysql %s.query', objType)
+
+      var span = ins.createSpan(null, 'db', 'mysql', 'query')
+      if (!span) {
```
astorm, line 104: +1 for bailing out early if the span didn't start -- there doesn't appear to be any significant code path that follows in the function for an undefined span.

astorm (approved):
👍 does what is says, approving.

PR #2468, opened 2021-11-24
tests: fix benchmark config to not skip span.stacktrace collection

trentm:
When the `spanFramesMinDuration` config var was added, an unintended
side-effect is that it removed span.stacktrace collection from this
benchmark that was intended to measure that.

Refs: #2081

### Checklist

- [x] Implement code


trentm:
run benchmark tests

trentm:
The "transaction-and-span-overhead-realistic-size" benchmark value (the top green line) is back up closer to what is expected:
Screen Shot 2021-11-29 at 12 55 42 PM


Note that the benchmarked value heere isn't as high as it was before the accidental drop due to `spanFramesMinDuration`. Looking at the last year of data, it used to range between 2500 and 3000:

Screen Shot 2021-11-29 at 12 57 17 PM


Time will tell if the new value settles at something better than that. `spanFramesMinDuration` was added on 2021-06-03. Perf improvements in stacktrace collection was added soon after that in 2021-06-09. So I'm hoping this 2500-3000 drop to ~1400 is an indication of that performance win.

astorm (approved):

PR #2470, opened 2021-11-25
fix: ensure the span run-context for outgoing http requests does not spill into user code

trentm:
This fixes the 'http' and 'https' instrumentation for outgoing requests
to not have the 'http' span context be active in user code. This ensures
that user code cannot create a child span of the http span, which would
(a) be misleading and (b) cause problems for coming exit span and
compressed span work.

Refs: #2430

### Checklist

- [x] Implement code
- [x] Add tests


trentm:
The tests were failing on this (with some added `console.log('XXX ...')` to try to show what is going on):

```
% node  test/central-config-enabled.test.js
TAP version 13
# remote config enabled
XXX making proto.request in http-request
XXX making proto.request in http-request
XXX making proto.request in http-request
ok 1 mock apm-server got central config request
ok 2 agent conf for key transactionSampleRate was updated to expected value
ok 3 agent conf for key transactionMaxSpans was updated to expected value
ok 4 agent conf for key captureBody was updated to expected value
ok 5 agent conf for key transactionIgnoreUrls was updated to expected value
ok 6 agent conf for key logLevel was updated to expected value
# remote config enabled: receives comma delimited
XXX making proto.request in http-request
XXX instrumentation not yet started
/Users/trentm/el/apm-agent-nodejs6/lib/instrumentation/http-shared.js:172
      const spanRunContext = parentRunContext.enterSpan(span)
                                              ^

TypeError: Cannot read property 'enterSpan' of null
    at Object.request (/Users/trentm/el/apm-agent-nodejs6/lib/instrumentation/http-shared.js:172:47)
    at httpRequest (/Users/trentm/el/apm-agent-nodejs6/lib/http-request.js:117:21)
    at getMetadataGcp (/Users/trentm/el/apm-agent-nodejs6/lib/cloud-metadata/gcp.js:28:15)
    at /Users/trentm/el/apm-agent-nodejs6/lib/cloud-metadata/index.js:73:9
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
```

The issue is that we are accidentally instrumenting the outgoing http request attempts being done for cloud metadata fetching... *and* doing so before the Instrumentation has finished starting. This is because of this ordering in Agent#start:

```js
  this._transport = this._conf.transport(this._conf, this)

  this._instrumentation.start()
```

and `this._conf.transport()` creation (i.e. the creation of the elastic-apm-http-client) synchronously starts the cloud metadata fetching. Options to fix this are:

1. Sequencing changes in Agent.start() -- might have other impact.
2. Start the cloud metadata fetching asynchronously -- would be straightforward.
3. Tweak lib/http-request.js (as we already did for the elastic-apm-http-client in https://github.com/elastic/apm-nodejs-http-client/pull/161) to get early references to `http.request` and `https.request` so that the agent doesn't instrument itself. This kills two birds with one stone: we will also no longer get the following occasional abortive attempts to instrument the cloud-metadata fetching:

```
[2021-11-25T00:29:20.480Z] DEBUG (elastic-apm-node): intercepted call to http.request {"id":null}
[2021-11-25T00:29:20.481Z] DEBUG (elastic-apm-node): no active transaction found - cannot build new span
```

astorm (approved):
:+1: Approving.

I see here that we've taken steps to ensure the request methods used by the agent are uninstrumented functions.  

I also see that we're binding a request's callback to the run context that was active _when the request was made_.  

I see we're invoking the original request using a new run context created via the parent run context's `enterSpan` method.

Finally, in the response event handler, I see we're ensuring the response is bound to the parent's run context.

Issue #2443, opened 2021-11-16
Only 30 tests are reported when running in the CI


v1v:
**Describe the bug**

We have noticed that since https://github.com/elastic/apm-agent-nodejs/pull/2138 the main pipeline of the APM Agent Node.js is reporting only 30 tests to the CI, before of that change it was reporting about few thousands of tests.

![image](https://user-images.githubusercontent.com/2871786/141962586-0098dfd5-d9a2-4e3c-bb88-99abac593129.png)

**Expected behavior**

The CI reports the executed tests as expected, this will help with all the notifications, flaky tests analyser and other bits and pieces we have automated.

For instance, before the above changes then the number of tests reported in the CI can be seen below

![image](https://user-images.githubusercontent.com/2871786/141962916-ac939f19-8f2c-4c60-b34e-c3cd6bc37120.png)

https://github.com/elastic/apm-agent-nodejs/pull/2094#issuecomment-851672524

trentm:
I don't see how this could be associated with the changes from #2138.

@v1v Am I right that the Jenkins "Test Results" is a representation of the junit XML file created from https://github.com/elastic/apm-agent-nodejs/blob/bf760484ad80a2a26fb5f46a9795f2e5145b94c5/.ci/Jenkinsfile#L374 ?

Is there a way to have these `*-junit.xml` be build artifacts so we can look at them?

For what it is worth, I tried to mimic what `tap2Junit.groovy` is doing, using a "test-suite-output.tap" test result output file that I had kicking around in my dev area (https://gist.github.com/trentm/d0e56068ff614f07c1977d6cd9b09248) using this bash:

```bash
set -e
set -x
docker run --rm \
    -v $(pwd):/usr/src/app \
    -w /usr/src/app \
    node:12-alpine \
    sh -c 'export HOME=/tmp ; mkdir ~/.npm-global; npm config set prefix ~/.npm-global ; npm install tap-xunit -g ; for i in "*-output.tap" ; do echo "i is $i"; ls ${i}; cat ${i} | /tmp/.npm-global/bin/tap-xunit --package="Node.js" > ${i%.*}-junit.xml ; done'
```

I added `echo "i is $i"; ls ${i}; ` to try to see what was going on in there.
I don't know if I have the quoting correct, but this is the result I got:

```
% bash foo.sh
++ pwd
[...]

PR #2471, opened 2021-11-25
ci: support for tap specific nodejs implementation

v1v:
### What

Use the extended support for the TAP 2 JUnit transformation.

### Why

The existing TAP file is not one file per test file execution but an aggregation, `tap2xunit` NPM library does not support it. Therefore, the number of reported tests in the CI is ridicule small compared to the ones that were executed.

This is the temporary fix until the changes in the test engine in this project supports one file per TAP execution.

### Issues

Relates to https://github.com/elastic/apm-agent-nodejs/issues/2443
Requires https://github.com/elastic/apm-pipeline-library/pull/1414 

### Test

There are >200.000 executed UTs:

image

### Tasks

- [x] 38 JUnit files are empty :/  maybe some issues with the parser? Or maybe empty TAP executions 

Solved by moving from `tap-xunit` to `tap-junit`

<details><summary>Expand to view</summary>
<p>

```
# SKIP apollo-server-express@3.5.0 does not support node v8.6.0
running (cwd: ./): node --require /app/test/_promise_rejection.js test/instrumentation/modules/elasticsearch.test.js
```

produced an empty JUnit

image

image

</p>
</details>


.ci/Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env groovy
-@Library('apm@current') _
+@Library('apm@feature/tap-with-multiple-entries-for-nodejs-specific') _
```
v1v, line 2: This should be changed to `current`. Though, still pending to be merged the required PR in the apm-pipeline-library and a new release. 

In any case, I think this particular PR is ready to be reviewed.

mdelapenya (approved):

trentm:
https://github.com/elastic/apm-pipeline-library/pull/1414 seems a lot of complexity added to apm-pipeline-library just for the Node.js agent repo. This implements "solution option 2" from https://github.com/elastic/apm-agent-nodejs/issues/2443#issuecomment-971942426 right?

If I showed that "solution option 3" could be implemented fairly easily in apm-agent-nodejs.git, would that be acceptable (even preferable)?  If that seems reasonable to you, I'd like to take a quick stab at this today.  If I can't get this working, then let's take what you have here. Thanks for the work to code around the quirks in this repo.

(As a separate issue, I would also be happy to have this repo (apm-agent-nodejs.git) take on the responsibility of producing test output in xunit format -- i.e. for now taking on the tap-xunit dependency, so the separate Docker container and `tap2Junit` support from apm-pipeline-library.git is no longer necessary.)

trentm:
Here is my proof of concept: https://github.com/elastic/apm-agent-nodejs/pull/2478

It changes the Jenkins test run to write a separate ".tap" file to "test_output/" for each test file. Then it runs `tap-xunit` on each of those to convert to a "test_output/*.junit.xml" file. Then it uses the Jenkins `junit()` command directly -- i.e. it no longer needs `tap2Junit` from apm-pipeline-library.git.

Here is an example run:
https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2478/3/testReport/
It shows **86,897 tests**, which is more representative.

Does this seem reasonable to you to pursue instead of this PR?  If so, I can work through finishing it. The main current issues are:
- There are 22 failures for test file runs that don't actually generate TAP output, but instead just have empty output. In some cases that is because a particular module test isn't supported for a particular node version and it is just skipped.
- The TAP output is no longer in the test run log output. This should be straightforward to fix.

trentm:
> `tap-junit` is a more up-to-date library

Thanks! I'll try that one.

trentm:
> I wonder how I managed to produce 200.000 UTs

If you think it is important to dig into, I can look.  Do you have a jenkins build that showed 200k?   

My test run was only the "Test" section, and not the "Edge Test" which might have included more. I don't think "Edge Tests" would have pushed mine up to 200k, however.

v1v:
> If you think it is important to dig into, I can look. Do you have a jenkins build that showed 200k?

Maybe it's how the `xunit-tap` worked, and `junit-tap` might report even more?


This PR contains 200.000 UTs

See the build -> https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-nodejs%2Fapm-agent-nodejs-mbp%2FPR-2471/detail/PR-2471/1/tests

v1v:
As discussed recently, we agreed to use a different approach, so I'll close this now

PR #2472, opened 2021-11-25
fix: ensure correct run-context for instrumented aws-sdk client usage

trentm:
This fixes aws-sdk (S3, SQS, SNS) instrumentation to ensure the
automatically created AWS spans do not spill into user code. This also
ensures that any captured APM error for a failed AWS command is a child
of the AWS span.

Refs: #2430

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/.ci/jobs/apm-agent-nodejs-mbp.yml
+++ b/.ci/jobs/apm-agent-nodejs-mbp.yml
@@ -25,7 +25,7 @@
             ignore-tags-newer-than: -1
         - regular-branches: true
         - change-request:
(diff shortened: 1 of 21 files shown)


astorm (approved):
👍 Approving -- also, re:
> This also
ensures that any captured APM error for a failed AWS command is a child
of the AWS span.

Just to say this a bit more loudly, with the callback bound correct bound, the `_setOutcomeFromErrorCapture` code is no longer needed captureError will now set this value on the correct span.

Issue #2493, opened 2021-12-02
'mssql' instrumentation improvements


trentm:
My understanding is that a Node.js dev would use https://www.npmjs.com/package/mssql as a client for SQL Server. We instrument tedious (https://www.npmjs.com/package/tedious) which `mssql` uses by default and is "actively maintained and production ready. Platform independent, runs everywhere Node.js runs. Officially supported by Microsoft."

Some potential improvements:
- [ ] Test with `mssql` module usage. It has await, promises, callback, streaming styles (https://github.com/tediousjs/node-mssql#examples). Does our wrapping of the underlying tedious suffice?
- [ ] There is an optional binary driver other than `tedious` -- https://github.com/tediousjs/node-mssql#microsoft--contributors-node-v8-driver-for-nodejs-for-sql-server -- that can be used via `require('mssql/msnodesqlv8')`. Is it important to instrument usage via this driver? If so, do we instrument `'mssql/msnodesqlv8'` specifically to not conflict with `tedious` instrumentation.

Issue #2103, opened 2021-06-09
[META 448] Auto-infer destination.service.resource and adapt public API


elastic-apm-tech:
As described in https://github.com/elastic/apm/issues/448, the agent needs to apply the following changes to comply with the current SPEC for destination fields:

- [ ] Infer destination.service.resource based on other fields
- [ ] Add a public API to set destination.service.resource
- [ ] As destination.service.name and destination.service.type are not used anywhere in the product, I propose to deprecate and remove them also in the public API (if already added there)
- [ ] Add a public API to start an exit span. This auto-infers the destination.service.resource based on the subtype if not set otherwise. Avoids creating nested exit spans and makes the service map work with those custom spans by default, without having to set any destination fields.

trentm:
> Add a public API to set destination.service.resource

Go has a `func (c *SpanContext) SetDestinationService(service DestinationServiceSpanContext) {`.

Java has a `Span setDestinationService(String)`

Python checks on serialization of a span if `nested_key(self.context, "destination", "service", "resource")` is already set.

PR #2494, opened 2021-12-02
fix: ensure correct run-context for "tedious" instrumentation

trentm:
Refs: #2430

### Checklist

- [x] Implement code
- [x] Add tests


lib/instrumentation/modules/tedious.js
```diff
@@ -69,7 +69,7 @@ module.exports = function (tedious, agent, { version, enabled }) {
         if (!request.parametersByName) {
           return super.makeRequest(...arguments)
         }
-        const span = agent.startSpan(null, 'db', 'mssql', 'query')
+        const span = ins.createSpan(null, 'db', 'mssql', 'query')
```
astorm, line 72: Just noting out loud that, unlike some of the other instrumentations, it looks like we're now creating this span but never binding it to a particular run context.  I'm presuming this is intentional.

test/instrumentation/modules/tedious.test.js
```diff
@@ -1,11 +1,7 @@
 'use strict'
 
-// Don't test on travis, there is no mssql service
-if (process.env.TRAVIS) process.exit()
```
astorm, line 4: Last mention of travis in this repo's active code, leaving only a few stray comments. [test/instrumentation/modules/restify/basic.test.js](https://github.com/elastic/apm-agent-nodejs/blob/cf2b0c59bfde974ff5ad2403f9fc86f9b3a502fe/test/instrumentation/modules/restify/basic.test.js#L53)

astorm (approved):
👍

lib/instrumentation/modules/tedious.js
trentm: Yes. 
- The tedious API's user callback was already wrapped in a `ins.bindFunction` which by default picks up the current run context, and
- currently the wrapped callback doesn't handle captureError or setting span.outcome, so there is no *current* need to have that run in the *span's* run-context as has been done for other instrumentations. I've open https://github.com/elastic/apm-agent-nodejs/issues/2493 to improve the MSSQL instrumentation here.

One minor thing is that the `super.makeRequest(...arguments)` is *not* currently wrapped in `ins.withRunContext` to make it run in the span's run context. That *could* be done, but given that for #2103 this will be marked an exit span and nothing runs in the span context, there can't be any children. In other words, I'm not aware of any harm in not wrapping that `super.makeRequest` in the span's run context, rather than in its parent (the current run context).

test/instrumentation/modules/tedious.test.js
trentm: Cool

Issue #2489, opened 2021-12-01
Memory leak in version > 3.20.0


kurigashvili-godaddy:
**Describe the bug**

We have upgraded to 3.24.0 all our team services (10) and all of them started running out of memory.

All of them are running in K8S. It would take between 30 min to 3 hours of a pod run out of memory and restart.

We downgraded to 3.20.0 and the issue disappeared. 

We were not seeing issue in a staging environment so I believe the issue can be noticed under higher loads.

**To Reproduce**

Steps to reproduce the behavior: 
Run http service with high load and observe that memory slowly increases until eventually node process dies with "JavaScript heap out of memory" error.

**Expected behavior**

Do not run our of memory like version 3.20.0

**Environment (please complete the following information)**

- OS: [Linux]
- Node.js version: 12.22.1
- APM Server version: 7.15.2
- Agent version: 3.24.0

**How are you starting the agent? (please tick one of the boxes)**

- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**

- Agent config options 
  <details>
    <summary>Click to expand</summary>

    ```
    replace this line with your agent config options
    ```
  </details>
- `package.json` dependencies:
  <details>
    <summary>Click to expand</summary>

    ```
    replace this line with your dependencies section from package.json
    ```
  </details>

astorm:
Thanks for reporting @kurigashvili-godaddy, that's definitely not the intended behavior.

Usually the fastest path to resolution on these sorts of issues is an isolated bit of code that reproduces the problem.  Do you happen to have an isolated reproduction of this problem (GitHub repository, Dockerfile)?  If not, could you share what these services are doing?  i.e. what web framework you're using, what libraries you're using to query databases/datastores, any third party promise libraries you might be using, etc. 

Basically any information that would help us reproduce the problem so we can diagnose what's going on.

kurigashvili-godaddy:
I will try to create  an isolated reproduction of this problem.
The simplest service that had the problem:
crud service based on express 4.17.1 talking to dynamo db (aws-sdk 2.983.0).

kurigashvili-godaddy:
* All of the services are using DynamoDB
* I will see if I can run v3.21.0 and v3.22.0. This issue only happened in prod so it is a little tricky to be testing in prod.

[...]

PR #2528, opened 2022-01-12
perf: reduce memory retained by async tasks that outlive their transaction/span

trentm:
With the RunContext work, the "current transaction or span" is managed
by tracking a RunContext object for relevant Node async tasks. This
RunContext holds a reference to the bound transaction and, if any,
spans. Those references are not freed until the *async task* ends.

In versions of the agent before v3.24.0, references to the current APM
transaction and spans would be released when the *transaction* ended.
The newer RunContext tracking can keep APM Transaction and Span
references longer. This can have two visible impacts:

1. If user code creates async tasks in the context of a transaction/span,
   and the task outlives the transaction/span, then such an app under
   sustained load can use more memory: the extra memory of those ended
   transactions and spans that are still referenced until the async task
   ends. In a memory-constrained container with an OOM killer, this can 
   look exactly like a memory leak. An example is code that calls
   `setTimeout` or creates a `Promise` in an HTTP request handler, but
   responds to the HTTP request before the timeout or promise completes.
2. Any leak of async tasks (e.g. Promises) in a user's application can
   be exacerbated by the APM agent.

This change mitigates the impact by removing references in *ended*
Transactions and Spans to minimize their memory footprint. For developer
debugging, the internal `runContextFromAsyncIdSize` stat has been added
to `agent._getStats()`; it can indicate if there are elevated
RunContexts in memory.

Refs: #2489

### Checklist

- [x] Implement code
- [x] Add tests: test full span/trans API on *ended* spans and transactions

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -126,6 +126,9 @@ Agent.prototype._getStats = function () {
   const stats = {
     frameCache: frameCacheStats
   }
+  if (this._instrumentation._runCtxMgr && this._instrumentation._runCtxMgr._runContextFromAsyncId) {
+    stats.runContextFromAsyncIdSize = this._instrumentation._runCtxMgr._runContextFromAsyncId.size
+  }
   if (this._transport && typeof this._transport._getStats === 'function') {
     stats.apmclient = this._transport._getStats()
   }
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -437,6 +437,12 @@ Agent.prototype.captureError = function (err, opts, cb) {
     : Date.now() * 1000)
   const trans = this._instrumentation.currTransaction()
   const traceContext = (span || trans || {})._context
+  const req = (opts.request instanceof IncomingMessage
+    ? opts.request
+    : trans && trans.req)
+  const res = (opts.response instanceof ServerResponse
+    ? opts.response
+    : trans && trans.res)
 
   // As an added feature, for *some* cases, we capture a stacktrace at the point
   // this `captureError` was called. This is added to `error.log.stacktrace`.
@@ -466,12 +472,6 @@ Agent.prototype.captureError = function (err, opts, cb) {
   //    etc.).
   setImmediate(() => {
     // Gather `error.context.*`.
-    const req = (opts.request instanceof IncomingMessage
-      ? opts.request
-      : trans && trans.req)
-    const res = (opts.response instanceof ServerResponse
-      ? opts.response
-      : trans && trans.res)
     const errorContext = {
       user: Object.assign(
         {},
--- a/lib/instrumentation/generic-span.js
+++ b/lib/instrumentation/generic-span.js
@@ -23,6 +23,7 @@ function GenericSpan (agent, ...args) {
 
   this.timestamp = this._timer.start
   this.ended = false
+  this._duration = null // Duration in milliseconds. Set on `.end()`.
 
   this.outcome = constants.OUTCOME_UNKNOWN
 
@@ -82,6 +83,16 @@ Object.defineProperty(GenericSpan.prototype, 'traceparent', {
     return this._context.toString()
   }
 })
+
+// The duration of the span, in milliseconds.
+GenericSpan.prototype.duration = function () {
+  if (!this.ended) {
+    this._agent.logger.debug('tried to call duration() on un-ended transaction/span %o', { id: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type })
+    return null
+  }
+
+  return this._duration
+}
 
 // The 'stringify' option is for backward compatibility and will likely be
 // removed in the next major version.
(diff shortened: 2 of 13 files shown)


trentm:
run module tests for ALL

lib/agent.js
```diff
@@ -126,6 +126,9 @@ Agent.prototype._getStats = function () {
   const stats = {
     frameCache: frameCacheStats
   }
+  if (this._instrumentation._runCtxMgr && this._instrumentation._runCtxMgr._runContextFromAsyncId) {
+    stats.runContextFromAsyncIdSize = this._instrumentation._runCtxMgr._runContextFromAsyncId.size
+  }
```
trentm, line 131: REVIEW NOTE: This internal state might be helpful in manually debugging with users that are seeing high memory usage. If this value is and stays very high, then it can indicate async tasks that outlive current transactions and possibly indicate async task leaks in the application code.

lib/agent.js
```diff
@@ -434,6 +437,12 @@ Agent.prototype.captureError = function (err, opts, cb) {
     : Date.now() * 1000)
   const trans = this._instrumentation.currTransaction()
   const traceContext = (span || trans || {})._context
+  const req = (opts.request instanceof IncomingMessage
+    ? opts.request
+    : trans && trans.req)
+  const res = (opts.response instanceof ServerResponse
+    ? opts.response
+    : trans && trans.res)
```
trentm, line 446: REVIEW NOTE: The `trans.req` and `trans.res` need to be grabbed in the sync part of `captureError` because a possible `trans.end()` in the next tick will remove those references.  One of the existing test cases covers this case.

lib/instrumentation/generic-span.js
```diff
@@ -83,16 +83,6 @@ Object.defineProperty(GenericSpan.prototype, 'traceparent', {
...
 
-// duration is returned as an int representing the time in milliseconds
-GenericSpan.prototype.duration = function () {
-  if (!this.ended) {
-    this._agent.logger.debug('tried to call duration() on un-ended transaction/span %o', { id: this.id, parent: this.parentId, trace: this.traceId, name: this.name, type: this.type })
-    return null
-  }
-
-  return this._timer.duration
-}
```
trentm, line 94: REVIEW NOTE: Having this `.duration()` be a lazy fetch on `this._timer` means that we could not remove the `this._timer` reference on `Transaction#end()` and `Span#end()` to free up some memory. I instead changed this to be a `duration` value on the Transaction and Span objects.  The `.duration()` is API is not documented. However, Alan, if you feel this is a breaking change, I can change this implementation to still have a `.duration()` method. Let me know.

lib/instrumentation/span.js
```diff
@@ -191,6 +206,39 @@ Span.prototype._encode = function (cb) {
 
   if (!this.ended) return cb(new Error('cannot encode un-ended span'))
 
+  const payload = {
```
trentm, line 209: REVIEW NOTE: Moved `const payload = ...` to the synchronous part of this method to gather all these fields before some refs (e.g. to `this._timer`, which is used for `self.traceId` here) are removed in `Span#end()`.

lib/instrumentation/transaction.js
```diff
@@ -262,11 +267,32 @@ Transaction.prototype.end = function (result, endTime) {
...
   this._agent._instrumentation.addEndedTransaction(this)
   this._agent.logger.debug('ended transaction %o', { trans: this.id, parent: this.parentId, trace: this.traceId, type: this.type, result: this.result, name: this.name })
+
+  // Reduce this transaction's memory usage by dropping references except to
+  // fields required to support `interface Transaction`.
+  // Transaction fields:
+  this._customName = this.name // Short-circuit the `name` getter.
+  this._defaultName = ''
+  this.req = null
+  this.res = null
```
trentm, line 283: REVIEW NOTE: These were the big wins. These refs to the HTTP request and response objects were most of the memory issue.

trentm (commented):
Some review notes.

astorm (approved):
Looks OK from here -- it preserves existing functionality (other than `.duration()` -- see extended comment below) and if it clears up one our memory pressure scenarios then that's a Good Thing™.

lib/instrumentation/generic-span.js
```diff
@@ -83,16 +83,6 @@ Object.defineProperty(GenericSpan.prototype, 'traceparent', {
   }
 })
 
-// duration is returned as an int representing the time in milliseconds
-GenericSpan.prototype.duration = function () {
```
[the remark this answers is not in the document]
trentm: From separate discussion, I've restored the `.duration()` method so there is no possible breakage in usage of that undocumented API.  Its implementation uses `this._duration` now to not need a ref to `this._timer` after the span/trans has ended.

astorm (approved):

PR #2529, opened 2022-01-13
docs: refactor troubleshooting docs

trentm:
Refactor the troubleshooting doc (https://www.elastic.co/guide/en/apm/agent/nodejs/current/troubleshooting.html) somewhat inspired by the ordering/organization of the Java agent's troubleshooting doc (https://www.elastic.co/guide/en/apm/agent/java/current/trouble-shooting.html).

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -20,99 +20,24 @@ so that we can analyze the problem.
 Logs should include everything from when the application starts up until the first request executes.
 See <<debug-mode>> for more information.
 
-* <<no-data-sent>>
-* <<missing-performance-metrics>>
-* <<debug-mode>>
-* <<disable-agent>>
 
 [float]
-[[no-data-sent]]
-=== No data is sent to the APM Server
+[[use-latest-agent]]
+=== Updating to latest agent version
 
-If neither errors nor performance metrics are being sent to the APM Server,
-it's a good idea to first check your log file and look for output just as the app starts.
+The Elastic Node.js APM Agent is updated frequently and releases are not
+strongly tied to other components in the Elastic Stack.  Therefore, trying to
+update to the most recently released agent version is often the recommended
+first troubleshooting step.
 
-[float]
-[[error-messages]]
-==== Explanation of error messages
+See <<upgrading,upgrading documentation>> for more details.
 
-[float]
-[[message-agent-inactive]]
-===== Agent Inactive
-
-[source,text]
-----
-Elastic APM agent is inactive due to configuration
-----
-
-You've most likely set the `active` option or the `ELASTIC_APM_ACTIVE` environment variable to `false`.
-Ensure that it's set to `true` and try again.
-
-See the API documentation about <<active,`active`>> for details.
-
-[float]
-[[message-missing-service-name]]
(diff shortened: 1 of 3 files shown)


trentm:
The page's table of contents is a little bit nicer with the "problems" section more obvious under a "Common problems" header:
Screen Shot 2022-01-13 at 2 07 13 PM

trentm:
A lot of the churn in this change is from moving the "Debug mode" section up from the bottom to near the top (similar to what is in the Java troubleshooting docs).

docs/troubleshooting.asciidoc
```diff
@@ -20,122 +20,105 @@ so that we can analyze the problem.
...
-[[no-data-sent]]
-=== No data is sent to the APM Server
+[[use-latest-agent]]
+=== Updating to latest agent version
 
-If neither errors nor performance metrics are being sent to the APM Server,
-it's a good idea to first check your log file and look for output just as the app starts.
+The Elastic Node.js APM Agent is updated frequently and releases are not
+strongly tied to other components in the Elastic Stack.  Therefore, trying to
+update to the most recently released agent version is often the recommended
```
bmorelli25, line 30:
```suggestion
strongly tied to other components in the Elastic Stack.  Therefore,
updating to the most recently released agent version is often the recommended
```

bmorelli25 (dismissed):
This looks great! We recently updated the **On this Page** (OTP) widget on the right-hand side of the page to show level three headings—I like the idea of removing the manual jump list in favor of the automated OTP.

One small nit about "trying to update". Is there a reason a user wouldn't be able to just update? Either way, looks good.

docs/troubleshooting.asciidoc
trentm: > Is there a reason a user wouldn't be able to just update?

Probably not, no. :)  I had actually copied this language from https://www.elastic.co/guide/en/apm/agent/java/current/trouble-shooting.html#trouble-shooting-use-latest-agent

Eventually, when we have a new major version of the Node.js APM agent that drops support for some older versions of node, then the user might not be able to use the latest APM agent version. But that's a stretch. I agree the "trying to" doesn't need to be in there. Thanks.

trentm: Damn. Missed opportunity to commit your suggestion with the commit message: "do or do not, there is no try"

trentm:
> We recently updated the On this Page (OTP) widget on the right-hand side of the page to show level three headings

Ah. That perhaps explains why in a local build I didn't see the 3rd-level, but on www.elastic.co/guide I *did*.

bmorelli25 (approved):
LGTM. That's curious that you can't see the OTP L3 headings locally. I wonder if it's just a case of needing to update your local clone of elastic/docs? I'll need to test that locally as I'm the one that added that feature 😨

trentm:
> That's curious that you can't see the OTP L3 headings locally.

@bmorelli25  False alarm. I just rebuilt and ran the docs locally and it looks good:

Screen Shot 2022-01-19 at 9 21 10 AM

I *had* updated my local clone of elastic/docs, but I think I had not done a full/clean rebuild of the local docs. Not sure.

Issue #2455, opened 2021-11-18
[META 545] Drop unsampled transactions when connected to APM Server 8.0+


AlexanderWert:
[![issue details](https://img.shields.io/endpoint?label=meta-issue:&url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F545)](https://github.com/elastic/apm/issues/545)

[![issue details](https://img.shields.io/endpoint?label=spec-issue:&url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F552)](https://github.com/elastic/apm/issues/552)

Drop unsampled transactions when the agent is connected to an APM Server of version 8.0 or higher.

* * *

- [ ] Does APM server back to the [supported](https://www.elastic.co/guide/en/apm/guide/7.16/agent-server-compatibility.html) version 6.6 have this "version" in the JSON response body of "GET /"?

trentm:
FYI: The java agent PR: https://github.com/elastic/apm-agent-java/pull/2329/files
If I read that correctly, it doesn't block processing on having gotten a first version value from the APM server... instead it defaults to saying "yes, this is possibly an older APM server that should recent unsampled transactions" up until it has a value.

https://github.com/elastic/apm-agent-java/blob/8e15859b583fab19c12aeef25be2fc8e11227d62/apm-agent-core/src/main/java/co/elastic/apm/agent/report/ApmServerClient.java might be useful for comparison.

trentm:
The Python agent seems to do possibly the opposite:
- It will "GET /" from APM server lazily when the first batch of events is ready to send to APM Server (setting `client.server_version`). If there is some error the `client.server_version` is left `None`.
- Subsequent checks whether the server is `>= 8.0` will default to passing if `server_version` couldn't be set, or hasn't yet been set.


Q:
- [ ] Does APM server back to the [supported](https://www.elastic.co/guide/en/apm/guide/7.16/agent-server-compatibility.html) version 6.6 have this "version" in the JSON response body of "GET /"?

trentm:
Spec ref: https://github.com/elastic/apm/blob/main/specs/agents/tracing-sampling.md#non-sampled-transactions

PR #2546, opened 2022-01-24
feat: Drop unsampled transactions when connected to APM Server 8.0+

trentm:
This updates to an elastic-apm-http-client that calls the APM Server
Information API on creation to get the APM Server version.

Closes: #2455

### Checklist

- [x] Update to `"elastic-apm-http-client": "^10.4.0"` when https://github.com/elastic/apm-nodejs-http-client/pull/176 is merged and published.
- [x] Implement code
- [x] Add tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "url": "https://github.com/elastic/apm-agent-nodejs/issues"
   },
   "homepage": "https://github.com/elastic/apm-agent-nodejs",
+  "XXX elastic-apm-http-client": "^10.4.0",
   "dependencies": {
     "@elastic/ecs-pino-format": "^1.2.0",
     "after-all-results": "^2.0.0",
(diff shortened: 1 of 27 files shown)


package.json
```diff
@@ -86,7 +87,7 @@
     "basic-auth": "^2.0.1",
     "cookie": "^0.4.0",
     "core-util-is": "^1.0.2",
-    "elastic-apm-http-client": "^10.3.0",
+    "elastic-apm-http-client": "elastic/apm-nodejs-http-client#trentm/issue2455-drop-unsampled-trans",
```
astorm: Just the obvious note update this before merging.

astorm (approved):
- I see that when we create the base configuration object for the HTTP transport that it has a new top level property named apmServerVersion, and that its default value comes from the agent's configuration
- I see that when ending a transaction we check both the sampled value and the new `supportsKeepingUnsampledTransaction` method on the HTTP client -- if sampled is false-ish (meaning we did not collect spans for this transaction) AND we're not using an older version of APM server (per supportsKeepingUnsampledTransaction), we exit the method before the transaction can be encoded and sent.
- One quirk of this implementation is that end-users will be able to override the value of `apmServerVersion` by providing a configuration value to the `start` method.  They could say they're on APM Server version 0.4.2, and the client would take them at their word.  This seems harmless (so long as we don't document `apmServerVersion` as something folks can use) but also seems at least worth mentioning out loud for googlers of the future. 

approving.

trentm:
Thanks for the reviews!

PR #2550, opened 2022-01-27
fix: ensure correct run context for '@elastic/elasticsearch' instrumentation

trentm:
This is a re-write of the @elastic/elasticsearch instrumentation that
stops using the ES client observability events
    https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/observability.html
and switches to patching Transport#request() instead. This is necessary
to be able to bind that `request()` invocation to the RunContext for the
Span we've created; without using `apm.startSpan(...)`, which bleeds the
RunContext out to user code. (Patching Transport#request() is what the
legacy 'elasticsearch' instrumentation is also doing.)

Refs: #2430

### Checklist

- [x] Implement code
- [x] Add tests
- [x] ensure TAV tests pass

### DIFF
--- a/.tav.yml
+++ b/.tav.yml
@@ -367,7 +367,7 @@ elasticsearch:
   commands: node test/instrumentation/modules/@elastic/elasticsearch.test.js
 '@elastic/elasticsearch-canary':
   name: '@elastic/elasticsearch-canary'
-  versions: '^8.0.0-canary.35'
+  versions: '>=8.0.0-canary.35 || >=8.1.0-canary.2'
   node: '>=12'
   commands: node test/instrumentation/modules/@elastic/elasticsearch-canary.test.js
 
--- a/.tav.yml
+++ b/.tav.yml
@@ -367,7 +367,7 @@ elasticsearch:
   commands: node test/instrumentation/modules/@elastic/elasticsearch.test.js
 '@elastic/elasticsearch-canary':
   name: '@elastic/elasticsearch-canary'
-  versions: '>=8.0.0-canary.35 || >=8.1.0-canary.2'
+  versions: '>=8.0.0-canary.36 || >=8.1.0-canary.2'
   node: '>=12'
   commands: node test/instrumentation/modules/@elastic/elasticsearch-canary.test.js
 
--- a/lib/instrumentation/elasticsearch-shared.js
+++ b/lib/instrumentation/elasticsearch-shared.js
@@ -28,20 +28,13 @@ exports.pathIsAQuery = pathIsAQuery
 //    {"query":{"query_string":{"query":"pants"}}}
 //
 // `path`, `query`, and `body` can all be null or undefined.
-exports.setElasticsearchDbContext = function (span, path, query, body, isLegacy) {
+exports.setElasticsearchDbContext = function (span, path, query, body) {
   if (path && pathIsAQuery.test(path)) {
-    // From @elastic/elasticsearch: A read of Transport.js suggests query and
-    // body will always be serialized strings, however the documented
-    // `TransportRequestParams` (`ConnectionRequestParams` in version 8)
-    // allows for non-strings, so we will be defensive.
-    //
-    // From legacy elasticsearch: query will be an object and body will be an
-    // object, or an array of objects, e.g. for bulk endpoints.
     const parts = []
     if (query) {
       if (typeof (query) === 'string') {
         parts.push(query)
-      } else if (isLegacy && typeof (query) === 'object') {
+      } else if (typeof (query) === 'object') {
         const encodedQuery = querystring.encode(query)
         if (encodedQuery) {
           parts.push(encodedQuery)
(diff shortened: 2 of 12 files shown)


trentm:
run module tests for @elastic/elasticsearch,@elastic/elasticsearch-canary,elasticsearch

trentm:
run module tests for @elastic/elasticsearch,@elastic/elasticsearch-canary,elasticsearch

.tav.yml
```diff
@@ -367,7 +367,7 @@ elasticsearch:
   commands: node test/instrumentation/modules/@elastic/elasticsearch.test.js
 '@elastic/elasticsearch-canary':
   name: '@elastic/elasticsearch-canary'
-  versions: '^8.0.0-canary.35'
+  versions: '>=8.0.0-canary.36 || >=8.1.0-canary.2'
```
trentm, line 370: REVIEW NOTE: Kibana's HEAD is using `8.1.0-canary.2` now. I had to drop at least `8.0.0-canary.35` because our tests depend on https://github.com/elastic/elasticsearch-js/commit/76659b604f97993da84a5e3d12d9df7c3e0cbd52 which special-cased handling of the "sort" param to client commands such that it is always passed as a query param to ES.

lib/instrumentation/elasticsearch-shared.js
```diff
@@ -51,12 +44,18 @@ exports.setElasticsearchDbContext = function (span, path, query, body, isLegacy)
...
+        // https://github.com/elastic/elastic-transport-js/blob/069172506d1fcd544b23747d8c2d497bab053038/src/Transport.ts#L614-L618
+      } else if (Array.isArray(body)) {
+        try {
+          parts.push(body.map(JSON.stringify).join('\n') + '\n') // ndjson
+        } catch {}
+      } else if (typeof (body) === 'object') {
+        try {
           parts.push(JSON.stringify(body))
-        }
+        } catch {}
```
trentm, line 58: REVIEW NOTE: try/catch around these in case there is a surprise circular ref or something. I'll be opening a separate ticket to improve our instrumentation patching so that we can capture the body and querystring *as serialized by the client* instead of having to do it (again) in the agent.

test/instrumentation/modules/@elastic/elasticsearch.test.js
```diff
@@ -542,17 +541,11 @@ ${body.map(JSON.stringify).join('\n')}
 
           const err = data.errors
             .filter((e) => e.exception.type === 'RequestAbortedError')[0]
-          if (semver.satisfies(esVersion, '7.14.x')) {
-            // https://github.com/elastic/elasticsearch-js/issues/1517 was fixed
```
trentm, line 546: REVIEW NOTE: This issue was only relevant when using the diagnostics events from the ES client. Given that the instrumentation re-write is no longer using those events, we no longer need to have the tests workaround the issue.

test/instrumentation/modules/@elastic/elasticsearch.test.js
```diff
@@ -686,8 +673,8 @@ function checkSpanOutcomesFailures (t) {
   return function (data) {
     data.spans.sort((a, b) => { return a.timestamp < b.timestamp ? -1 : 1 })
     if (semver.gte(esVersion, '7.14.0') && semver.satisfies(esVersion, '7.x')) {
-      // Remove leading ES span and HTTP span from product check.
-      data.spans = data.spans.slice(2)
+      // Remove the product check spans for subsequent assertions.
+      data.spans = data.spans.slice(0, 1).concat(data.spans.slice(3))
```
trentm, line 677: REVIEW NOTE: The order of product-check spans has changed due to the change in the implementation of the instrumentation. We create a Span on `Transport#request()` now, rather than later when the "request" diagnostic event is sent. That means that the first Span is the one we want to keep.

1. ES span for the actual client command (want to keep this one)
2. ES span for the product check
3. HTTP span for the product check
4. HTTP span for the actual client command  (want to keep this one)

test/instrumentation/modules/elasticsearch.test.js
```diff
@@ -125,7 +125,7 @@ if (semver.satisfies(pkg.version, '>= 13')) {
       }
     ]
 
-    var statement = body.map(JSON.stringify).join('\n')
+    var statement = body.map(JSON.stringify).join('\n') + '\n'
```
trentm, line 128: REVIEW NOTE: This is for the slight change in `setElasticsearchDbContext()`'s handling of the `Array.isArray(body)` case to match how this is serialized by the `@elastic/elasticsearch` client.

trentm:
run module tests for @elastic/elasticsearch,@elastic/elasticsearch-canary,elasticsearch

trentm:
run module tests for @elastic/elasticsearch,@elastic/elasticsearch-canary,elasticsearch

trentm:
run module tests for @elastic/elasticsearch,@elastic/elasticsearch-canary,elasticsearch

astorm (approved):

Issue #2554, opened 2022-01-31
Override sampled state from traceparent header


spalger:
**Is your feature request related to a problem? Please describe.**
I'm running a service using Google Cloud Run. Distributed tracing is a feature of cloud run if you use their tooling, and they automatically set a `traceparent` header for each request that comes into the service with a 0.1 trace sampling rate applied. APM sees this traceparent header and ignores the `transactionSampleRate` config I have provided and only samples transactions that Google Cloud Run deems "sampled". I'd really like to be able to control this behavior so that I can store APM transactions regardless of what Google Cloud Run deems noteworthy.

**Describe the solution you'd like**
I'd like a way to override the "sample" state that come from the `traceparent` header.

**Describe alternatives you've considered**
Wrapping the `server.emit()` method to delete the traceparent header before APM gets a chance to read it, requesting that Google Cloud Run allow customizing of the sample rate (which their docs explicitly state is not configurable)

astorm:
great work tracking this one down @spalger, and thanks for reporting.  

Possibly of interest -- this isn't the first time a vendor's header has stomped on what we want to do -- if you're interested there's further reading [over here](https://github.com/elastic/apm/issues/286)

trentm:
@spalger FYI: https://github.com/elastic/apm-agent-nodejs/pull/2555#issuecomment-1027208321

PR #2555, opened 2022-02-01
feat: add traceContinuationStrategy=restart_always config option

trentm:
This adds a `traceContinutationStrategy` configuration option to allow
some control over how the APM Agent uses incoming trace-context headers
for context propagation. The default 'continue_always' value results in
the current behaviour of using 'traceparent' and 'tracestate' per the
W3C spec. A value of 'restart_always' will result in incoming
'traceparent' and 'tracestate' headers being *ignored*. This can be
useful for a service that is receiving problematic tracing headers from
an upstream service that is unmonitored (without root transactions in
ES, the APM UI has limitations) or otherwise out of the user's control
(sampling rate may not be what the user wants).

This is currently *experimental* while a design for
`trace_continuation_strategy` is being discussed at
https://github.com/elastic/apm/issues/286. This currently implements a
*subset* of the proposed design there -- `restart_external` is not yet
implemented.

Closes: #2554
Refs: https://github.com/elastic/apm/issues/286

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation
- [ ] Update for agreed-upon APM Agents spec (per https://github.com/elastic/apm/issues/286) 

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1124,3 +1124,28 @@ require('elastic-apm-node').start({
   ]
 })
 ----
+
+
+[[trace-continuation-strategy]]
+==== `traceContinuationStrategy`
+* *Type:* String
+* *Default:* `'continue_always'`
+* *Env:* `ELASTIC_APM_TRACE_CONTINUATION_STRATEGY`
+
+Warning: This configuration option is **experimental** and may change in
+non-major versions of the agent.
+
+This option allows some control on how the APM agent handles W3C trace-context
+headers on incoming requests. By default, the `traceparent` and `tracestate`
+headers are used per W3C spec for distributed tracing. However, in certain cases
+(typically where an Elastic-monitored service is receiving requests with
+`traceparent` headers from *unmonitored* services) it can be helpful to *not*
+use the incoming `traceparent` header.
+
+Valid values are:
+
+- `'continue_always'` - The current and default behavior. Every incoming
+  `traceparent` header determines the sampling decision and the trace-id.
+- `'restart_always'` - Always ignores the `traceparent` header of incoming
+  requests. A new trace-id will be generated and the sampling decision will be
+  made based on <<transaction-sample-rate,`transactionSampleRate`>>.
--- a/index.d.ts
+++ b/index.d.ts
@@ -280,6 +280,7 @@ declare namespace apm {
     sourceLinesSpanLibraryFrames?: number;
     spanFramesMinDuration?: string;
     stackTraceLimit?: number;
+    traceContinuationStrategy?: TraceContinuationStrategy;
     transactionIgnoreUrls?: Array<string>;
     transactionMaxSpans?: number;
     transactionSampleRate?: number;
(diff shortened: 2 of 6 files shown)


trentm:
To try this out, put this in your app's package.json:

```json
...
  "dependencies": {
    "elastic-apm-node": "elastic/apm-agent-nodejs#trentm/issue2554-override-incoming-traceparent",
...
```

and then configure the APM Agent via one of:

```js
const apm = require('elastic-apm-node').start({
  // ...
  traceContinuationStrategy: 'restart_always'
})
```

or:

```shell
ELASTIC_APM_TRACE_CONTINUTATION_STRATEGY=restart_always node myapp.js
```

spalger:
So, correct me if I'm wrong but with these changes I would have to choose between RUM + Cloud Run traceparents or neither, right? I'm don't have plans to implement RUM at this time but it's an interesting side-effect that this would prevent users from implementing RUM.

Do you think there's a chance that, if it comes up in the future, another config options could be added to enable more advanced `traceparent` filtering so that RUM could be used without accepting every arbitrary `traceparent` header?

trentm:
The current proposal for the `trace_continuation_strategy` config option at https://github.com/elastic/apm/issues/286 includes a third value `restart_external` that will use incoming traceparent/tracestate headers *if* `tracestate` includes the "es" vendor key that we add.  So, assuming that the RUM agent *is* adding that "es" vendor key to tracestate, then when the Node.js Agent adds support for `traceContinuationStrategy=restart_external`, then I think this would work as you wanted:

- `traceparent` headers from Google Cloud Run would be ignored
- `traceparent` and `tracestate` from the RUM agent would be used (the `transactionSampleRate` config value in the RUM agent would apply)


Side note: I'm slightly curious if Google Cloud Run always honours the "recording" flag on incoming `traceparent` headers, or if it further applies its own "0.1 requests per second for each container instance" limit.

trentm:
> assuming that the RUM agent _is_ adding that "es" vendor key to tracestate

My read is that it *is* doing so:
https://github.com/elastic/apm-agent-rum-js/blob/68cdd0938e6a0501d1d2fd2d08775681eaf1383e/packages/rum-core/src/common/utils.js#L120-L132

spalger:
That's great to hear, also confirming that this is working for ci-stats, thanks!

image

trentm:
@spalger Things are moving, albeit slowly. The cross-APM-agent spec for the `trace_continuation_strategy` config var (https://github.com/elastic/apm/pull/600) was approved on Monday (2d ago) and will go in early next week.  That means that the Node.js APM agent can now settle on the config name (hasn't changed) and its values (the string values have changed a bit from what this PR is using).

The target for this work (along with the related Span Links work) will generally be targeted for 8.4, but I should be able to add support for this config var sooner now that that spec is settled. I.e. it is just prioritization now.

spalger:
Thanks for the update!

spalger:
Closed by https://github.com/elastic/apm-agent-nodejs/pull/2692

trentm:
@spalger Thanks. I was going to let you know about the new functionality on "main" now, but you beat me to it. :) We don't yet have a released version with this new `traceContinuationStrategy`.  Also note that the supported values for the config var **have changed from what this branch used**. The `_always` was dropped.

Let me know if/when you are no longer using the [trentm/issue2554-override-incoming-traceparent](https://github.com/elastic/apm-agent-nodejs/tree/trentm/issue2554-override-incoming-traceparent) branch anywhere, and I will delete it.

spalger:
Yeah, you can go ahead and delete it. I'm going to switch to main for now and hopefully the latest version as soon as a new one is released.

trentm:
There is an elastic-apm-node@3.34.0 release with this now. I'm deleting the old PR branch now.

PR #2566, opened 2022-02-08
perf: fix benchmark 003 to actually have span.stacktrace

trentm:
This benchmark is named 'transaction-and-span-with-stack-trace'
but since inception incorrectly has span stacktrace collection
*disabled*.


astorm (approved):
I see we're explicitly setting a value of -1ms for the `spanFramesMinDuration` value rather than rely on the default behavior.  Per our docs [a negative value means always record stack traces](https://www.elastic.co/guide/en/apm/agent/nodejs/master/configuration.html#span-frames-min-duration).

Seems like the right thing to do and since we're the main consumer of these benchmarks I don't think we'll hear from anyone on the change in behavior. 

Approving.

PR #2573, opened 2022-02-11
feat: lambda config and transaction & metadata collection changes

trentm:
Closes: #2531

### Changes

- [x] API Gateway: `transaction.name` update
    - [x] Respect setting for `use_path_as_transaction_name`
- [x] API Gateway: `transaction.context.service.origin.name = event.requestContext.domainName`
- [x] Respect `service_name` and `service_version` agent config options
- [x] `s/metadata.service.id/transaction.faas.id/`
- [x] default `cloudProvider` config to `none` when in a AWS Lambda environment
- [x] default `centralConfig` to `false` when in a AWS Lambda environment

### Checklist

- [x] Implement code
- [x] Add tests


lib/lambda.js
```diff
@@ -5,19 +5,19 @@ const shimmer = require('./instrumentation/shimmer')
 
 // https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#deriving-cold-starts
 let isFirstRun = true
+let faasId // Set on first invocation.
```
astorm, line 8: nit: Small nit on the naming -- we've got a module global created here (`faasId`) that shares a name with some function parameters in the same module.  There's a school of thought that says this makes it harder to know which `faasId` you're setting/using and that outer scope variables shouldn't share names with inner scope variables.

astorm (dismissed):
👍 does the thing -- one optional style nit below but it looks good for a merge as is to me. Approving.

lib/lambda.js
trentm: Thanks. Great point. Shadowing vars is a smell. I'll update this.

I'll take a look at https://eslint.org/docs/rules/no-shadow as well. However, I wonder if adding this might add a whole slew of lint warnings I don't want to handle in this PR.

trentm: An attempt at using the eslint no-shadow rule.  
tl;dr: It *does* catch this shadowing, but `✖ 398 problems (398 errors, 0 warnings)`. :) So I'll skip this for now.

https://github.com/elastic/apm-agent-nodejs/pull/2579 is a draft PoC of switching from `standard` to `eslint` so that we can easily add/try other eslint rules. For example, with that PR I can add this change to try the 'no-shadow' rule:

```diff
% gd
diff --git a/package.json b/package.json
index 4ddb8e38..7c058530 100644
--- a/package.json
+++ b/package.json
@@ -196,7 +196,9 @@
   },
   "eslintConfig": {
     "extends": "standard",
-    "rules": {}
+    "rules": {
+      "no-shadow": ["error", { "allow": [] }]
+    }
   },
   "eslintIgnore": [
     "/test/sourcemaps/fixtures/lib",
```

Most of the "398 errors" above are in the "test/..." area. It gets down to "179 errors" if we allow shadowing of "t" in the test files. And "lib/..." has just "23 errors".  I think it would be reasonable/useful to consider adding this rule. I'll add a separate issue for it.

Issue #2574, opened 2022-02-14
Lambda: Layer ARN for Deploying an Agent


astorm:
If users want to use Elastic's AWS Lambda Extension with the Node.js Agent, they're currently required to manually include the agent source along side their application/service source and `require` it in their handler function.

We should build a Lambda Layer that includes an agent build and uses [a wrapper script](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-modify.html) to ensure the agent's loaded via the `-r` option.

This will be a two stage process.  

First, we'll need to create a build script that will package a released agent into an AWS layer, publish it to a region, and produce an ARN.  After users configure their function with this ARN they will have access to the Node.js agent via their service _and_ a wrapper script will be in place to ensure that the agent's preloaded with the `-r` option.

Second, once this script it built, it will need to be included in our agent release process.  This may mean one additional manual step, or may mean jenkins automation.  At the end of this step we'll have the ability to publish this layer to _all_ regions and produce a table of ARNs that users may choose from.

trentm:
- We should (maybe "must") ensure that the right thing happens if a user uses this layer *and* also ships the apm agent in their function layer and manually `require('elastic-apm-node')`s for manual instrumentation or just because that's what they had laying around.
- I think we should consider adding a package-lock.json or similar to get closer to repeatable builds. That is separate work and shouldn't block this issue, though.

astorm:
> We should (maybe "must") ensure that the right thing happens if a user uses this layer and also ships the apm agent in their function layer and manually require('elastic-apm-node')s for manual instrumentation or just because that's what they had laying around.

Good to call this out.

The layer works because the Amazon Node.js Lambda Runtime automatically sets the `NODE_OPTIONS` env variable to something like this

```
"NODE_PATH": "/opt/nodejs/node14/node_modules:/opt/nodejs/node_modules:/var/runtime/node_modules:/var/runtime:/var/task",
```

So in the case of someone having added the agent to their deployed Lambda source _and_ having added the layer, Node.js will prefer the version of the agent that's been added to the deployed lambda source (`/var/runtime/node_modules` above) over the layer's source (`/opt/nodejs/node_modules` above)

Talking naively, the right thing to do here seems to be to honor the Lambda path settings. While this could definitely get weird if there's two sources that are different versions in place, it could be equally weird to change these path settings and alter the overall behavior of the runtime. WDYT?

trentm:
[...]

Issue #2543, opened 2022-01-24
Lambda: Layer ARN for Installing Agent and Automatic Handler Wrapping


astorm:
If users want to use Elastic's AWS Lambda Extension with the Node.js Agent, they're currently required to manually instrument their handler function.

We should ensure that the agent _automatically_ patches the function indicated by the [`_HANLDER`](https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html) env variable when we're running in a Lambda environment.   

We should also take steps to ensure a handler is not double wrapped (allowing existing functions that manually wrap the handler to continue working).

astorm:
Renamed issue to reflect scope of work, moved ARN scope of work to new issue: https://github.com/elastic/apm-agent-nodejs/issues/2574

PR #2577, opened 2022-02-15
feat: auto wrapper for lambda handler functions

astorm:
Implements https://github.com/elastic/apm-agent-nodejs/issues/2543

This PR adds code to ensure that a user's handler function will be automatically patched/instrumented when using the AWS Lambda Node runtimes. Usage covered in the docs.  Main lambda docs will be updated when https://github.com/elastic/apm-agent-nodejs/issues/2574 lands.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


astorm:
jenkins run the tests please

lib/instrumentation/index.js
```diff
@@ -164,6 +167,13 @@ Instrumentation.prototype.start = function () {
     }
   }
 
+  const lambdaHandlerInfo = getLambdaHandler(process.env)
+  if (lambdaHandlerInfo) {
+    this.addPatch(
+      lambdaHandlerInfo.filePath,
+      require('./modules/_lambda-handler')
+    )
+  }
```
trentm, line 176: Was there a particular reason to do this in `.start()` rather than in the constructor where `addPatch` is called on all the static `MODULES`?

lib/instrumentation/index.js
```diff
@@ -164,6 +167,13 @@ Instrumentation.prototype.start = function () {
     }
   }
 
+  const lambdaHandlerInfo = getLambdaHandler(process.env)
+  if (lambdaHandlerInfo) {
+    this.addPatch(
+      lambdaHandlerInfo.filePath,
+      require('./modules/_lambda-handler')
```
trentm, line 174:
```suggestion
    (...args) => {
      return require(`./modules/_lambda-handler.js`)(...args)
    }
```

This *could* use the lazy-requiring technique used in the constructor, which would avoid loading _lambda-handler.js in the (admittedly rare) case where (a) lambda handler info can be loaded, but (b) the Lambda Runtime never gets to loading the user's function file, e.g. for handler name errors or some other runtime error.

lib/instrumentation/index.js
```diff
@@ -164,6 +167,13 @@ Instrumentation.prototype.start = function () {
     }
   }
 
+  const lambdaHandlerInfo = getLambdaHandler(process.env)
```
trentm, line 170: It would be good to save this `lambdaHandlerInfo` on the Instrumentation object to not have to call `getLambdaHandler()` multiple times below.

lib/instrumentation/index.js
```diff
@@ -216,7 +226,8 @@ Instrumentation.prototype._startHook = function () {
     var enabled = self._agent._conf.instrument && !disabled.has(name)
     var pkg, version
 
-    if (basedir) {
+    const lambdaHandlerInfo = getLambdaHandler(process.env)
+    if (!lambdaHandlerInfo && basedir) {
```
trentm, line 230: This logic feels wrong. Doesn't this change the version for, e.g., 'express'? I.e. for *all* non-core modules.

lib/instrumentation/index.js
```diff
@@ -234,7 +245,14 @@ Instrumentation.prototype._startHook = function () {
...
 Instrumentation.prototype._patchModule = function (exports, name, version, enabled) {
   this._agent.logger.debug('shimming %s@%s module', name, version)
-  var patches = this._patches.get(name)
+  const lambdaHandlerInfo = getLambdaHandler(process.env)
+  let patches
+  if (!lambdaHandlerInfo) {
+    patches = this._patches.get(name)
+  } else if (name === lambdaHandlerInfo.module) {
+    patches = this._patches.get(lambdaHandlerInfo.filePath)
+  }
```
trentm, line 254: Doesn't this break all patching of other modules in a lambda env? `patches` will be undefined for them.

lib/lambda.js
```diff
@@ -453,10 +453,30 @@ function isLambdaExecutionEnvironment () {
   return !!process.env.AWS_LAMBDA_FUNCTION_NAME
 }
 
+function getLambdaHandler (env) {
+  if (!isLambdaExecutionEnvironment() || !env._HANDLER || !env.LAMBDA_TASK_ROOT) {
+    return
+  }
+
+  const [handlerModule, handlerFunction, extra] = env._HANDLER.split('.')
```
trentm, line 461: Lambda allows "myModules.some.nestedpathto.myHandler". https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/c31c41ffe5f2f03ae9e8589b96f3b005e2bb8a4a/src/utils/UserFunction.ts#L39-L58

lib/instrumentation/modules/_lambda-handler.js
```diff
@@ -0,0 +1,11 @@
+'use strict'
+const { getLambdaHandler } = require('../../lambda')
+module.exports = function (module, agent, { version, enabled }) {
```
trentm, line 3: It would be nice to support `enabled` so a user could put "lambda" in `disableInstrumentations`.

trentm (commented):
First pass. I haven't had a chance to try it out yet.

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,90 @@
...
+    t.end()
+  })
+
+  suite.end()
+  // t.end()
+})
+
+tape.test('integration test', function (t) {
+  if (process.platform === 'win32') {
+    t.pass('skipping for windows')
```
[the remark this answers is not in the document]
astorm: Tests were failing on windows -- probably due to the path handling.  Lambda VMs are linux only so it didn't seem worth tracking down.

lib/instrumentation/index.js
astorm: Right you are -- looks like this inadvertently just checks for the existence of the handler instead of checking for both the existence of the handler AND whether the instrumentation currently being handled is the lambda instrumentation.  Test added to make sure the existing instrumentation are applied when lambda's detected and the correct (one hopes) logic's been added here: https://github.com/elastic/apm-agent-nodejs/pull/2577/commits/2e38f10fc9c4c4e6777af5fe6a74129cd411cd78#diff-413f1f1acaa8b1d97f8cae0884c443a777ef6eb822092e6483b4fe59a4936221R108

lib/instrumentation/index.js
astorm: Right you are -- looks like this inadvertently just checks for the existence of the handler instead of checking for both the existence of the handler AND whether the instrumentation currently being handled is the lambda instrumentation.  Test added to make sure the existing instrumentation are applied when lambda's detected and the correct (one hopes) logic's been added here: https://github.com/elastic/apm-agent-nodejs/pull/2577/commits/2e38f10fc9c4c4e6777af5fe6a74129cd411cd78#diff-413f1f1acaa8b1d97f8cae0884c443a777ef6eb822092e6483b4fe59a4936221R108

lib/instrumentation/modules/_lambda-handler.js
astorm: It was quick so I added the enabled check in cdbaa1d696e8974f7d67b9c58e59981139c8feab no harm in having it there regardless of which direction we head. As for supporting using the identifier "lambda" in `disableInstrumentations` -- I'm not sure I understand what you're proposing. 

Are you suggesting we special case `lambda` as an identifier for `disableInstrumentations` and use it to determine what to pass to this handler function?  If so I'm open to that but there's a real module named lambda: https://www.npmjs.com/package/lambda -- doubtful we'd ever instrument it but still feels a little weird to use its identifier.  WDYT?

If that's not what you're suggesting -- then what are you suggesting? ;)

lib/lambda.js
```diff
@@ -453,10 +453,30 @@ function isLambdaExecutionEnvironment () {
   return !!process.env.AWS_LAMBDA_FUNCTION_NAME
 }
 
+function getLambdaHandler (env) {
```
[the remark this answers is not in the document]
astorm: sure thing -- dd43c765572b879aa65693e73c5ba7450aff2444

lib/instrumentation/index.js
astorm: The task at hand (wrapping a module dynamically, and wrapping a module that's required in as a full file path outside of the application's node_modules folder) felt like more of a special case that belonged with applying the other special cases from configuration.

lib/instrumentation/index.js
astorm: @trentm I'm not quite sure I understand what you're proposing here. Could you show me some pseudo code of what you'd like to see in the constructor?

lib/instrumentation/index.js
astorm: We could, but I don't like the idea of the Instrumentation object duplicating state that's available from the environment.  I also don't like the idea that we'll have an instrumentation object that will behave differently when that state's set vs. when it's not set.  

@trentm What's the goal/benefit of removing those function calls and replacing it with that data set on the instrumentation object?

lib/instrumentation/modules/_lambda-handler.js
trentm: > Are you suggesting we special case `lambda` as an identifier for `disableInstrumentations` and use it to determine what to pass to this handler function?

That *had* been what I was suggesting, but I hadn't thought it through. I agree with you that this would be weird. I take back that "it would be nice to" suggestion.

lib/instrumentation/index.js
trentm: Okay. I was asking in case I was missing some reason this depended on being run after agent.start(...) had been called.

lib/instrumentation/index.js
trentm: The constructor already uses a lazy-requiring technique:

```js
    this.addPatch(mod, (...args) => {
      return require(`./modules/${pathName}.js`)(...args)
    })
```

That was initially added in #957 to avoid `require`ing files in "lib/instrumentation/modules/" unless they were required. The same thing could also be done in the `this.addPatch(...)` call that you've added here so that ".../modules/_lambda-handler.js" is only `require`d when it is actually necessary.

No need to bother, however. The only time there will be `lambdaHandlerInfo` and *not* a need to `require('./modules/_lambda-handler.js')` is when there is some early Lambda Runtime error.

lib/instrumentation/index.js
trentm: >  I also don't like the idea that we'll have an instrumentation object that will behave differently when that state's set vs. when it's not set.

It would always be set if you fetched it once in the constructor, or in Instrumentation#start.

> I don't like the idea of the Instrumentation object duplicating state that's available from the environment.

It feels equivalent (to me) to having `agent._conf` be an object that (at least partially) duplicates config state that was specified in the environment.

> What's the goal/benefit

My main concern was that the `getLambdaHandler()` processing is being done potentially many times:
- once in Instrumentation#start()
- and twice for every hooked module the way it is called in _startHook and _patchModule.

lib/lambda.js
```diff
@@ -453,10 +453,30 @@ function isLambdaExecutionEnvironment () {
...
+  if (!isLambdaExecutionEnvironment() || !env._HANDLER || !env.LAMBDA_TASK_ROOT) {
+    return
+  }
+
+  const [handlerModule, handlerFunction, extra] = env._HANDLER.split('.')
+  if (!handlerModule || !handlerFunction || extra) {
+    return
+  }
+
+  const filePath = `${env.LAMBDA_TASK_ROOT}/${handlerModule}.js`
```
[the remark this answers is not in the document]
astorm: > Note that this repo is somewhat out of date: it doesn't have the recentish ESM support that AWS announced earlier this year

tangent: possibly relevant to your interests -- this small lambda function will read the actual runtime in the execution environment and output each file: https://gist.github.com/astorm/d8f188e23511d3d3cefbeae904a7d4c8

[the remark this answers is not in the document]
astorm: Added code to check for the `js` file, and if it doesn't exist, assume a `.csj` file.

https://github.com/elastic/apm-agent-nodejs/blob/ce0f8d6ddbe19af507750dd49ac6f37331b6dd6b/lib/lambda.js#L496

Chose to just use the `.cjs` file without checking for its existence.

lib/instrumentation/index.js
astorm: Hmmm -- something about changing the base instrumentation object like this still bugs me but it's probably not worth tracking down right now.  I've added the `getLambdaHandlerInfo` and addPatch call to the constructor in cd54022a2ba42c1f3f0a71db30a88e68685594d5

lib/lambda.js
astorm: So they do -- tests and logic added to handle this a12ed5ab9dbdf21f3f98100bf98f445d08b7c260

lib/instrumentation/index.js
astorm: Ah, I see.  By "lazy loading" you meant the way that `addPatch` can accept _either_ the module to be patched, or a function that will be called in order to load the module to be patched (the later being lazy loading).

Seems simple enough.  Done in 9ffc95131bf33df3e49aa5b044148d632b8afcd8

lib/lambda.js
astorm: Unresolved this -- while testing it I'd misunderstood the assignment.

The _correct_ semantics of `myModules.some.nestedpathto.myHandler` are a _module_ named `myModules` that exports a handler that looks something like this

```
module.exports = {
  some:{
    nestedpathto: {
        myHandler:function() { } 
    }
  }
}
```

It does _not_ allow a nested file path. I intend to re-implement using this new information.

astorm: OK, got this one working. Wrapping's been manually tested with handlers  named `index.handler` and `science.testmore.handler` and data successfully reported in.

astorm:
OK @trentm that's bugs, features, style and nits addressed. Should be ready for another pass.

trentm (changes requested):
Looks great.  I have some minor comments/nits below.

This way of instrumenting a lambda without needing to touch the code at all is pretty great. I'll post some notes on cases that I tested work for me on a real lambda.

trentm:
### testing notes

I tested with a real Lambda, using the latest v0.0.3 extension layer and a layer of the APM Node.js agent code for this branch creates with [this script](https://gist.github.com/trentm/20f545e0aa8890a5865b20681cae3a52) and manually published; then with `NODE_OPTIONS` and the other required envvars setup.

Trace logging shows the handler module being instrumented, `shimming index@14.18.3 module`:

```
...
2022-03-01T20:04:01.519000+00:00 2022/03/01/[$LATEST]d1321d098b894877b5d674d6eff130de {"log.level":"debug","@timestamp":"2022-03-01T20:04:01.511Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"shimming index@14.18.3 module"}
...
```

- Confirmed I got trace data in APM UI.
- When loaded up a bit with `hey -c 2 -q 50 -z 60s https://REDACTED.execute-api.us-west-2.amazonaws.com/defaulttrentstage/trentm-play-fn1` I got success responses for all invocations. The APM UI throughput rate (~19tps) matched the "Requests/sec:	18.4032" reported by `hey`:
Screen Shot 2022-03-01 at 12 29 21 PM

- A number of Lambda instances were spun up. Some of those are from multiple tests with updated code.
Screen Shot 2022-03-01 at 12 27 29 PM

- I tested with a typical "index.handler" `_HANDLER` name and with a nested "index.my.deeply.nested.handlerNamedBob" handler name.

lib/instrumentation/modules/_lambda-handler.js
```diff
@@ -0,0 +1,33 @@
...
+  while (parts.length > 0) {
+    const prop = parts.shift()
+    if (parts.length === 0) {
+      object[prop] = agent.lambda(object[prop])
+    } else {
+      object = object[prop]
+    }
+    if (!object) {
+      break
+    }
```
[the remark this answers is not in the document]
astorm: done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

lib/instrumentation/modules/_lambda-handler.js
```diff
@@ -0,0 +1,33 @@
...
+  return module
+}
+
+module.exports = function (module, agent, { version, enabled }) {
+  if (!enabled) {
+    return module
+  }
+
+  const { field } = getLambdaHandlerInfo(process.env, Instrumentation.modules, agent.logger)
+  return wrapNestedPath(agent, field, module)
```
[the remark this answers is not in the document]
astorm: Done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

lib/lambda.js
```diff
@@ -464,7 +466,66 @@ function isLambdaExecutionEnvironment () {
   return !!process.env.AWS_LAMBDA_FUNCTION_NAME
 }
 
+function findHandlerNameInModules (handlerModule, modules) {
```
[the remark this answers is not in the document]
astorm: Done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

lib/lambda.js
```diff
@@ -464,7 +466,66 @@ function isLambdaExecutionEnvironment () {
...
+  const handlerModule = match[1]
+  const handlerFunctionPath = match[2]
+
+  // if there's a name conflict with an already instrumented module, skip the
+  // instrumentation of the lambda handle and log a message.
+  if (findHandlerNameInModules(handlerModule, modules)) {
+    logger.info(
+      'Unable to instrument Lambda handler due to name conflict with %s, please choose a different Lambda handler name',
+      handlerModule
+    )
```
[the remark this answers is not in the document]
astorm: 1 is done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

Re: 2 
> Is this a limitation we could overcome (later)?

Maybe/Probably? Most things are possible with code, it's just a matter of how long it would take.

The challenge her eis there's a lot of hooking and patching code that uniquely identifies a module by its name.  In order to allow a lambda handler's module name to be the same as an instrumented module's name we'd need to change how the agent's hooking/patching code identifies a module, and that might also mean changes to `require-in-the-middle`.  Non trivial planning, non trivial amount of work, and a high potential to break something.

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
+'use strict'
+const tape = require('tape')
+const path = require('path')
+
+const { getLambdaHandlerInfo } = require('../../lib/lambda')
+
+tape.test('unit tests for getLambdaHandlerInfo', function (suite) {
+  const logger = {
+    info: function () { }
+  }
```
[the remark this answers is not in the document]
astorm: Switched to the non-capturing logger in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

lib/lambda.js
```diff
@@ -464,7 +466,66 @@ function isLambdaExecutionEnvironment () {
...
+    for (const instrumentedModule of instrumentedModules) {
+      if (handlerModule === instrumentedModule) {
+        return true
+      }
+    }
+  }
+  return false
+}
+
+function getFilePath (taskRoot, handlerModule, pathParts) {
```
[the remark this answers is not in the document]
astorm: Added doc block in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
...
+      LAMBDA_TASK_ROOT: path.resolve(__dirname, 'fixtures')
+    }, MODULES, logger)
+    t.equals(handler.filePath, path.resolve(__dirname, 'fixtures', 'lambda.js'), 'extracted handler file path')
+    t.equals(handler.module, 'lambda', 'extracted handler module')
+    t.equals(handler.field, 'bar', 'extracted handler field')
+    t.end()
+  })
+
+  suite.test('returns no value if module name conflicts with already instrumented module', function (t) {
+    process.env.AWS_LAMBDA_FUNCTION_NAME = 'express'
```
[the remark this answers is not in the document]
astorm: Changed this in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
...
+    const handler = getLambdaHandlerInfo({
+      _HANDLER: 'express.bar',
+      LAMBDA_TASK_ROOT: '/var/task'
+    }, MODULES, logger)
+    t.equals(handler, undefined, 'no handler extracted')
+    t.end()
+  })
+
+  suite.test('no task root', function (t) {
```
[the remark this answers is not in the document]
astorm: Yup, they are.  I presume this was a request to have each test explicitly set a function name.  Done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
...
+    }, MODULES, logger)
+
+    t.equals(handler.filePath, path.resolve('/var', 'task', 'foo.cjs'), 'extracted handler file path')
+    t.equals(handler.module, 'foo', 'extracted handler module')
+    t.equals(handler.field, 'baz.bar', 'extracted handler field')
+    t.end()
+  })
+
+  suite.end()
+  // t.end()
```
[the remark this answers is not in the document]
astorm: Done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
...
+
+  // load express after agent (for wrapper checking)
+  const express = require('express')
+
+  // check that the handler fixture is wrapped
+  const handler = require(path.join(__dirname, '/fixtures/lambda')).foo
+  t.equals(handler.name, 'wrappedLambda', 'handler function wrapped correctly')
+
+  // did normal patching/wrapping take place
+  t.equals(express.static.name, 'wrappedStatic', 'handler function wrapped correctly')
```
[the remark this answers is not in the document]
astorm: Done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
...
+const tape = require('tape')
+const path = require('path')
+
+const { getLambdaHandlerInfo } = require('../../lib/lambda')
+
+tape.test('unit tests for getLambdaHandlerInfo', function (suite) {
+  const logger = {
+    info: function () { }
+  }
+  const MODULES = [
```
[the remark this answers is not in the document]
astorm: I presume this was a request to change this test fixture data to use the data exported by the instrumentation module.  Done in 7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

test/lambda/wrapper.test.js
```diff
@@ -0,0 +1,180 @@
...
+      LAMBDA_TASK_ROOT: path.resolve(__dirname, 'fixtures')
+    }, MODULES, logger)
+
+    t.equals(handler.filePath, path.resolve(__dirname, 'fixtures', 'handlermodule.cjs'), 'extracted handler file path')
+    t.equals(handler.module, 'handlermodule', 'extracted handler module')
+    t.equals(handler.field, 'lambda.bar', 'extracted handler field')
+    t.end()
+  })
+
+  suite.test('extracts info with expected env variables, cjs extension', function (t) {
```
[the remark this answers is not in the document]
astorm: probably.  changed in b44002c4d37096a1f6e58ab31b5888ae64ed3602

astorm:
@trentm latest requested changes are in.

lib/instrumentation/modules/_lambda-handler.js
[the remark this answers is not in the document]
trentm: Was this change missed. I don't see it in https://github.com/elastic/apm-agent-nodejs/commit/7b3d65f6a7b64382cd0c22100e21ff1912e8ec73

trentm:
@astorm Another Q at https://github.com/elastic/apm-agent-nodejs/pull/2577#discussion_r817250340

lib/instrumentation/modules/_lambda-handler.js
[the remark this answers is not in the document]
astorm: Yup, looks like I made it in my head: see 7967d8536ff0c21f55415bd43fdf783e1544b856

trentm (changes requested):

lib/instrumentation/modules/_lambda-handler.js
```diff
@@ -0,0 +1,32 @@
...
+function wrapNestedPath (agent, field, module) {
+  const parts = field.split('.')
+  let object = module
+  while (parts.length > 0) {
+    if (!object) {
+      break
+    }
+    const prop = parts.shift()
+    if (parts.length === 0) {
+      object[prop] = agent.lambda(object[prop])
```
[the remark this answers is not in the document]
astorm: Well that's annoying, especially since the version I had previously generally worked. And the edge case we want to fix, `index.nosuchprop`, would result in no lambda function being executed regardless of the agent's presence.  To be honest this doesn't seem worth the time we're taking with it just to preserve an error message in the logs -- that feeling's  probably why I biffed the change.

Regardless -- new version pushed.  The `wrapNestedPath` function has test coverage.  If the function's called with a property that doesn't exist the lambda function won't be wrapped. 

Tested in the live lambda environment and it works.  An `index.notthere` handler results in the following error message in the logs, as expected.

```
2022-03-02T23:43:18.222Z undefined ERROR Uncaught Exception {"errorType":"Runtime.HandlerNotFound","errorMessage":"index.notthere is undefined or not exported","stack":["Runtime.HandlerNotFound: index.notthere is undefined or not exported"," at Object.module.exports.load (/var/runtime/UserFunction.js:246:11)"," at Object.<anonymous> (/var/runtime/index.js:43:30)"," at Module._compile (internal/modules/cjs/loader.js:1085:14)"," at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)"," at Module.load (internal/modules/cjs/loader.js:950:32)"," at Function.Module._load (internal/modules/cjs/loader.js:790:12)"," at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:76:12)"," at internal/main/run_main_module.js:17:47"]} 
```

I'm leaving the lambda wrapper itself alone but wouldn't object if you want to add guard clause there.

If there's more cases you want `wrapNestedPath` to handle they should be pretty easy to add a new test case for.

trentm (approved):

Issue #2605, opened 2022-03-10
lambda: transaction.context.message.headers is broken if captureHeaders=true and an SNS or SQS event has message attributes


trentm:
## the bug

If a Lambda function configured to use the Node.js APM agent and with `captureHeaders: true` (the default) receives a SNS or SQS event with message attributes, it will result in a transaction object with bogus `context.message.headers` that breaks the APM server transaction schema and breaks on intake with an error like this:

```
APM server response body: {"accepted":0,"errors":[{"message":"decode error: data read error: v2.transactionRoot.Transaction: v2.transaction.SpanCount: v2.transactionSpanCount.Context: v2.context.Message: v2.contextMessage.Headers: invalid input for HTTPHeader: map[Type:String Value:this is my foo value]","document":"{\"transaction\":{\"name\":\"RECEIVE trentm-play-topic1\",\"type\":\"messaging\",\"result\":\"success\",\"id\":\"9176fe68d4e3b953\",\"trace_id\":\"208075b3b3fd2ac6b549711d94e877f2\",\"duration\":60.559,\"timestamp\":1646781029978068,\"sampled\":true,\"context\":{\"user\":{},\"tags\":{},\"custom\":{},\"service\":{\"origin\":{\"name\":\"trentm-play-topic1\",\"id\":\"arn:aws:sns:us-west-2:627286350134:trentm-play-topic1\",\"version\":\"1.0\"}},\"cloud\":{\"origin\":{\"provider\":\"aws\",\"region\":\"us-west-2\",\"service\":{\"name\":\"sns\"},\"account\":{\"id\":\"627286350134\"}}},\"message\":{\"queue\":{\"name\":\"trentm-play-topic1\"},\"age\":{\"ms\":1247},\"headers\":{\"foo\":{\"Type\":\"String\",\"Value\":\"this is my foo value\"}}}},\"span_count\":{\"started\":0},\"outcome\":\"success\",\"faas\":{\"id\":\"arn:aws:lambda:us-west-2:627286350134:function:trentm-play-fn1\",\"coldstart\":true,\"execution\":\"cc8565c8-a165-4b02-b606-8d477659a40a\",\"trigger\":{\"type\":\"pubsub\",\"request_id\":\"8d426c7c-fe30-5f4b-9a12-36dd1a940ae0\"}},\"sample_rate\":1}}"}]}
```

[...]

Issue #1831, opened 2020-10-15
TraceState Support for Amazon Lambda


astorm:
Once we take care of #1521, we'll want to make sure we add tracestate header parsing [to the lambda transaction starting code](https://github.com/elastic/apm-agent-nodejs/blob/4fe301d66a1ef748a77a0cde7584931e176c618b/lib/lambda.js#L76).

trentm:
Trace context *is* being picked up from `event.headers` to use as `childOf` for creating the lambda transaction. However there is some remaining work here:

- [ ] The current `event.headers` checking is being done for all trigger events, and not scoped to API Gateway events.
- [ ] Possible traceparent/tracestate from SNS and SQS single-event `record.messageAttributes` is not being picked up, but it is part of the spec.
- [ ] We don't have test cases for any of the above yet.

PR #2606, opened 2022-03-10
feat: use trace-context "headers" in SNS and SQS single event message attributes for lambda transaction

trentm:
- If a SNS or SQS single event trigger to an instrumented Lambda
  function includes message attributes with the name "traceparent" (and
  "tracestate"), case-insensitive, then those are used to continue the
  trace. This was already being done for API Gateway event headers.
- Also, fix a bug in the capturing of SNS and SQS message attributes as
  "transaction.context.message.headers". Note that only "String"-type
  message attributes are captured. Binary-type attributes are base64
  encoded. I'm not sure of the value in capturing them.

Closes: #1831
Fixes: #2605
Refs: https://github.com/elastic/apm/pull/614

### Checklist

- [x] Implement code
- [x] Add tests


lib/lambda.js
```diff
@@ -9,6 +9,32 @@ const path = require('path')
 let isFirstRun = true
 let gFaasId // Set on first invocation.
 
+// The trigger types for which we support special handling.
+// https://docs.aws.amazon.com/lambda/latest/dg/lambda-services.html
+const TRIGGER_GENERIC = 1
```
astorm, line 14: I see we've replaced the smaller functions that identify an individual payload type (ex. `isApiGatewayEvent`) with this larger/more-complex function `triggerTypeFromEvent` function that identifies an individual payload type, and then we call `triggerTypeFromEvent` once its result value is passed around where it needs to be.  The main tradeoffs here seem to be 

- the perf. of multiple calls to the older small function vs. the higher memory use of keeping the constant-ish `TRIGGER_` variables around. 

- Swapping where the complexity goes :) 

No objections, just noting the change out loud.

astorm, line 14: I see we've replaced the smaller functions that identify an individual payload type (ex. `isApiGatewayEvent`) with this larger/more-complex function `triggerTypeFromEvent` function that identifies an individual payload type, and then we call `triggerTypeFromEvent` once its result value is passed around where it needs to be.  The main tradeoffs here seem to be 

- the perf. of multiple calls to the older small function vs. the higher memory use of keeping the constant-ish `TRIGGER_` variables around. 

- Swapping where the complexity goes :) 

No objections, just noting the change out loud.

lib/lambda.js
```diff
@@ -199,7 +221,10 @@ function setSqsData (agent, trans, event, context, faasId, isColdStart) {
   }
 
   if (agent._conf.captureHeaders) {
-    messageContext.headers = Object.assign({}, record.messageAttributes)
+    const headers = headersFromSqsMessageAttributes(record.messageAttributes)
+    if (Object.keys(headers).length > 0) {
```
astorm, line 225: :+1: for `Object.keys(obj)` -- I'm pretty sure that's the fastest way to do this without needing to get into `for in` and `hasOwnProperty` territory. Unrelated to this PR/note-to-self -- a custom eslint linting rule that ensures we're using the more efficient Object.keys() for object looping/iteration (vs. one of the other myriad ways you can loop an object in javascript) would be a nice to have.

astorm, line 225: :+1: for `Object.keys(obj)` -- I'm pretty sure that's the fastest way to do this without needing to get into `for in` and `hasOwnProperty` territory. Unrelated to this PR/note-to-self -- a custom eslint linting rule that ensures we're using the more efficient Object.keys() for object looping/iteration (vs. one of the other myriad ways you can loop an object in javascript) would be a nice to have.

lib/lambda.js
```diff
@@ -559,10 +594,57 @@ function wrapNestedPath (agent, field, module) {
...
+
+// Reduce a SQS record message attributes object to a headers-like object of
+// string key/value-pairs. Some attribute types are ignored.
+// https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-message-metadata.html#message-attribute-data-types
+//
+// `record.messageAttributes` is of the form:
+//    { <attribute-name>: { dataType: <attr-type>, stringValue: <attr-value>, ... } }
+// For example:
+//    { traceparent: { dataType: 'String', stringValue: 'test-traceparent' } }
+function headersFromSqsMessageAttributes (messageAttributes) {
```
astorm, line 613: 👍 I see a new `headersFromSqsMessageAttributes` that simplifies the data from amazon's strong-ishly typed JSON data into something that APM Server can handle. (I also see a separate function for SNS messages)

astorm, line 613: 👍 I see a new `headersFromSqsMessageAttributes` that simplifies the data from amazon's strong-ishly typed JSON data into something that APM Server can handle. (I also see a separate function for SNS messages)

lib/lambda.js
```diff
@@ -411,7 +410,12 @@ function elasticApmAwsLambda (agent) {
     }
 
     return function wrappedLambda (event, context, callback) {
-      log.trace({ awsRequestId: context && context.awsRequestId }, 'lambda: fn start')
+      if (!(event && context && typeof callback === 'function')) {
```
astorm, line 413: 👍  for guarding against unknown datatypes here, per previous discussions on how things can go wrong without this.

astorm, line 413: 👍  for guarding against unknown datatypes here, per previous discussions on how things can go wrong without this.

lib/lambda.js
```diff
@@ -432,30 +436,61 @@ function elasticApmAwsLambda (agent) {
...
+          normedHeaders = lowerCaseObjectKeys(headersFromSqsMessageAttributes(record.messageAttributes))
         }
-        parentId = normalizedHeaders.traceparent ? normalizedHeaders.traceparent : normalizedHeaders['elastic-apm-traceparent']
-        tracestate = normalizedHeaders.tracestate
+      } else if (triggerType === TRIGGER_SNS_SINGLE_EVENT) {
+        const record = event.Records[0]
+        if (record.Sns && record.Sns.MessageAttributes) {
+          normedHeaders = lowerCaseObjectKeys(headersFromSnsMessageAttributes(record.Sns.MessageAttributes))
+        }
+      }
```
astorm, line 459: Not required for this PR, but I'm genuinely unsure if we should mention the agent's  automatic handling of reading the trace context headers in a Lambda environment in our message queue doc, or if that would confusing things more https://github.com/elastic/apm-agent-nodejs/blob/main/docs/message-queues.asciidoc

astorm, line 459: Not required for this PR, but I'm genuinely unsure if we should mention the agent's  automatic handling of reading the trace context headers in a Lambda environment in our message queue doc, or if that would confusing things more https://github.com/elastic/apm-agent-nodejs/blob/main/docs/message-queues.asciidoc

astorm (commented):
Nothing objectionable here. It looks like _the thing to do_ and the approach seems reasonable.  Happy to approve once a changelog entry gets added.

I noticed this was still in draft state -- if there's a specific part you wanted feedback  on or  a part you think needs more extensive testing from a second set of eyes just let me know.  Otherwise 👍

lib/lambda.js
trentm: I think it is worthwhile to mention this behaviour in the *lambda.asciidoc*. Likely a reference in the message-queues doc would be worthwhile as well, I'd have to re-read that doc.

trentm:
> I noticed this was still in draft state

Ah, thanks. I was going to pull out of draft after the changelog entry and merge from main.

trentm:
Also there is still some discussion on the apm spec: https://github.com/elastic/apm/pull/614#discussion_r823436714
So I'll wait for that to settle before finalizing this.

trentm:
Trying that failed "Integration Tests" again: https://apm-ci.elastic.co/job/apm-integration-tests-selector-mbp/job/main/2446/

astorm (approved):

Issue #2604, opened 2022-03-07
Span Compression: Replace isCompressionEligible method in Span Buffering Behavior, Enable By Default


astorm:
This issue represents the work to replace the implementation of `isCompressionEligible` done in #2600 with a version that's fully compliant with the [Compressing spans](https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md) spec.  This will be done via the work completed in #2601 and #2602

Specifically, we will replace the following behavior

> a span is compressible if it's type is in this list" behavior with

with 

> a span in compressible if it's an exit span, its context has not been propagated, and its outcome is not failure.

This will be the last bit of work to complete for span compression. Therefore, in addition to the above we'll want to set  `spanCompressionEnabled`'s default value to `true`.

astorm:
See if we can fix this flaky test while we're in there: https://github.com/elastic/apm-agent-nodejs/runs/6081847063?check_suite_focus=true#step:8:2934

Issue #2602, opened 2022-03-07
Span Compression: Implement TraceContext Injection API


astorm:
Full compliance with the [Compressing spans](https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md) spec requires us to know when a span's context has been propagated downstream.  It's been suggested the best way to do this is to implement an API for trace context injection/propagation, similar to the java agent's

> The Java agent defines an API for context propagation and marks the span as non-discardable if context has been propagated: https://github.com/elastic/apm-agent-java/blob/2d6933b09e144bcf4057822a6b2f6d19d4374cbd/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/AbstractSpan.java#L583-L587

> Example usage for an HTTP client: https://github.com/elastic/apm-agent-java/blob/2d6933b09e144bcf4057822a6b2f6d19d4374cbd/apm-agent-plugins/apm-apache-httpclient-plugin/src/main/java/co/elastic/apm/agent/httpclient/LegacyApacheHttpClientInstrumentation.java#L72

> This is the implementation that's responsible for injecting the context: https://github.com/elastic/apm-agent-java/blob/main/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/TraceContext.java#L559-L572

This PR represents that work.

Issue #2601, opened 2022-03-07
Span Compression: Fully Implement Exit Spans in Instrumentation


astorm:
Full compliance with the [Compressing spans](https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md) spec requires us to treat exit spans an non-exit spans differently.  This PR represents the work to decide which spans in the Node.js will be exit spans, and to make the necessary code changes to turn them into exit spans.

Issue #2600, opened 2022-03-07
Span Compression: Implement Span Buffering Behavior


astorm:
This PR represents the work needed to implement the span buffering behavior of the [Compressing spans](https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md) work.  We will need to 

- Implement the configuration gated span end buffering behavior described in the spec
- Implement the `tryToCompress` method fully
- Implement a preliminary `isCompressionEligible` method that, contrary to spec, uses the outcome AND a hard coded list of span types to determine if it's compressible

At the end of this work we will have an agent that compresses spans but is not 100% spec compliant.

PR #2623, opened 2022-03-24
feat: span compression

astorm:
Implements: https://github.com/elastic/apm-agent-nodejs/issues/2600

Implements an initial version of span compression that compresses/collapses both identical and same kind spans into a single "composite" span. 

This PR does _not_ identify compressible spans by the exit span and injected trace context rules -- instead a hard coded list of span types is referenced.   The work to use exit spans and injected trace context to identify compressible spans will be completed in 

- https://github.com/elastic/apm-agent-nodejs/issues/2601
- https://github.com/elastic/apm-agent-nodejs/issues/2602
- https://github.com/elastic/apm-agent-nodejs/issues/2604

Since this is an initial version we have set the default of `spanCompressionEnabled` to `false`.  We will make this `true` by default when finishing up in https://github.com/elastic/apm-agent-nodejs/issues/2604. (I went back and forth on this one -- so if you think we should enable by default just let me know)

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


docs/configuration.asciidoc
```diff
@@ -1181,3 +1181,57 @@ require('elastic-apm-node').start({
   ]
 })
 ----
+
```
astorm, line 1184: Review Note: docs taken from java agent's docs of the same field.

test/instrumentation/modules/@elastic/elasticsearch.test.js
```diff
@@ -7,7 +7,8 @@ const agent = require('../../../..').start({
   metricsInterval: 0,
```
astorm, line 7: Review Note: Span Compression can effect _when_ a span is recorded, which can create race-like conditions in the existing test suites when enabled.  To counter this, ` spanCompressionEnabled: false` has been set in these tests.

lib/instrumentation/span-compression.js
```diff
@@ -0,0 +1,176 @@
...
+
+    return false
+  }
+
+  // Attempts to compression the second span into the first
+  //
+  // @param Span compositeSpan
+  // @param Span spanToCompress
+  // @return boolean
+  tryToCompress (compositeSpan, spanToCompress) {
```
trentm, line 123: The `tryToCompress` implementation would be a lot easier to compare to the spec if it followed the pseudo-code at https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md#turning-compressed-spans-into-a-composite-span   Is there a particular reason to not have followed that?

trentm (changes requested):
I agree with setting `spanCompressionEnabled: false` by default until the implementation is complete. It would be helpful to mark that config var as "experimental" in configuration.asciidoc until the implementation is complete, in case we do releases in the interim.

lib/instrumentation/span-compression.js
astorm: That's interesting -- I generally don't implement pseudo code literally from specs and  the couple of other agent's I looked at drifted a bit from it as well.  I'm not really sure how to answer that question other than I simplified the implementation details of that pseudo code and created the compression service object because 1. it made it easier to reason about and test the end result as the feature was being developed and 2. because I thought the complexity might be objected to during review.  

That said with the end-to-end tests in place refactoring this shouldn't be a big issue other than the time it will take, so I'll get started on that.

lib/config.js
```diff
@@ -215,7 +222,18 @@ var DURATION_OPTS = [
     allowedUnits: ['ms', 's', 'm'],
     allowNegative: false
   },
-
+  {
+    name: 'spanCompressionExactMatchMaxDuration',
+    defaultUnit: 'ms',
```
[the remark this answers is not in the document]
astorm: No idea -- I don't intend to take on this additional work in this PR.

lib/instrumentation/index.js
```diff
@@ -345,6 +350,32 @@ Instrumentation.prototype.addEndedSpan = function (span) {
     return
   }
 
+  if (!this._agent._conf.spanCompressionEnabled) {
```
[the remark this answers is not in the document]
astorm: Seems out of scope for this PR and feature.  Is there a compelling reason we'd want this here?

lib/instrumentation/span-compression.js
```diff
@@ -0,0 +1,176 @@
...
+      // set initial composite context values
+      this.timestamp = compositeSpan.timestamp
+      this.composite.count = 1
+      this.composite.sum = compositeSpan._duration
+    } else {
+      // if so, compare with the compression strat and bail if mismatch
+      const strat = this._getCompressionStrategy(compositeSpan, spanToCompress)
+      if (strat !== this.composite.compression_strategy) {
+        return false
+      }
```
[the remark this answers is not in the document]
astorm: I disagree (or think I do) -- this appears to work correctly. Both the _test tryToCompress exact match_ and _test tryToCompress same kind_ include tests for "the strategy changed".  I also added a new test named _test tryToCompress same kind, then exact match_  with an additional scenario to confirm that things work as intended. 

Once a span becomes a same kind composite its name is going to be prefixed with the string _Calls to_ which means the names will no longer exact match. 

If I've misunderstood could you express the scenario you're talking about as a test case?

astorm:
@trentm Design's been changed to explicitly match the pseudo code, per your request. Also, the other feedback I'm addressing has been addressed. 

Finally -- I wasn't sure if your pseduo code request extended to the fact that the logic for everything was broken out into a `SpanCompression` object/class. If you think this logic is better off in/on the span object itself just let me know -- that shouldn't be too big a refactor with the tests in place.

lib/instrumentation/span-compression.js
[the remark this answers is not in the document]
trentm: Okay, understood. Here is a contrived example (creating a span with the name "Calls to mysql") that can break this, but seems unlikely :)

```js
var apm = require('./').start({
  metricsInterval: '0s',
  centralConfig: false,
  cloudProvider: 'none',
  spanCompressionEnabled: true,
  spanCompressionSameKindMaxDuration: '50ms'
})

function main () {
  var t0 = apm.startTransaction('t0')
  setImmediate(() => {
    var s1 = t0.startSpan('s1', 'db', 'mysql')
    s1.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
    setTimeout(() => {
      s1.end()
      var s2 = t0.startSpan('s2', 'db', 'mysql')
      s2.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
      setTimeout(() => {
        s2.end()

        // Create a span with a contrived `name` to fool `SpanCompression._initCompressionStrategy()`
        var s3 = t0.startSpan('Calls to mysql', 'db', 'mysql')
        s3.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
        setTimeout(() => {
          s3.end()
          t0.end()
        }, 10)
      }, 10)
    }, 10)
  })
}

main()
```

This results in:

```
    trace 2c2404
    `- transaction 004fa5 "t0" (41.486ms, outcome=unknown)
       `- span fb8e47 "Calls to mysql" (24.776ms, mysql, compressed 2 same_kind spans sum 23.366ms)
       `- span 7e5a8e "Calls to mysql" (11.518ms, mysql)
```

lib/instrumentation/span.js
```diff
@@ -248,6 +248,21 @@ Span.prototype._encode = function (cb) {
...
+  }
+
+  if (self.isCompositeSameKind()) {
+    // rename
+    let callsTo = self && self._destination && self._destination.service && self._destination.service.resource
+    if (!callsTo) {
+      callsTo = 'unknown'
+    }
+    payload.name = `Calls to ${callsTo}`
+  }
```
trentm, line 264: This results in this incorrect span name for the first span:

    trace 8666e4
    `- transaction 37e147 "t0" (30.227ms, outcome=unknown)
       `- span ca9816 "Calls to mysql" (11.748ms, mysql)
       `- span e21dc6 "s2" (11.564ms, mysql

using this example code:

```js
var apm = require('./').start({
  metricsInterval: '0s',
  centralConfig: false,
  cloudProvider: 'none',
  spanCompressionEnabled: true
})

// Create two sibling mysql spans that *would* be compressed with "same_kind"
// but don't because they are shorter than the 50ms default threshold.
function main () {
  var t0 = apm.startTransaction('t0')
  setImmediate(() => {
    var s1 = t0.startSpan('s1', 'db', 'mysql')
    s1.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
    setTimeout(() => {
      s1.end()
      var s2 = t0.startSpan('s2', 'db', 'mysql')
      s2.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
      setTimeout(() => {
        s2.end()
        t0.end()
      }, 10)
    }, 10)
  })
}

main()
```

Is the "Calls to ..." naming in `span._encode()` necessary given that naming is also being done in `SpanCompression.prototype.tryToCompressRegular`? If I comment that block out, then the above example no longer fails, and the tests pass.

trentm (changes requested):

lib/instrumentation/span.js
astorm: That's definitely some old/dead code in the `_encode` method, so I've removed that. 

However I'm not quite sure I understood your example and wasn't able to get it to fail in the way you indicated (at least with the feature as it was Monday).  

Here's what I mean -- in the code above I see two spans, each of which should be around 10ms long.

The spans have different names but otherwise look similar, so if compression is enabled they should be compressed as same kind.

Each of the spans is less than 10ms which means they _should_ be compressed into a single span with the _Calls to_ name since 10ms is less than the 50ms default `spanCompressionSameKindMaxDuration`. 

The current code appears to do this correctly. I've adapted your code into a test named _create two sibling mysql spans ..._ that demonstrates this.

I was not able to reproduce the behavior you describe where you end up with a tree that looks like this.

```
    trace 8666e4
    `- transaction 37e147 "t0" (30.227ms, outcome=unknown)
       `- span ca9816 "Calls to mysql" (11.748ms, mysql)
       `- span e21dc6 "s2" (11.564ms, mysql
```

Also I don't understand quite understand this comment (unless it's describing undesirable behavior you saw last week?)

```
// Create two sibling mysql spans that *would* be compressed with "same_kind"
// but don't because they are shorter than the 50ms default threshold.
```      

So -- are you still seeing problems with the above code with the up-to-date branch?  Or have I. made some dreadful presumption about something in the spec that's incorrect?

trentm: First, sorry: I botched that *comment* and made this all confusing.  Let me try again.

I'm running this example with commit 8979b453be7b893150ca9a9975cf51b5f39134c6 (i.e. before your https://github.com/elastic/apm-agent-nodejs/pull/2623/commits/06c06efd1eebbd3e691b16b1a372d0e619b0d21a to remove the defunct code in Span.prototype._encode).

The new code example is as follows. I've only updated the comments from the above code example:

```js
var apm = require('./').start({
  metricsInterval: '0s',
  centralConfig: false,
  cloudProvider: 'none',
  spanCompressionEnabled: true
})

function main () {
  var t0 = apm.startTransaction('t0')
  setImmediate(() => {
    var s1 = t0.startSpan('s1', 'db', 'mysql')
    s1.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
    setTimeout(() => {
      // `s1` is compression eligible, so we expect this to get buffered.
      s1.end()

      var s2 = t0.startSpan('s2', 'db', 'mysql')
      s2.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })
      setTimeout(() => {
        // `s2` is a `same_kind` match to s1. However,
        // `spanCompressionSameKindMaxDuration` defaults to 0ms and this is,
        // obviously, longer than that. So we expect *no compression*.
        s2.end()
        t0.end()
      }, 10)
    }, 10)
  })
}

main()
```

My expection is a transaction `t0` with separate spans `s1` and `s2` -- i.e. **no compression ** -- because they are `same_kind` match, but have duration ~10ms which is longer than the default `spanCompressionSameKindMaxDuration: '0ms'`.

What I observe is:

```
    {
        "span": {
            "name": "Calls to mysql",
            "type": "db",
            "id": "ece21bbe8bc46317",
            "transaction_id": "650eb8284d6c4831",
            "parent_id": "650eb8284d6c4831",
            "trace_id": "363fd5101e87cbac9ce72b3a29be63b9",
            "subtype": "mysql",
            "action": null,
            "timestamp": 1649183613812077,
            "duration": 11.721,
            "context": {
                "destination": {
                    "service": {
                        "name": "mysql",
                        "resource": "mysql",
                        "type": "db"
                    }
                }
            },
            "sync": false,
            "outcome": "success",
            "sample_rate": 1
        }
    }
    {
        "span": {
            "name": "s2",
            "type": "db",
            "id": "dd9eeda8bdbf45a4",
            "transaction_id": "650eb8284d6c4831",
            "parent_id": "650eb8284d6c4831",
            "trace_id": "363fd5101e87cbac9ce72b3a29be63b9",
            "subtype": "mysql",
            "action": null,
            "timestamp": 1649183613824160,
            "duration": 11.388,
            "context": {
                "destination": {
                    "service": {
                        "name": "mysql",
                        "resource": "mysql",
                        "type": "db"
                    }
                }
            },
            "sync": false,
            "outcome": "success",
            "sample_rate": 1
        }
    }
```

- Good: The spans were not compressed.
- Bad: The name of the first span was incorrectly changed to `Calls to mysql`.

If I update to the latest commit, then this is fixed and I observe what is expected. In other words, all good now with the latest changes.

lib/instrumentation/span-compression.js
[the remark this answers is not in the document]
trentm: Unresolving this in case you didn't see my follow-up comment. Feel free to resolve again if you don't think the contrived case is worth guarding against.

test/instrumentation/span-compression.test.js
```diff
@@ -0,0 +1,385 @@
...
+    resetAgent(function (data) {
+      t.equals(data.length, 2)
+      const span = data.spans.shift()
+      t.equals(span.name, 'name1')
+      t.equals(span.composite.compression_strategy, constants.STRATEGY_EXACT_MATCH)
+      t.equals(span.composite.count, 3)
+      if (!(span.composite.sum > 30)) {
+        console.log(span.composite)
+        process.exit(1)
+      }
```
trentm, line 37: Is this left over debugging code?

test/instrumentation/span-compression.test.js
```diff
@@ -0,0 +1,385 @@
...
+      t.true(span.composite.sum > 30)
+      t.equals(span.duration, (finalSpan._endTimestamp - firstSpan.timestamp) / 1000)
+      t.end()
+    })
+
+    agent.startTransaction(agent)
```
trentm, line 43:
```suggestion
    agent.startTransaction('trans')
```

test/instrumentation/span-compression.test.js
```diff
@@ -0,0 +1,385 @@
...
+      const span = data.spans.shift()
+      t.equals(span.name, 'Calls to foo')
+      t.equals(span.composite.compression_strategy, constants.STRATEGY_SAME_KIND)
+      t.equals(span.composite.count, 3)
+      t.true(span.composite.sum > 30)
+      t.equals(span.duration, (finalSpan._endTimestamp - firstSpan.timestamp) / 1000)
+      t.end()
+    })
+
+    agent.startTransaction(agent)
```
trentm, line 85:
```suggestion
    agent.startTransaction('trans')
```

trentm (commented):
(This may be something for a separate change and I think it points to an issue in the span-compression *spec*. I recall Felix mentioning something about the span-compression spec needing to deal with the case of an ended parent.)

If you run the following script with the current PR code:

```js
// A bug in span-compression where a child span gets buffered to an
// *already ended* parent, which results in the child span never being sent.

var apm = require('./').start({
  metricsInterval: '0s',
  centralConfig: false,
  cloudProvider: 'none',
  spanCompressionEnabled: true
})

function main () {
  const t0 = apm.startTransaction('t0')

  const s1 = apm.startSpan('SELECT FROM a', 'db', 'mysql', { exitSpan: true })
  s1.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })

  setTimeout(() => {
    s1.end()
  }, 100)

  // Create an async task which captures the current async run context
  // (i.e. apm.currentSpan === s1)
  setTimeout(() => {
    // s2 is a child of s1 *even if s1 may have already ended*.
    const s2 = apm.startSpan('SELECT FROM b', 'db', 'mysql', { exitSpan: true })
    s2.setDestinationContext({ service: { name: 'mysql', resource: 'mysql', type: 'db' } })

    // When s2 is ended, it is compression eligible, so it is buffered on its
    // parent (s1). However, s1 has already ended, so the current span
    // compression code will never send the buffered s2.
    s2.end()
  }, 200)

  setTimeout(() => {
    t0.end()
  }, 300)
}

main()
```

Then you get a trace like this:

```
trace 153dcb
`- transaction 139abe "t0" (303.727ms, outcome=unknown)
    `- span b57c98 "SELECT FROM a" (103.077ms, mysql)
```

This is broken because the "SELECT FROM b" span is missing. The issue is that we've created a case where a child span *ends after its parent has already ended*. This results in the span potentially (if eligible for compression) being buffered on that parent, and then never sent. (While the example above looks contrived, it isn't actually too unexpected in Node.js code. There is also a case with async functions where a user doing manual instrumentation can have this happen unexpectedly: https://github.com/elastic/apm-agent-nodejs/issues/2611.)

With some debug-prints, this is what happens in the example above:

```
% node compressionplay-break-children.js
XXX -- Instrumentation.addEndedSpan(<Span "SELECT FROM a">)
XXX parent has no buffered span: buffer <Span "SELECT FROM a"> on <Span "t0">
XXX -- Instrumentation.addEndedSpan(<Span "SELECT FROM b">)
XXX parent has no buffered span: buffer <Span "SELECT FROM b"> on <Span "SELECT FROM a">    <-- The bug is here, span "SELECT FROM a" has already ended.
XXX -- Instrumentation.addEndedTransaction(<Transaction "t0">)
XXX send buffered <Span "SELECT FROM a">
```

If you set `spanCompressionEnabled: false` in this script, you get the expected two spans.

## spec issue

I think this points to an oversight in the spec in the `onChildEnd` pseudo-code at https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md#span-buffering. Perhaps that could change to:

```
...
void onChildEnd(Span child) {
    if (ended || !child.isCompressionEligible()) {
...
```

Here is a potential fix in this PR:

```diff
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -355,25 +358,31 @@ Instrumentation.prototype.addEndedSpan = function (span) {
       this._encodeAndSendSpan(span.getBufferedSpan())
     }

-    if (!span.isCompressionEligible()) {
+    if (span.getParentSpan().ended || !span.isCompressionEligible()) {
       // span is not compressible -- send
       const buffered = span.getBufferedSpan()
       if (buffered) {
```

I can open a spec PR for discussion.

trentm:
> I can open a spec PR for discussion.

https://github.com/elastic/apm/pull/632

test/instrumentation/span-compression.test.js
astorm: It's empty bytes now.

astorm:
> (This may be something for a separate change and I think it points to an issue in the span-compression spec. I recall Felix mentioning something about the span-compression spec needing to deal with the case of an ended parent.)

Implemented the suggested fix and adopted the reproduction code as a test in the repository.

trentm (approved):

Issue #2626, opened 2022-03-28
start using package-lock.json


trentm:
The Lambda Layer builds and coming https://github.com/elastic/apm-agent-nodejs/issues/2622 means we should start using a package-lock for repeatable builds.

Also, turning on using dependabot for the repo (and should do so for the http-client repo as well) will probably help us keep package-lock.json updated.

Also, can we start using [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) instead of `npm install` in our CI builds to speed them up? I'm not sure if `npm ci` was there for the npm version in node v8.6 (our current base node version).

IIUC, a slight downside of using package-lock.json is that our regular testing will not test with the same (latest published versions of) deps that a user would get when using `"elastic-apm-node": "..."` in their project deps. But, then again, that not running with the current latest is the point. :)

# checklist

- [x] Add a package-lock.json file.
- [x] See if we can use `--ignore-scripts` for the Lambda Layer build. 
- [ ] Decide on, and if "yes", setup dependabot to help with maintaining package-lock.json -> https://github.com/elastic/apm-agent-nodejs/issues/2689
- [ ] Create a follow-up ticket, once we have a package-lock file, to work through https://github.com/elastic/apm-agent-nodejs/security/dependabot items.
- [x] See if we can use `npm ci` in some places for CI tests. `npm ci` isn't supported in the npm v5 that is part of our Node 8.6 runs. However, we *could* consider `npm install -g npm` for node v8 runs, or just fallback to `npm install` given this is about speeding up the install step.  Does TAV use `npm ci` if possible?
  *Moved this to https://github.com/elastic/apm-agent-nodejs/issues/2682 to do after we drop node v8 support.*
- [x] Use a supported version of npm for building release packages so we can be sure we get security fixes. Note that what npm versions get security fixes is not exactly clear: [Current npm docs](https://docs.npmjs.com/about-npm-versions) just say "use the latest version", which I suppose is an option, but I've generally be hesitant or didn't feel the need to use a version of npm other than the one that came as part of that *node* version install. An [old npm blog](https://blog.npmjs.org/post/162451604515/the-npm-clis-long-term-support-lts-policy.html) on LTS policy said:

    > In the event of a security issue, the npm CLI project will back port security patches to any version of npm currently shipping with a [supported Node.js version](https://github.com/nodejs/LTS#lts-schedule1), that is to say, any Node.js version still in its maintenance window.
    
[...]

PR #2627, opened 2022-03-28
chore: start using package-lock.json for repeatable Lambda Layer builds

trentm:
We should have repeatable (strict dep versions) builds for our Lambda
Layer released packages.

Closes: #2626 
 
### Checklist

- [ ] Implementation: there are a few task list items on #2626

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -307,6 +307,8 @@ pipeline {
               setEnvVar('RELEASE_NOTES_URL', getReleaseNotesUrl())
               deleteDir()
               unstash 'source'
+              // This should use a currently supported Node.js version to ensure
+              // and npm version that gets security patches is used.
               withNodeJSEnv(version: 'v14.17.5'){
                 dir("${BASE_DIR}"){
                   cmd(label: 'make dist', script: 'make -C .ci dist')
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # Files to ignore
 .DS_Store
 /.vscode
-/package-lock.json
 /test/babel/out.js
 /test/types/transpile/index.js
 /test/types/transpile-default/index.js
--- a/.npmrc
+++ b/.npmrc
@@ -1,5 +1,3 @@
-package-lock=false
-
 # Workaround unresolvable peerDependencies between express-graphql, graphql,
 # and apollo-server-express. npm v7 (included with node v15) makes these
 # peerDependencies issues an install error. Until the community catches up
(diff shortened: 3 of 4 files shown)


.ci/Jenkinsfile
```diff
@@ -17,7 +17,7 @@ pipeline {
     NOTIFY_TO = 'build-apm+apm-agent-nodejs@elastic.co'
     NPMRC_SECRET = 'secret/jenkins-ci/npmjs/elasticmachine'
     TOTP_SECRET = 'totp/code/npmjs-elasticmachine'
-    NODE_VERSION = 'v14.17.5'
+    BUILD_NODE_VERSION = 'v16.15.0'
```
trentm, line 20: REVIEW NOTE: 
1. Switch to v16 for npm v8 which has `lockfileVersion: 2`.
2. Rename to `BUILD_NODE_VERSION` to differentiate from other unrelated usages of `NODE_VERSION` in this and other files in the repo.

.ci/Jenkinsfile
```diff
@@ -383,7 +383,7 @@ pipeline {
       environment {
         HOME = "${env.WORKSPACE}"
         RESULT_FILE = 'apm-agent-benchmark-results.json'
-        NODE_VERSION = '14'
+        BENCH_NODE_VERSION = '14'
```
trentm, line 386: REVIEW NOTE: Rename to `BENCH_NODE_VERSION` to differentiate from other unrelated usages of `NODE_VERSION` in this and other files in the repo.

.ci/Makefile
```diff
@@ -1,23 +1,21 @@
...
-	rm -rf $(AWS_FOLDER) || true
-	mkdir -p $(AWS_FOLDER)
-	cd $(AWS_FOLDER); \
-		npm init -y; \
-		npm install --global-style https://github.com/elastic/apm-agent-nodejs#$(BRANCH_NAME); \
-		mkdir nodejs; \
-		mv node_modules nodejs
+.PHONY: dist
+dist: validate-branch-name
+	../dev-utils/make-lambda-layer-zip.sh
```
trentm, line 10: REVIEW NOTE: Move the authoritative logic for building the lambda layer zip to a separate script that can be used separate from CI (useful for local dev as well), and to not have the same logic in two places.

trentm:
I tested the created lambda layer with my play/dev Lambda briefly.

astorm (approved):
No objection to this in principle.  Using Node 16 makes sense to me and having the logic for doing build extracted from the CI Makefile so it can be used in different contexts seems like a reasonable step.  

Approving, but my understanding of the `.ci` folder is very black box so I can't really comment on the more ground level changes made there -- so file this review under caveat emptor.

PR #2629, opened 2022-03-30
🚧   test pre-docker upgrade

v1v:
🚧 

test pre-docker upgrade


v1v:
I'll close this now, 2 times out of 16 got stuck and only one failed.

image


image

I don't know how flaky the project is but those figures could help to understand that three times per month we will see those failures if we use the above average failures.

PR #2631, opened 2022-03-31
feat: inject/propagate headers api for context

astorm:
Implements https://github.com/elastic/apm-agent-nodejs/issues/2602

Seeking early feedback on this draft PR.  

This is the simplest version of a new method on our TraceContext object that gives it the ability to inject/propgate headers into a carrier object _and_ set a boolean such that we can check later if the span's had headers propagated.  The end goal is to create that `boolean` so we can implement a spec compliant version of the "`isCompressionEligable`" method.

In addition to the usual feedback free for all -- there's two specific bits of feedback I'm soliciting. 

- @felixbarny you had previously suggested this sort of trace context API during the initial phases of span compression.  Is this what you had in mind, or was there more you think needed to happen here? (asking in part I saw the java agent has the concept of a "non-discardable" span and I'm not familiar with everything that entails)

- Is the verb we want to use here _inject_, _propagate_, or something else?


trentm (changes requested):

lib/instrumentation/http-shared.js
```diff
@@ -152,21 +153,15 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
...
-        if (agent._conf.useElasticTraceparentHeader) {
-          newHeaders['elastic-apm-traceparent'] = headerValue
-        }
-        options.headers = newHeaders
+      if (parent && parent._context && span) {
+        options.headers = span.propagateTraceContextHeaders(options.headers, parent._context)
+      } else if (parent && parent._context && !span) {
+        // headers still need to propagate
+        options.headers = propagateTraceContextHeaders(options.headers, parent._context)
+      }
```
felixbarny, line 161: This is the kind of API I'd suggest:
```suggestion
      if (parent) {
         // uses the the TraceContext from parent._context
        parent.propagateTraceContextHeaders(
          // this argument (may be named `carrier`) is passed down to the header manipulating function
          req,
          // allows to pass in a technology specific function that handles the act of adding headers to a "header carrier" object
          // In the case of http, this will be the http request object
          // For other technologies, like SQS, it will be something more specific
          // This has the benefit that we don't need to allocate an intermediate map that holds the headers,
          // instead, we can directly set the headers on the request.
          // Note that we don't have to know about the names of the headers being added. Only the `TraceContext` class knows about the header names.
          // The thing that `TraceContext` doesn't have to know about is how to add a header for a specific technology.
          // Because the function receives the request object as a parameter, it allows us to store this function as a constant, avoiding having to allocate an anonymous function every time
          (req, key, value) => req.headers[key] = value
        )
      }
```

lib/instrumentation/http-shared.js
```diff
@@ -152,21 +153,15 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
       // If so, then fallback to the current run context's span or transaction,
       // if any.
       var parent = span || parentRunContext.currSpan() || parentRunContext.currTransaction()
```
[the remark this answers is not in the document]
astorm: Good to note this -- if it keeps happening we'll consider that.

lib/instrumentation/http-shared.js
astorm: That makes sense. Just to say it back, the semantics for "propagating" should change from "add the headers from this random context to an object" to "add _my_ headers to this object".  Also, rather than directly setting header values the new method will accept a callback so that the caller can define how headers are set.

Re: that callback -- I'll definitely build it into the API _but_ we'll still be allocating a new object for request headers.  We've [run into issues](https://github.com/elastic/apm-agent-nodejs/commit/5170e0f9ec4c40a02772a9c7d4fe291692c6516f) where modifying the headers object in place causes issues with AWS v4 signatures.

lib/tracecontext/index.js
```diff
@@ -63,8 +65,32 @@ class TraceContext {
...
+
+  const newHeaders = Object.assign({}, currentHeaders)
+  newHeaders.traceparent = headerValue
+  if (traceStateValue) {
+    // Do not set the header to the empty string. Though RFC 7230 allows
+    // empty header values, some servers apparently do not.
+    newHeaders.tracestate = traceStateValue
+  }
+
+  return newHeaders
```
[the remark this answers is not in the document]
astorm: Felix's suggestion above will take care of this. The propagate method will get a callback args that will allow the caller to define how headers are set which will, effectively, move this logic.

lib/instrumentation/http-shared.js
```diff
@@ -152,23 +153,25 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
...
-        options.headers = newHeaders
       }
-
+      const newHeaders = Object.assign({}, options.headers)
+      if (parent && parent._context && span) {
+        parent.propagateTraceContextHeaders(newHeaders, setter)
+      } else if (parent && parent._context && !span) {
+        // headers still need to propagate
+        propagateTraceContextHeaders(newHeaders, parent._context, setter)
+      }
```
felixbarny, line 173: Isn't the else branch redundant? The `span` variable also gets assigned to the `parent` variable, if present.
```suggestion
      if (parent) {
        parent.propagateTraceContextHeaders(newHeaders, setter)
      }
```

lib/instrumentation/http-shared.js
```diff
@@ -152,23 +153,25 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
...
-        newHeaders.traceparent = headerValue
-        if (traceStateValue) {
-          // Do not set the header to the empty string. Though RFC 7230 allows
-          // empty header values, some servers apparently do not.
-          newHeaders.tracestate = traceStateValue
+
+      const setter = function (carrier, name, value) {
+        if (!value) {
+          return
         }
```
felixbarny, line 160: Instead of checking for empty values in the setter, can TraceContext guarantee to only call the setter for non-empty values? This makes the life of all setters easier.

lib/instrumentation/http-shared.js
```diff
@@ -152,23 +153,25 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
...
-        if (agent._conf.useElasticTraceparentHeader) {
-          newHeaders['elastic-apm-traceparent'] = headerValue
+        carrier[name] = value
+
+        if (agent._conf.useElasticTraceparentHeader && name === 'traceparent') {
+          carrier['elastic-apm-traceparent'] = value
         }
```
felixbarny, line 165: The setter should be called by TraceContext for each header it wants to set. So instead of hard-coding knowledge about the header names in each setter, can this if block be part of TraceContext?

lib/tracecontext/index.js
```diff
@@ -63,8 +63,17 @@ class TraceContext {
...
+
+  propagateTraceContextHeaders (carrier, contextFrom, setter) {
+    return propagateTraceContextHeaders(carrier, contextFrom, setter)
+  }
+}
+
+function propagateTraceContextHeaders (carrier, contextFrom, setter) {
+  setter(carrier, 'traceparent', contextFrom.toTraceParentString())
+  setter(carrier, 'tracestate', contextFrom.toTraceStateString())
 }
```
felixbarny, line 75: This encapsulates the knowledge of the header names to this class and only calls setters if the values are non-empty so that the setters don't have to check for that.

I'm not sure how TraceState can access `agent._conf.useElasticTraceparentHeader`. Probably best to pass in `agent._conf.useElasticTraceparentHeader` as a constructor argument.

```suggestion
  propagateTraceContextHeaders (carrier, contextFrom, setter) {
    const traceparent = contextFrom.toTraceParentString()
    if (traceparent) {
      setter(carrier, 'traceparent', traceparent)
    }
    if (agent._conf.useElasticTraceparentHeader) {
      setter(carrier, 'elastic-apm-traceparent', traceparent)
    }
    const tracestate = contextFrom.toTraceStateString()
    if (tracestate) {
      setter(carrier, 'tracestate', contextFrom.toTraceStateString())
    }
  }
}
```

lib/instrumentation/http-shared.js
felixbarny: > I'll definitely build it into the API but we'll still be allocating a new object for request headers. 

As APIs are harder to change and can limit the efficiency of all implementations, I think getting the API right and efficient is definitely more important than optimizing all implementations right now.

lib/instrumentation/http-shared.js
astorm: Seems reasonable -- done in 587e1d4395071a34f0604d22300e9be790a52e3c.

lib/instrumentation/http-shared.js
astorm: Seems reasonable -- (I think I avoided doing this because I tend to think of this specific header as a legacy thing that we'll eventually not want to carry forward _and_ it can sometimes be tricky for us to get the agent conf injected into the place we need to check a config) done in 7cbbec6c64f2db5393c46f83c20dcd7373d689ef

lib/instrumentation/http-shared.js
astorm: cbd1cd80fe46cc43c0fd89828a20e72874cc4d08

astorm: Hmm -- I guess with the current semantics that branch is redundant and the tests seem happy with the change.  Getting rid of it also means we can avoid exporting that logic as a stand alone function.  

Also, not explicitly mentioned, but re: the `_context` check you removed we've been cargo culting that `parent._context` check along for a while now, but it seems reasonable to remove it (as along as `currSpan` or `currTransaction` keep returning something that has GenericSpan as a base object)

Done in cbd1cd80fe46cc43c0fd89828a20e72874cc4d08

lib/tracecontext/index.js
```diff
@@ -6,6 +6,8 @@ class TraceContext {
   constructor (traceparent, tracestate) {
     this.traceparent = traceparent
     this.tracestate = tracestate
+
+    this._hasPropagatedTraceContextHeaders = false
```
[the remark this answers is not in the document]
astorm: Moved this to the GenericSpan

lib/instrumentation/http-shared.js
```diff
@@ -152,21 +153,15 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
...
+      if (parent && parent._context && span) {
+        options.headers = span.propagateTraceContextHeaders(options.headers, parent._context)
+      } else if (parent && parent._context && !span) {
+        // headers still need to propagate
+        options.headers = propagateTraceContextHeaders(options.headers, parent._context)
+      }
+
+      if (agent._conf.useElasticTraceparentHeader && options.headers.traceparent) {
+        options.headers['elastic-apm-traceparent'] = options.headers.traceparent
       }
```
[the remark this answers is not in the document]
astorm: Done in multiple commits based on other feedback -- elastic-apm-traceparent will now be sent to the setter function and the logic for whether to do this or not lives in the `TraceContext` object (where we're not storing a reference to the agent configuration, per Felix's suggestion.

lib/tracecontext/index.js
```diff
@@ -63,6 +64,22 @@ class TraceContext {
   toTraceParentString () {
     return this.traceparent.toString()
   }
+
+  propagateTraceContextHeaders (carrier, contextFrom, setter) {
```
felixbarny, line 68: The methods [`toTraceParentString()`](https://github.com/elastic/node-traceparent/blob/4d97700fca219dfde202c0bb7a839f7792cb8b7a/index.js#L147-L149) and [`contextFrom.toTraceStateString()`](https://github.com/elastic/apm-agent-nodejs/blob/f7e34a62771d192f7bdb138224beb6fcba6743d2/lib/tracecontext/tracestate.js#L210-L227) appear to be rather expensive. Probably better to store them in a local variable than calling them several times.

test/tracecontext/tracecontext.test.js
```diff
@@ -0,0 +1,126 @@
...
+    const headers = {}
+
+    const newHeaders = Object.assign({}, headers)
+    context.propagateTraceContextHeaders(
+      newHeaders,
+      fromContext,
+      function (carrier, name, value) {
+        if (!value) {
+          return
+        }
```
felixbarny, line 62: context.propagateTraceContextHeaders should always make sure that the value is not empty
```suggestion
```

felixbarny (approved):
After caching the traceparent and tracestate strings, LGTM!

lib/tracecontext/index.js
astorm: Feels a little like a premature optimization but it's easy enough to do.  Done in 05ed2ed6790b96705490a86a63461713d688fc6c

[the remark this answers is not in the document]
astorm: Fair enough since it's easy enough to do.  Done in 006cb5f5158e2749ff0730f3a3656e1645399717

test/tracecontext/tracecontext.test.js
astorm: Seems harmless in a test, but also harmless enough to fix.  I've done in in a separate commit -- the setter callbacks no longer have this guard.

trentm (approved):

trentm (changes requested):
Actually, this change is missing from the earlier creation of test/tracecontext to have it picked up by the test runner:

```diff
diff --git a/test/test.js b/test/test.js
index de4edee5..fe702518 100644
--- a/test/test.js
+++ b/test/test.js
@@ -176,6 +176,7 @@ var directories = [
   'test/sanitize-field-names',
   'test/sourcemaps',
   'test/stacktraces',
+  'test/tracecontext',
   'test/uncaught-exceptions'
 ]

```

astorm:
> Actually, this change is missing from the earlier creation of test/tracecontext to have it picked up by the test runner:

well that's annoying.  Fixed -- also PR incoming to build that list of test folders dynamically.

trentm:
@astorm 

> also PR incoming to build that list of test folders dynamically.

Cool. If/when we convert to node-tap, this'll become a non-thing.

Perhaps use the "glob" package:

```
> var glob = require('glob')
> glob.sync('test/**/*.test.js')
[
  'test/agent.test.js',
  'test/central-config-disabled.test.js',
  'test/central-config-enabled.test.js',
  ...
  
```

trentm (approved):

PR #2653, opened 2022-04-20
feat: Allow starting a span on a transaction that has already ended

trentm:
This was a restriction that dated back to the v1 APM Server intake
API that required all a transaction's spans to be sent along with the
transaction.

The Java APM agent, FWIW, allows this.  

This came up in discussion about some edge cases with the OTel Bridge. As we move closer to OTel semantics, the distinction between transactions and spans gets to be less. Currently can start a child span of a span that has already ended, so it is nice to be able to do the same for a transaction.

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,6 +43,9 @@ Notes:
   child spans from being added to the exit spans.  See issue for a full list of
   spans types that will be treated as exit spans. ({issues}2601[#2601])
 
+* Allow a new span to be created/started even if its transaction has ended.
+  This is expected to be a very rare use case. ({pull}2653[#2653])
+
 [float]
 ===== Bug fixes
 
--- a/lib/instrumentation/transaction.js
+++ b/lib/instrumentation/transaction.js
@@ -133,10 +133,6 @@ Transaction.prototype.createSpan = function (...args) {
   if (!this.sampled) {
     return null
   }
-  if (this.ended) {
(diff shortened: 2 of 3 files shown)


trentm:
Here is an example of a trace (in APM UI v7.15.0) that shows a span ("s6") that started after its transaction ended:

Screen Shot 2022-04-20 at 12 33 41 PM

This is from running script https://github.com/elastic/apm-agent-nodejs/blob/trentm/otel-bridge/test/opentelemetry-sdk/fixtures/start-span-with-context.js from the ongoing OTel Bridge work.

astorm (approved):
Seems reasonable -- especially if java does it. Also, the implementation is reasonable.  

Just to say it out loud (not a blocker) I am slightly curious what, if any, effect this will have on features that rely on the transaction's `span_count` -- it seems like it's possible for these late spans to not be counted (depending on whether an ended transaction's been reported).

A question for future us, I suppose :). Approving, presuming the current CI failures as a :jenkins: hiccup.

trentm:
> it seems like it's possible for these late spans to not be counted

Thanks for mentioning it. Yes. We'd discussed this in chat briefly. Felix's opinion was:

[Trent]
> > Do you think that `transaction.span_count.*` values being possibly less than the number of span documents will surprise/break any APM UI code?

[Felix]
> The Java agent already supports spans that end after their transaction. I don’t think it will break the UI. It shows a warning if it can find fewer spans than expected but not if it finds more than expected.

PR #2667, opened 2022-05-02
chore:Publish to npm on release

### LINKED ISSUE #2625: automate releases
Work with the robots team to get automated releases.

The current release process is documented here: https://github.com/elastic/apm-agent-nodejs/blob/main/CONTRIBUTING.md#releasing

#2585 added a "Release" stage for publishing Lambda Layers when a tag is pushed. 

- Is the release flow for the other agents based on a developer making/pushing a release tag? If not, what triggers an automatic release process?

pazone:
Fixes #2625 
However there is some space for improvements

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -17,6 +17,7 @@ pipeline {
     NOTIFY_TO = 'build-apm+apm-agent-nodejs@elastic.co'
     NPMRC_SECRET = 'secret/jenkins-ci/npmjs/elasticmachine'
     TOTP_SECRET = 'totp/code/npmjs-elasticmachine'
+    NODE_VERSION = 'v14.17.5'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
@@ -290,7 +291,7 @@ pipeline {
               setEnvVar('RELEASE_NOTES_URL', getReleaseNotesUrl())
               deleteDir()
               unstash 'source'
-              withNodeJSEnv(version: 'v14.17.5'){
+              withNodeJSEnv(version: "${env.NODE_VERSION}"){
                 dir("${BASE_DIR}"){
                   cmd(label: 'make dist', script: 'make -C .ci dist')
                 }
@@ -328,11 +329,11 @@ pipeline {
         }
         stage('Publish to npm') {
           steps {
-            withNodeJSEnv(version: 'v14.17.5') {
+            withNodeJSEnv(version: "${env.NODE_VERSION}") {
               withNpmrc(secret: "${env.NPMRC_SECRET}", path: "${env.WORKSPACE}/${env.BASE_DIR}") {
                 withTotpVault(secret: "${env.TOTP_SECRET}", code_var_name: 'TOTP_CODE') {
                   dir("${BASE_DIR}") {
-                    sh(label: 'npm-publish', script: 'npm publish --otp="${env.TOTP_CODE}"')
+                    sh(label: 'npm-publish', script: "npm publish --otp=${env.TOTP_CODE}")
                   }
                 }
               }
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -15,6 +15,8 @@ pipeline {
     RELEASE_URL_MESSAGE = "(<https://github.com/elastic/${env.REPO}/releases/tag/${env.TAG_NAME}|${env.TAG_NAME}>)"
     SLACK_CHANNEL = '#apm-agent-node'
     NOTIFY_TO = 'build-apm+apm-agent-nodejs@elastic.co'
+    NPMRC_SECRET = 'secret/jenkins-ci/npmjs/elasticmachine'
+    TOTP_SECRET = 'totp/code/npmjs-elasticmachine'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
@@ -315,7 +317,7 @@ pipeline {
             }
           }
         }
-        stage('Publish') {
+        stage('Publish to AWS') {
           steps {
             withGithubNotify(context: "Publish") {
               withGoEnv(){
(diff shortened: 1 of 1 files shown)


.ci/Jenkinsfile
```diff
@@ -343,6 +345,19 @@ pipeline {
             }
           }
         }
+        stage('Publish to npm') {
+          steps {
+            withNodeJSEnv(version: 'v14.17.5') {
+              withNpmrc(secret: "${env.NPMRC_SECRET}", path: "${env.WORKSPACE}/${env.BASE_DIR}") {
+                withTotpVault(secret: "${env.TOTP_SECRET}", code_var_name: 'TOTP_CODE') {
+                  dir("${BASE_DIR}") {
+                    sh(label: 'npm-publish', script: 'npm publish --otp ${TOTP_CODE}')
```
trentm, line 354: I notice when comparing to https://github.com/elastic/synthetics/blob/9019b2b03b91170afa2d54cea25d790f601ffe56/.ci/Jenkinsfile#L194-L196 that that uses `env.TOTP_CODE` rather than bare `TOTP_CODE`. Do you know if this needs to change to the former?

```suggestion
                    sh(label: 'npm-publish', script: 'npm publish --otp="${env.TOTP_CODE}"')
```

.ci/Jenkinsfile
```diff
@@ -343,6 +345,19 @@ pipeline {
             }
           }
         }
+        stage('Publish to npm') {
```
trentm, line 348: The `stage('Opbeans')` above indirectly depends on the package already having been published to npm because the [`bump-version.sh` script it is running does `npm install elastic-apm-node@$THE-PUBLISHED-VERSION`](https://github.com/elastic/opbeans-node/blob/f2cdbe3a68c1ba4ea5f10907bac549320218173f/.ci/bump-version.sh#L14). So I think this should move the 'Opbeans' stage to the bottom of this `stage('Release')` block.

.ci/Jenkinsfile
trentm, line 354: I opened https://github.com/elastic/apm-agent-nodejs/issues/2668 for a future concern with tags/releases for maintenance releases that are not for the current major version.

Do you think it would be worthwhile to change:

```
    stage('Release') {
      options { skipDefaultCheckout() }
      when {
        beforeAgent true
        tag pattern: 'v\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
```

to:

```
    stage('Release') {
      options { skipDefaultCheckout() }
      when {
        beforeAgent true
        tag pattern: 'v3\\.\\d+\\.\\d+', comparator: 'REGEXP'
```

to lock this "Release" stage down to new tags of the *current* major only?

If when we have a new 4.x major version and do occasional 3.x maint releases, we won't want all of these release stages to run, or at least not in the same way.

trentm (changes requested):

.ci/Jenkinsfile
v1v: That's a cosmetic preference, I like to use `env.VARIABLE` to help with what's the variable coming from, while `VARIABLE`, it might  happen from either a groovy variable or an environment variable. No strong opinion, but the context about what's the `env` map coming from.

v1v (approved):

v1v (commented):
changes LGTM, missed the one that Trent mentioned regarding the `opbeans` stage to be the last stage in the release to happen

.ci/Jenkinsfile
pazone: I will work as you expect. What steps you wish to exclude for maintenance releases?

.ci/Jenkinsfile
```diff
@@ -343,6 +345,19 @@ pipeline {
...
+          steps {
+            withNodeJSEnv(version: 'v14.17.5') {
```
cachedout, line 350: Nit: It might be a bit cleaner to set this as a variable at the top of the pipeline around L19 or so.

cachedout (approved):

trentm:
> it's a bit odd that the CLA bot didn't recognize that @pazone is a member of the Elastic org. @trentm have you ever seen this behavior before?

@cachedout No, I haven't. Do you know if/how we can retrigger the CLA check?

trentm:
cla/check

.ci/Jenkinsfile
trentm: I'd like to hear an opinion from @v1v (who initially added the 'tag pattern' usage) if restricting this to the just tags for the current major version sounds reasonable.

v1v: I wonder if this proposal should be done only when those potential release stages for the 3.x maint releases are in place? Otherwise, there is a chance we don't enable them and then `v3` will be hardcoded in the pipeline when `v4` is branched off.

v1v:
> cla/check

https://github.com/elastic/apm-agent-nodejs/pull/2667/commits/9c020d1f1fd6aff2e924706bacc595f0f9a4baa8 is not tracked to a GitHub user, so I'd say the email linked to the commit has not been configured to the GitHub user.

trentm:
> [9c020d1](https://github.com/elastic/apm-agent-nodejs/commit/9c020d1f1fd6aff2e924706bacc595f0f9a4baa8) is not tracked to a GitHub user ...

ah, yes, it was commited with `Pavel <pavel@Pavels-MacBook-Pro.local>`, so that isn't going to work. :)

pazone:
Fixed my git email. 💌

.ci/Jenkinsfile
trentm: @pazone There is still this to do.

.ci/Jenkinsfile
trentm: A question about Jenkinsfile usage: If/when we have a "3.x" maintenance branch and "main" is used for 4.x releases of this agent, the only relevant ".ci/Jenkinsfile" will be the one on "main", right? If, for example, I push a "v3.42.45" tag that points to a commit on the "3.x" branch, will it still be the ".ci/Jenkinsfile" on the **"main" branch** that is used by Jenkins to configure the build?

trentm: Thanks for clarifying my misunderstanding. That answers my concerns for this point on the `tag pattern` usage.

v1v (approved):

.ci/Jenkinsfile
pazone: moved to the end of release stage

pazone:
@trentm  I was need to revert your proposed change about env variable interpolation, if you don't mind. We gotta be consistent because as I can see string interpolation is used everywhere in the Jenkinsfile.

.ci/Jenkinsfile
pazone: done

trentm:
I'm going to merge this so I can try it out in a release I want to do later today.

PR #2672, opened 2022-05-03
test: incomplete usage of @cucumber/cucumber for testing gherkin-specs

trentm:
(This builds on OTel Bridge work coming in https://github.com/elastic/apm-agent-nodejs/pull/2641)

This change includes an *incomplete* start at using @cucumber/cucumber
to test shared gherkin-specs files from apm.git for APM agent testing.
This is meant to show the current state and then move on, because there
are issues here.

1. @cucumber/cucumber's usage of 'stack-chain' conflicts with the
   agent's usage of 'error-callsites':

    ```
    % ./node_modules/.bin/cucumber-js
    TypeError: Cannot redefine property: prepareStackTrace
        at Function.defineProperty (<anonymous>)
        at Object.<anonymous> (/Users/trentm/el/apm-agent-nodejs8/node_modules/error-callsites/index.js:11:8)
        at Module._compile (internal/modules/cjs/loader.js:1085:14)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
        at Module.load (internal/modules/cjs/loader.js:950:32)
        at Function.Module._load (internal/modules/cjs/loader.js:790:12)
        at Module.require (internal/modules/cjs/loader.js:974:19)
        at require (internal/modules/cjs/helpers.js:101:18)
        at Object.<anonymous> (/Users/trentm/el/apm-agent-nodejs8/lib/stacktraces.js:28:24)
        at Module._compile (internal/modules/cjs/loader.js:1085:14)
    ```

2. There is an issue where async context is not carried through
   cucumber's steps. This means that setting a current transaction
   (which in the APM agent is stored in async context) does not
   carry through as a current transaction for subsequent steps.
   This makes the "OTel span kind <kind> for spans & default span type &
   subtype" scenario in "otel_bridge.feature" impossible to test.

   My guess is this is an issue with @cucumber/cucumber code losing
   async context -- which isn't unheard of in libraries -- but I have
   not dug into it.

### DIFF
--- a/cucumber.js
+++ b/cucumber.js
@@ -1,0 +1,16 @@
+// https://github.com/cucumber/cucumber-js/blob/main/docs/configuration.md
+module.exports = {
+  default: {
+    paths: [
+      // We don't support all of the gherkin specs from apm.git, so list each
+      // one individually.
+      'test/cucumber/gherkin-specs/otel_bridge.feature'
+    ],
+    require: ['test/cucumber/*.js'],
+    publishQuiet: true,
+    backtrace: true, // Enable this to see stacktraces into cucumber library code.
+    // `parallel > 1` should be fine, but avoid it for now to avoid
+    // multi-process execution of tests. Linear should be more debuggable.
+    parallel: 1
+  }
+}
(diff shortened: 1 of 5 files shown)


trentm:
This PR was added as a starting point if we come back to attempt to using `@cucumber/cucumber` for testing.

```
$ ./node_modules/.bin/cucumber-js
...
```

Issue #2670, opened 2022-05-03
Disconnected HTTP2 sessions cause a crash when reporting transactions


chinthakagodawita:
**Describe the bug**

We're using the agent to instrument grpc calls. Since `@grpc/grpc-js` uses the internal Node HTTP/2 server, it's able to auto-instrument all incoming requests, we've added additional instrumentation information ourselves.

We've found that very intermittently, some requests are causing process crashes. We've had to disable APM as a result.

Error thrown:
```
Error [ERR_HTTP2_SOCKET_UNBOUND]: The socket has been disconnected from the Http2Session
    at new NodeError (node:internal/errors:371:5)
    at Object.get (node:internal/http2/core:867:17)
    at Object.getContextFromRequest (/opt/app/node_modules/elastic-apm-node/lib/parsers.js:24:32)
    at Transaction.toJSON (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:206:41)
    at Transaction._encode (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:237:15)
    at Instrumentation.addEndedTransaction (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/index.js:315:63)
    at Transaction.end (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/transaction.js:288:32)
    at ServerHttp2Stream.<anonymous> (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/modules/http2.js:80:17)
    at ServerHttp2Stream.f (/opt/app/node_modules/once/once.js:25:25)
    at ServerHttp2Stream.onend (/opt/app/node_modules/end-of-stream/index.js:36:27)
    at /opt/app/node_modules/elastic-apm-node/lib/instrumentation/run-context/AbstractRunContextManager.js:76:49
    at AsyncHooksRunContextManager.with (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/run-context/BasicRunContextManager.js:49:17)
    at ServerHttp2Stream.wrapper (/opt/app/node_modules/elastic-apm-node/lib/instrumentation/run-context/AbstractRunContextManager.js:76:23)
    at ServerHttp2Stream.emit (node:events:538:35)
    at ServerHttp2Stream.emit (node:domain:475:12)
    at endReadableNT (node:internal/streams/readable:1345:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
```

[...]

PR #2674, opened 2022-05-03
fix: possible crash in instrumentation of an incoming HTTP/2 request

trentm:
This also fixes a crash that was possible in instrumentation if an
HTTP/2 server called `stream.respond()` after the stream had been
destroyed (and the transaction ended):

    ../lib/instrumentation/modules/http2.js:149
          trans.res.statusCode = status
                               ^
    TypeError: Cannot set properties of null (setting 'statusCode')

Fixes: #2670

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/instrumentation/modules/http2.js
+++ b/lib/instrumentation/modules/http2.js
@@ -61,9 +61,15 @@ module.exports = function (http2, agent, { enabled }) {
         // core `http.IncomingMessage` and `http.ServerResponse` objects,
         // sufficient for `parsers.getContextFromRequest()` and
         // `parsers.getContextFromResponse()`, respectively.
+        // `remoteAddress` is fetched now, rather than at stream end, because
+        // this `stream.session.socket` is a proxy object that can throw
+        // `ERR_HTTP2_SOCKET_UNBOUND` if the Http2Session has been destroyed.
         trans.req = {
           headers,
-          socket: stream.session.socket,
+          // socket: stream.session.socket,
+          socket: {
+            remoteAddress: stream.session.socket.remoteAddress
+          },
           method: headers[':method'],
           url: headers[':path'],
           httpVersion: '2.0'
@@ -141,7 +147,7 @@ module.exports = function (http2, agent, { enabled }) {
 
   function updateHeaders (headers) {
     var trans = agent._instrumentation.currTransaction()
-    if (trans) {
+    if (trans && !trans.ended) {
       var status = headers[':status'] || 200
       trans.result = 'HTTP ' + status.toString()[0] + 'xx'
       trans.res.statusCode = status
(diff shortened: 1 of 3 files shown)


lib/instrumentation/modules/http2.js
```diff
@@ -141,7 +147,7 @@ module.exports = function (http2, agent, { enabled }) {
 
   function updateHeaders (headers) {
     var trans = agent._instrumentation.currTransaction()
-    if (trans) {
+    if (trans && !trans.ended) {
```
astorm, line 150: FYI -- I didn't notice any difference in the behavior of the test suite with the additional `trans.ended` check here.  Presuming its here to skip the extra work if the work isn't needed.

lib/instrumentation/modules/http2.js
```diff
@@ -61,9 +61,15 @@ module.exports = function (http2, agent, { enabled }) {
         // core `http.IncomingMessage` and `http.ServerResponse` objects,
         // sufficient for `parsers.getContextFromRequest()` and
         // `parsers.getContextFromResponse()`, respectively.
+        // `remoteAddress` is fetched now, rather than at stream end, because
+        // this `stream.session.socket` is a proxy object that can throw
+        // `ERR_HTTP2_SOCKET_UNBOUND` if the Http2Session has been destroyed.
         trans.req = {
           headers,
-          socket: stream.session.socket,
+          // socket: stream.session.socket,
```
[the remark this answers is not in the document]
trentm: Ah, duh. Thanks. Removed in commit b7d2d4f8

lib/instrumentation/modules/http2.js
trentm: This is a case similar to this guard lower down on `wrapEnd`:
https://github.com/elastic/apm-agent-nodejs/blob/f6c4ee8a978b6edfee31b74f2e410c1a86206b38/lib/instrumentation/modules/http2.js#L171-L177

There is actually a possible crash here:

```
/Users/trentm/el/apm-agent-nodejs8/lib/instrumentation/modules/http2.js:149
      trans.res.statusCode = status
                           ^

TypeError: Cannot set properties of null (setting 'statusCode')
    at updateHeaders (/Users/trentm/el/apm-agent-nodejs8/lib/instrumentation/modules/http2.js:149:28)
    at ServerHttp2Stream.<anonymous> (/Users/trentm/el/apm-agent-nodejs8/lib/instrumentation/modules/http2.js:159:7)
    at Timeout._onTimeout (/Users/trentm/el/apm-agent-nodejs8/play-http2-insecure-server.js:27:12)
```

...if the server calls stream.respond() *after* the stream has been destroyed (which ends the transaction, which nulls out `trans.res`).

However, I (lazily) didn't add a test case for this. It was an unrelated crasher I hit when initially attempting to repro the issue (#2670).

Issue #2302, opened 2021-08-25
[META 497] Collecting statistics about dropped spans


AlexanderWert:
As described in [this spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md), agents should collect statistics on dropped spans.

Meta issue: [![issue details](https://img.shields.io/endpoint?url=https%3A%2F%2Fgiss.app.elstc.co%2Fapi%2Fstatus%2Felastic%2Fapm%2F497)](https://github.com/elastic/apm/issues/497)

astorm:
Looking at [the spec](https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md) and using the java agent's implementation as a reference, it looks like span statistics make the assumption that even a span that's not kept still exists for the lifetime of a transaction.  Specifically, the statistics include data that's only avaiable at span end (the `duration.sum.us` property).  The Node Agent's current implementation doesn't share this assumption.

[...]

PR #2694, opened 2022-05-15
feat: span statistics -- ensure dropped span objects are still created

astorm:
Part of  https://github.com/elastic/apm-agent-nodejs/issues/2302 (2 of 3): 

This intent of this PR is to ensure that in cases when a span will be dropped that the span is still created (matching the behavior of the java agent).   This is necessary for the span statistics work, as the collected statistics includes capturing the sum of time for dropped spans -- meaning that even a dropped span needs to start and end. 

Prior to this work the node agent "dropping" a span meant that the span would not be created. Now, the span object is created but the span's context object has its recorded property set to false.  Non-recorded spans are then excluded from sending, but otherwise have a full start/end lifecycle. 

The next (and final) step after this PR is to implement the logic for capturing the dropped span statistics and sending them on to APM Server. 

### Checklist

- [x] Implement code
- [x] Edit tests
- ~[ ] Update TypeScript typings~
- ~[ ] Update documentation~
- ~[ ] Add CHANGELOG.asciidoc entry~


lib/instrumentation/transaction.js
```diff
@@ -153,8 +149,15 @@ Transaction.prototype.createSpan = function (...args) {
     return null
   }
 
+  const span = new Span(this, ...args, opts)
+
+  if (this._builtSpans >= this._agent._conf.transactionMaxSpans) {
```
astorm, line 154: Note: The meaning of `_buildSpans` ends up changing slightly in this new world.  The `_buildSpans` property is _always_ incremented regardless of whether a span will be dropped or not.  To get the number of started spans we subtract the number of dropped spans from built spans.

test/opentelemetry-bridge/fixtures.test.js
```diff
@@ -162,15 +162,11 @@ const cases = [
   {
```
astorm, line 162: Note: I spent a day poking at the OTel Bridge work behind this test and I don't quite understand the reasoning behind when a span vs. a transaction is created here or the odd nesting -- but `createSpan` changing to always returning spans seems to have had a dramatic effect on the objects created.  I've modified the test to capture the current behavior, but if there's a better/different thing we should be doing here I'm open to hearing it.

lib/instrumentation/transaction.js
```diff
@@ -172,7 +175,7 @@ Transaction.prototype.toJSON = function () {
     sampled: this.sampled,
     context: undefined,
     span_count: {
-      started: this._builtSpans
+      started: this._builtSpans - this._droppedSpans
```
astorm, line 178: The `_builtSpans` property now tracks all span objects that were created.  To se the number of spans started we subtract the number dropped. (see above)

lib/opentelemetry-bridge/OTelTracer.js
```diff
@@ -74,7 +74,7 @@ class OTelTracer {
       }
       newTransOrSpan = trans.createSpan(name, spanOpts)
 
-      // There might be no span, e.g. if transactionMaxSpans was hit. We have
+      // There might be no span, e.g. if the span is an exit span. We have
```
astorm, line 77: Note: transactionMaxSpans being reached now creates a span -- but sets its recorded value to false.

astorm:
jenkins run the tests please

test/opentelemetry-bridge/fixtures.test.js
trentm: This took a while to see what was going on here. It has to do with what the `sampled` flag should be in the `traceparent` header of an outgoing HTTP request that is part of a transaction that *has hit `transaction_max_spans`*.  This PR changes that behaviour in the agent. It isn't specific to the OTel Bridge, it is just that this is the only test case in the agent that tested (a) hitting transaction max spans *and* (b) testing distributed trace headers.

Let's take the OTel Bridge out of it and use this "play-hit-trans-max-spans.js":

```js
  serviceName: 'play-hit-trans-max-spans',
  usePathAsTransactionName: true
  // transactionMaxSpans: 3
})
const http = require('http')

const server = http.createServer(function onRequest (req, res) {
  console.log('server request: %s %s %s', req.method, req.url, req.headers)
  req.resume()
  req.on('end', function () {
    const resBody = JSON.stringify({ ping: 'pong' })
    res.writeHead(200, {
      'content-type': 'application/json',
      'content-length': Buffer.byteLength(resBody)
    })
    res.end(resBody)
  })
})

async function makeAClientRequest (i, port) {
  console.log('\n-- make client request %d', i)
  return new Promise(resolve => {
    http.get({
      host: 'localhost',
      port: port,
      path: `/ping-${i}`
    }, (cRes) => {
      console.log('client response status:', cRes.statusCode)
      console.log('client response headers:', cRes.headers)
      const body = []
      cRes.on('data', (chunk) => body.push(chunk))
      cRes.on('end', () => {
        console.log('client response body:', body.toString())
        // s.end()
        resolve()
      })
    })
  })
}

server.listen(async () => {
  const port = server.address().port
  const myTrans = apm.startTransaction('myTrans')
  for (let i = 0; i < 5; i++) {
    await makeAClientRequest(i, port)
  }
  myTrans.end()
  server.close()
})
```

This is a script that runs:
- Runs an HTTP server that always responds with HTTP 200, body=`{"ping":"pong"}`.
- Starts a transaction and makes 5 sequential HTTP client requests to that HTTP server.

# first run: do not hit `transaction_max_spans`

If run as is, we expect and get this trace:

```
    trace c4c5af
    `- transaction e4a2ef "myTrans"
       `- span 57adde "GET localhost:65305"     <-- the HTTP client request
         `- transaction 840699 "GET /ping-0"    <-- the HTTP server incoming request
       `- span 3c452f "GET localhost:65305"
         `- transaction 36a19b "GET /ping-1"
       `- span 997321 "GET localhost:65305"
         `- transaction 1c75d5 "GET /ping-2"
       `- span 5e36d9 "GET localhost:65305"
         `- transaction c79494 "GET /ping-3"
       `- span 018830 "GET localhost:65305"
         `- transaction 442565 "GET /ping-4"
```

This shows transactions as children of spans to show a distributed trace. For
example that `transaction 840699 "GET /ping-0"` is an incoming HTTP request
with this header: `traceparent: 00-c4c5af3aaeca172931ae1a1f11a4d604-57adde3b7722ac14-01`.

# second run: use "main" branch, hit `transactionMaxSpans=3`

Using the current main branch code (i.e. *not* this PR), re-add the
`transactionMaxSpans: 3` config and re-run. We get this trace:

```
    trace 00a45a
    `- transaction b77fa0 "myTrans"
       `- span 67a683 "GET localhost:65349"    // 1
         `- transaction c23cb1 "GET /ping-0"
       `- span 8c942d "GET localhost:65349"    // 2
         `- transaction 93cc5f "GET /ping-1"
       `- span 325e1d "GET localhost:65349"    // 3, no more spans after this one
         `- transaction f605d3 "GET /ping-2"
       `- transaction 91d6b8 "GET /ping-3"
       `- transaction 3bdb85 "GET /ping-4"
```

When the agent starts the span for the 4th HTTP request, it hits `transactionMaxSpans`, and returns null.
When the `traceparent` is being set for the outgoing HTTP request, it finds the parent `transaction b77fa0 "myTrans"`, and uses *that* trace-context. In this example the `transaction 91d6b8 "GET /ping-3"` gets this header: `traceparent: '00-00a45a24ed082e230d1206816748b90e-b77fa0011efc4804-01`.

# third run: use this PR, hit `transactionMaxSpans=3`

Now run with your branch with `transactionMaxSpans: 3`. We get this trace:

```
    trace a94e03
    `- transaction 3a989b "myTrans" (34.346ms, outcome=unknown)
       `- span 3f2c28 "GET localhost:65384" (19.679ms, http)
         `- transaction df6a4b "GET /ping-0" (4.21ms, parent_id=3f2c285aa0d280f8)
       `- span 15a344 "GET localhost:65384" (3.609ms, http)
         `- transaction 09149b "GET /ping-1" (0.9ms, parent_id=15a34418bc16b76e)
       `- span 725234 "GET localhost:65384" (2.48ms, http)
         `- transaction 80aec1 "GET /ping-2" (0.549ms, parent_id=72523430c9b04778)
```

This time, when the agent starts the span for the 4th HTTP request, a span is returned with `isRecording()` false. When the `traceparent` is set for the outgoing HTTP request that span's trace-context is used, and the HTTP server request sees this header: `traceparent: '00-a94e03dfa502523f5519c97f83c0d062-d0499b244f3a5e74-00'`. **Notice that the sampled flag if the traceparent header is unset.**

The change in behaviour is that outgoing traceparent headers on transactions that have hit `transaction_max_spans`, have sampled=false. I.e. it stop sampling on the distributed trace. Perhaps this is a good thing, but it is a change in behaviour. I'd like to know what the other agents do.

trentm (commented):
This looks mostly good.

- I think it is worth discussing the behaviour change with other agent teams.
- Then if we are good with that change, I'd like to update or remove the OTel Bridge "hit-transaction-max-spans.js" test case, because it will no longer be testing what it was meant to be testing.

I think it might be worth creating a separate ticket to investigate the performance impact of this change when the agent does hit `transactionMaxSpans`. (Or perhaps we just stay aware of it, if it comes up.) For example, `transactionMaxSpans` is no longer any help in reducing the CPU load of calculating breakdown metrics, or of `captureSpanStackTraces` if that happens to be enabled (hopefully not).

astorm:
jenkins run the tests please

test/opentelemetry-bridge/fixtures.test.js
trentm: From discussion in the team meeting today my understanding is:
- we want the sampled flag to be false in this case (i.e. the behaviour of your PR here, so that is good); and
- the behaviour of what 'parent-id' should be propagated in the non-recording span is not specified for our APM agents and likely differs. Therefore it isn't something to block this PR.

Now that hitting `transactionMaxSpans` no longer triggers the code path where the OTelTracer needs to cope with `transaction.createSpan()` returning null, I've replaced this test case with a new "createSpan-returns-null.js" fixture test case in commit e0825e50fd9be0f3305251d3a69414e106f0a4a4.

lib/instrumentation/index.js
```diff
@@ -385,6 +385,10 @@ Instrumentation.prototype.addEndedSpan = function (span) {
 }
 
 Instrumentation.prototype._encodeAndSendSpan = function (span) {
+  if (!span.isRecorded()) {
+    return
+  }
+
```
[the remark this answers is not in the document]
astorm: 152df0c05eca23b0e21f2a64566bf710ab0066fc

trentm (approved):

PR #2698, opened 2022-05-17
chore: deprecate 'hapi' package instrumentation

trentm:
The old 'hapi' package was itself deprecated about 2 years ago.  The APM
agent now deprecates instrumenting it. The instrumentation code remains,
but it is no longer tested and will be removed in the next major.

Note that '@hapi/hapi' is still supported, this is just about the old
package name and versions. Any versions starting with 17.9.0, 18.2.0,
19 and later are the newer '@hapi/hapi'.


trentm:
run module tests for hapi,@hapi/hapi

astorm (approved):
Appears to catch all instances of the old hap tests.  Approving.

PR #2707, opened 2022-05-19
feat: dropped span stats

astorm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/2302 (part 3 of 3)

Dependent on https://github.com/elastic/apm-agent-nodejs/pull/2694

This PR creates a class/object for collecting statistic on the dropped spans, uses that new object to collect statistics on non-sampled spans and ensures this data is serialized/encoded along with the transaction object.  

The implementation is based on [the same implementation in the java agent](https://github.com/elastic/apm-agent-java/blob/main/apm-agent-core/src/main/java/co/elastic/apm/agent/impl/transaction/DroppedSpanStats.java).  

Spec: https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md

### Checklist

- [x] Implement code
- [x] Add tests
- ~[ ] Update TypeScript typings~
- ~[ ] Update documentation~


lib/instrumentation/dropped-span-stats.js
```diff
@@ -0,0 +1,52 @@
...
+    const resource = span && span._destination && span._destination.service && span._destination.service.resource
+    if (!resource || !span.outcome || this.statsMap.size === 128) {
+      return
+    }
+
+    const stats = this.getOrCreateStats(resource, span.outcome)
+
+    stats.duration.count++
+    stats.duration.sum.us += (span._duration * 1000)
+    return true
```
trentm, line 17: Why return a true/false for this `captureDroppedSpan()`? (FWIW the Java impl doesn't.)

lib/instrumentation/index.js
```diff
@@ -356,6 +356,10 @@ Instrumentation.prototype.addEndedSpan = function (span) {
     return
   }
 
+  if (!span.isRecorded()) {
+    return
```
trentm, line 360:
```suggestion
    span.transaction.captureDroppedSpan(span)
    return
```

Otherwise dropped span stats won't ever be collected.

test/instrumentation/dropped-span-stats.js
```diff
@@ -0,0 +1,133 @@
+'use strict'
```
trentm, line 1: This filename needs to be "*.test.js".

test/instrumentation/dropped-span-stats.js
```diff
@@ -0,0 +1,133 @@
...
+    span.setOutcome(OUTCOME_FAILURE)
+    span.end()
+
+    test.ok(!transaction.captureDroppedSpan(span))
+    transaction.end()
+    test.equals(transaction._droppedSpanStats.statsMap.size, 128)
+    test.end()
+  })
+
+  suite.end()
```
trentm, line 132: It would be good to have a test case for `transaction.captureDroppedSpan(span)` automatically being called. So, a case where `transactionMaxSpans` is hit and the resulting serialized transaction includes the expected `dropped_spans_stats: [...]`.   A quick place to do this might be in this existing transaction.test.js case that exercises `transactionMaxSpans`: https://github.com/elastic/apm-agent-nodejs/blob/main/test/instrumentation/transaction.test.js#L493-L529

trentm (changes requested):

lib/instrumentation/index.js
astorm: ha -- guess I didn't hit ⌘s before committing that one.

lib/instrumentation/dropped-span-stats.js
```diff
@@ -0,0 +1,52 @@
+'use strict'
+class DroppedSpanStats {
+  constructor () {
+    this.statsMap = new Map()
+  }
+
+  captureDroppedSpan (span) {
+    const resource = span && span._destination && span._destination.service && span._destination.service.resource
+    if (!resource || !span.outcome || this.statsMap.size === 128) {
```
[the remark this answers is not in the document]
astorm: Well -- with what we have I don't see catastrophic problems, and I presume minor problems will work themselves out after we release and we can adjust.  But I'm happy to make a change here if you think it's necessary. What change are you asking for here @trentm?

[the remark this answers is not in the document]
astorm: a3a823f107fe5de2afbe89a6dbdde06fd69383f8

lib/instrumentation/dropped-span-stats.js
astorm: Sends a signal as to whether the operation succeeded in order to enable test driven development.

lib/instrumentation/transaction.js
```diff
@@ -225,10 +230,11 @@ Transaction.prototype.toJSON = function () {
     sampled: this.sampled,
     context: undefined,
     span_count: {
-      started: this._builtSpans
+      started: this._builtSpans - this._droppedSpans
     },
     outcome: this.outcome,
-    faas: this._faas
+    faas: this._faas,
+    dropped_spans_stats: this._droppedSpanStats.encode()
```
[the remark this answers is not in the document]
astorm: 2b341240083c554aacd108673110912ef895a10c

test/instrumentation/dropped-span-stats.js
astorm: 9a2d1a3a02e7b6469f182b204af2f193fa7b1d48

test/instrumentation/dropped-span-stats.js
astorm: 8437013896f655dbbe973c3fddbcfb54b3f29a10

trentm: I don't see where that commit adds a test case for this.

lib/instrumentation/dropped-span-stats.js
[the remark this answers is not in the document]
astorm: 33b5ebc0c08991ca246ac761839255140778ba1d

test/instrumentation/dropped-span-stats.js
astorm: New test code is here: https://github.com/elastic/apm-agent-nodejs/commit/8437013896f655dbbe973c3fddbcfb54b3f29a10#diff-115ec49fff221477451b17e1fca2a5208448489ccbb84c4eb7b53c6c4016f3caR536

We ensure that the serialized payload contains data for dropped_spans_stats.

```
    t.equals(payload.dropped_spans_stats.length, 1)
```

If this isn't what you wanted to see please let me know what would make this an acceptable test to you.

trentm:
> [astorm](https://github.com/astorm) requested a review from [trentm](https://github.com/trentm) 2 days ago

Was there another change since my last review? If so, I don't see it.

test/instrumentation/dropped-span-stats.js
trentm: Thanks. I had been trying to find it in the [dropped-span-stats.test.js](https://github.com/elastic/apm-agent-nodejs/commit/8437013896f655dbbe973c3fddbcfb54b3f29a10#diff-bb4011f7f490e34ded1fbef4a570aedc94fd1a3c2bb563791678f413ca0c4927) file part of the diff.

lib/instrumentation/transaction.js
```diff
@@ -280,6 +282,10 @@ Transaction.prototype.toJSON = function () {
     payload.links = this._links
   }
 
+  // dropped_spans_stats: this._droppedSpanStats.encode()
```
trentm, line 285: nit: left over dev/debugging comment?

lib/instrumentation/dropped-span-stats.js
```diff
@@ -0,0 +1,52 @@
+'use strict'
+const LIMIT_STATS = 128
+class DroppedSpanStats {
+  constructor () {
+    this.statsMap = new Map()
+  }
+
+  captureDroppedSpan (span) {
+    const resource = span && span._destination && span._destination.service && span._destination.service.resource
+    if (!resource || !span._exitSpan || this.statsMap.size >= LIMIT_STATS) {
```
trentm, line 10: Sorry for this to come late in the review process. I think that checking the `statsMap.size` against the limit here is in the wrong place. If we hit the 128 size limit, with the current code we will stop incrementing `stats.duration.*` values *for existing entries*.

trentm (changes requested):

trentm (approved):
LGTM.

PR #2684, opened 2022-05-10
feat: use require.main to search for package.json for serviceName/Version

trentm:
Closes: #2420

This also drops the `read-pkg-up` dep that was being used for this. It was big:

```
% npm install read-pkg-up
...
added 39 packages, and audited 40 packages in 1s
...

% du -sh node_modules
1.6M	node_modules

% npm ls -a
b@1.0.0 /Users/trentm/tmp/b
└─┬ read-pkg-up@9.1.0
  ├─┬ find-up@6.3.0
  │ ├─┬ locate-path@7.1.0
  │ │ └─┬ p-locate@6.0.0
  │ │   └─┬ p-limit@4.0.0
  │ │     └── yocto-queue@1.0.0
  │ └── path-exists@5.0.0
  ├─┬ read-pkg@7.1.0
  │ ├── @types/normalize-package-data@2.4.1
  │ ├─┬ normalize-package-data@3.0.3
  │ │ ├─┬ hosted-git-info@4.1.0
  │ │ │ └─┬ lru-cache@6.0.0
  │ │ │   └── yallist@4.0.0
  │ │ ├─┬ is-core-module@2.9.0
  │ │ │ └─┬ has@1.0.3
  │ │ │   └── function-bind@1.1.1
  │ │ ├─┬ semver@7.3.7
  │ │ │ └── lru-cache@6.0.0 deduped
  │ │ └─┬ validate-npm-package-license@3.0.4
  │ │   ├─┬ spdx-correct@3.1.1
  │ │   │ ├── spdx-expression-parse@3.0.1 deduped
  │ │   │ └── spdx-license-ids@3.0.11
  │ │   └─┬ spdx-expression-parse@3.0.1
  │ │     ├── spdx-exceptions@2.3.0
  │ │     └── spdx-license-ids@3.0.11 deduped
  │ ├─┬ parse-json@5.2.0
  │ │ ├─┬ @babel/code-frame@7.16.7
  │ │ │ └─┬ @babel/highlight@7.17.9
  │ │ │   ├── @babel/helper-validator-identifier@7.16.7
  │ │ │   ├─┬ chalk@2.4.2
  │ │ │   │ ├─┬ ansi-styles@3.2.1
  │ │ │   │ │ └─┬ color-convert@1.9.3
  │ │ │   │ │   └── color-name@1.1.3
  │ │ │   │ ├── escape-string-regexp@1.0.5
  │ │ │   │ └─┬ supports-color@5.5.0
[...]

PR #2727, opened 2022-05-26
fix: avoid spurious warning about serviceName in Lambda

trentm:
Recently in #2684 a change was made so that the agent uses
`require.main` to look for "package.json" to infer serviceName.
That doesn't work for a Lambda Node.js env, because `require.main`
is in the Lambda Runtime ("/var/runtime/...") and not the same tree
where the lambda handler is ("/var/task/..."). The result is a spurious
log.warn in lambda logs:
    could not infer serviceName: could not find package.json up from /var/runtime

Per spec, we prefer the AWS_LAMBDA_FUNCTION_NAME envvar for serviceName
so won't use the package.json value at all. This changes to avoid
looking for package.json at all.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -339,6 +339,8 @@ class Config {
     this.sanitizeFieldNamesRegExp = []
     this.ignoreMessageQueuesRegExp = []
 
+    const isLambda = isLambdaExecutionEnvironment()
+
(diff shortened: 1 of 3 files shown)


trentm:

astorm (approved):

Issue #2622, opened 2022-03-24
Kubernetes attachment


felixbarny:
- [x] Publish a docker image that contains the agent distribution, similar to what the Java agent is already doing (see [build](https://github.com/elastic/apm-agent-java/blob/3187b4fbbf349041f95080e49b194848b1c33098/.ci/release/Jenkinsfile#L245-L257) and the [observability/apm-agent-java](https://container-library.elastic.co/r/observability/apm-agent-java) image in our Docker registry)
- [x] Integrate with the webhook in APM Server (https://github.com/elastic/apm-server/issues/7386) to copy the agent from the image to a volume via an init container and add environment variables to attach the agent.

AlexanderWert:
Note: This can also be reused to support Elastic APM with [container-based AWS Lambda functions](https://github.com/elastic/apm-aws-lambda/issues/107).

trentm:
For comparison, my (very naive) read of the "OTel Operator" is that it sets (or appends to) `NODE_OPTIONS=" --require /otel-auto-instrumentation/autoinstrumentation.js"`
https://github.com/open-telemetry/opentelemetry-operator/blob/main/pkg/instrumentation/nodejs.go#L24-L27

astorm:
Request made to productivity team for a published Dockerfile: https://github.com/elastic/observability-robots/issues/1229

I'm proceeding to making the adjustments to the mutating webhook to enable attachment.

trentm:
see https://github.com/elastic/apm-agent-nodejs/pull/2742

astorm:
Completed the minor changes needed to the mutating webhook, see https://github.com/elastic/apm-mutating-webhook/pull/31.  @trentm per last week's planning I decided to go absolute `NODE_OPTIONS` path and left out `NODE_PATH` for now.

astorm:
Final task to finish this will be fore the Node.js Docker image to be listed as a _public_ docker image.  Right now it appears to require some sort of elastic auth to fetch. See: https://github.com/elastic/sre/issues/194

trentm:
@astorm This can be closed now, right?

PR #2742, opened 2022-05-31
Build docker image

AlexanderWert:
Adds a step in the CI to create a Docker image for the agent

### DIFF
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -5,6 +5,7 @@ DIST_DIR = /build/dist/nodejs
 LAMBDA_LAYER_ZIP_FILE = ../build/dist/elastic-apm-node-lambda-layer.zip
 DOCKER_IMAGE_NAME = observability/apm-agent-nodejs
 DOCKER_REGISTRY = docker.elastic.co
+AGENT_VERSION = $(shell echo $${BRANCH_NAME} | cut -f 2 -d 'v')
 
 export AWS_FOLDER
 
@@ -22,12 +23,12 @@ env:
 	env
 
 # Create Docker image
-build-docker: validate-branch-name
-	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME):$(BRANCH_NAME) \
+build-docker: validate-version
+	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME):$(AGENT_VERSION) \
   --build-arg AGENT_DIR=$(DIST_DIR) ..
 
 push-docker: build-docker
-	../dev-utils/push-docker.sh $(DOCKER_REGISTRY) $(DOCKER_IMAGE_NAME) $(BRANCH_NAME)
+	../dev-utils/push-docker.sh $(DOCKER_REGISTRY) $(DOCKER_IMAGE_NAME) $(AGENT_VERSION)
 
 # List all the AWS regions
 get-all-aws-regions:
@@ -89,6 +90,11 @@ github-release: validate-branch-name validate-suffix-arn-file
 		create $(BRANCH_NAME) \
 		--title '$(BRANCH_NAME)' \
 		--notes-file $(AWS_FOLDER)/$(SUFFIX_ARN_FILE)
+
+validate-version:
+ifndef AGENT_VERSION
+	$(error AGENT_VERSION is undefined)
+endif
 
 validate-branch-name:
 ifndef BRANCH_NAME
--- a/.ci/Makefile
+++ b/.ci/Makefile
@@ -1,13 +1,16 @@
 SHELL = /bin/bash -eo pipefail
 
 AWS_FOLDER = ../build/aws
-LAMBDA_LAYER_ZIP_FILE = ../build/lambda-layer-zip/elastic-apm-node-lambda-layer.zip
+DIST_DIR = /build/dist/nodejs
+LAMBDA_LAYER_ZIP_FILE = ../build/dist/elastic-apm-node-lambda-layer.zip
+DOCKER_IMAGE_NAME = observability/apm-agent-nodejs
+DOCKER_REGISTRY = docker.elastic.co
 
 export AWS_FOLDER
 
 .PHONY: dist
 dist: validate-branch-name
-	../dev-utils/make-lambda-layer-zip.sh
+	../dev-utils/make-distribution.sh
 	@mkdir -p $(AWS_FOLDER)
 	cp $(LAMBDA_LAYER_ZIP_FILE) $(AWS_FOLDER)/elastic-apm-node-lambda-layer-$(BRANCH_NAME).zip
 
@@ -17,6 +20,14 @@ clean:
 
 env:
 	env
+
+# Create Docker image
+build-docker: validate-branch-name
+	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME):$(BRANCH_NAME) \
+  --build-arg AGENT_DIR=$(DIST_DIR) ..
+
+push-docker: build-docker
+	../dev-utils/push-docker.sh $(DOCKER_REGISTRY) $(DOCKER_IMAGE_NAME) $(BRANCH_NAME)
 
 # List all the AWS regions
 get-all-aws-regions:
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,6 +2,5 @@
 # For a complete list of hashes, see:
 # https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
 FROM alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178
-RUN mkdir /usr/agent
 ARG AGENT_DIR
-COPY ${AGENT_DIR} /usr/agent/nodejs
+COPY ${AGENT_DIR} /opt/nodejs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,7 @@
+# Pin to Alpine 3.9
+# For a complete list of hashes, see:
+# https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
+FROM alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178
+RUN mkdir /usr/agent
+ARG AGENT_DIR
+COPY ${AGENT_DIR} /usr/agent/nodejs
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -18,6 +18,8 @@ pipeline {
     NPMRC_SECRET = 'secret/jenkins-ci/npmjs/elasticmachine'
     TOTP_SECRET = 'totp/code/npmjs-elasticmachine'
     BUILD_NODE_VERSION = 'v16.15.0'
+    DOCKER_REGISTRY = 'docker.elastic.co'
+    DOCKER_SECRET = 'secret/apm-team/ci/docker-registry/prod'
   }
   options {
     timeout(time: 3, unit: 'HOURS')
@@ -295,6 +297,16 @@ pipeline {
                 dir("${BASE_DIR}"){
                   cmd(label: 'make dist', script: 'make -C .ci dist')
                 }
+              }
+            }
+          }
+        }
+        stage('Push Docker Image') {
+          steps {
+            withGithubNotify(context: "Push-Docker-Image") {
+              dir("${BASE_DIR}"){
+                dockerLogin(secret: "${env.DOCKER_SECRET}", registry: "${env.DOCKER_REGISTRY}", role_id: 'apm-vault-role-id', secret_id: 'apm-vault-secret-id')
+                cmd(label: 'make push-docker', script: 'make -C .ci push-docker')
               }
             }
           }
--- a/dev-utils/make-distribution.sh
+++ b/dev-utils/make-distribution.sh
@@ -23,7 +23,7 @@ function fatal {
 # ---- mainline
 
 TOP=$(cd $(dirname $0)/../ >/dev/null; pwd)
-BUILD_DIR="$TOP/build/lambda-layer-zip"
+BUILD_DIR="$TOP/build/dist"
 
 # Guard against accidentally using this script with a too-old npm.
 if [[ $(npm --version | cut -d. -f1) -lt 8 ]]; then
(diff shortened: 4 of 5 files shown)


.ci/Makefile
```diff
@@ -18,6 +21,14 @@ clean:
 env:
 	env
 
+# Create Docker image
+build-docker: validate-branch-name
+	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME):$(BRANCH_NAME) \
+  --build-arg AGENT_DIR=$(DIST_DIR) ..
+
+push-docker: build-docker
+	../dev-utils/push-docker.sh $(DOCKER_REGISTRY) $(DOCKER_IMAGE_NAME) $(BRANCH_NAME)
```
trentm, line 30: Forgot to 'git add dev-utils/push-docker.sh'?

.ci/Makefile
```diff
@@ -18,6 +21,14 @@ clean:
 env:
 	env
 
+# Create Docker image
+build-docker: validate-branch-name
```
felixbarny, line 25: Is `validate-branch-name` a copy/paste leftover?

.ci/Makefile
```diff
@@ -18,6 +21,14 @@ clean:
...
+	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME):$(BRANCH_NAME) \
```
felixbarny, line 26: Does the BRANCH_NAME include the `v` prefix, as in `v3.1.2`? If so, we probably want to remove it.

felixbarny: Probably beyond the scope of this but we might also want to publish a `latest` tag and a tag for the latest major version, such as `3.x`. Not that `latest` doesn't necessarily refer to the most recent release, but the one that has the highest semver version number. For example `3.1.0 > 3.0.1`, even if `3.0.1` has been released after `3.1.0`.

Dockerfile
```diff
@@ -0,0 +1,7 @@
+# Pin to Alpine 3.9
+# For a complete list of hashes, see:
+# https://github.com/docker-library/repo-info/tree/master/repos/alpine/remote
+FROM alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178
+RUN mkdir /usr/agent
+ARG AGENT_DIR
+COPY ${AGENT_DIR} /usr/agent/nodejs
```
trentm, line 7: For comparison the Java APM agent docker container is using "/usr/agent/elastic-apm-handler{,.jar}". With this the Node.js APM agent will be at "/usr/agent/nodejs/node_modules/elastic-apm-node/...", I think (I haven't run this build yet :).

trentm (commented):
Thanks!

.ci/Makefile
AlexanderWert: yes, forgot it, added now

.ci/Makefile
trentm: Nope, that's good to use. It is defined below and ensures the `$(BRANCH_NAME)` var is set for this target.

.ci/Makefile
trentm: In current usage the `BRANCH_NAME` does *not* include the `v` prefix.

.ci/Makefile
```diff
@@ -1,13 +1,16 @@
 SHELL = /bin/bash -eo pipefail
 
 AWS_FOLDER = ../build/aws
-LAMBDA_LAYER_ZIP_FILE = ../build/lambda-layer-zip/elastic-apm-node-lambda-layer.zip
+DIST_DIR = /build/dist/nodejs
```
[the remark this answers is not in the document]
AlexanderWert: In Dockerfile it's always relative to the Dockerfile.
I first had the Dockerfile in the .ci sub dir and tried with relative `../build/dist/nodejs` but that does not work because all things added in the Dockerfile need to be on the same or below level in the directory structure relative to the Dockerfile

.ci/Makefile
trentm: +1 to "latest". We won't have much need for a "3.x" tag until/if we do a "4.x" major version of the APM agent.

AlexanderWert: > Does the BRANCH_NAME include the v prefix, as in v3.1.2? If so, we probably want to remove it.

Yes, it does. Will try to remove it.

> Probably beyond the scope of this but we might also want to publish a latest tag ...

I added pushing of the `latest` tag already in `dev-utils/push-docker.sh`, but without the advanced logic, yet. 
Do you know how we can query the tags on a docker repo (e.g. from something like this https://container-library.elastic.co/r/observability/apm-agent-java ) to implement this advanced logic?

AlexanderWert: | In current usage the BRANCH_NAME does not include the v prefix.

@trentm With my local testing it did. But I guess it behaves different on CI?

trentm: If it helps with the logic, I don't think we've *ever* done the equivalent of a "3.0.1" release after having already released "3.1.0" or later. I.e. I don't think this code needs to worry about whether the current release *is* the "latest". Perhaps just have a "IS_LATEST" var hardcoded to true. If/when we have a "3.x" maintenance branch and "main" is for version 4.x of the agent, then we can set `IS_LATEST=false` for the 3.x branch (or just delete the block of code that adds the "latest" tag).

Dockerfile
trentm: I don't know exactly how the k8s attachment will work.  I'm guessing it will mount one of these directories in a target container and add a `NODE_OPTIONS="--require $mountPath"` envvar.

I *think* that would work without the `/node_modules/` component in the path. In other words, I think it might be possible to have the agent Docker image use "/usr/agent/elastic-apm-node/..." for its path, if you want.

trentm: https://github.com/elastic/apm-agent-nodejs/issues/2622#issuecomment-1131573905 mentions perhaps wanting this Docker image to be also usable for container-based Lambda functions. That might dictate what path we want the APM agent to be installed into.

.ci/Makefile
felixbarny: > Do you know how we can query the tags on a docker repo (e.g. from something like this https://container-library.elastic.co/r/observability/apm-agent-java ) to implement this advanced logic?

This script returns the latest version number:
```
git tag --list "v*" | cut -c2- | sort --version-sort | tail -n 1
```

You could just check whether the current version is the same.

felixbarny: Another alternative would be `git tag --list --sort=version:refname "v*" | cut -c2- | tail -n1`

If the tags may contain suffix that indicates pre-releases, such as `1.2.3-pre1`, or `1.2.3.RC1` (the latter is used in the Java agent) we can use this: `git -c 'versionsort.suffix=.RC' tag --list --sort=version:refname "v*" | cut -c2- | tail -n1`

Dockerfile
astorm: TL;DR; -- the path the agent's copied to in the container is arbitrary and its path doesn't need a particular name.  

Re: k8s attachment

> I don't know exactly how the k8s attachment will work. I'm guessing it will mount one of these directories in a target container and add a NODE_OPTIONS="--require $mountPath" envvar.

That's almost it.  The additional nuance is that the agent's source will be copied out of the Docker container and into a mounted volume at a fixed path.  The command that currently does this copying is here: https://github.com/elastic/apm-mutating-webhook/blob/2c6a4cecf33fb8022eb1f3a365e6705345ff908a/patch.go#L150.  (part of the work of K8s attachment is changing this copy to be able to work with multiple files instead of a single file `jar`) The some `NODE_OPTIONS` and `NODE_PATH` setting will happen in order to load the agent from this spot.

trentm: Thanks.
We coudl consider *just* setting/updating `NODE_OPTIONS` and not bothering with `NODE_PATH`, right?

astorm: > We could consider just setting/updating NODE_OPTIONS and not bothering with NODE_PATH, right?

That would be ideal -- but without a `NODE_PATH` set (it doesn't seem like?) manual instrumentation in the user's code would able to find the agent module.  Is that incorrect?

astorm (commented):
Thanks for putting this together @AlexanderWert -- overall it looks great but I have few questions for you (or maybe for @v1v) that's more about me not knowing how our jenkins systems work. 

Context: I tried building the container (without pushing) locally with the following invocation.

    % cd .ci
    % DOCKER_IMAGE_NAME=observability/apm-agent-nodejs DOCKER_REGISTRY='docker.elastic.co' BRANCH_NAME='v3.34.0' make build-docker

However, the docker build failed.
    
    docker build -t docker.elastic.co/observability/apm-agent-nodejs:3.34.0 \
      --build-arg AGENT_DIR=/build/dist/nodejs ..
    [+] Building 2.1s (7/7) FINISHED                                                                                                                                     
     => [internal] load build definition from Dockerfile                                                                                                            0.0s
     => => transferring dockerfile: 335B                                                                                                                            0.0s
     => [internal] load .dockerignore                                                                                                                               0.0s
     => => transferring context: 2B                                                                                                                                 0.0s
     => [internal] load metadata for docker.io/library/alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178                               1.9s
     => [1/3] FROM docker.io/library/alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178                                                 0.0s
     => => resolve docker.io/library/alpine@sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178                                                 0.0s
     => => sha256:115731bab0862031b44766733890091c17924f9b7781b79997f5f163be262178 1.64kB / 1.64kB                                                                  0.0s
     => => sha256:ab3fe83c0696e3f565c9b4a734ec309ae9bd0d74c192de4590fd6dc2ef717815 528B / 528B                                                                      0.0s
     => => sha256:82f67be598ebc8d968137c18521fe174ca6afc9b542aa5773c32b3a3970e647c 1.51kB / 1.51kB                                                                  0.0s
     => [internal] load build context                                                                                                                               0.0s
     => => transferring context: 27B                                                                                                                                0.0s
     => CACHED [2/3] RUN mkdir /usr/agent                                                                                                                           0.0s
     => ERROR [3/3] COPY /build/dist/nodejs /usr/agent/nodejs                                                                                                       0.0s
    ------
     > [3/3] COPY /build/dist/nodejs /usr/agent/nodejs:
    ------
    failed to compute cache key: "/build/dist/nodejs" not found: not found
    make: *** [build-docker] Error 1

It seemed to be trying to copy files from `/build/dist/nodejs` into the container but no such files existed.  I also tried running `BRANCH_NAME='v3.34.0' make dist` to make the distribution and it was built into `../build/dist` and not `/build/dist/nodejs`.  I was finally able to get things built locally by running

    DOCKER_IMAGE_NAME=observability/apm-agent-nodejs DOCKER_REGISTRY='docker.elastic.co' BRANCH_NAME='v3.34.0' DEST_DIR='/Users/astorm/Documents/github/elastic/apm-agent-nodejs/dist/nodejs' make build-docker

In other words -- by running `make dist` and then pointing `DEST_DIR` at the correct build directory I was able to get this working.

All of this raised two questions in my mind:

1. Is there any quick way to make this build locally in one step while maintaining the needed functionality in our Jenkins environments?

2. Does the `stage('Push Docker Image')` section need to run `make dist` as well, or can we safely assume that the source will be there for this stage (this is a very naive "how does our Jenkins work" question)

Dockerfile
trentm: > manual instrumentation in the user's code would able to find the agent module. Is that incorrect?

You are correct. I think there are some finer points around that that might be easier to discuss on a call (or better on https://github.com/elastic/apm-agent-nodejs/issues/2622). Basically whether we *want* the user to be able to `require('elastic-apm-node')` and get this externally installed package. My (limited) understanding of the OTel equivalent is that they are not modifying `NODE_PATH`.  If the programmer is adding manual instrumentation then they are likely better served having their own `"elastic-apm-node"` in their package.json dependencies.

AlexanderWert:
@astorm 
> 1. Is there any quick way to make this build locally in one step while maintaining the needed functionality in our Jenkins environments?
> 2. Does the `stage('Push Docker Image')` section need to run `make dist` as well, or can we safely assume that the source will be there for this stage (this is a very naive "how does our Jenkins work" question)

To run it locally you need execute the following:
```
cd .ci
make dist && make push-docker
```

The `push-docker` stage assumes that the `make dist` stage has been executed before.
Our current `Publish to AWS` stage has the same assumption. 
In the Jenkinsfile those steps are executed in sequence as part of the `Release` stage.

````

stage('Release') {
      options { skipDefaultCheckout() }
      when {
        beforeAgent true
        tag pattern: 'v\\d+\\.\\d+\\.\\d+', comparator: 'REGEXP'
      }
      environment {
        SUFFIX_ARN_FILE = 'arn-file.md'
      }
      stages {
        stage('Dist') {
          steps {
            withGithubNotify(context: "Dist") {
              ...
              withNodeJSEnv(version: "${env.BUILD_NODE_VERSION}"){
                dir("${BASE_DIR}"){
                  cmd(label: 'make dist', script: 'make -C .ci dist')
                }
              }
            }
          }
        }
        stage('Push Docker Image') {
          steps {
            withGithubNotify(context: "Push-Docker-Image") {
              dir("${BASE_DIR}"){
                dockerLogin(secret: "${env.DOCKER_SECRET}", registry: "${env.DOCKER_REGISTRY}", role_id: 'apm-vault-role-id', secret_id: 'apm-vault-secret-id')
                cmd(label: 'make push-docker', script: 'make -C .ci push-docker')
              }
            }
          }
        }
        stage('Publish to AWS') {
          steps {
            withGithubNotify(context: "Publish") {
              withGoEnv(){
                withAWSEnv(secret: 'secret/observability-team/ci/service-account/apm-aws-lambda', forceInstallation: true, version: '2.4.10') {
                  dir("${BASE_DIR}"){
                    cmd(label: 'make publish-in-all-aws-regions', script: 'make -C .ci publish-in-all-aws-regions')
                    cmd(label: 'make create-arn-file', script: 'make -C .ci create-arn-file')
                  }
                }
              }
            }
          }
````

Does it answer your questions?

astorm:
> Does it answer your questions?

Yup -- this part in particular

> The push-docker stage assumes that the make dist stage has been executed before. Our current Publish to AWS stage has the same assumption. In the Jenkinsfile those steps are executed in sequence as part of the Release stage.

I tend to try and make each build not rely on artifacts from previous steps -- but as long as there's prior art and the steps are sequential it sounds good to me. 👍

Issue #2741, opened 2022-05-31
transaction.name not being captured in express in case of path parameter.


agarwalarjun123:
**Describe the bug**

transaction.name does not include the base router url in express.

**To Reproduce**

1. Call API http://localhost:3000/v1/vital/:id
2. See error
{"log.level":"debug","@timestamp":"2022-05-31T12:46:51.832Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"ended transaction {\"trans\":\"xxxxxxxxxxxx\",\"trace\":\"xxxxxxxxxxxxx\",\"type\":\"request\",\"result\":\"HTTP 2xx\",\"name\":\"GET /:id\"}"}

**Expected behavior**

transaction.name should come as /v1/vital/:id

**Environment**

- OS: Linux
- Node.js version: 14
- APM Server version: 7.17.3
- Agent version: 3.34.0

**How are you starting the agent? (please tick one of the boxes)**

- [X] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**
- Agent config options
  environment: config.NODE_ENV,
  serviceName: config.APM.SERVICE_NAME,
  serverUrl: _.get(config, 'APM.SERVICE_URI'),
  captureBody: 'all',
  captureHeaders: true,
  logUncaughtExceptions: true,
  errorOnAbortedRequests: true,
  captureErrorLogStackTraces: 'always',
  usePathAsTransactionName: true,
  logLevel: 'trace',

trentm:
Hi @agarwalarjun123. Thanks for the issue.

Here is the code in the APM agent that is determining the default transaction name from the incoming HTTP request info:

```js
function getPathFromRequest (req, useBase, usePathAsTransactionName) {
  if (req[symbols.staticFile]) {
    return 'static file'
  }

  var path = getStackPath(req)
  var route = routePath(req.route)

  if (route) {
    return path ? join([path, route]) : route
  } else if (path && (path !== '/' || useBase)) {
    return path
  }

  if (usePathAsTransactionName) {
    const parsed = parseUrl(req)
    return parsed && parsed.pathname
  }
}
```

What web framework are you using? Express? or some other one?

[...]

PR #2750, opened 2022-06-02
fix: Express route calc when next(err-that-is-not-Error-instance) is used

trentm:
Express considers any value passed to a handler's next() handler
to be an error, except the special case 'route' and 'router' strings.
Before this change the agent's Express route handling (used for setting
transaction.name) would only consider the arg to be an error if
`instanceof Error`.

Refs: #2741

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -55,6 +55,12 @@ Notes:
 [float]
 ===== Bug fixes
 
+- Fix Express route tracking (used for `transaction.name`) when an argument
+  is passed to the `next(arg)` callback of a request handler. Before this
+  change passing `next(<some object not an instance of Error>)` would be
+  considered an error by Express, but not by the APM agent's route
+  tracking. ({pull}2750[#2750])
+
 - Fixes automatic Lambda handler wrapping to work with handlers that point to
   subfolders (ex. `_HANDLER=path/to/folder.methodName`) ({issues}2709[#2709])
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,31 @@ Notes:
 === Node.js Agent version 3.x
 
 
+==== Unreleased
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
+- Adds https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md[dropped span statistics]
+  to transaction payloads allowing APM Server to calculate more accurate
+  throughput metrics. ({issues}2302[#2302])
+
+[float]
+===== Bug fixes
+
+- Fix Express route tracking (used for `transaction.name`) when an argument
+  is passed to the `next(arg)` callback of a request handler. Before this
+  change passing `next(<some object not an instance of Error>)` would be
+  considered an error by Express, but not by the APM agent's route
+  tracking. ({pull}2750[#2750])
+
+[float]
+===== Chores
+
+
 [[release-notes-3.35.0]]
 ==== 3.35.0 2022/06/01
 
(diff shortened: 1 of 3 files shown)


trentm:
If it helps, here is an Express server that plays with routing cases to explore how routing works and how our transaction.name calculation works:

```js
  serviceName: 'play-express-next-values',
  metricsInterval: '0s',
  centralConfig: false,
  cloudProvider: 'none'
})
apm.addTransactionFilter(trans => {
  console.log('-- transaction.name="%s"', trans.name)
  return trans
})

const app = express()

const apiRouter = express.Router()
apiRouter.get('/foo',
  function fooHandler1 (req, res, next) {
    console.log('[fooHandler1] query', req.query)
    if (req.query.next === 'route') {
      // This skips over `fooHandler2`, to the next matching route, if any.
      next('route')
    } else if (req.query.next === 'router') {
      // This skips out of the whole router.
      next('router')
    } else {
      next()
    }
  },
  function fooHandler2 (req, res, next) {
    console.log('[fooHandler2]')
    res.send('response from foo')
  })
apiRouter.get('/boom', function boomHandler (req, res, next) {
  throw new Error('boom')
})
apiRouter.get('/weird', function weirdHandler (req, res, next) {
  // This passes a "weird" value to `next()`. Weird in that it is not an Error
  // instance or either of the special case "route" or "router" values that
  // Express accepts.
  next({ weird: true })
})
apiRouter.use(function apiMiddleware (req, res, next) {
  console.log('[apiMiddleware]')
  next()
})
apiRouter.use(function apiErrHandler (err, req, res, next) {
  console.log('[apiErrHandler] err=%s', err.message || err)
  res.status(500)
    .send('there was an API error: ' + err)
})

app.use('/top', function topHandler (req, res, next) {
  console.log('[topHandler]')
  res.send('response from top-level')
})
app.use('/api', apiRouter)
app.use(function topLevelMiddleware (req, res, next) {
  console.log('[topLevelMiddleware]')
  next()
})
app.get('/:any/foo', function weirdTopLevelAnyFooHandler (req, res, next) {
  console.log('[weirdTopLevelAnyFooHandler]')
  res.send('response from weirdTopLevelAnyFooHandler')
})
app.use(function appErrHandler (err, req, res, next) {
  console.log('[appErrHandler] err=%s', err.message || err)
  res.status(500)
    .send('there was an error: ' + err)
})

app.listen(3000, () => console.log('server running on port 3000'))
```

If you run that and then call its `/api/weird` endpoint (`curl -i localhost:3000/api/weird`), you get:

```
% node play-express-next-values.js
server running on port 3000
[apiErrHandler] err={ weird: true }
-- transaction.name="GET /api/weird"
```

Before the fix in this PR the transaction.name set by the APM agent was "GET /weird" because the `/api` part was incorrectly popped from `req[symbols.expressMountStack]`.

trentm:
Note that this drops the `req.router` guard that was in the previous code. That was added in https://github.com/elastic/apm-agent-nodejs/pull/938 (to fix https://github.com/elastic/apm-agent-nodejs/issues/935), but I don't know why that guard. It wasn't discussed on the issue or PR.

AFAICT Express (for all versions that we currently support: 4.0.0 ... 4.18.1) the `Route.prototype.dispatch` method has always set `req.router`, e.g. https://github.com/expressjs/express/blob/4.18.1/lib/router/route.js#L112  The full TAV tests are passing.

trentm:
run module tests for express

astorm (approved):
Fix looks good and look like is results in more correct behavior.  Approving.  

Since we are changing a long standing (if incorrect) behavior this might be worth calling out explicitly in the changelog so users and support folks can find it.

trentm:
> Since we are changing a long standing (if incorrect) behavior this might be worth calling out explicitly in the changelog

:+1:

PR #2764, opened 2022-06-09
chore: include NOTICE file in npm published package and in dist

trentm:
The NOTICE file in the distribution includes the license for all
the bundled packages (i.e. add dependencies, excluding devDependencies).

Creating the distribution (dev-utils/make-distribution.sh) will fail
if one of the deps has no license or an unknown license.


trentm:
Here is an example generated NOTICE file: https://gist.github.com/trentm/057ec6b14669a6b02562581756a362d9
It is the current NOTICE.md with the generated part below: https://gist.github.com/trentm/057ec6b14669a6b02562581756a362d9#notice-for-distributed-packages

dev-utils/gen-notice.sh
```diff
@@ -0,0 +1,120 @@
...
+            "ISC": true,
+            "MIT": true,
+            "WTFPL OR ISC": true // oddball from is-integer package
+        }
+        const licFileNames = [
+            "LICENSE",
+            "LICENSE.txt",
+            "license.md",
+            "LICENSE-MIT",
+            "LICENSE-MIT.txt"
```
astorm, line 63: Is this going to run on linux build machines?  If so (I think?) we'll need the alternate casing of these file names as well or we'll need something more complex that [the`existsSync`](https://github.com/elastic/apm-agent-nodejs/pull/2764/files#diff-57cd11bbaa52fcd7b495d503a57bd99e109b3a489322b031815d7883cfde6928R104) here.

astorm, line 63: The script worked great when I ran it on my case insensative Mac filesystem.  Will this   script ever run on linux build machines with case sensitive file systems?  

If so (I think?) we'll need the alternate casing of these file names (`license.md`, `License.md`, `LICENSE.md`) or we'll need something more complex that [the`existsSync` check here](https://github.com/elastic/apm-agent-nodejs/pull/2764/files#diff-57cd11bbaa52fcd7b495d503a57bd99e109b3a489322b031815d7883cfde6928R104) here.

astorm (commented):
Looks good -- one question on case sensitivity and I think we'll be good to go.

dev-utils/gen-notice.sh
trentm: Nice catch. There were a few that didn't match that set when on a case-sensitive file system.

A little rabbit hole: npm uses [npm-packlist](https://www.npmjs.com/package/npm-packlist) to try to ensure a license file in a repo is included with `npm publish|pack`, which uses this logic: https://github.com/npm/npm-packlist/blob/da1ba4a3051ea822d9625347e00c186677247a45/lib/index.js#L20-L30

However that pattern does *not* include "LICENSE-MIT.*", which some of our current runtime deps use. I've changed the licFile handling to search the dir for a matching license file name that covers all these cases.

I've also added 'npm run lint:license-files' (added to `npm run lint`) that will run `gen-notice.sh ...` as part of linting. This means we'll find out if a new dep has a license issue for release before we get to release time.

commit 01de760b

trentm:
@astorm CI passing again. I've added a `npm run lint:license-files`. This is ready for a second review.

astorm (approved):

PR #2783, opened 2022-06-23
docs: drop potentially confusing language in config precendence

trentm:
The "(higher overwrites lower)" for config precendence can be
confusing because "4" is higher than "1" in the numbered bullets,
but that's not what is intended.


bmorelli25 (approved):
Thanks!

PR #2785, opened 2022-06-24
chore: update to got@10 and stop there

trentm:
We use 'got' in our tests only to test its usage by mimic-response.
got@11 dropped usage of mimic-response, so there is no point having
anything later than got@10.

* * *

This PR will stop dependabot suggesting updates to got beyond got@10, e.g. no more PRs like https://github.com/elastic/apm-agent-nodejs/pull/2784


astorm (approved):
👍

PR #2787, opened 2022-06-27
chore: bump to got@11.8.5 for vuln fix

trentm:
got@11.8.5 includes a fix for CVE-2022-33987. Ours is a dev-dep so the
vuln shouldn't affect us. This update is effectively the same thing
as the recent #2784 from dependabot.

This also removes the change from #2785 to lock us to got@10, and
clarifies that we want to continue testing with recent versions of 'got'
as a sanity check that the agent works with it. At least for now.

* * *

Alan, initially I closed the [dependabot PR to update us to got@11](https://github.com/elastic/apm-agent-nodejs/pull/2784) and did #2785 instead to lock us on got@10, based on the arguments that (a) testing with got is *only* for testing the mimic-response@1.0.0 issue and (b) got@11 no longer uses mimic-response.

I had missed that there is a vuln in got@10 that is fixed in latest got@11 and got@12. We shouldn't be affected by the vuln, but it is helpful to keep the [alerts](https://github.com/elastic/apm-agent-nodejs/security/dependabot) clean, so options were to:
1. drop the `got` dev-dep and its testing altogether, or
2. update to `got@11`; we still test earlier versions with TAV.

I've opted for option 2. The argument is that the existing "test/instrumentation/modules/http/github-423.test.js" is a reasonable sanity check that the agent basically works with `got` usage, so there is some value in keeping those TAV tests for now.  I don't have a strong opinion here, however.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,9 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
-    ignore:
-      # Lock at current got@10. See comment in .tav.yml.
-      - dependency-name: "got"
-        versions: [">=11"]
     reviewers:
       - "elastic/apm-agent-node-js"
--- a/.tav.yml
+++ b/.tav.yml
@@ -10,23 +10,24 @@ mimic-response:
   commands:
     - node test/instrumentation/modules/mimic-response.test.js
     - node test/instrumentation/modules/http/github-179.test.js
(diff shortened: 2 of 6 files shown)


astorm (approved):
> I've opted for option 2.

Works for me, and anything that can keep our alerts down and not make us have to think about whether a vulnerability affects us or not is a win in my book.  Approving.

PR #2796, opened 2022-06-30
feat: replace metrics impl with one based on @otel/sdk-metrics-base

trentm:
OpenTelemetry JS Metrics is nearing 1.0. (The Java, Python, and Go
OTel projects have, I think, reached 1.0.) This SDK might be a good
replacement for the node-measured libs we are using for handling
metrics. It could be a path to working on current limitations in our
metrics support.

This PR uses `@opentelemetry/sdk-metrics-base` for the agent's metrics generation.

# status

This is incomplete, and currently is just a PoC.

Supported so far:
- The `apm.registerMetric()` only public API for adding custom metrics.
- Runtime metrics (`nodejs.*`).

Not yet:
- breakdown metrics
- system mertics
- queue metrics for SQS instrumentation
- perf comparison
- etc.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -19,7 +19,12 @@ var errors = require('./errors')
 const { InflightEventSet } = require('./InflightEventSet')
 var Instrumentation = require('./instrumentation')
 var { elasticApmAwsLambda } = require('./lambda')
-var Metrics = require('./metrics')
+var Metrics // XXX
+if (process.env.OLD_METRICS) {
+  Metrics = require('./metrics')
+} else {
+  Metrics = require('./metrics2')
+}
 var parsers = require('./parsers')
 var symbols = require('./symbols')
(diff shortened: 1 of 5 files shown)


trentm:
Current status: There is a test hang:

    cd test/start/env && node --unhandled-rejections=strict test.test.js

Guess: There is an unref'd setInterval in @otel/sdk-metrics-base that is keeping the event loop running.
From a quick look, it looks like this is the case.

What to do? Bug on core? Probably.

For comparison, measured-reporting/lib/reporters/Reporter.js will unref (optionally), which was added by us: https://github.com/yaorg/node-measured/pull/55

trentm:
This OnWeek is done. Some status in https://github.com/elastic/apm-dev/issues/781 if we pick up this work again. I think it is promising and I hope we do pick it up later.

PR #2805, opened 2022-07-05
ci: send github actions and tests as distributed traces

v1v:
### What

Enable traces for the steps/jobs executed in the GitHub actions and also the test results.

For such I used:
- https://github.com/inception-health/otel-upload-test-artifact-action 
- https://github.com/inception-health/otel-export-trace-action

Although, it's now hidden in the composite action defined in https://github.com/elastic/apm-pipeline-library/tree/main/.github/actions/opentelemetry

### Why

I've been playing a bit and it looks promising, though to make this work it's required to merge this PR since the actions will not work until then

### Actions

- [x] secrets for the opentelemetry backend


.github/workflows/opentelemetry.yml
```diff
@@ -0,0 +1,19 @@
+name: OpenTelemetry Export Trace
+
+on:
+  workflow_run:
+    workflows: [Lint, Test]
```
v1v, line 5:
```suggestion
```

.github/workflows/opentelemetry.yml
```diff
@@ -0,0 +1,18 @@
+name: OpenTelemetry Export Trace
+
+on:
+  workflow_run:
+    types: [completed]
```
v1v, line 5:
```suggestion
    workflows: [Test]
    types: [completed]
```

.github/workflows/test.yml
```diff
@@ -140,4 +140,15 @@ jobs:
     - run: docker ps  # show the services against which we'll be testing
     - run: npm install
     - run: npm ls --all || true
-    - run: npm test
+    - name: npm test
+      run: npm test
+    - name: debug test_output
+      run: ls test_output/*.* || true
+    - uses: v1v/otel-upload-test-artifact-action@v2
```
v1v, line 147: This is my forked repo, waiting for the fix I put regarding JUnit to be merged https://github.com/inception-health/otel-upload-test-artifact-action/pull/9

So we can use it mine meanwhile

cachedout (commented):
I wonder if it is possible to test these changes on a fork of this repo before pushing them into the main branch on the primary?

v1v:
> I wonder if it is possible to test these changes on a fork of this repo before pushing them into the main branch on the primary?

I tested it out in some other projects.

I could split the change in two PRs:
1) `.github/workflows/opentelemetry.yml` in a new PR
2) remain this one, so we can verify everything works as expected.

.github/workflows/test.yml
```diff
@@ -140,4 +140,15 @@ jobs:
...
+      if: always()
+      with:
```
v1v, line 149: Should we continue if something bad happened when generating the test traces?

```suggestion
      continue-on-error: true
      with:
```

cachedout:
> I tested it out in some other projects.

@v1v  OK, cool. The description of this made it sound like this was totally experimental but if you've already been looking at it that makes me feel a bit better. No need to split it up IMHO.

.github/workflows/test.yml
cachedout: I would presume that we would never want a telemetry error to cause a PR not to pass a GH check.

kuisathaverat (approved):

test/script/run_tests.sh
```diff
@@ -47,12 +47,10 @@ run_test_suite () {
...
-  if [ -z "$COVERAGE" ]
-  then
-    node test/test.js
-  else
-    nyc node test/test.js
-  fi
+  rm -rf ./test_output
+  mkdir ./test_output
+  nyc node test/test.js -o ./test_output
+  ls test_output/*.tap | while read f; do cat $f | ./node_modules/.bin/tap-junit > $f.junit.xml; done
```
trentm, line 53: Changes to this script will change the behaviour of `npm test`, which is used for local development. I don't think we want to have (a) coverage and (b) generation of .tap -> .junit.xml files in the general case. 

1. Currently the "Test" action does *not* run with coverage (i.e. does not run with `nyc ...`). Is there a particular reason to add coverage testing for these runs?
2. How about we put the `-o ./test_output` usage and conversion to junit format handling behind a guard on the "CI" envvar being defined? This would mean that `npm test` for local dev would not be changed, but when run in GitHub Actions (which [defines CI=true always](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables)) the Junit files would be defined for the OTel export.

Something like this (untested):

```diff
--- a/test/script/run_tests.sh
+++ b/test/script/run_tests.sh
@@ -47,11 +47,21 @@ run_test_suite () {

   npm run test:deps

-  if [ -z "$COVERAGE" ]
-  then
-    node test/test.js
+  # If running in CI, then output .tap files and covert them to JUnit
+  # format for test reporting.
+  local testArgs=""
+  if [[ -n "$CI" ]]; then
+    rm -rf ./test_output
+    mkdir ./test_output
+    testArgs="-o ./test_output"
+  fi
+  if [[ -n "$COVERAGE" ]]; then
+    nyc node test/test.js $testArgs
   else
-    nyc node test/test.js
+    node test/test.js $testArgs
+  fi
+  if [[ -n "$CI" ]]; then
+    ls test_output/*.tap | while read f; do cat $f | ./node_modules/.bin/tap-junit > $f.junit.xml; done
   fi

   npm run test:types
```

trentm (changes requested):

v1v:
I already enabled a similar set of actions in `elastic/beats.git` and https://github.com/elastic/beats/pull/32267 and it's the one that contains the example with the traces (github action + test results)

.github/workflows/opentelemetry.yml
```diff
@@ -0,0 +1,19 @@
...
+    workflows: [Test]
+    types: [completed]
+
+jobs:
+  otel-export-trace:
+    name: OpenTelemetry Export Trace
+    runs-on: ubuntu-latest
+    steps:
+      - name: Export Workflow Trace
+        uses: inception-health/otel-export-trace-action@latest
```
v1v, line 14: The existing workflow does not support to set the name of the service, so I'll change it to use the open PR https://github.com/inception-health/otel-export-trace-action/pull/13

```suggestion
        uses: v1v/otel-export-trace-action@v2
```

.github/workflows/opentelemetry.yml
```diff
@@ -0,0 +1,19 @@
...
+        with:
+          otlpEndpoint: "${{ secrets.APM_SERVER }}"
+          otlpHeaders: "Authorization=Bearer ${{ secrets.APM_TOKEN }}"
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
```
v1v, line 18:
```suggestion
          otelServiceName: ${{ github.repository }}
          githubToken: ${{ secrets.GITHUB_TOKEN }}
```

v1v (commented):
Let's use the version from my forked repo, so we can benefit from some new features

v1v:
> Where are these traces going to end up? Is it in an internal cluster?

Yes, that's one of the cluster we have been using for quite sometime, for tracking all the CI/CD events, I'll send you the URL, I guess I could expose it as it's secured and only accessible with SSO, but to play secure, I won't do it unless it's required

.github/workflows/test.yml
```diff
@@ -140,4 +140,16 @@ jobs:
     - run: docker ps  # show the services against which we'll be testing
     - run: npm install
     - run: npm ls --all || true
-    - run: npm test
+    - name: npm test
+      run: npm test
+    - name: debug test_output
+      run: ls test_output/*.* || true
```
trentm, line 146: Was this a debugging step that can be dropped?

v1v, line 146:
```suggestion
```

.github/workflows/test.yml
v1v:
```suggestion
    - uses: inception-health/otel-upload-test-artifact-action@v1
```

trentm:
@v1v *Where* does this data end up? I see that `OTLP_ENDPOINT` is used in the apm-pipeline-library action, but I don't see where that value is set.

Issue #2834, opened 2022-07-19
benchmarks reporting is failing: Action/metadata line [1] contains an unknown parameter [_type]


trentm:
Any recent Jenkins CI builds that run the "Benchmarks" step are failing with (e.g. from https://apm-ci.elastic.co//blue/rest/organizations/jenkins/pipelines/apm-agent-nodejs/pipelines/apm-agent-nodejs-mbp/pipelines/main/runs/297/steps/1161/log/?start=0):

```
httpRequest: Failure connecting to the service https://51156728e02f4196a2e96b218ec71781.us-central1.gcp.cloud.es.io:9243/_bulk : httpRequest: Failure connecting to the service https://51156728e02f4196a2e96b218ec71781.us-central1.gcp.cloud.es.io:9243/_bulk : 
Code: 400
Error: {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Action/metadata line [1] contains an unknown parameter [_type]"}],"type":"illegal_argument_exception","reason":"Action/metadata line [1] contains an unknown parameter [_type]"},"status":400}
```

From https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/main/ the first failure was: 
[#289](https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/main/289/) 14-Jul-2022 7:39 PM

trentm:
The java agent hit the same error recently: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-java%2Fapm-agent-java-mbp/detail/main/303/pipeline
But their build 304 seems to have resolved that: https://apm-ci.elastic.co/job/apm-agent-java/job/apm-agent-java-mbp/job/main/304/
Fixed by: https://github.com/elastic/apm-agent-java/pull/2710

PR #2835, opened 2022-07-19
test: Fix benchmarks reporting to an ES 8.x

trentm:
Our benchmarks deployment was updated from 7.x to 8.x, which means no
mapping types, which means no `_type` field, which is an easy fix
because we weren't using that type distinction for anything.
    https://www.elastic.co/guide/en/elasticsearch/reference/7.17/removal-of-types.html

Fixes: #2834

* * *

This is the same fix the Java agent did in https://github.com/elastic/apm-agent-java/pull/2710

### DIFF
--- a/test/benchmarks/utils/analyzer.js
+++ b/test/benchmarks/utils/analyzer.js
@@ -66,7 +66,7 @@ function storeResult () {
     bench
   }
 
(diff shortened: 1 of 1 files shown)


astorm (approved):

trentm:
^^ The Benchmarks step in run 2 at https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2835/ passed.

Issue #1967, opened 2021-02-05
improve agent docs for use with TypeScript, bundlers, and perhaps ESM


trentm:
There are a number of surprises/gotchas/not obvious hurdles to getting APM working with TypeScript, bundlers (like webpack), and ESM. The agent docs could be improved a lot for some of these cases. Comments below are collecting some Qs, gotchas, notes for this. (I realize that makes this issue a little fuzzy for "completion.")

## initial description

See https://github.com/elastic/apm-agent-nodejs/issues/1962#issuecomment-774236997 about it being hard to find the suggestion to use the `npm -r elastic-apm-node/start app.js` technique for those using bundlers (like webpack) or transpilers (like TypeScript) that might remove or re-order the "elastic-apm-node" import.

https://github.com/elastic/apm-agent-nodejs/issues/608 was a similar issue a while back, for which https://github.com/elastic/apm-agent-nodejs/pull/638 added a note to https://www.elastic.co/guide/en/apm/agent/nodejs/current/agent-api.html#apm-start

trentm:
Note https://github.com/elastic/apm-agent-nodejs/issues/2318

trentm:
- TypeScript will elide the following import:

    ```js
    import apm from 'elastic-apm-node/start'
    ```

    One can avoid that elision with:

    ```js
    import 'elastic-apm-node/start'
    ```

    which is subtle enough in my experience to surprise users, because it means that commenting out the second line (which may be 100s of lines into the file) has a huge impact:

    ```js
    import apm from 'elastic-apm-node/start'
    ...
    console.log('curr trans is:', apm.currentTransaction)
    ```

    Any use of the `apm` import will do. So we could suggest this usage to be more explicit:

    ```js
    import apm from 'elastic-apm-node/start'; apm // Ensure import is not elided by tsc.
    ```

    The canonical TypeScript reference is https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-imports-being-elided-in-my-emit

- I am not yet positive if this at the top of a TypeScript file, or used in ESM (e.g. with node v16) suffices to have that `Agent#start()` setup hooks before subsequent imports:

    ```js
    import apm from 'elastic-apm-node/start'
    import {Client} from 'pg'
    ```

    Or if the sometimes-suggestion to have a separate "apm.{js,mjs,ts}" file is preferred:

    ```js
    // app.ts
    import apm from './apm'

    // apm.ts
[...]

Issue #2578, opened 2022-02-15
could we log.warn if `require.cache` already has modules we typically instrument at `agent.start()` time?


trentm:
A reasonably common mistake when using the APM agent is to `.start()` it *after* having `require()`d some modules that the agent *would* have instrumented.  We could probably look through `require.cache` at `Agent#start()`-time to see if it already had imported modules that the agent *would* have instrumented (modulo `active` and `disableInstrumentations`), and if so, then emit a `log.warn('yo, you need to apm.start() earlier')`. Then that log message could be in the troubleshooting docs to explain the issue.

trentm:
This could be interesting in combination with https://github.com/elastic/apm-agent-nodejs/issues/1967

trentm:
If/when this is done, having a reference to this in https://www.elastic.co/guide/en/apm/agent/nodejs/current/starting-the-agent.html might be worthwhile as well.

trentm:
OTel does this in `@otel/instrumentation`:

```js
    _warnOnPreloadedModules() {
        this._modules.forEach((module) => {
            const { name } = module;
            try {
                const resolvedModule = require.resolve(name);
                if (require.cache[resolvedModule]) {
                    // Module is already cached, which means the instrumentation hook might not work
                    this._diag.warn(`Module ${name} has been loaded before ${this.instrumentationName} so it might not work, please initialize it before requiring ${name}`);
                }
            }
            catch (_a) {
                // Module isn't available, we can simply skip
            }
        });
    }
```

trentm:
# first take



# modules we instrument that we also load in the agent

Here is a quick patch to index.js to list the modules that we instrument that we also load as part of the APM agent.

```diff
diff --git a/index.js b/index.js
index 04711888..8899840d 100644
--- a/index.js
+++ b/index.js
@@ -6,6 +6,44 @@

 'use strict'

+function intersection (setA, setB) {
+  const _intersection = new Set()
+  for (const elem of setB) {
+    if (setA.has(elem)) {
+      _intersection.add(elem)
+    }
+  }
+  return _intersection
+}
+
+var modPathsBefore = new Set(Object.keys(require.cache))
+
 var Agent = require('./lib/agent')

 module.exports = new Agent()
+
+var modPathsAfter = new Set(Object.keys(require.cache))
+function difference (setA, setB) {
+  const _difference = new Set(setA)
+  for (const elem of setB) {
+    _difference.delete(elem)
+  }
+  return _difference
[...]

PR #2837, opened 2022-07-22
docs: improve docs around bundlers/transpilers, methods of starting the agent

trentm:
Closes: #1967

## review notes

Here are some highlights of what changes this PR makes to the docs:

- A re-write of the ['starting-the-agent' doc page](https://apm-agent-nodejs_2837.docs-preview.app.elstc.co/guide/en/apm/agent/nodejs/master/starting-the-agent.html) to:
    - Discuss all the ways to require/import and start the agent.
    - Discuss all the confusing gotchas with different tech. There are subtle diffs between them.
- Update [supported-technologies doc page](https://apm-agent-nodejs_2837.docs-preview.app.elstc.co/guide/en/apm/agent/nodejs/master/supported-technologies.html):
    - Add the latest node release schedule svg (we had an ancient one). Also add a nuance to the "Node.js versions" section stating that we still *do* test with 8.6 and will only drop node version support in a major.
    - Add an ESM section describing the current limitation, and how to configure compilers/bundlers to emit CommonJS module code.
- *Remove* the "es-modules.html" page, which was incomplete and misleading. It has [a redirect page](https://apm-agent-nodejs_2837.docs-preview.app.elstc.co/guide/en/apm/agent/nodejs/master/es-modules.html) to the relevant new content mentioned above.
- Add a ["Get started with TypeScript" doc](https://apm-agent-nodejs_2837.docs-preview.app.elstc.co/guide/en/apm/agent/nodejs/master/typescript.html), and an "[./examples/typescript/](https://github.com/elastic/apm-agent-nodejs/tree/trentm/docs-bundlers-ts-etc/examples/typescript)" using the APM agent.
- Add an "[./examples/esbuild/](https://github.com/elastic/apm-agent-nodejs/tree/trentm/docs-bundlers-ts-etc/examples/esbuild)" using the APM agent. (Doing similar for Webpack would perhaps be helpful, but I haven't added one here.)
- Ensure 'npm run test:babel' fails if it doesn't use `import apm from 'elastic-apm-node/start'` to start the agent before other imports.


docs/set-up.asciidoc
```diff
@@ -39,32 +40,278 @@ include::./custom-stack.asciidoc[]
...
+
+[source,js]
+----
+var apm = require('elastic-apm-node');
+var http = require('http');
+apm.start() // This is started too late.
+// ...
+----
+
+All imports are "hoisted" to the top of a module, properly following ECMAScript module (ESM) semantics. However, the `apm.start()` is now too late: *after* the `http` module has been imported. This is why <<start-option-require-start-module,the `elastic-apm-node/start` module>> was added as a start option. The following will work:
```
bmorelli25, line 171: I think an em dash works better than a colon here.

```suggestion
All imports are "hoisted" to the top of a module, properly following ECMAScript module (ESM) semantics. However, the `apm.start()` is now too late—*after* the `http` module has been imported. This is why <<start-option-require-start-module,the `elastic-apm-node/start` module>> was added as a start option. The following will work:
```

docs/set-up.asciidoc
```diff
@@ -39,32 +40,278 @@ include::./custom-stack.asciidoc[]
...
+
+[[start-gotchas]]
+==== Start Gotchas
+
+This section shows some sometimes subtle surprises starting the APM agent with some technologies. A general troubleshooting tip for using the Elastic Node.js APM agent with any build tool/system that produces compiled JavaScript is to **look at the compiled JavaScript** to see what is actually being executed by `node`.
+
+[[start-esm-imports]]
+===== Hoisted ES module imports
+
+When using a compiler/transpiler (such as https://babeljs.io/docs/en/[Babel] or https://esbuild.github.io[esbuild]) to translate from code using ES modules (`import ...` statements) to CommonJS modules (`require(...)`) there is a common surprise for starting the Elastic Node.js APM agent. Running Babel on the following produces code that does *not* work as expected:
```
bmorelli25, line 150: So the "gotcha" here is that imports are hoisted to the top of a module. But in the intro, we only call it a "common surprise". I wonder if the hoisted explanation should be moved to the top so that we immediately explain the meat of the issue. For example:


> When using a tool like Babel or esbuild to translate/transpile from code using ES modules (i.e. `import ...` statements) to code using CommonJS (i.e. `require(...)`), all imports are "hoisted" to the top of a module, properly following ECMAScript module (ESM) semantics. This means the `apm.start()` method is called too late—*after* the `http` module has been imported.

> For example, running Babel on the following code does not initiate APM early enough:

> **example and equivalent example**

> The <<start-option-require-start-module,the `elastic-apm-node/start` module>> fixes this problem. The following will work:

> **example**

> A more complete example...

docs/set-up.asciidoc
```diff
@@ -39,32 +40,278 @@ include::./custom-stack.asciidoc[]
...
+----
+
+One can avoid that elision with:
+
+[source,js]
+----
+import 'elastic-apm-node/start';
+----
+
+or with something like this:
```
bmorelli25, line 204: Further down you capitalize the Or between code examples.

```suggestion
Or with something like this:
```

docs/supported-technologies.asciidoc
```diff
@@ -7,19 +7,74 @@ endif::[]
...
+[[compatibility-node]]
+=== Node.js versions
+
+Support for the Elastic APM Node.js agent follows the https://nodejs.org/en/about/releases/[support schedule of Node.js itself]
+to the end-of-life period of each version after its maintenance term.
 Versions of Node.js past their end-of-life date are not supported.
 
 image::./images/node_release_schedule.svg[Node.js release schedule]
 
+Having said that, the current APM agent version (3.x) works with Node.js versions back to v8.6. We will only break support for older Node.js versions with a *major version release* of the APM agent.
```
bmorelli25, line 23: Do we need this?

```suggestion
The current APM agent version (3.x) works with Node.js versions back to v8.6. We will only break support for older Node.js versions with a *major version release* of the APM agent.
```

bmorelli25: I'm also not sure that the bold adds anything here. The purpose of the sentence is to explain when support breaks, but only bolding *major version release* doesn't emphasize that fact.

docs/supported-technologies.asciidoc
```diff
@@ -7,19 +7,74 @@ endif::[]
...
+
+[float]
+[[compatibility-esm]]
+=== ES Modules
```
bmorelli25, line 27: I like how you've explained import vs require in this section.

docs/typescript.asciidoc
```diff
@@ -0,0 +1,85 @@
...
+// ...
+----
+<1> This start method requires you to use environment variables to configure the agent. See <<starting-the-agent>> for all the ways to start the agent.
+
+Pay special attention to <<start-typescript,the possible surprise gotcha>> where the TypeScript compiler can throw away your import in the generated JavaScript.
+
+
+[float]
+[[typescript-full-documentation]]
+==== Full documentation
```
bmorelli25, line 78: What about "Next steps" or something like that? "Full documentation" doesn't really explain this section.

docs/set-up.asciidoc
```diff
@@ -39,32 +40,278 @@ include::./custom-stack.asciidoc[]
 
 include::./lambda.asciidoc[]
 
+include::./typescript.asciidoc[]
```
bmorelli25, line 43: What do you think of swapping the order of these includes so that all "Get started with..." pages are grouped together? 

Screen Shot 2022-07-28 at 2 30 07 PM

docs/agent-api.asciidoc
```diff
@@ -7,22 +7,10 @@ endif::[]
...
-
-The Elastic APM agent for Node.js is a singleton.
-You get the agent instance by either requiring `elastic-apm-node` or `elastic-apm-node/start`.
-For details on the two approaches,
-see the <<advanced-setup,Setup and Configuration>> guide.
-
-The agent is also returned by the `start()` function,
-which allows you to require and start the agent on the same line:
+NOTE: This is API reference documentation for the Elastic APM Node.js Agent.
+For getting started, we recommend that you <<set-up,start here>>.
```
bmorelli25, line 11: What about something like this instead:

```
To get started, see <set up the agent>.
New to Elastic APM? <Start here>.
```

docs/troubleshooting.asciidoc
```diff
@@ -115,9 +115,7 @@ Make sure that the agent is *both required and started at the very top of your m
 It's important that the agent is started before any other modules are
 `require`d.  If not, the agent will not be able to hook into any modules and
 will not be able to measure the performance of your application.
-
-IMPORTANT: If you are using Babel / ES modules in your application,
-make sure you have read about the <<es-modules,Babel / ES Module support>>.
+See <<start-gotchas>> for a number of possible surprises in correctly starting the APM agent when used with some compilers or bundlers.
```
bmorelli25, line 118: Are there other reasons users would end up on the troubleshooting page and need to see the start gotchas? Is this link prominent enough?

docs/set-up.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[set-up]]
 == Set up the Agent
 
-To get you off the ground, we've prepared guides for setting up the Agent with a few different popular web frameworks:
+To get you off the ground, we've prepared guides for setting up the Agent with a few different popular web frameworks and technologies:
```
bmorelli25, line 4: This problem existed before this PR:

A link to starting the agent is buried under "other useful documentation." Do we need to explain the starting the agent page here? Maybe how it relates to the "Get started with..." content?

bmorelli25 (commented):
This is looking really good! Here are some nitpicks, questions, and ideas for you to consider.

docs/set-up.asciidoc
```diff
@@ -39,32 +40,278 @@ include::./custom-stack.asciidoc[]
...
 [source,js]
 ----
-var apm = require('elastic-apm-node').start({
-  // add configuration options here
+const apm = require('elastic-apm-node/start');
+
+// Application main code goes here.
+----
+
+When authoring code using CommonJS (i.e. using `require(...)`), this method is rarely used. However, when using a tool (such as Babel or esbuild) to translate/transpile from code using ES modules (i.e. `import ...` statements) to code using CommonJS, this start method is necessary to ensure that the APM agent is _started before other imports in the same file_. See <<start-esm-imports>> below for details.
```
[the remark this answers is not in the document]
trentm: > Also, what do you mean when you say "rarely used"?

There isn't a use case for this beyond saving a couple keystrokes over `require('elastic-apm-node').start()` at the cost of not having a way to pass options to that `.start()` call.

@bmorelli25  How about I just drop the reference to CommonJS usage completely? There is no harm if a `require`-using programmer uses this form. So something like this:

> This start method exists for those that use a tool like Babel or esbuild to translate/transpile from code using ES modules (as in the following example) to code using CommonJS. This start method ensures that the APM agent is _started before other imports in the same file_. See ... below for details.

docs/set-up.asciidoc
```diff
@@ -39,32 +40,278 @@ include::./custom-stack.asciidoc[]
...
+import 'initapm'
+
+// Application code starts here.
+----
+
+
+[[start-gotchas]]
+==== Start Gotchas
+
+This section shows some sometimes subtle surprises starting the APM agent with some technologies. A general troubleshooting tip for using the Elastic Node.js APM agent with any build tool/system that produces compiled JavaScript is to **look at the compiled JavaScript** to see what is actually being executed by `node`.
```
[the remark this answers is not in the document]
trentm: I might ahve been shouting at you. ;)

> Also, This section shows some sometimes subtle surprises starting...

Thanks for noticing! I am (unreasonably?) proud of that alliteration. :)

docs/typescript.asciidoc
trentm: Sounds good to me. I'll change that.

FWIW, this section was originally derived from the "Full documentation" sections copied into the many "Get started with ..." docs that we have here. E.g.: https://www.elastic.co/guide/en/apm/agent/nodejs/current/restify.html#restify-full-documentation  https://www.elastic.co/guide/en/apm/agent/nodejs/current/fastify.html#fastify-full-documentation  etc.

astorm (approved):
Approving. Overall it says true things and gives us a better place to point folks working with TypeScript and other compiled languages.  I didn't go full tech review on all the examples but what's there seems good.  Approving presuming the more _written word_ advice/feedback from @bmorelli25 in incorporated/resolved.

docs/set-up.asciidoc
trentm: Yup, I'm fine with that. I'll move it up.

A couple notes on this.

1. The "Monitoring AWS Lambda ..." entry is named that in the few APM agents that added Lambda support. E.g. https://www.elastic.co/guide/en/apm/agent/python/current/index.html   I guess that page title has better Google juice than "Get started with AWS Lambda". Dunno. It *is* basically still a "getting started" page.

2. I had originally started making larger TOC changes here, but moved that to separate work (if I get to it). I was thinking a TOC something like this:

```
TOC                     # add brief intro prose to this page
(Getting started)
Introduction
Get started    # fastest 1-2-3 get started, with links to tech-specific quick start pages
    Get started with Express
    Get started with hapi
    Get started with Koa
    Get started with Fastify
    Get started with Restify
    *Get started with TypeScript    # ???
    *Get started with ...           # ??? Next.js, etc.
    ~Get started with a custom Node.js stack~  # maybe drop this
Supported technologies
Features        # new page that gives a one pager with a short pitch for defining features, inspiration https://tailscale.com/kb/1169/features/, I'm not married to doing this

(Reference)
Starting the agent          # maybe incorporate the "ES Modules support" (incl. TypeScript) docs in this page? Isn't quite right. Hrm.
Configuration               # merge "Configuration", "Configuring the agent", and "Configuration options" into one page here. *Consider* separate subpages for related-groups of config (a la Java), though I *really* like having one searchable doc page for all config. Perhaps just reconsider the TOC organization of the page and still do in a single page.
Tracing APIs                # I like some of the extra content here from Java docs
    Agent API
    Transaction API
    Span API
    OpenTelemetry Bridge
    OpenTracing Bridge

(Feature References, Howtos, etc.)    XXX perhaps re-order these.
Metrics
Log correlation
Source map support
Distributed tracing
Message queues
Manual instrumentation      # Merge "custom transactions" and "Custom spans" into this

(Extra stuff)
Performance Tuning
Troubleshooting
Upgrading
    ...
Release notes
    ...
```

docs/set-up.asciidoc
trentm: Sounds good to me. Interesting (well, "interesting" if you are text converter pedant, nerd, or just really bored) that asciidoc (and GitHub markdown) will convert double-dash, `--`, to an emdash. However asciidoc has a bug where if there are no spaces around it, then it will not do that conversion.

docs/set-up.asciidoc
trentm: Perhaps.  I agree that it is somewhat buried. I think the "starting the agent" doc page is *almost* really just reference material and that it should hopefully only need to be referred to for the minority of users. It can be linked to from each of the "getting started" guides.

I'd prefer to punt any of this to a separate change if that is okay... perhaps as part of some of the TOC shuffling mentioned above at https://github.com/elastic/apm-agent-nodejs/pull/2837#discussion_r932728937

docs/troubleshooting.asciidoc
trentm: tl;dr: The only side-effect I can think of right now of starting the agent too early (i.e. hitting a start gotcha) is that the user doesn't get expected auto-instrumentation. So, mostly, I think "no". This is the main reason to end up here.

However, fair question. I think this section of the troubleshooting page could be improved a lot.  Truth be told I ran out of steam on this PR (I was many more days into than I had planned for).  I was going for the minimum change to replace the `<<es-modules>>` link that was dropped.

Longer term I'd like to attempt https://github.com/elastic/apm-agent-nodejs/issues/2578 to have a relatively prominent log warning if we can detect a too-late agent start.  As well (and perhaps after that), this troubleshooting doc could quote examples of those log warnings so theoretically a google search for particular warnings would hit this page that could have a better explanation.

docs/agent-api.asciidoc
trentm: Thoughts on just dropping the `NOTE: ...` completely?  The TOC sidebar should be sufficient to guide them to the top of the docs.

trentm: Dropped in commit c9fb5c34

docs/set-up.asciidoc
trentm: Moved up in commit c9fb5c34

docs/typescript.asciidoc
trentm: Done in commit c9fb5c34

docs/set-up.asciidoc
trentm: Done in commit c9fb5c34

docs/configuration.asciidoc
```diff
@@ -33,7 +33,6 @@ will be normalized to the allowed characters. If the name cannot be inferred
 from package.json, then a fallback value of "unknown-nodejs-service" is used.
 
 
-[float]
 [[service-node-name]]
```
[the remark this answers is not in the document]
trentm: Yah, the float was making the entry indented in the "On this page" TOC sidebar:

Screen Shot 2022-07-28 at 5 30 22 PM

docs/set-up.asciidoc
```diff
@@ -1,7 +1,7 @@
 [[set-up]]
```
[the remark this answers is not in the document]
trentm: Yah, it drives me crazy. And an `<<advanced-setup>>` anchor that is labelled "Configuration", but a separate `<<configuration>>` anchor that is labelled "Configuration options".  Neither of these are in the "Set up the Agent" section.

docs/set-up.asciidoc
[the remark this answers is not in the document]
bmorelli25: Okay, can I backtrack here? I think you should keep this in. It's fun.

docs/agent-api.asciidoc
bmorelli25: ++ Love it

docs/set-up.asciidoc
bmorelli25: That's fine 👍

docs/redirects.asciidoc
```diff
@@ -41,4 +41,12 @@ This endpoint has moved. Please see <<transaction-add-labels>>.
 [role="exclude",id="get-started"]
```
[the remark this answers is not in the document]
bmorelli25: Yeah! They're perfect for this use case.

docs/troubleshooting.asciidoc
bmorelli25: Fair enough. I think this is fine for now.

docs/set-up.asciidoc
```diff
@@ -35,36 +36,284 @@ include::./restify.asciidoc[]
...
+
+For the Node.js APM agent to be able to fully function, it *must be started before `require(...)` statements for other modules*. The APM agent automatically instruments modules by interposing itself in the import process. If a given module is imported before the APM agent has started, then it won't be able to instrument that module.
+
+
+==== Start methods
+
+[[start-option-require-and-start]]
+===== `require('elastic-apm-node').start(...)`
+
+The most common way to start the APM agent is to require the `elastic-apm-node` module and call the <<apm-start,`start`>> function at the top of your main module. This allows you to use any of the methods to <<configuring-the-agent,configure the agent>>.
```
bmorelli25, line 61: How do you feel about updating this to `.start()`? This would (1) match [this line](https://github.com/elastic/apm-agent-nodejs/pull/2837/files#diff-23427ac409c61f7de0225443c0c7c77aca90b732e73957bad73edb357c30addeR10), and (2) provide just slightly more information to the user to hopefully prevent them from clicking away (unless that's what they actually want to do). Also, does function vs method matter? We use method above and function here.

```suggestion
The most common way to start the APM agent is to require the `elastic-apm-node` module and call the <<apm-start,`.start()`>> function at the top of your main module. This allows you to use any of the methods to <<configuring-the-agent,configure the agent>>.
```

docs/set-up.asciidoc
```diff
@@ -35,36 +36,284 @@ include::./restify.asciidoc[]
...
+// Application main code goes here.
+----
+
+A limitation of this approach is that you cannot configure the agent with an options object, but instead have to rely on <<configuring-the-agent,one of the other methods of configuration>>, such as setting `ELASTIC_APM_...` environment variables.
+
+
+[[start-option-node-require-opt]]
+===== `node -r elastic-apm-node/start ...`
+
+Another way to start the agent is with the `-r elastic-apm-node/start` https://nodejs.org/api/cli.html#-r---require-module[command line option to `node`]. This will import and start the APM agent before your application code starts. This method allows you to enable the agent _without touching any code_. This is the recommended start method for <<lambda,monitoring AWS Lambda functions>>.
```
bmorelli25, line 100: This anchor doesn't work for me—I get linked to the bottom of the page. Weirdly, it does seem correct, and it does work when clicked in the Node.js docs.

bmorelli25: Hmm. It works in incognito. Must be something on my end.

bmorelli25 (approved):

docs/set-up.asciidoc
trentm: Done in commit d410b09a and commit 0e305c1c.

It is a method (in that it is a function that is tied to a "class", inasmuch as JS has classes). Thanks.

docs/set-up.asciidoc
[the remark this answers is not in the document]
bmorelli25: [`e76ab76`](https://github.com/elastic/apm-agent-nodejs/pull/2837/commits/e76ab76f6e4e8bac1b8e56ae78ead352f2f1e767) 😍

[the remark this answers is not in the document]
trentm: Ti esrever dna ti pilf nwod gnaht ym tup i (commit e76ab76f)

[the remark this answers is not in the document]
bmorelli25: uɐɟ ƃᴉq ɐ ɯ,I

PR #2843, opened 2022-07-25
fast exit spans

astorm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/2301

Refs: https://github.com/elastic/apm/issues/496

Introduces the `exitSpanMinDuration` configuration value.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -56,6 +56,7 @@ var DEFAULTS = {
   disableSend: false,
   environment: process.env.NODE_ENV || 'development',
   errorOnAbortedRequests: false,
+  exitSpanMinDuration: '0ms',
   filterHttpHeaders: true,
   globalLabels: undefined,
   ignoreMessageQueues: [],
@@ -122,6 +123,7 @@ var ENV_TABLE = {
   disableInstrumentations: 'ELASTIC_APM_DISABLE_INSTRUMENTATIONS',
   disableSend: 'ELASTIC_APM_DISABLE_SEND',
   environment: 'ELASTIC_APM_ENVIRONMENT',
+  exitSpanMinDuration: 'ELASTIC_APM_EXIT_SPAN_MIN_DURATION',
   ignoreMessageQueues: ['ELASTIC_IGNORE_MESSAGE_QUEUES', 'ELASTIC_APM_IGNORE_MESSAGE_QUEUES'],
   errorMessageMaxLength: 'ELASTIC_APM_ERROR_MESSAGE_MAX_LENGTH',
   errorOnAbortedRequests: 'ELASTIC_APM_ERROR_ON_ABORTED_REQUESTS',
@@ -228,6 +230,12 @@ var DURATION_OPTS = [
     name: 'apiRequestTime',
     defaultUnit: 's',
     allowedUnits: ['ms', 's', 'm'],
+    allowNegative: false
+  },
+  {
+    name: 'exitSpanMinDuration',
+    defaultUnit: 'ms',
+    allowedUnits: ['us', 'ms', 's', 'm'],
     allowNegative: false
   },
   {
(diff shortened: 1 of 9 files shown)


astorm:
jenkins run the jobs please

lib/config.js
```diff
@@ -122,6 +123,7 @@ var ENV_TABLE = {
   disableInstrumentations: 'ELASTIC_APM_DISABLE_INSTRUMENTATIONS',
   disableSend: 'ELASTIC_APM_DISABLE_SEND',
   environment: 'ELASTIC_APM_ENVIRONMENT',
+  exitSpanMinDuration: 'ELASTIC_APM_EXIT_SPAN_MIN_DURATION',
```
trentm, line 126: The spec says this is a config var that could come from central config, so an entry should be added to `CENTRAL_CONFIG` and an addition made to test/central-config-enabled.test.js.

trentm (changes requested):

lib/config.js
astorm: ✅

lib/instrumentation/index.js
```diff
@@ -381,6 +381,12 @@ Instrumentation.prototype.addEndedSpan = function (span) {
     return
   }
 
+  const duration = span.isComposite() ? span.getCompositeSum() : span.duration()
+  if (span.discardable && duration / 1000 < this._agent._conf.exitSpanMinDuration) {
+    span.transaction.captureDroppedSpan(span)
+    return
+  }
```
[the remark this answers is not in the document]
astorm: You know -- as I was adding the ternary for grabbing the composite sum a small voice said "you should probably write a test for the compressed case" and another voice said "nah, it'll be fine".  I probably should have listened to that first voice. 

Moved the logic, added a test.

test/instrumentation/drop-fast-exit-spans.test.js
```diff
@@ -66,6 +66,48 @@ tape.test('end to end test', function (t) {
   }, 20) // longer than exitSpanMinDuration
 })
 
+tape.test('end to end test with compression', function (t) {
+  resetAgent(function (data) {
+    // would normally be a single compressed span, but here there's
+    // no spans because min duration was not hit
+    t.equals(data.spans.length, 0)
```
trentm, line 73: Could we change this test (or add a separate one) to check the case of a compressed span *not* being dropped, even if all of its included spans each have a duration that is less than `exitSpanMinDuration`?

test/instrumentation/drop-fast-exit-spans.test.js
```diff
@@ -0,0 +1,73 @@
...
+const agent = require('../..').start({
+  serviceName: 'test-fast-exit-span',
+  breakdownMetrics: false,
+  captureExceptions: false,
+  metricsInterval: 0,
+  centralConfig: false,
+  cloudProvider: 'none',
+  exitSpanMinDuration: '10ms',
+  spanStackTraceMinDuration: 0, // Always have span stacktraces.
+  transport () { return new CapturingTransport() }
```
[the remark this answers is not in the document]
trentm: It was a nit, so totally fine to skip it, but is there a reason to keep those two config vars on the agent in this test file? They are slightly misleading for a future maintainer of this test file.

trentm (approved):

astorm:
> It was a nit, so totally fine to skip it, but is there a reason to keep those two config vars on the agent in this test file? They are slightly misleading for a future maintainer of this test file.

No real intent here @trentm -- those configs are just part of whatever boilerplate I copy/pasted to get started with the tests and I didn't really want to think through whether removing them about have any impact.  I don't think that's all that confusing -- (or no more so than other things in this codebase) -- but I don't have any objections if you wanted to remove them.

test/instrumentation/drop-fast-exit-spans.test.js
astorm: No objections if you wanted to take that on and think there's value there.

PR #2852, opened 2022-08-02
docs: correct the envvar name to use for spanStackTraceMinDuration

trentm:


astorm (approved):
✅ not now/this PR, but another possible low hanging fruit for a docs linter would be to validate that the env var in docs exist in the agent.

Issue #2383, opened 2021-10-20
add undici HTTP client instrumentation


trentm:
Add instrumentation support for [undici](https://github.com/nodejs/undici).
`@elastic/elasticsearch@8` is (optionally) using it now and it is gaining traction in the community.

trentm:
@mcollina Thanks! Your GitHub-notification-fu is strong. :)   @delvedor Has been encouraging and suggesting diagnostics_channel as well.

trentm:
Here is a trace of a small undici-using script that makes four calls to a local HTTP server, using `pipelining: 2`:

<img width="1265" alt="Screen Shot 2022-08-04 at 2 54 49 PM" src="https://user-images.githubusercontent.com/46866/182959855-68653e2a-114f-43cf-a7a6-41d9ddc671b2.png">

PR #2858, opened 2022-08-04
feat: undici and `fetch()` instrumentation

trentm:
Closes: #2383


trentm:
run module tests for undici,@elastic/elasticsearch

trentm:
run module tests for undici,@elastic/elasticsearch

test/context-propagation-only.test.js
```diff
@@ -191,7 +191,7 @@ tape.test('contextPropagationOnly', function (suite) {
       // stacktrace collection when contextPropagationOnly=true. It isn't a
       // perfect way to test that.
       const durationMs = duration[0] / 1e3 + duration[1] / 1e6
-      const THRESHOLD_MS = 3 // Is this long enough for slow CI?
+      const THRESHOLD_MS = 5 // Is this long enough for slow CI?
```
trentm, line 194: REVIEW NOTE: This is an unrelated change. I hit a spurious test failure with this tight threshold, so I bumped the threshold to make it less likely to hit it in CI.

trentm: For example this just failed on "main": https://github.com/elastic/apm-agent-nodejs/runs/7771135311?check_suite_focus=true

```
not ok 16 captureError is fast (<3ms): 3.015229ms
```

lib/instrumentation/index.js
```diff
@@ -209,6 +213,11 @@ Instrumentation.prototype.start = function (runContextClass) {
 
   this._runCtxMgr.enable()
   this._startHook()
+
+  if (nodeHasInstrumentableFetch && this._isModuleEnabled('undici')) {
+    this._log.debug('instrumenting fetch')
+    undiciInstr.instrumentUndici(this._agent)
```
astorm, line 219: (I'm afraid to ask but) -- what's driving the need for this manual call to `instrumentUndici`.  Naively -- it seems like adding undichi to `MODULES` will always load `lib/instrumentation/modules/undici.js` which also calls `instrumentUndici`.  What's the weird case that leads us to  need this second, earlier, call to `instrumentUndici`?

astorm (approved):
I ran this through some rudimentary tests -- spans are made, tracecontext is propagated/distributed tracing functions -- things seem to work and having this in place puts us ahead of most folks w/r/t `undici` and any bleeding edge folks bold enough to incorporate the new HTTP client into their projects (:wave: @ elastic client team)

Fine with this going in as is but I do have a question re: what looks like double instrumentation below, and I'm not sure if the lack of a `CHANGELOG` entry is deliberate or just an oversight.

lib/instrumentation/index.js
trentm: > adding undichi to MODULES will always load lib/instrumentation/modules/undici.js 

It will load `lib/instrumentation/modules/undici.js`, yes, but the shimming (`shimUndici`) isn't called unless the require-in-the-middle hook sees a `require('undici')`.

> What's the weird case that leads ...

To instrument node's core `fetch()` implementation we need to enable the undici instrumentation **even if there is no `require('undici')`** anywhere. This is because node's core `fetch()` uses undici as an esbuild bundle: https://github.com/nodejs/node/blob/main/deps/undici/undici.js
That we can even instrument that is down to the hooks being via diagnostics_channel rather than needing to monkey-patch.

And to avoid double-instrumenting, there is a guard at the start of `instrumentUndici`:

```js
function instrumentUndici (agent) {
  if (isInstrumented) {
    return
  }
  isInstrumented = true
...
```


> I'm not sure if the lack of a CHANGELOG entry is deliberate or just an oversight.

Oh, that is totally an oversight. Thanks for pointing that out!

PR #2866, opened 2022-08-09
fix: an S3 ListBuckets span was invalid for APM server intake

trentm:
The APM agent spec for S3 spans says
`span.context.destination.service.resource` should be the bucket name
or access point, if available. That isn't available for ListBuckets,
which doesn't operate on a bucket. In this case the APM agent was emiting:

    span.context.destination.service = { name: ..., type: ... }

which fails APM server schema validation with

    missingProperty: 'resource'

and results in APM server intake logs with:

    Error: validation error: span: context: destination: service: 'resource' required

The fix, for now, is to skip `span.context.destination.service` entirely
for ListBuckets.

----

Dropping `span.context.destination.service` for `ListBuckets` spans will mean the span will be accepted by APM server (good), but won't result in "S3" being recognized as a backend for the service (not the best, but any other S3 spans will make that connection).

Longer term, the [coming](https://github.com/elastic/apm-agent-nodejs/issues/2822) replacement of `span.context.destination.service.*` fields with `span.context.service.target.*` fields provides a mechanism to handle this better.

### DIFF
--- a/test/_validate_schema.js
+++ b/test/_validate_schema.js
@@ -1,0 +1,52 @@
+/*
+ * Copyright Elasticsearch B.V. and other contributors where applicable.
+ * Licensed under the BSD 2-Clause License; you may not use this file except in
+ * compliance with the BSD 2-Clause License.
+ */
+
+'use strict'
+
+// Validate an APM server intake object against its schema.
+//
+// Usage example:
+//    const { validateSpan } = require('./_validate_schema')
+//    const errs = validateSpan(mySpanObj)
+// `errs` is null if mySpanObj is valid, else it is an array of ajv ErrorObject.
+
+const fs = require('fs')
+const { join } = require('path')
+
+const Ajv = require('ajv').default
+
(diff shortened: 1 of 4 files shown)


test/_validate_schema.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+'use strict'
+
+// Validate an APM server intake object against its schema.
+//
+// Usage example:
+//    const { validateSpan } = require('./_validate_schema')
+//    const errs = validateSpan(mySpanObj)
+// `errs` is null if mySpanObj is valid, else it is an array of ajv ErrorObject.
```
trentm, line 14: REVIEW NOTE: Note that this file somewhat duplicates facilities in "test/integration/api-schema/_utils.js". However those are a little baroque, undocumented, and being async (unnecessarily), means its `fooValidator` functions are hard to use in synchronous test code.  I think this `_validate_schema.js` utility module will be a more convenient basis for doing more schema validation in our tests.

astorm (approved):

PR #2869, opened 2022-08-10
chore: convert issue templates to use unix line-endings

trentm:
I was surprised that our issue templates use Windows (`\r\n`) line endings.


astorm (approved):
Ah vbcrlf -- you're looking well.  Surprised this came up at all, most text editors handle this sort of thing seamlessly these days. Approving.

Issue #2019, opened 2021-03-29
Reconsider structure of `context.db.statement`


gingerwizard:
We utilise the Node JS agent to capture Elasticsearch queries from Kibana. For this we need the full content body which is currently captured in the field `context.db.statement`. This value, however, contains the request body and URL (with parameters) delimited by two new lines. In order for us to extract the ES body + index name + any parameters, we currently parse this payload client side.

This obviously works it makes our code susceptible to changes in the agent. Whilst fields are governed by ECS and a change control process, are  changes in the structure of field values also considered potentially breaking changes? If so, could such changes be made in a minor or could we rely on their consistency across majors? 
https://www.elastic.co/guide/en/apm/agent/nodejs/current/release-notes-3.x.html#release-notes-3.10.0 suggests that changes can be made in a minor which is concerning.

This is further complicated by the fact we use the agent version distributed with Kibana, so we would need to track the agent version this uses.

One proposal is that these values be stored in separate fields i.e. only the request body be stored in `context.db.statement` and the params and index names be stored in separate fields. The agent would in effect be responsible for parsing these values. Whilst we would prefer this solution, we appreciate that is a decision beyond a single agent.

@trentm

trentm:
Some background for others. Dale and I discussed this a little bit before this ticket.

Currently the structure of `context.db.statement` for Elasticsearch spans is defined by this code:
https://github.com/elastic/apm-agent-nodejs/blob/c64a089e6d4a814f15c2f111850a9d36e771b39b/lib/instrumentation/elasticsearch-shared.js#L18-L56
That structure is mention in the changelog (Dale linked to it above), but not elsewhere in docs.

Here is the section of the current apm-server spec that governs what fields are allowed in `context.db`: https://github.com/elastic/apm-server/blob/141ee87f27b41f1dc6b03f278fb1afa88ead747e/docs/spec/v2/span.json#L31-L83

That Elasticsearch allows material fields for the query to be in either the query params and/or the body is what led to changing the form in v3.10.0 of the agent. Earlier versions of the Node.js APM agent would only include the query params if the body was empty. The content/structure of `context.db.statement` is not in the specs. My understanding of its intent is only that it shows relevant query data when viewed in the APM UI in Kibana. The structure being using in the Node.js APM agent now matches what the Python APM agent uses. For contrast, the Ruby APM agent uses a serialized JSON object with "params" (the query params) and "body" fields.


[...]

PR #2873, opened 2022-08-12
feat: 'elasticsearchCaptureBodyUrls' config option

trentm:
This adds a new config option, per the
'elasticsearch_capture_body_urls' cross-agent spec, to control for which
ES client requests the request body is captured. By default it is just
the body of search-related APIs. The use case for configuring this is
to capture for *all* requests for reply in Kibana performance work. This
option is not expected to be used often.

Closes: #2880
Closes: #2019
Refs: https://github.com/elastic/apm/issues/665
Refs: https://github.com/elastic/apm/pull/670

# checklist

- [x] finish implementation (docs, some tests)
- [x] CHANGELOG entry
- [x] TAV tests: https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2873/7/
- [ ] perhaps wait for the [spec](https://github.com/elastic/apm/pull/670) to be merged


trentm:
run module tests for elasticsearch,@elastic/elasticsearch

trentm:
# some premature optimization perf notes

I digressed slightly to sanity check that the change in ES URL path checking here (from a custom `pathIsAQuery` regexp, to a WildcardMatcher-created array of regexps) will not be a performance problem.

tl;dr: The WildcardMatcher-based comparison is much slower. However, it is on the order of 1 *microsecond* per path check on my MacBook Pro (2.4 GHz 8-Core Intel Core i9). Given that ES queries are remote calls and on the order of milliseconds, it would generally be premature optimization to bother getting fancy to improve performance of this.

### details

Here is my benchmark script: https://gist.github.com/trentm/ba426e5a08317b18d40e897aeb536c0c

And a run:

```
% node bench-shouldCaptureBody.js
WilcardMatcher array:
	[/^.*\/_search$/i, /^.*\/_search\/template$/i, /^.*\/_msearch$/i, /^.*\/_msearch\/template$/i, /^.*\/_async_search$/i, /^.*\/_count$/i, /^.*\/_sql$/i, /^.*\/_eql\/search$/i]
Current pathIsAQuery RegExp: /\/(_search|_msearch|_count|_async_search|_sql|_eql)(\/|$)/
Naive WildcardMatcher in a single RegExp:
	/^(.*\/_search|.*\/_search\/template|.*\/_msearch|.*\/_msearch\/template|.*\/_async_search|.*\/_count|.*\/_sql|.*\/_eql\/search)$/i
customEquivReIgnoreCase:
	/\/(_search|_search\/template|_msearch|_msearch\/template|_async_search|_count|_sql|_eql\/search)$/i
customEquivRe:
	/\/(_search|_search\/template|_msearch|_msearch\/template|_async_search|_count|_sql|_eql\/search)$/

WildcardMatcher array x 719,208 ops/sec ±0.99% (90 runs sampled)
Current pathIsAQuery RegExp x 15,700,534 ops/sec ±0.97% (90 runs sampled)
Naive WildcardMatcher in a single RegExp x 856,094 ops/sec ±1.10% (88 runs sampled)
customEquivReIgnoreCase x 17,477,871 ops/sec ±1.37% (88 runs sampled)
customEquivRe x 17,942,769 ops/sec ±0.91% (88 runs sampled)
Fastest is customEquivRe
```

1. The WildcardMatcher-based array of regexps that this PR changes to is ~20x slower than the current `pathIsAQuery` RegExp.
2. A naive potential optimization might be to make a WildcardMatcher array config into a *single* regexp -- `^(pattern1|pattern2|...)$`. This turns out to be not much faster, in this case.
3. The major perf difference is that the (a) the WildcardMatcher necessarily always has a `^` anchor and (b) every pattern is leading with `*` to avoid that anchor.  A possible optimization of WildcardMatcher usage would be to drop the `^` anchor and the leading `*` **if and only if every pattern had that leading `*`.**  An equivalent optimization could be done for the '$' anchor and a trailing `*` on every pattern.
4. There is a small perf win with dropping the case-insensitive flag. Arguably the default `elasticsearch_capture_body_urls` values *could/should* be case-sensitive. I don't believe the Elasticsearch REST API would accept `POST /my-index/_SeArCh`.

However, this is all fast enough, so I'm not proposing doing any WildcardMatcher perf improvements right now.

trentm:
@astorm the [spec](https://github.com/elastic/apm/pull/670) for this will get merged tomorrow, so feel free to review this when you get a chance. Thanks.

examples/trace-elasticsearch7.js
```diff
@@ -18,6 +18,8 @@ const apm = require('../').start({ // elastic-apm-node
   logUncaughtExceptions: true
 })
 
+// Note that version *8* is installed by default. To use v7 you'll need to:
```
astorm, line 21: nit: Might be better to say _version 7 is not installed by default_ instead of _Note that version *8* is installed by default_, since the later will inevitably be not true one day but the former will almost certainly be true into the future.

index.d.ts
```diff
@@ -249,6 +249,7 @@ declare namespace apm {
     contextPropagationOnly?: boolean;
     disableInstrumentations?: string | string[];
     disableSend?: boolean;
+    elasticsearchCaptureBodyUrls?: Array<string>;
```
astorm, line 252: No objection, but why `Array<string>` instead of `string[]`?

lib/instrumentation/elasticsearch-shared.js
```diff
@@ -10,64 +10,50 @@
 // - elasticsearch - the legacy Elasticsearch JS client
 // - @elastic/elasticsearch - the new Elasticsearch JS client
 
-const querystring = require('querystring')
```
astorm, line 13: 👍 dings the "two `require('querystring')` references left before we can drop another dep." bell 🛎️

astorm (approved):

index.d.ts
trentm: Honestly just copying syntax from elsewhere in this file. https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#arrays says both are fine and doesn't claim a preference, so I'm fine with either way.  Both are used in index.d.ts.

examples/trace-elasticsearch7.js
trentm: Sounds good to me, thanks.

Note, if interested: For a while I'd had both v7 and v8 installed using an alias for v7 -- `"@elastic/elasticsearch7": "npm:@elastic/elasticsearch@7.x"`. However the default npm in node v8 doesn't support that alias syntax, so I dropped it.

trentm: commit f8631c07

trentm:
elastic-apm-node@3.39.0 is release with this now

PR #2879, opened 2022-08-15
perf: disable Knex instrumentation if spanStackTraceMinDuration=-1

trentm:
Knex instrumentation exists only for better span stack traces. It should
be disabled when not collecting span stack traces. This should be a
perf win for default usage of Knex because we avoid an expensive
`Error.captureStackTrace(obj)` for each wrapped knex query.

Obsoletes: #506

* * *

Here I'm finishing up work started by @watson in #506. I updated for the newer config var (`spanStackTraceMinDuration`), added a test, and tweaked the knex-related docs.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -51,6 +51,10 @@ Notes:
 [float]
 ===== Chores
 
+- Disable knex instrumentation when not collecting span stack traces
+  (because there is no point). This is a performance improvement for
+  Knex usage in the default configuration.
+
 
 [[release-notes-3.38.0]]
 ==== 3.38.0 2022/08/11
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -53,7 +53,7 @@ Notes:
 
 - Disable knex instrumentation when not collecting span stack traces
   (because there is no point). This is a performance improvement for
-  Knex usage in the default configuration.
+  Knex usage in the default configuration. ({pull}2879[#2879])
 
 
 [[release-notes-3.38.0]]
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -153,19 +153,18 @@ so those should be supported as well.
 [[compatibility-better-stack-traces]]
 ==== Better Stack Traces
 
-When viewing a span in Elastic APM,
-you'll see where in your code you initiated the span (e.g. a database query).
+The APM agent <<span-stack-trace-min-duration,can be configured>> to capture
+span stack traces, to show where in your code a span (e.g. for a database query)
+was initiated.
 
-Given the async nature of Node.js,
-it's not possible for us to see further back than the last async boundary.
-We therefore make sure to monitor as close to your code as possible.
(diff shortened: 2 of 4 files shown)


astorm (approved):
Seems a little fiddly but sure, approved.

Any idea of what expensive here means? 1ms a query? 1μs? Or a big old it depends?

trentm:
> Any idea of what expensive here means?

Very small, according to:

```js
var Benchmark = require('benchmark')
var suite = new Benchmark.Suite()

suite

  .add('Error.captureStackTrace', function () {
    var obj = {}
    Error.captureStackTrace(obj)
    // obj.stack
  })

  .on('cycle', function (event) {
    console.log(String(event.target))
  })
  // .on('complete', function () {
  //   console.log('Fastest is ' + this.filter('fastest').map('name'))
  // })
  .run({ async: true })
```

A run of that:

```
% node bench-Error-captureStackTrace.js
Error.captureStackTrace x 199,227 ops/sec ±4.88% (54 runs sampled)
```

means about ~5us, so not significant.

If that `obj.stack` is referenced then that rises to 20us, but `obj.stack` is not referenced when `spanStackTraceMinDuration = -1`.

That cost can rise if the stack depth is larger. In this example it is 9 levels deep:

```
Error
    at Benchmark.<anonymous> (/Users/trentm/el/apm-agent-nodejs3/bench-Error-captureStackTrace.js:14:11)
    at Benchmark.eval [as compiled] (eval at createCompiled (/Users/trentm/el/apm-agent-nodejs3/node_modules/benchmark/benchmark.js:1725:16), <anonymous>:5:138)
    at clock (/Users/trentm/el/apm-agent-nodejs3/node_modules/benchmark/benchmark.js:1658:29)
    at cycle (/Users/trentm/el/apm-agent-nodejs3/node_modules/benchmark/benchmark.js:2007:49)
    at Benchmark.run (/Users/trentm/el/apm-agent-nodejs3/node_modules/benchmark/benchmark.js:2114:13)
    at execute (/Users/trentm/el/apm-agent-nodejs3/node_modules/benchmark/benchmark.js:860:74)
    at Timeout._onTimeout (/Users/trentm/el/apm-agent-nodejs3/node_modules/lodash/lodash.js:2792:43)
```

As https://v8.dev/docs/stack-trace-api says:

> We collect 10 frames because it is usually enough to be useful but not so many that it has a noticeable negative performance impact. You can control how many stack frames are collected by setting the variable
>
>     Error.stackTraceLimit

Unfortunately the APM agent sets `Error.stackTraceLimit = 50` by default.

So yah, fair. This is likely premature optimization. The arguable "fussy" part of the change is this 5 line guard (and the added test file):

```js
  if (agent._conf.spanStackTraceMinDuration < 0) {
    agent.logger.trace('not instrumenting knex because not capturing span stack traces (spanStackTraceMinDuration=%s)',
      agent._conf.spanStackTraceMinDuration)
    return Knex
  }
```

The rest of the patch is about improving some docs around the Knex instrumentation.

OCD-me kinda likes the "if this module isn't going to do anything useful, then turn it off" effect of adding this.  Let me know if you think we should not bother with the guard and I can drop that part (and the test file).

Issue #1897, opened 2020-11-26
"mongodb-core" module instrumentation is missing "destination" context


trentm:
Similar to #1892 the instrumentation for mongodb-core is missing "destination" context.

Issue #1494, opened 2019-10-31
mongo needs span.db.instance support


Qard:
See: elastic/apm-agent-nodejs#1482

trentm:
Dropping `.link` per subsequent discussion on #1482.

PR #2458, opened 2021-11-18
feat: exit spans, infer destination.service.resource

trentm:
This adds the explicit concept of "exit spans" -- spans the represent an
outgoing call to a downstream service. An exit span doesn't have child
spans. (Technically it can have spans of the same type+subtype, but this
agent doesn't have any cases of that.) An exit span is explicitly marked
at creation time by setting the `exitSpan` option to true.
https://github.com/elastic/apm/blob/master/specs/agents/tracing-spans.md#exit-spans

For exit spans the 'span.context.destination.service.resource' is
automatically inferred from other context. This can be overriden with
the new `span.setDestinationService(resource)`.
https://github.com/elastic/apm/blob/master/specs/agents/tracing-spans-destination.md

Fixes: #2103

### Checklist

- [ ] Implement code
    - [ ] inferred http resource from URL (avoid re-parsing)
    - [x] no-child-span handling
    - [ ] perhaps include http context in ES and aws-sdk spans as part of this PR
    - [ ] this likely depends on parent/child run-context fixes from #2430 
- [ ] Add tests
- [x] Update TypeScript typings
- [x] Update documentation
- [ ] Add CHANGELOG.asciidoc entry
- [x] Commit message follows [commit guidelines](https://github.com/elastic/apm-agent-nodejs/blob/master/CONTRIBUTING.md#commit-message-guidelines)

PR #2882, opened 2022-08-17
feat: `service.target.*` to improve backend granularity

trentm:
This adds the new `span.context.service.target.{type,name}` fields for improved granularity on backend services data in exit spans. This data is used for "Service Maps" and "Dependencies" in the Kibana APM app.

All instrumentations have been updated to set appropriate service target values.  `service.target.*` is typically inferred automatically from other span data, so much of the instrumentation work was in adding other span fields, most commonly `span.context.db.instance`. The one current exception to the general inference algorithm is S3 spans, for which the spec'd `service.target.name` doesn't follow the general pattern.

A new, public `span.setServiceTarget(type, name)` API has been added (this is a "SHOULD" in the spec).

The new fields will replace the (now deprecated) `span.context.destination.service.*` fields. In the current stage of transition:
- `destination.service.{type,name}` are set to the empty string. They are no longer used, but the intake API v2 schema up to 7.13 require them to be set.
- Setting `destination.service.resource` directly is discouraged. Typically it is inferred from `service.target.*` values when a span is ended. Again the one exception is S3 spans.

The not-public-but-available-because-JavaScript `span.setDestinationContext` has been deprecated (using it will `process.emitWarning()`) and replaced with an internal `span._setDestinationContext()`.

As part of this change, improvements have been made to some module instrumentations:

- `redis` and `ioredis`: `span.type` has changed from "cache" to "db" per spec (https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#redis)
- `mongodb`: `span.action` used to be "query", now it will be the mongodb command name, e.g. "find", "insert".
- `mongodb` and `mongodb-core`: `span.db.instance` is now set to the database name (#1494)
- `mysql` and `mysql2`: `span.db.{instance,user}` are now populated.
- `@elastic/elasticsearch`: The cluster name is heuristically determined for Elastic Cloud deployments and used for `db.instance`.
- `sqs`: `span.destination.{address,port}` are now populated.
- `pg`: `span.db.{instance,user}` are now populated.
- `cassandra-driver`: the Cassandra keyspace is captured for service target data, if available.
- OpenTelemetry Bridge: OTel spans with kind PRODUCER and CLIENT are now handled as exit spans (e.g. span compression could apply).

Closes: #2621
Closes: #2822
Closes: #1494
Closes: #1897
Closes: #2103
Obsoletes: #2458

# Notes for reviewers

This changes a lot of files. The following might help break that down to reasonable chunks:

- The main part of the change is the general algorithm to set `span.context.service.target.*` in `Span.end()`, and then the algorithm to infer `span.context.destination.service.resource` from the service.target value, also in `Span.end()`.  This is all in "lib/instrumentation/span.js".
- Those two algorithms are being clarified in this spec PR: https://github.com/elastic/apm/pull/674  Mostly that PR is putting pseudo-code in the specs to match what is implemented in the Java agent.
- Changes to specific instrumentations are listed above, but most of the changes are:
  - Because `destination.service.resource` is inferred automatically, instrumentations for all exit spans changed to no longer specify `destination.service` in the call to `span.setDestinationContext()`.
  - Also, the algo to set `service.target.*` uses `context.db.instance` if available, so many of the instrumentations were updated to determine `db.instance` according to spec.
- "lib/instrumentation/span-compression.js" and "lib/instrumentation/dropped-span-stats.js" were updated per spec to use `service.target.*` fields.
- The tests. Most test file updates are just updating `service.target.*` and `destination.service.*` expectations. "Interesting" test files are:
    - "elasticsearch.test.js" testing the handling of the 'x-found-handling-cluster' header for our partial implementation of https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#cluster-name
    - "test/instrumentation/modules/cassandra-driver/_utils.js" improves the Cassandra test utils to separate "keyspace" and "table" for more cleaning testing those. This is more important now that `keyspace` is used for `db.instance` if available.
    - "test/opentelemetry-bridge/otel-bridge-feature.test.js" is a painstaking manual update to the latest "otel_bridge.feature" gherkin spec from apm.git that we cannot use directly
    - "test/service-resource-inference.test.js" is new and adds testing of the shared "json-specs/service_resource_inference.json" from apm.git


trentm:
run module tests for ALL

trentm:
run module tests for @elastic/elasticsearch

trentm:
run module tests for ALL

trentm:
> run module tests for ALL

That didn't take that I could see. I've manually started https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/PR-2882/30/ to do a full TAV run.

lib/constants.js
```diff
@@ -9,6 +9,9 @@
  */
 
 module.exports = {
+  // The default span or transaction `type`.
+  DEFAULT_SPAN_TYPE: 'custom',
```
trentm, line 13: REVIEW NOTE: The spec says the default span.type should be 'custom'. We had been doing that in Span serialization. However, when inferring `span.context.service.target.type` in `Span.end()`, the algorithm needs that `span.type` value. One way was to have a shared constant -- which I've done here. Another way would be to actually set `span.type = 'custom'` as a default.

lib/instrumentation/span.js
```diff
@@ -137,8 +212,86 @@ Span.prototype.setHttpContext = function (context) {
...
+
+/**
+ * The internal method for setting "destination" context.
+ *
+ * "destination.service.resource" should only ever be included for special
+ * cases. It is typically inferred from other fields via a general algorithm.
+ */
+Span.prototype._setDestinationContext = function (destCtx) {
+  this._destination = Object.assign(this._destination || {}, destCtx)
+}
```
astorm, line 246: This is a longer term discussion that goes beyond this PR, but we don't seem to be super consistent on the semantic meaning of method names with a leading underscore. 

This usually means the method is meant to be analogous to a protected or private method in a language with access modifiers. 

However,  we're calling `_setDestinationContext` _as_ a public method in the instrumentation modules so in _this_ instance we seem to be saying that `_setDestinationContext` is a public method but just not a public method that's part of our API.

If we're going to adopt the leading underscore as a convention is would be good to get on the same page as to what it means and more generally how we think about enforcing API boundaries between our own modules.

lib/instrumentation/span.js
```diff
@@ -99,6 +102,78 @@ Span.prototype.end = function (endTime) {
...
+        //   used.
+        // - Use `span.type === "external"`. This, at least currently corresponds.
+        //   Let's use this one.
+        this._destination.service.resource = this._serviceTarget.name
+      } else {
+        this._destination.service.resource = `${this._serviceTarget.type}/${this._serviceTarget.name}`
+      }
+    }
+  }
+
```
astorm, line 176: Dropping 70+ lines into an already busy method seems like -- a lot. This new logic might be better off in its own  method or private module helper function.

lib/instrumentation/span.js
```diff
@@ -137,8 +212,86 @@ Span.prototype.setHttpContext = function (context) {
...
+/**
+ * This is deprecated and will be dropped in a future version. This was always
+ * an internal method, but possibly used by enterprising users of manual
+ * instrumentation.
+ *
+ * @deprecated Users should use the public `setServiceTarget()`.
+ *    Internal APM agent code should use `_setDestinationContext()`.
+ */
+Span.prototype.setDestinationContext = function (destCtx) {
+  process.emitWarning(
```
astorm, line 224: I don't know how necessary this warning is since `setDestinationContext` was never documented as a part of our public API. This is another example of us not being clear about what is and isn't a public API.  

By emitting this warning we seem to be saying this _was_ a part of our public API.

However, by not documenting it we were saying it's not part of our public API. 

By not having a CHANGELOG entry for this change we seem to be saying it was never a part of our public API.

It we want to start emitting these sorts of warnings when we deprecate something I don't object -- I'm just ring the "if we're going to be this fiddly we should come up with some rules" rather than making adhoc decisions about how we're doing things on a PR by PR basis.

astorm (approved):
This is pretty unwieldy and large. I don't think I can digest it for a full top-to-bottom review in a reasonable amount of time.  I've taken a few passes and left some notes. The changes to a both a large number of implementation and a large number of tests at the same time make me a little uncertain of how to ensure we're not creating some small side effects that are going to cause behavior changes for users.

That said -- from a stability point of view the changes appear solid.  The only place where I see potential for new null references are the changes to `getDBDestination` (where we can now return a null value) -- but our own use of this function is to set a span's `_destination` value, and we appear to have appropriate guard clauses on `_destination` to ensure no null references.  I don't think these changes will cause any new crashes and should be OK to release as is, so I'm approving.

lib/instrumentation/span.js
trentm: > I don't know how necessary this warning is

My goal in adding this warning is to *attempt* to help the accidental user of this never-public method to (a) notice and (b) have time to migrate to the new, public `.setServiceTarget()`.

While these `<span>.set<Something>Context()` methods have never been part of our public API, they sure look appealing for those attempting to do manual instrumentation:
- Those users might not pay attention to the TypeScript-y `index.d.ts`,
- those users might not rely solely on the API docs for what APIs they can use,
- there *aren't* any public APIs to use to set some of these context attributes on spans to access features like the "Service Map" and "Dependencies" in APM UI, and
- that these methods are not prefixed with an underscore **hints** that perhaps they are fair game to use.

On that last point, perhaps I am showing my [Python heritage in using a leading underscore as a hint that an attribute is non-public](https://peps.python.org/pep-0008/#method-names-and-instance-variables). Though that leading-underscore practice is used frequently in Node.js-core code as well.

Also, I feel some duty to the customer to whom I offered this code to use for manual instrumentation of Oracle (which we don't support yet): https://gist.github.com/trentm/8c8fcecbec1a99ce0fbb415ef87ae2db
That code uses `.setDbContext()` and `.setDestinationContext()`.

> By not having a CHANGELOG entry for this change we seem to be saying it was never a part of our public API.

I could certainly add a message to the changelog about this. I'll do that. I had been intending to add the following to the commit message (which is already in the PR description) -- but in the changelog is good too:

```
The not-public-but-available-because-JavaScript `span.setDestinationContext` has been deprecated (using it will `process.emitWarning()`) and replaced with an internal `span._setDestinationContext()`.
```


> By emitting this warning we seem to be saying this _was_ a part of our public API.

I'm changing to warning message to the following to attempt to alleviate that concern slightly:

```
'<span>.setDestinationContext() was never a public API and will be removed, use <span>.setServiceTarget().',
```

> "if we're going to be this fiddly we should come up with some rules" rather than making adhoc decisions 

Sounds good. I'll bring up a discussion separately. Perhaps it can lead to developer guide content in DEVELOPMENT.md

lib/instrumentation/span.js
trentm: Good call. I'll move those out to a separate function or method.

lib/instrumentation/span.js
trentm: I responded a little bit below (https://github.com/elastic/apm-agent-nodejs/pull/2882#discussion_r971162895) as well.

Mostly this use of leading underscore to mean "internal to the agent" seemed inline with current common uses of the following in the instrumentation modules:

```
agent._instrumentation
agent._conf
span._setOutcomeFromHttpStatusCode
span._setOutcomeFromErrorCapture
```

However, documenting naming conventions would be good.
I'll bring up a separate discussion on conventions for this.

PR #2883, opened 2022-08-18
ci: use github commit status

v1v:
### What

Enable GitHub commit status for the GitHub checks

### Why

As long as the mask-pass is still 3.3 based, we cannot use GitHub check notifications but GitHub commit status.

GitHub check notifications are enabled with the feature environment variable, by unsetting it then the previous behaviour regarding GitHub commit status will be enabled, GitHub check notifications are the ones using the regex in the mask-pass, as part of the JWT.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -10,7 +10,7 @@ pipeline {
     JOB_GCS_BUCKET = credentials('gcs-bucket')
     GITHUB_CHECK_ITS_NAME = 'Integration Tests'
     ITS_PIPELINE = 'apm-integration-tests-selector-mbp/main'
-    GITHUB_CHECK = 'true'
+    GITHUB_CHECK = 'false'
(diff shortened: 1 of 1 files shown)


trentm:
Is there some infra issue for the mask-password 3.3 breakage that we can link to?

trentm (approved):
This is meant to be temporary, until the mask-passwords thing is solved?

I am somewhat lost in the apm-pipeline-library groovy code and vars involved here. My understanding is that this will fallback to using the (older?) `githubStatus()` instead of `githubCheck()` for showing status on GitHub for `withGithubNotify() { ... }` blocks in our ".ci/Jenkinsfile".  I'm not sure how those map to the "apm-ci/pr-merge" and/or other GitHub checknames, but I don't need to know that now.  `githubStatus()` uses `githubNotify()` ... which is some internal or special mechanism that doesn't require mask-password?

trentm:
The Jenkins mask-password plugin has been downgraded to v3.1 on apm-ci, which has fixed the "main" build. I think we no longer need this change.

Issue #2065, opened 2021-04-29
"metadata requests did not finish, possible deadlock" log.error is slightly alarmist


trentm:
Cloud metadata collection in the agent can log this:

```
{"log.level":"error","@timestamp":"2021-04-29T04:56:27.104Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"metadata requests did not finish, possible deadlock"}
```

if either (a) the [CallbackCoordination](https://github.com/elastic/apm-agent-nodejs/blob/v3.14.0/lib/cloud-metadata/callback-coordination.js#L33-L44) for cloud metadata collection has a bug where it actually deadlocks (which won't actually deadlock or crash the node.js app/process, or (b) it just hit the 3s timeout given for cloud metadata collection.

The latter (b) is most likely. Kibana was hitting this in their tests recently, so that timeout can be hit -- presumably on a very CPU-busy process (like Kibana can be on startup).

That message, and log level, is a bit alarmist. Perhaps log.warn and mention the timeout case specifically. Possibly drop the work "deadlock".

nicomabs:
Same issue here

JojiePalahang:
Any workaround or permanent solution on this message

trentm:
This message was mentioned in an SDH recently. The message was initially read as a likely cause of an observed issue, but actually was unrelated.

PR #2904, opened 2022-08-31
feat: change log message

astorm:
Closes:  #2065

Change the log message to be less alarmist.

### Checklist

- [x] Implement code

### DIFF
--- a/lib/cloud-metadata/callback-coordination.js
+++ b/lib/cloud-metadata/callback-coordination.js
@@ -40,7 +40,7 @@ class CallbackCoordination extends EventEmitter {
       this.timeout = setTimeout(() => {
         if (!this.done) {
           this.complete()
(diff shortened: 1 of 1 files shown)


lib/cloud-metadata/callback-coordination.js
```diff
@@ -40,7 +40,7 @@ class CallbackCoordination extends EventEmitter {
       this.timeout = setTimeout(() => {
         if (!this.done) {
           this.complete()
-          this.logger.error('metadata requests did not finish, possible deadlock')
+          this.logger.error('metadata requests timed out, using default values instead')
```
trentm, line 43:
```suggestion
          this.logger.warn('cloud metadata requests timed out, using default values instead')
```

What do you think about making this a warning? I don't think a user typically needs to act on this.

trentm (approved):

PR #2761, opened 2022-06-08
fix: graphql v16 support

jgillick:
Adds support for graph16.

From what I could tell, this broke because graphql dropped support for [positional arguments](https://github.com/graphql/graphql-js/pull/2904), and the APM agent was forcing the arg object to be converted to positional args.

This PR simply passes the args straight through to graphql, just as they are passed to it.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update TypeScript typings
- [ ] Update documentation
- [x] Add CHANGELOG.asciidoc entry
- [x] Commit message follows [commit guidelines](https://github.com/elastic/apm-agent-nodejs/blob/main/CONTRIBUTING.md#commit-message-guidelines)

PR #2927, opened 2022-09-14
fix: graphql instrumentation for versions <=0.9 were broken

trentm:
The changes in #2761 to support graphql@16 accidentally broken
instrumentation for really old graphql (versions before 0.10.0).
This wasn't noticed because the .tav.yml block for older graphql
versions didn't have `name: graphql`, so it was never run in CI.

### DIFF
--- a/test/instrumentation/modules/graphql.test.js
+++ b/test/instrumentation/modules/graphql.test.js
@@ -26,7 +26,11 @@ var test = require('tape')
 
 var mockClient = require('../../_mock_http_client')
 
-if (semver.lt(graphqlVer, '16.0.0')) {
+// See explanation of these in "lib/instrumentation/modules/graphql.js".
+const onlySupportsPositionalArgs = semver.lt(graphqlVer, '0.10.0')
+const onlySupportsSingleArg = semver.gte(graphqlVer, '16.0.0')
+
+if (!onlySupportsSingleArg) {
   // graphql@16 dropped support for positional arguments.
   test('graphql.graphql(...) - positional args', function (t) {
     resetAgent(done(t))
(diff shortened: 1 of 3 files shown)


trentm:
run module tests for graphql

test/instrumentation/modules/graphql.test.js
```diff
@@ -26,7 +26,11 @@ var test = require('tape')
 
 var mockClient = require('../../_mock_http_client')
 
-if (semver.lt(graphqlVer, '16.0.0')) {
+// See explanation of these in "lib/instrumentation/modules/graphql.js".
+const onlySupportsPositionalArgs = semver.lt(graphqlVer, '0.10.0')
```
astorm, line 30: if we end up needing to do this in a third module it might be worth getting this logic into some sort of helper function.  Just a comment, fine for this PR.

astorm (approved):
I see a name fields been added to the TAV tests for both the `graphql-v0.7-v16` and `graphql` sections, ensuring both run when invoked with `TAV=graphql`. I see bullets have been changed to *'s in the ASCIIDOC for consistency, although both are speced to work as bullets. I see we have two new two boolean to indicate what style of arguments are supported. I see we use the boolean to ensure we're getting correct value from arguments when intercepting function calls. I see out tests use similar booleans to test different paths as well.

Changes seem reasonable and do the things -- approving.

trentm:
>  I see bullets have been changed to *'s in the ASCIIDOC for consistency, although both are speced to work as bullets. 

@astorm  The nested-bullets in the "backend granularity" changelog item only rendered correctly in our docs build when I used `*` rather than `-` for the bullets. I didn't dig into why.  I am cowed. I personally prefer `-` for bullets because I don't have to hit the `Shift` key, but <shrug/>.

PR #2939, opened 2022-09-20
fix avoids a possible 'RangeError: Maximum call stack size exceeded' in maybeTime(...)

trentm:
This avoids a possible 'RangeError: Maximum call stack size exceeded' as
described in #2929.

The case the user hit is in `maybeTime(...)` used in Span timer handling.
Fixed by dropping the recursion.  A `Timer` instance's `_timer` is always the
same object as its parent's `_timer` if it has a parent, so there is no
need to walk the chain.

Refs: #2929 

---

See the discussion at https://github.com/elastic/apm-agent-nodejs/pull/2929 for why this doesn't include tests. We don't have a perfect understanding of how this is happening so don't have a good test to add.  If this passes the current breakdown metrics tests ("test/metrics/breakdown.test.js"), then I think this should be good as a workaround for the user's issue.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -101,6 +101,9 @@ paths are considered for body capture. ({pull}2873[#2873])
 [float]
 ===== Bug fixes
 
+* Avoid a possible `RangeError: Maximum call stack size exceeded` in
+  Span timer handler for exceedingly deep Span trees. ({pull}2939[#2939])
+
 * Fix instrumentation of (very old) 'graphql' module versions <=0.9.6.
   Instrumentation of these older graphql versions was broken in v3.36.0.
   ({pull}2927[#2927])
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -34,6 +34,13 @@ function Span (transaction, ...args) {
     : {}
   const [name, ...tsaArgs] = args // "tsa" === Type, Subtype, Action
 
+  if (opts.timer) {
+    process.emitWarning(
+      'specifying the `timer` option to `new Span()` was never a public API and will be removed',
+      'DeprecationWarning',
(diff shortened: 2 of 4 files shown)


astorm (approved):
Approving, but just to say this out loud, that approval hinges on this being true

>  A Timer instance's _timer is always the same object as its parent's _timer if it has a parent

I only dug into this shallowly, but looking at the place a span-or-transaction's `Timer` object gets instantiated 

https://github.com/elastic/apm-agent-nodejs/blob/main/lib/instrumentation/generic-span.js#L25

```
```

The timer passed to `Timer`'s constructor is an object that could be set by a user passing in options to `startSpan` or `startTransaction`, which means making assumptions about its structure is trusting that end users aren't getting too wild with what they're doing.    

This isn't a documented part of the user facing API so not a blocker, but if we start getting reports of weird timings from our most power of power users we'll want to remember this change has been made.

trentm:
> Approving, but just to say this out loud, that approval hinges on this being true
> 
> > A Timer instance's _timer is always the same object as its parent's _timer if it has a parent
> ...

@astorm Good point, thanks. What do you think about adding a warning if a user passing in `opts.timer` to new Span or new Transaction? Using `process.emitWarning` like this https://github.com/elastic/apm-agent-nodejs/pull/2939/commits/9c4e22f36c55737f6059e9c3d43b15c6aeccb5c1 looks like the following on the CLI:

```
(node:65143) [ELASTIC_APM_SPAN_TIMER_OPTION] DeprecationWarning: specifying the `timer` option to `new Span()` was never a public API and will be removed
```

astorm (approved):
The log makes sense 👍 approving.

trentm:
https://github.com/elastic/apm-agent-nodejs/issues/2990 added to drop the option in a future major rev.

Issue #2938, opened 2022-09-20
[META 689] Ignore *principal* headers by default


elastic-apm-tech:
See meta issue for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/689

PR #2982, opened 2022-10-12
feat: Add '*principal*' to `sanitizeFieldNames` to redact possible auth headers

trentm:
Closes: #2938

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -98,6 +98,10 @@ paths are considered for body capture. ({pull}2873[#2873])
   setting of an `Error.prepareStackTrace` handler until necessary for stacktrace
   collection. ({issues}2833[#2833] {pull}2906[#2906])
 
+* Add `*principal*` pattern to default value for `sanitizeFieldNames` config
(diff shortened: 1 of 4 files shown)


CHANGELOG.asciidoc
```diff
@@ -280,6 +280,10 @@ specifying links during span creation (with the limitation that span link
   Lambda layers now that
   https://aws.amazon.com/blogs/compute/node-js-16-x-runtime-now-available-in-aws-lambda/[this runtime is available on AWS].
 
+- Add `*principal*` pattern to default value for `sanitizeFieldNames` config
```
astorm, line 283: Slightest of nits with the weight of a photon -- but did you want `*` instead of `-` here for the list? (based on previous commit feedback)

astorm (approved):
✅ does the thing, approving

CHANGELOG.asciidoc
trentm: You are right. Thanks for pointing it out.

(Also I added this bullet in totally the wrong section (it was low down in a section for an earlier release).

trentm: commit 040cc180

PR #3013, opened 2022-10-31
test: fix two TAV=next test failures with Next.js v11

trentm:
1. next.test.js was failing in Docker/CI with node v14 and v16
   because https://nextjs.org/telemetry automatically turned on
   which resulted in APM spans for `POST telemetry.nextjs.org`
   during tests, which broke assertions.
2. `next build` with Next.js v11 and Node.js >=17 crashes due to
   https://github.com/vercel/next.js/issues/30188
   We just skip testing that older Next.js with newer Node.js.


trentm:
run module tests for next

astorm (approved):

Issue #2419, opened 2021-11-04
Lambda: fill the HTTP context for API Gateway-triggered functions


astorm:
https://github.com/elastic/apm-agent-nodejs/pull/2194#discussion_r742328035

> Also the spec says:
>> The agent should use the information in the request and response objects to fill the HTTP context (context.request and context.response) fields in the same way it is done for HTTP transactions.
>Perhaps a separate issue to add `transaction.context.{request,response}` fields? We won't be able to re-use the `parsers.getContextFrom{Request,Response}()` functions that take core http objects.

trentm:
Relevant spec section: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#api-gateway-v1--v2

trentm:
See https://github.com/elastic/apm-agent-nodejs/issues/3000 that I think might just be handled by doing this one.

PR #3043, opened 2022-11-24
feat: capture HTTP context for API Gateway-triggered Lambda functions

trentm:
https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#api-gateway--lambda-urls
> The agent should use the information in the request and response
> objects to fill the HTTP context (context.request and context.response)
> fields in the same way it is done for HTTP transactions.

Closes: #2419

# Checklist

- [x] `context.response` as well (initial commit just has `context.request` handling
- [x] changelog entry

### DIFF
--- a/test/lambda/trans-setDefaultName.test.js
+++ b/test/lambda/trans-setDefaultName.test.js
@@ -87,6 +87,7 @@ tape.test('lambda instrumentation does not break trans.setDefaultName', function
         cb(null, 'hi')
       })
 
+      server.clear()
       lambdaLocal.execute({
         event: eventCase.event,
         lambdaFunc: {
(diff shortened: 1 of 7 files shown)


test/lambda/trans-setDefaultName.test.js
```diff
@@ -87,6 +87,7 @@ tape.test('lambda instrumentation does not break trans.setDefaultName', function
         cb(null, 'hi')
       })
 
+      server.clear()
```
trentm, line 90: REVIEW NOTE: This was an unrelated fix for this test. It wasn't testing what it thought it was testing and was passing by accident.

delvedor (approved):
Supporting so many versions is no fun, LGTM!
If it helps, we did some work with lambda in the fastify community, you can see it here: https://github.com/fastify/aws-lambda-fastify

PR #3053, opened 2022-12-02
feat: support instrumentation of restify@10

trentm:
Restify v10 was recently released to support node >=18. No changes were
required for our instrumentation, other than allowing v10 to be
instrumentated.


trentm:
run module tests for restify

Issue #3072, opened 2022-12-15
Next.js instrumentation doesn't honour `usePathAsTransactionName` config var


trentm:
As discussed at https://discuss.elastic.co/t/next-js-and-dynamic-routes/321302 the current Next.js instrumentation that sets the `transaction.name` for a Dynamic Route does not honour the `usePathAsTransactionName: true` config var. That means that a request `GET /a-dynamic-route/42` to this dynamic route https://github.com/elastic/apm-agent-nodejs/blob/main/test/instrumentation/modules/next/a-nextjs-app/pages/a-dynamic-page/%5Bnum%5D.js results in `transaction.name = 'GET /a-dynamic-route/[num]` even if `usePathAsTransactionName: true`.

Note that there is also naming of the `Next.js _data/route ...` transactions for possibly dynamic routes -- e.g. "Next.js _next/data route /a-dynamic-page/[num]". However, I don't think those should honour `usePathAsTransactionName`. At least not without user feedback.

trentm:
Note to self: a start at this is:

```diff
diff --git a/lib/instrumentation/modules/next/dist/server/dev/next-dev-server.js b/lib/instrumentation/modules/next/dist/server/dev/next-dev-server.js
index ab1e7052..d02b5da8 100644
--- a/lib/instrumentation/modules/next/dist/server/dev/next-dev-server.js
+++ b/lib/instrumentation/modules/next/dist/server/dev/next-dev-server.js
@@ -143,7 +143,11 @@ module.exports = function (mod, agent, { version, enabled }) {
       // specific route wrapper hasn't already done so.
       if (trans && !trans[kSetTransNameFn]) {
         trans[kSetTransNameFn] = (req, pathname) => {
-          trans.setDefaultName(`${req.method} ${pathname}`)
+          if (agent._conf.usePathAsTransactionName) {
+            trans.setDefaultName(`${req.method} ${req.url}`)
+          } else {
+            trans.setDefaultName(`${req.method} ${pathname}`)
+          }
           // Ensure only the first `findPageComponents` result sets the trans
           // name, otherwise a loaded `/_error` for page error handling could
           // incorrectly override.
diff --git a/lib/instrumentation/modules/next/dist/server/next-server.js b/lib/instrumentation/modules/next/dist/server/next-server.js
index 0431e260..428309fe 100644
--- a/lib/instrumentation/modules/next/dist/server/next-server.js
[...]

PR #3073, opened 2022-12-15
fix: honour 'usePathAsTransactionName' config var in Next.js instrumentation

trentm:
Fixes: #3072

# Checklist

- [ ] add a test case


trentm:
Try this out via:

```
npm install 'elastic-apm-node@elastic/apm-agent-nodejs#trentm/nextjs-usePathAsTransactionName'
```

PR #3076, opened 2022-12-19
fix: span compression bug where a buffered span would not be sent when an incompressible sibling ended

trentm:
https://github.com/elastic/apm/blob/main/specs/agents/handling-huge-traces/tracing-spans-compress.md#span-buffering says:
> A buffered span gets reported when
>  1. its parent ends
>  2. a non-compressible sibling ends

Before this change, case 2 was not being handled. Further, a possibly
buffered span on the ending sibling would incorrectly get sent *twice*.
The scenario is as follows:
```
   trans t0
     span s1: _bufferdSpan=s2
       span s2: ended, compressible -> buffered on s1
       span s3: incompressible, _bufferdSpan=s4
         span s4: ended, compressible -> buffered on s3
```
What happens when s3 ends? We expect:
- s4 is encoded and sent
- s2 is encoded and sent
- s3 is encoded and sent

Before this change the agent would send: s4, s4, s3

Refs: https://discuss.elastic.co/t/apm-agent-send-duplicate-record-in-mysql/321440

### DIFF
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -417,18 +417,19 @@ Instrumentation.prototype.addEndedSpan = function (span) {
       span.setBufferedSpan(null)
     }
 
-    if (!span.isCompressionEligible() || span.getParentSpan().ended) {
-      const buffered = span.getParentSpan().getBufferedSpan()
+    const parentSpan = span.getParentSpan()
+    if ((parentSpan && parentSpan.ended) || !span.isCompressionEligible()) {
+      const buffered = parentSpan && parentSpan.getBufferedSpan()
       if (buffered) {
         this._encodeAndSendSpan(buffered)
-        span.getParentSpan().setBufferedSpan(null)
+        parentSpan.setBufferedSpan(null)
       }
       this._encodeAndSendSpan(span)
-    } else if (!span.getParentSpan().getBufferedSpan()) {
+    } else if (!parentSpan.getBufferedSpan()) {
       // span is compressible and there's nothing buffered
       // add to buffer, move on
       span.getParentSpan().setBufferedSpan(span)
-    } else if (!span.getParentSpan().getBufferedSpan().tryToCompress(span)) {
+    } else if (!parentSpan.getBufferedSpan().tryToCompress(span)) {
       // we could not compress span so SEND bufferend span
       // and buffer the span we could not compress
       this._encodeAndSendSpan(span.getParentSpan().getBufferedSpan())
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -414,13 +414,14 @@ Instrumentation.prototype.addEndedSpan = function (span) {
     // if I have ended and I have something buffered, send that buffered thing
     if (span.getBufferedSpan()) {
       this._encodeAndSendSpan(span.getBufferedSpan())
+      span.setBufferedSpan(null)
     }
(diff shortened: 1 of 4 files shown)


lib/instrumentation/index.js
```diff
@@ -414,24 +414,26 @@ Instrumentation.prototype.addEndedSpan = function (span) {
     // if I have ended and I have something buffered, send that buffered thing
     if (span.getBufferedSpan()) {
       this._encodeAndSendSpan(span.getBufferedSpan())
+      span.setBufferedSpan(null)
```
trentm, line 417: REVIEW NOTE: This clearing avoids the bug where we would encode and send a span *twice* -- the first time here, the second time in the next if-block.

lib/instrumentation/index.js
```diff
@@ -414,24 +414,26 @@ Instrumentation.prototype.addEndedSpan = function (span) {
...
 
-    if (!span.isCompressionEligible() || span.getParentSpan().ended) {
-      const buffered = span.getBufferedSpan()
+    const parentSpan = span.getParentSpan()
+    if ((parentSpan && parentSpan.ended) || !span.isCompressionEligible()) {
+      const buffered = parentSpan && parentSpan.getBufferedSpan()
       if (buffered) {
         this._encodeAndSendSpan(buffered)
-        span.setBufferedSpan(null)
+        parentSpan.setBufferedSpan(null)
```
trentm, line 425: REVIEW NOTE: This corrects the logic to make the spec pseudo code: this block is meant to be able looking at the possibly buffered span on the *parent*.   This fixes the case where a buffered span would not get sent if it was followed by an *incompressible* sibling span.

trentm:
Here is a snippet of the output of a test run showing the newly added test failing before the fix:

```
# ensure a span buffered on parent is sent when incompressible span ends
ok 18 s2 is a child of s1
ok 19 s2 is buffered on s1
ok 20 s3 is a child of s1
ok 21 s4 is a child of s3
ok 22 s4 is buffered on s3
ok 23 3 events sent
ok 24 3 spans sent
not ok 25 sent spans are s4,s2,s3; in that order
  ---
    operator: equal
    expected: 's4,s2,s3'
    actual:   's4,s4,s3'
    at: <anonymous> (/Users/trentm/el/apm-agent-nodejs6/test/instrumentation/span-compression.test.js:185:9)
    stack: |-
      Error: sent spans are s4,s2,s3; in that order
          at Test.assert [as _assert] (/Users/trentm/el/apm-agent-nodejs6/node_modules/tape/lib/test.js:312:48)
          at Test.bound [as _assert] (/Users/trentm/el/apm-agent-nodejs6/node_modules/tape/lib/test.js:95:17)
          at Test.strictEqual (/Users/trentm/el/apm-agent-nodejs6/node_modules/tape/lib/test.js:476:7)
          at Test.bound [as equals] (/Users/trentm/el/apm-agent-nodejs6/node_modules/tape/lib/test.js:95:17)
          at /Users/trentm/el/apm-agent-nodejs6/test/instrumentation/span-compression.test.js:185:9
          at Timeout._onTimeout (/Users/trentm/el/apm-agent-nodejs6/test/_mock_http_client.js:80:7)
  ...

1..25
# tests 25
# pass  24
# fail  1
```

trentm:
elastic-apm-node@3.41.1 is published with this fix now.

Issue #2382, opened 2021-10-20
outgoing http.request instrumentation exacerbates node issue with ipv6 addresses


trentm:
I have an express server with a 'GET /ping' endpoint running on localhost (port 3000) on macOS. When I run this (using node v16):

```js
var http = require('http')
theUrl = 'http://[::1]:3000/ping'
http.get(theUrl, function (res) {
  console.log('res:', res.statusCode, res.headers)
  res.on('data', (chunk) => { console.log('data: %s', chunk) })
  res.on('end', () => { console.log('end') })
})
```

I get:

```
% node foo.js
res: 200 {
  'x-powered-by': 'Express',
  'content-type': 'text/html; charset=utf-8',
  'content-length': '4',
  etag: 'W/"4-DlFKBmK8tp3IY5U9HOJuPUDoGoc"',
  date: 'Wed, 20 Oct 2021 16:59:25 GMT',
  connection: 'close'
}
data: pong
end
```

All good. However, when I run this:

```js
var http = require('http')
theUrl = {
  href: 'http://[::1]:3000/ping',
  pathname: '/ping',
  path: '/ping',
  protocol: 'http:',
  host: '[::1]:3000',
  port: '3000',
  hostname: '[::1]',
  hash: '',
  search: '',
  headers: {}
}
http.get(theUrl, function (res) {
  console.log('res:', res.statusCode, res.headers)
  res.on('data', (chunk) => { console.log('data: %s', chunk) })
  res.on('end', () => { console.log('end') })
})
```

I get this:

```
% node foo.js
node:events:368
      throw er; // Unhandled 'error' event
      ^

Error: getaddrinfo ENOTFOUND [::1]
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:71:26)
Emitted 'error' event on ClientRequest instance at:
    at Socket.socketErrorListener (node:_http_client:447:9)
    at Socket.emit (node:events:390:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: '[::1]'
}
```

That this behaviour differs in node strikes me as a bug, but I haven't looked into it yet.

### the APM agent exacerbates this

[...]

Issue #2044, opened 2021-04-13
instrumentation of `http{s}.request()` has a few edge case issues


trentm:
In lib/instrumentation/http-shared.js the `traceOutgoingRequest` function that handles instrumenting `http.request` and `https.request` has a few issues:

```
      var options = {}
      var newArgs = [options]
      for (const arg of args) {
        if (typeof arg === 'function') {
          newArgs.push(arg)
        } else {
          Object.assign(options, ensureUrl(arg))
        }
      }
```

1. That iteration through all `args` changes the signature of `http.request()` such that you can pass any number of objects and URLs to it -- before and after a callback function. Not a biggie.
2. That `ensureUrl` uses a `formatURL` function that is meant to be (or should be) an equivalent to node core's handling (https://github.com/nodejs/node/blob/8d9d8236b79ea91640f973cc8c1423603694b482/lib/internal/url.js#L1288-L1311) but misses a few things:
  - it doesn't restore options.auth, so users of APM and `http.request` and HTTP Basic auth in a first URL arg are being broken
  - it handles "port" slightly differently that could surprise, I think
  - it doesn't have changes for a couple node core issues over the years: https://github.com/nodejs/node/pull/26226 https://github.com/nodejs/node/pull/19720

* * *

I wonder if it would be possible to avoid full processing of the original args, only try to sniff out and update an "options.headers" object. Then the agent doesn't have to try to track change made to node's internal "urlToHttpOptions" handling.

One side-effect of ^^ is that the potential handling for extracting the "url", see https://github.com/elastic/apm-agent-nodejs/pull/2039, may have to change to handle inspecting those original `args` itself. I believe that would be preferable to changing the behaviour of `http.request`.

astorm:
> I wonder if it would be possible to avoid full processing of the original args, only try to sniff out and update an "options.headers" object. Then the agent doesn't have to try to track change made to node's internal "urlToHttpOptions" handling.

My thinking is similar.

Looking at the code here, it appears that that only property of the extracted options variable that our function uses is `options.headers`.  It is not, to my knowledge, possible to extract a headers value from a URL string or object.  In other words, it's not necessary to merge the URL data to get the value we want (`options.headers`).  Sniffing out the options object, and then passing the original arguments (with a modified options object) seems like the best course of action, with one caveat.

Here's all the method signatures of `http.request` 

      // Three Argument:
      // http.request(string|String|URL url, object options, function callback)
      //
      // Two Argument:
[...]

PR #3090, opened 2023-01-14
fix: instrumentation of `http.request` et al has a few edge cases

trentm:
This change updates the argument munging that is done for the wrapping
of `http.request` et al to fix some cases where the instrumentation
would *change the behaviour* of the function.  The main fix here is
that before this change the conversion of a URL instance to request
options would accidentally drop basic auth info (the 'username' and
'password' fields).

This issue has existed since v2.17.0. It was introduced in commit dd60b12
when switching directly from `url.parse()` to `new URL()`. Recent
versions of node have a `url.urlToHttpOptions()` to help with this
conversion.

Fixes: #2044
Fixes: #2382
Refs: https://discuss.elastic.co/t/issue-apm-agent-with-backend-requests-username-password-url/322903


trentm:
The cleaning up `http.request` options handling also fixes #2382.

### repro

Run this server:

```bash
% cat issue2382-server.example.js
const app = express()
app.get('/ping', (req, res) => {
  res.send('pong')
})
app.listen(3000)

% node issue2382-server.example.js
```

Then run this client request script (`issue2382.example.js`):

```js
require('.').start({ // elastic-apm-node
  serviceName: 'example-outgoing-http-ipv6-boom',
  cloudProvider: 'none',
  metricsInterval: '0s',
  centralConfig: false,
  captureExceptions: false,
  logUncaughtExceptions: true,
  disableSend: true
})

var http = require('http')
var theUrl = 'http://[::1]:3000/ping'
http.get(theUrl, function (res) {
  console.log('res:', res.statusCode, res.headers)
  res.on('data', (chunk) => { console.log('data: %s', chunk) })
  res.on('end', () => { console.log('end') })
})
```

without the fix:

```
% node issue2382.example.js
node:events:491
      throw er; // Unhandled 'error' event
      ^

Error: getaddrinfo ENOTFOUND [::1]
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:109:26)
Emitted 'error' event on ClientRequest instance at:
    at Socket.socketErrorListener (node:_http_client:494:9)
    at Socket.emit (node:events:513:28)
    at emitErrorNT (node:internal/streams/destroy:157:8)
    at emitErrorCloseNT (node:internal/streams/destroy:122:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: '[::1]'
}
```

and with this PR:

```
% node issue2382.example.js
res: 200 {
  'x-powered-by': 'Express',
  'content-type': 'text/html; charset=utf-8',
  'content-length': '4',
  etag: 'W/"4-DlFKBmK8tp3IY5U9HOJuPUDoGoc"',
  date: 'Sat, 14 Jan 2023 00:35:18 GMT',
  connection: 'close'
}
data: pong
end
```

PR #3126, opened 2023-01-25
action: edge

v1v:
### What

Migrate the edge stage to GitHub actions and will run based on the below events:

1) merge commits in `main`
2) manually

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### Test

See https://github.com/elastic/apm-agent-nodejs/actions/workflows/edge.yml

### Follow ups

- when github command `/test-nightly` or `/test-rc`


reakaleek (approved):

trentm:
@v1v  Is there a pattern that other APM projects use for the "/test..." commands for rebuilds that we could follow for consistency? I notice that apm-agent-python, for example uses:

```
/test : Re-trigger the build.

/test linters : Run the Python linters only.

/test full : Run the full matrix of tests.

/test benchmark : Run the APM Agent Python benchmarks tests.

... plus the `run` `elasticsearch-ci/docs` one that is out of that repo's control
```

Ah, and Java and others seem to have "/ test" and "run ..." things.

For what it is worth I like the `/ test [OPTIONS]` or `/ test-OPTION` pattern if that is workable. Probably because it hints towards IRC and Slack commands.

.github/workflows/edge.yml
```diff
@@ -0,0 +1,69 @@
...
+# candidate (RC) and "nightly" node.js builds.
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - '**.asciidoc'
```
trentm, line 15:
```suggestion
      - '**.asciidoc'
      - 'docs/**'
      - 'examples/**'
```

Could perhaps add these as we have in "test.yml".

trentm (approved):
This is great.

.github/workflows/edge.yml
```diff
@@ -0,0 +1,69 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        node: [ "20" ]
+        contextManager: [ 'patch', '' ]
+    steps:
+      - uses: actions/checkout@v3
+      - run: .ci/scripts/test.sh -b "nightly" "${{ matrix.node }}"
+        env:
+          VERSION: ${{ matrix.node }}
```
trentm, line 44: Is the VERSION envvar needed?

v1v: good catch, let me double check

v1v:
> @v1v Is there a pattern that other APM projects use for the "/test..." commands for rebuilds that we could follow for consistency?

`run ...` was the first github command pattern then we moved to
`/test ...` but kept the `run` for backward compatibility. Some of those github commands are documented in the `oblt-dev` repo and https://github.com/elastic/apm-pipeline-library/blob/15602fdc2deeb020c9b67c0ae24bc88f0a2b5fcd/src/co/elastic/NotificationManager.groovy#L491

I think we are in a good position to normalise those names. 

OTOH, I tried to configure the github commands within this PR but it seems it won't work until it's available in the default branch :/ For now I'll not implement this now as I'd like to think about your proposal https://github.com/elastic/apm-pipeline-library/issues/1904
so we can use something similar shortly.

.github/workflows/edge.yml
```diff
@@ -0,0 +1,71 @@
...
```
v1v, line 46:
```suggestion
```

.github/workflows/edge.yml
```diff
@@ -0,0 +1,69 @@
...
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        node: [ "20" ]
+        contextManager: [ 'patch', '' ]
+    steps:
+      - uses: actions/checkout@v3
+      - run: .ci/scripts/test.sh -b "nightly" "${{ matrix.node }}"
+          ELASTIC_APM_CONTEXT_MANAGER: ${{ matrix.contextManager }}
```
v1v, line 45:
```suggestion
        env:
          ELASTIC_APM_CONTEXT_MANAGER: ${{ matrix.contextManager }}
```

PR #3127, opened 2023-01-25
action: tav

v1v:
### What

Migrate TAV to run as GitHub actions when
- a merge commit in the `main` branch
- manually triggered

TAV matrix is declared in the GH workflow hence no more .tav files to be maintained.

### Checklist

- [x] Implement code

### Caveats

We agreed recently that It will not run on PRs for now, if needed we can add that feature in a follow up, maybe using the smarter jenkins implementation if it still works.


.github/workflows/tav.yml
```diff
@@ -0,0 +1,52 @@
...
+    - 'examples/**'
+  pull_request:
+    branches:
+    - main
+    paths-ignore:
+    - '*.md'
+    - '*.asciidoc'
+    - 'docs/**'
+    - 'examples/**'
+
```
v1v, line 21:
```suggestion

concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}

```

.github/workflows/tav.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    outputs:
+      matrix: ${{ steps.generate.outputs.matrix }}
+    steps:
+      - id: generate
+        uses: elastic/apm-pipeline-library/.github/actions/version-framework@current
+        with:
+          versionsFile: .ci/.jenkins_tav_nodejs.yml
+          frameworksFile: .ci/.jenkins_tav.yml
```
trentm, line 42: How about putting the TAV module names and Node version name data in *this* file rather than having to have this separate action (that is using Python) to do this processing? Would that be possible/easier?  I think it would be nice to see the matrix values right here in the "tav.yml" file.

v1v: No strong opinion, just trying to migrate without changing much the existing process. So, let me add them

reakaleek (approved):

.github/workflows/tav.yml
```diff
@@ -0,0 +1,56 @@
+name: TAV
+
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
trentm: Ah, nevermind. I gather this is for triggering manually via `gh workflow run ...` or something.

.github/workflows/tav.yml
```diff
@@ -0,0 +1,78 @@
...
+      matrix:
+        node:
+          - "19"
+          - "18"
+          - "16"
+          - "14"
+          - "12"
+          - "10"
+          - "8"
+        framework:
```
trentm, line 38: nit: Maybe call this "module" instead of "framework". This is minor, however.

trentm (commented):
How can one trigger a TAV run on a PR?

- a full TAV run
- just for a particular module
- just for a particular node version

v1v:
> How can one trigger a TAV run on a PR?
a full TAV run
just for a particular module
just for a particular node version

I probably misunderstood what we discussed online the other day, somehow I thought the smart jenkins stuff was never used hence I kept it simple to run only on the main branch or manually when needed.

If the above is needed, then let's discuss about each individual scenario:

* a full TAV run on a PR basis

  - Use a GitHub command, though we don't have that particular functionality in place yet. It requires some permissions to know who triggered the build through a comment to avoid anyone outside of the Elastic org to run them. Last time when I checked the org permissions were not accessible through the GitHub temporary token, but I can double check. 
  - Use a GitHub label, though we used something like that in the APM Agent java, I'm not convinced yet as it required several `on: ` types, see https://github.com/elastic/apm-agent-java/blob/a6be6fc83c802215fb449706cd7bb016558fbfc2/.github/workflows/main.yml#L43-L49

* just for a particular module and  just for a particular node version

  - Same as above, either specific labels or using a github command, unless we wanna use the git changeset and compare what modules have been changed.

A few other questions:
1) What's your preference in terms of usability?
2) What's the one of the above options that we can discard?
3) Any other ideas?


Thanks

.github/workflows/tav.yml
v1v, line 38:
```suggestion
        module:
```

.github/workflows/tav.yml
```diff
@@ -0,0 +1,78 @@
...
+          - pg
+          - pug
+          - redis
+          - restify
+          - tedious
+          - undici
+          - ws
+    steps:
+      - uses: actions/checkout@v3
+      - run: .ci/scripts/test.sh -b "release" -t "${{ matrix.framework }}" "${{ matrix.node }}"
```
v1v, line 76:
```suggestion
      - run: .ci/scripts/test.sh -b "release" -t "${{ matrix.module }}" "${{ matrix.node }}"
```

trentm:
> > How can one trigger a TAV run on a PR?
> > ...
> 
> I probably misunderstood what we discussed online the other day, somehow I thought the smart jenkins stuff was never used hence I kept it simple to run only on the main branch or manually when needed.

@v1v I think we only need the ability to *manually* start a TAV test run. I don't think we need any smart automatic triggering. I also don't have any requirements for how we do that -- any of the following would be fine: a GitHub comment, clicking some GitHub checks UI, running some `gh workflow ...` command. Whatever is cleanest/easiest.

---

Should we be able to see some "TAV / ..." GitHub checks on this PR?

v1v:
> Should we be able to see some "TAV / ..." GitHub checks on this PR?

I'm afraid new GH actions coming from forked repositories are not executed. Let me create a branch in the upstream and cherry-pick the changes so you can see how it looks like.

trentm:
> I'm afraid new GH actions coming from forked repositories are not executed. Let me create a branch in the upstream and cherry-pick the changes so you can see how it looks like.

Oh, right. Thanks for doing that.  This looks very good.

---

Some TAV comparisons between GH Actions and Jenkins:

- " Total duration 1h 19m 9s " (from https://github.com/elastic/apm-agent-nodejs/actions/runs/4540484765?pr=3221) vs. about 1h15m for a Jenkins TAV run (https://apm-ci.elastic.co/job/apm-agent-nodejs/job/apm-agent-nodejs-mbp/job/main/)

- Both have a way to look at all the separate tests run:

Screenshot 2023-03-28 at 8 41 18 AM
Screenshot 2023-03-28 at 8 41 11 AM

- To find TAV tests that were taking very long to run (to know if we should look into reducing the number of versions tested), I had written https://github.com/elastic/apm-agent-nodejs/blob/main/dev-utils/jenkins-build-slow-steps.sh   At some point I'll see about writing a replacement for these GH Actions.  That definitely is not a requirement for this work.

trentm:
@v1v Do you know if there will be a way to *manually* run just a part of the TAV tests on a PR? For example, say I have a PR that is working on the "fastify" module. Can I run the TAV tests for module=fastify for either all node versions, or just a subset of the node versions?

This doesn't work:

```
% gh workflow run tav.yml --ref test/action-tav -f node=16 -f module=fastify
could not create workflow dispatch event: HTTP 422: Unexpected inputs provided: ["module", "node"] (https://api.github.com/repos/elastic/apm-agent-nodejs/actions/workflows/46437793/dispatches)
```

because the `matrix` values are not the same as [workflow dispatch inputs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_dispatchinputs).   I haven't played with workflow dispatch inputs in GH Actions, so I don't know if what I'm asking is possible.

This isn't a requirement for this PR. We could look into this separately.

trentm:
I notice this from https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix

> A matrix will generate a maximum of 256 jobs per workflow run. This limit applies to both GitHub-hosted and self-hosted runners.

Currently we have 35 modules * 7 node versions = 245 jobs.  So we are very close to the maximum currently. :)
When node v20 is released (expected 2023-04-18) and there are initial node v21 releases (I'm not sure when), then we'll be over the limit.
Or, if/when we add two new instrumented modules, then we'll be over the limit.
This is just something to be aware of. We may have a new major version of the APM agent by then that drops support for some of the older ancient node versions. Or we could stop TAV testing for the deprecated node v8. Or stop testing some older, less useful modules like `got` or `bluebird`.

v1v:
> @v1v Do you know if there will be a way to manually run just a part of the TAV tests on a PR? For example, say I have a PR that is working on the "fastify" module. Can I run the TAV tests for module=fastify for either all node versions, or just a subset of the node versions?

I'd like to work on this in a follow up, so we can use one of the above proposals (https://github.com/elastic/apm-agent-nodejs/pull/3127#issuecomment-1484811610)

> Currently we have 35 modules * 7 node versions = 245 jobs. So we are very close to the maximum currently. :)

We managed to use buckets in the apm-agent-python:
- https://github.com/elastic/apm-agent-python/blob/28ecb8e804f38458d07dfad515d8eb3f828abcec/.github/workflows/test.yml#L35-L81

It's not ideal, otherwise if possible we could create combinations of two modules per runner, though I don't know whether that could cause any other issues (tearing down the environment might be required)


> To find TAV tests that were taking very long to run (to know if we should look into reducing the number of versions tested), I had written https://github.com/elastic/apm-agent-nodejs/blob/main/dev-utils/jenkins-build-slow-steps.sh At some point I'll see about writing a replacement for these GH Actions. That definitely is not a requirement for this work.

You can use the existing opentelemetry data :)  https://ela.st/oblt-ci-cd-stats. (though it's only accessible for Elasticians and the new TAV GH action won't be available until this PR gets merged)

Our plan is to leverage the existing GitHub metadata and provide insights that are meaningful. We are at the moment with the migration stage, but definitely it's something we will invest effort so we can help to answer those kind of questions.

PR #3128, opened 2023-01-25
action: lint and ELASTIC_APM_CONTEXT_MANAGER=patch

v1v:
### What

Migrate linting and ELASTIC_APM_CONTEXT_MANAGER=patch to GitHub actions

I tried to reuse the same runners by running the same command with the ELASTIC_APM_CONTEXT_MANAGER set as default and patch, but it failed, likely the env context needs to be clean, so it works like a charm by adding the context axis in the matrix.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


reakaleek (approved):

.github/workflows/lint.yml
```diff
@@ -28,3 +28,5 @@ jobs:
         node-version: '16'
     - run: npm install
     - run: npm run lint
+    - run: .ci/scripts/test_basic.sh
+    - run: .ci/scripts/test_types_babel_esm.sh
```
trentm, line 32: These two scripts effectively do:

```
npm install
npm run lint
npm run test:deps

npm install
npm run test:types
npm run test:babel
npm run test:esm
```

The repeated `npm install`s are wasteful. As well, all those other `npm run test:*` targets are already being run in regular test matrix here: https://github.com/elastic/apm-agent-nodejs/blob/main/test/script/run_tests.sh#L40-L74

Can we remove these two lines and delete those two ".ci/scripts/" files?

.github/workflows/test.yml
```diff
@@ -133,6 +132,7 @@ jobs:
           - '10.0'
           - '8'
           - '8.6'
+        context: [ 'patch', '' ]
```
trentm, line 135: Nice!

```suggestion
        contextManager: [ 'patch', '' ]
```

Can we call this `contextManager` to match the var in https://github.com/elastic/apm-agent-nodejs/pull/3126/files#diff-a8e31e174e3ddce38cb2351c978bb5d63bd46c41235a13c9dc12763e26afef05R63 doing the same thing?

.github/workflows/test.yml
```diff
@@ -144,6 +144,8 @@ jobs:
     - run: npm ls --all || true
     - name: npm test
       run: npm test
+      env:
+        ELASTIC_APM_CONTEXT_MANAGER: ${{ matrix.context }}
```
trentm, line 148:
```suggestion
        ELASTIC_APM_CONTEXT_MANAGER: ${{ matrix.contextManager }}
```

trentm (changes requested):
Looking great.

.github/workflows/test.yml
v1v: See 413f021b630f33dddbd520483ebaea22662f3d35

.github/workflows/test.yml
v1v: See 413f021b630f33dddbd520483ebaea22662f3d35

.github/workflows/lint.yml
v1v: See 7dd45f74418382c3e3c3300edb704fc6ca4999b3

trentm (approved):

Issue #2954, opened 2022-09-28
[META 691] Reporter for OTel Metrics


elastic-apm-tech:
See meta issue and spec for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/691
- Spec issue: https://github.com/elastic/apm/issues/692

trentm:
I'll edit this comment to give a summary of this PoC work. This is a placeholder for now.

trentm:
# vanilla OTel JS Metrics usage

Basic OTel JS Metrics usage looks like this (adapted from [this example](https://github.com/open-telemetry/opentelemetry-js/blob/main/examples/otlp-exporter-node/metrics.js)):

```js
const api = require("@opentelemetry/api-metrics");
// const { OTLPMetricExporter } = require('@opentelemetry/exporter-metrics-otlp-grpc'); // OTLP/gRPC
// const { OTLPMetricExporter } = require('@opentelemetry/exporter-metrics-otlp-http'); // OTLP/HTTP+JSON
const { OTLPMetricExporter } = require('@opentelemetry/exporter-metrics-otlp-proto'); // OTLP/HTTP+protobuf
const { MeterProvider, PeriodicExportingMetricReader } = require('@opentelemetry/sdk-metrics');
const { Resource } = require('@opentelemetry/resources');
const { SemanticResourceAttributes } = require('@opentelemetry/semantic-conventions');

// Setup a MeterProvider.
const metricExporter = new OTLPMetricExporter({ 
    // url: <the collector or APM server>
});
const meterProvider = new MeterProvider({
  // metadata
  resource: new Resource({
    [SemanticResourceAttributes.SERVICE_NAME]: 'basic-metric-service',
  }),
});
meterProvider.addMetricReader(new PeriodicExportingMetricReader({
  exporter: metricExporter,
  exportIntervalMillis: 1000,
}));
api.metrics.setGlobalMeterProvider(meterProvider);

// User code to create metrics (called "Instruments" in OTel-land).
const meter = api.metrics.getMeter('my-metrics-play');
const requestCounter = meter.createCounter('my_requests', {
  description: 'Example of a Counter',
});
const upDownCounter = meter.createUpDownCounter('my_up_down_counter', {
  description: 'Example of a UpDownCounter',
});
const histogram = meter.createHistogram('my_histogram', {
  description: 'Example of a Histogram',
});

// Then some artifical code showing updating of those metrics.
setInterval(() => {
[...]

PR #3152, opened 2023-02-06
feat: Reporter for OTel Metrics

trentm:
With this change the APM agent adds support for using OTel Metrics in two ways:

1. The `@opentelemetry/api` package is instrumented so that usage of
   `.metrics.getMeterProvider(...)` will use one provided by the APM agent,
   if the user hasn't explicitly provided one of their own. This implementation
   uses a version of the OTel Metrics SDK included with the APM agent. It
   is configured to sent metrics to APM server by default. This allows a user
   to use the OTel Metrics API and get metrics set to APM server without any
   configuration.
2. Alternatively, the `@opentelemetry/sdk-metrics` package is instrumented so
   that when a user creates a `MeterProvider`, the APM agent will added a
   `MetricReader` that sends metrics to APM server.  This allows a user to
   configure metrics as they wish using the OTel Metrics SDK and then
   automatically get those metrics sent to APM server.

Closes: https://github.com/elastic/apm-agent-nodejs/issues/2954
Refs: https://github.com/elastic/apm/issues/691
Refs: https://github.com/elastic/apm/pull/742 (spec PR)


lib/agent.js
```diff
@@ -749,3 +793,22 @@ Agent.prototype.registerMetric = function (name, labelsOrCallback, callback) {
...
 }
+
+/**
+ * Return true iff the given metric name is "disable", according to the
+ * `disableMetrics` config var.
+ *
+ * @returns {boolean}
+ */
+Agent.prototype._isMetricNameDisabled = function (name) {
+  if (!this._conf.disableMetricsRegExp) {
```
david-luna, line 804: This property is initialised as an empty array in `config.js` file. I think if we can ensure that the type stays the same (array) we can check on the length of it for early returns.

trentm: There is this one example of a "mock" Agent._conf object that doesn't have `.disableMetricsRegExp` set:
https://github.com/elastic/apm-agent-nodejs/blob/trentm/otel-metrics-poc/test/lambda/mock/agent.js#L19-L23
Ideally that example would have a more complete default `Config` instance from "lib/config.js".

However, tests pass when I comment out the `if (!this._conf.disableMetricsRegExp) {` guard, so things are fine right now.

trentm: I dropped the guard in commit 3feae2ba.

PR #3153, opened 2023-02-08
Add update-spec updatecli workflow

reakaleek:

Moves the update specs workflow from apm-server to the consumer using updatecli to have a "dependabot"-like experience.

## How to test

```
GITHUB_TOKEN=<token>\
GIT_USER=<username> \
GIT_EMAIL=j<email> \
updatecli diff -c .ci/update-specs.yml
```

## Related issues

- https://github.com/elastic/observability-robots/issues/1516


trentm (commented):
1. Should this PR also handle removing the now obsolete (I think) https://github.com/elastic/apm-agent-nodejs/blob/main/test/integration/api-schema/apm-server-schema/log.json file? It was removed from apm-server.git (where these spec files used to be sync'd from) in https://github.com/elastic/apm-server/pull/9349

2. I guess `updatecli` doesn't support mapping a full directory or a glob pattern?

reakaleek:
> 1. Should this PR also handle removing the now obsolete (I think) https://github.com/elastic/apm-agent-nodejs/blob/main/test/integration/api-schema/apm-server-schema/log.json file? It was removed from apm-server.git (where these spec files used to be sync'd from) in [elastic/apm-server#9349](https://github.com/elastic/apm-server/pull/9349)
> 2. I guess `updatecli` doesn't support mapping a full directory or a glob pattern?

1. Since, the log.json file is also not part of https://github.com/elastic/apm-data/tree/main/input/elasticapm/docs/spec/v2, I guess it's safe to remove it. Will do that in this PR.

2. Yes, unfortunately there is no native way to sync/manipulate multiple files at once. We concluded, that having hardcoded list is feasible in this case because the amount of files rarely changes.

reakaleek:
@trentm 

FYI, there is an issue to bundle the schemas into one file.

https://github.com/elastic/apm-data/issues/15

Hence, the mentioned issues will be solved there.

trentm (approved):

PR #3167, opened 2023-02-24
feat: Make Agent.flush() return a Promise if no callback is passed as param

### LINKED ISSUE #2857: `await`able `apm.flush()`
To use `apm.flush(callback)` effectively in an async function one has to resort to promisify'ing it like this:

```js
const { promisify } = require('util')
await promisify(apm.flush.bind(apm))().then(function (err) {
  ...
})
```

It would be helpful if `apm.flush()` could return a Promise if no callback is given, then it could be `await`d directly.

adamyeats:
Fixes #2857.

This PR modifies `Agent.flush()` to return a `Promise` if no `callback` parameter is supplied to the function. This allows the method to be usable with the `await` keyword.

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -50,7 +50,7 @@ Notes:
 * Extend Lambda instrumentation to capture details for Lambda function URL
   and ELB-triggered Lambdas. ({issues}2901[#2901])
 
-* Make `Agent.flush()` return a `Promise` if no callback is passed as param ({issues}2857(#2857)
+* Make `Agent.flush()` return a `Promise` if no callback is passed as param. ({issues}2857(#2857))
 
 [float]
 ===== Bug fixes
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -50,6 +50,8 @@ Notes:
 * Extend Lambda instrumentation to capture details for Lambda function URL
   and ELB-triggered Lambdas. ({issues}2901[#2901])
 
+* Make `Agent.flush()` return a `Promise` if no callback is passed as param ({issues}2857(#2857)
+
 [float]
 ===== Bug fixes
 
--- a/test/agent.test.js
+++ b/test/agent.test.js
@@ -1148,6 +1148,21 @@ test('#flush()', function (t) {
     })
   })
 
+  t.test('flush can be used without a callback to return a Promise', function (t) {
+    t.plan(1)
+
+    const agent = new Agent()
+
+    agent.flush().then(function () {
+      t.pass('should resolve the Promise for agent.flush')
+    }).catch(function (err) {
+      t.error(err, 'no error passed to agent.flush callback')
+    }).finally(function () {
+      agent.destroy()
+      t.end()
+    })
+  })
+
   t.end()
 })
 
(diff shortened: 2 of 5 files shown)


trentm:
Hi @adamyeats. Nice. This is still "draft", so I'm not sure if you wanted review on it yet. The implementation looks perfect to me.

1. Would you be willing to add a test? A `test(...)` in this block should suffice: https://github.com/elastic/apm-agent-nodejs/blob/main/test/agent.test.js#L932    That test file can be run directly via `node test/agent.test.js`.
2. To pass the current (semicolon-hostile) style (`npm run lint:eslint`, aka `make check`) you'll need to drop the semicolons.

adamyeats:
Hey @trentm, great to make contact with you! 👋 As you noticed, I haven't added tests yet thus it was still in draft, but thank you for the review on what was already there.

- Test is incoming for this, I'll add something ASAP!
- Thanks for picking up on the formatting, looks like ESlint integration was broken in my copy of VS Code. This should be resolved in 5776936.

Quick question: is there anywhere in the documentation that will need to be edited? If so, can you point me in the right direction?

trentm:
> is there anywhere in the documentation that will need to be edited?

Yah, look for "apm.flush" in https://github.com/elastic/apm-agent-nodejs/blob/main/docs/agent-api.asciidoc
Thanks.

adamyeats:
@trentm This should now be ready for review, thanks!

CHANGELOG.asciidoc
```diff
@@ -50,6 +50,8 @@ Notes:
 * Extend Lambda instrumentation to capture details for Lambda function URL
   and ELB-triggered Lambdas. ({issues}2901[#2901])
 
+* Make `Agent.flush()` return a `Promise` if no callback is passed as param ({issues}2857(#2857)
```
trentm, line 53:
```suggestion
* Make `Agent.flush()` return a `Promise` if no callback is passed as param. ({issues}2857(#2857))
```

nit

test/agent.test.js
```diff
@@ -1148,6 +1148,21 @@ test('#flush()', function (t) {
...
+    const agent = new Agent()
+
+    agent.flush().then(function () {
+      t.pass('should resolve the Promise for agent.flush')
+    }).catch(function (err) {
+      t.error(err, 'no error passed to agent.flush callback')
+    }).finally(function () {
+      agent.destroy()
+      t.end()
+    })
```
trentm, line 1163: Unfortunately we still support and test with node v8, which [doesn't support Promise.finally](https://node.green/#ES2018-features-Promise-prototype-finally). I think it would be fine to put the `agent.destroy(); t.end()` in the `.then(...)` handler and call it a day.

trentm (approved):
@adamyeats This is great, thanks!  One nit, and one thing in the test. I'm guessing the test with node v8 will fail. We shall see.  I'm "approving" because, IIRC, our Jenkins CI guards require an approval before running contributed code.

trentm (approved):

test/agent.test.js
adamyeats: Ah yes, those were the days! Agreed, tweaked this in 6a6afea.

trentm:

PR #3178, opened 2023-02-28
fix: transaction name for Next.js API routes in next@13.2 was broken

trentm:
Before this change internal changes in next@13.2.0 resulted in
transactions for Next.js API routes being `{method} unknown route`.

### DIFF
--- a/test/instrumentation/modules/next/next.test.js
+++ b/test/instrumentation/modules/next/next.test.js
@@ -29,6 +29,7 @@ const os = require('os')
 const path = require('path')
 const semver = require('semver')
 const tape = require('tape')
+const treekill = require('tree-kill')
 
 const { MockAPMServer } = require('../../../_mock_apm_server')
 const { formatForTComment } = require('../../../_utils')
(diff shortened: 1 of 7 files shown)


trentm:
run module tests for next

trentm:
run module tests for next

trentm:
run module tests for next

test/instrumentation/modules/next/next.test.js
```diff
@@ -774,7 +775,7 @@ tape.test('-- dev server tests --', suite => {
       waitForServerReady(t, waitErr => {
         if (waitErr) {
           t.fail(`error waiting for Next.js server to be ready: ${waitErr.message}`)
-          nextServerProc.kill('SIGKILL')
+          treekill(nextServerProc.pid, 'SIGKILL')
```
trentm, line 778: REVIEW NOTE: The `treekill` usage was added because on macOS I was running into a failure in `TAV=next npm run test:tav` when the *second* `next` version was run. The used port (port 3000) was still busy when the `next start` server was started for the second run. It was as if the process for the `next dev` server from the previously tested version was still running.  The `treekill` usage proved useful for a similar case of a process *tree* in the azure-functions tests.

Issue #3179, opened 2023-03-01
ci: getting occasional 429 errors in GH Actions


trentm:
An error I'm noticing recently is:

```
/usr/bin/git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +ec5049de109c35acf9fd647189a43da006b841bc:refs/remotes/pull/3178/merge
  Error: fatal: unable to access 'https://github.com/elastic/apm-agent-nodejs/': The requested URL returned error: 429
  Error: The process '/usr/bin/git' failed with exit code 128
```

For example in: https://github.com/elastic/apm-agent-nodejs/actions/runs/4306737009/jobs/7510930064

Basically the first step in each test runner (`actions/checkout@v2` to clone the repo) is occasionally getting a 429 (Too Many Requests) error. Usually, if I manually re-run the tests it works fine. However, this has happened a few times now.

I wonder if the large number of jobs (26 for the "test" workflow) is contributing to the issue. 
https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting

Some suggestions/ideas:

1. Try using a personal access token, which theoretically can increase rate limits. Note that https://github.com/actions/checkout/issues/775 discusses this issue and a commenter there suggested using an access token didn't help.
2.  Try `actions/checkout@v3`. Suggested by Victor. Perhaps this retries use exponential backoff with jitter that could theoretically help?
3. Try limited parallelism of the workflow via `max-parallel` (https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#defining-the-maximum-number-of-concurrent-jobs). Suggested by Victor.

v1v:
I've just requested some help to the GitHub support

v1v:
GitHub supported suggested to bump the version v2 to v3

<img width="1041" alt="image" src="https://user-images.githubusercontent.com/2871786/223067268-8da07912-bd05-4eb5-b717-d26cbe987554.png">


as there was a [known issue](https://github.com/actions/checkout/issues/417)

PR #3180, opened 2023-03-01
ci: limit parallel GH Action test runs to try to avoid 429 errors on checkout

trentm:
Fixes: #3179

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,8 +23,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
-    - uses: actions/setup-node@v2
+    - uses: actions/setup-node@v3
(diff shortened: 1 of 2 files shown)


trentm:
Currently with "lint" and "test" workflows we have 27 jobs. It would be nice if we only *doubled* the execution time with `max-parallel`. Therefore I'll try 15 to start (more than half of 27).

trentm:
I'm not sure this will actually help. To *know* we'd ideally get some rate limiting headers from the failing `actions/checkout` requests. Also we'd want to understand what the rate limits actually are in this case. I'm not sure from https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting which case applies here. Is a workflow in a GitHub Action a "GitHub App"? Etc.

PR #3246, opened 2023-04-05
action: support github commands for the tav modules

v1v:
### What

Support `/test tav`, `/test tav module1,module2`, `/test tav module1,module2 18,16`, `/test tav all` and `/test tav all 16`

### Why

Help with running specific modules for specific node versions, in addition, the default behaviour will read the versions and modules from `.ci/tav.json`

### Test

[`/test tav`](https://github.com/elastic-robots/gh-actions-interactions/pull/3#issuecomment-1498127791) triggered https://github.com/elastic-robots/gh-actions-interactions/actions/runs/4622685839

[`/test tav all`](https://github.com/elastic-robots/gh-actions-interactions/pull/3#issuecomment-1498131397) triggered https://github.com/elastic-robots/gh-actions-interactions/actions/runs/4622723259 (similarly to `/test tav`)

[`/test tav acme,foo`](https://github.com/elastic-robots/gh-actions-interactions/pull/3#issuecomment-1498128207) triggered https://github.com/elastic-robots/gh-actions-interactions/actions/runs/4622688577

[`/test tav all 1,2`](https://github.com/elastic-robots/gh-actions-interactions/pull/3#issuecomment-1498128426) triggered https://github.com/elastic-robots/gh-actions-interactions/actions/runs/4622689567

[`/test tav acme,foo 2`](https://github.com/elastic-robots/gh-actions-interactions/pull/3#issuecomment-1498131397) triggered https://github.com/elastic-robots/gh-actions-interactions/actions/runs/4622709699

### Update

Changed to use the `pull_request_review` event so we can warranty every run will be attached to a particular sha commit. https://github.com/elastic-robots/gh-actions-interactions/pull/4 can be used for testing a similar workflow.


.github/workflows/tav-command.yml
```diff
@@ -18,35 +18,66 @@ jobs:
...
+            }
+            const matrix = await f()
+            let modules, versions
+            try {
+              versions = matrix.versions
+              modules = matrix.modules
+            } catch (err) {
+              core.setFailed("Error parsing JSON string")
+              return
+            }
```
trentm, line 57:
```suggestion
            const fs = require('fs')

            let modules, versions
            try {
              const matrix = JSON.parse(fs.readFileSync('./.ci/tav.json'))
              versions = matrix.versions
              modules = matrix.modules
            } catch (err) {
              core.setFailed(`Error loading './.ci/tav.json': ${err}`)
              return
            }
```

Could perhaps simplify this block with `fs.readFileSync` and just a single error catch.

.github/workflows/tav.yml
```diff
@@ -24,21 +24,41 @@ jobs:
...
+            let modules, versions
+            try {
+              versions = matrix.versions
+              modules = matrix.modules
+            } catch (err) {
+              core.setFailed("Error parsing JSON string")
+              return
+            }
+            core.setOutput('modules', modules)
+            core.setOutput('versions', versions)
```
trentm, line 54:
```suggestion
            const fs = require('fs')
            let matrix
            try {
              matrix = JSON.parse(fs.readFileSync('./.ci/tav.json'))
            } catch (err) {
              core.setFailed(`Error loading './.ci/tav.json': ${err}`)
              return
            }
            core.setOutput('modules', matrix.modules)
            core.setOutput('versions', matrix.versions)
```

Perhaps this.

.github/workflows/tav-command.yml
```diff
@@ -18,35 +18,66 @@ jobs:
...
+            const comment = context.payload.comment.body
+            if (comment !== '/test tav') {
+              const regex = /\/test tav ([^\s\\]+)(\s*)([^\s\\]*)/
```
trentm, line 60: I'm curious why `\\` is included in the `[^\s\\]` parts.

.github/workflows/tav-command.yml
```diff
@@ -18,35 +18,66 @@ jobs:
...
+              const match = comment.match(regex)
+              console.log(`parse comment - ${match}`)
+              if (match[1]) {
+                if (match[1] !== 'all') {
+                  modules = match[1].split(',')
+                }
+              }
+              if (match[3]) {
+                versions = match[3].split(',')
+              }
```
trentm, line 70: This should guard against the `match` possibly being `null`, otherwise `match[1]` could throw an error.

```js
> const regex = /\/test tav ([^\s\\]+)(\s*)([^\s\\]*)/

> comment = '/test tavern'
'/test tavern'
> match = comment.match(regex)
null
> match[0]
Uncaught TypeError: Cannot read properties of null (reading '0')

> comment = '/test tav\nblah blah blah\n'
'/test tav\nblah blah blah\n'
> match = comment.match(regex)
null
```

trentm (commented):
Thanks! This is looking good.

.github/workflows/tav-command.yml
v1v: Unrequired, any char but whitespaces rather than whitespaces/back-slashes

v1v:
```suggestion
              const regex = /\/test tav ([^\s]+)(\s*)([^\s]*)/
```

.github/workflows/tav-command.yml
v1v: https://github.com/elastic/apm-agent-nodejs/pull/3246/commits/9e3c64a7bd1458a927caf6cf6cf684373cb3ba2d is the guard and 
https://github.com/elastic-robots/gh-actions-interactions/actions/runs/4626017105 was caused with https://github.com/elastic-robots/gh-actions-interactions/pull/3#issuecomment-1498544288

.github/workflows/tav.yml
```diff
@@ -24,21 +24,34 @@ jobs:
...
+            core.setOutput('modules', matrix.modules)
+            core.setOutput('versions', matrix.versions)
 
   test-tav:
     needs: prepare-matrix
     runs-on: ubuntu-latest
     timeout-minutes: 40
     strategy:
-      max-parallel: 30
+      max-parallel: 15
```
v1v, line 54: There is no need to run in parallel more since it's a merge build, so it's not happening on PRs.

.github/workflows/tav.yml
```diff
@@ -24,21 +24,34 @@ jobs:
...
-    - uses: actions/checkout@v3
-    ## TODO: use elastic/apm-pipeline-library/.github/actions/version-framework
-    ##       as soon as https://github.com/elastic/apm-pipeline-library/issues/2171 is done.
-    - id: generate-matrix
-      run: .ci/generate-github-tav-matrix.sh
+
+      - uses: actions/checkout@v3
+
+      - id: transform
+        name: Transform comment to the supported matrix
```
trentm, line 34: nit: There is no comment in use for this workflow, so perhaps a name like "load matrix from tav.json" or something. However, I can understand if you want to keep the same `id` and `name` as the step in the "tav-command.yml" workflow.

trentm (approved):

.github/workflows/tav-command.yml
```diff
@@ -0,0 +1,103 @@
+name: tav-command
+
+on:
+  pull_request_review:
```
v1v, line 4: [pull_request_review](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review) worked like a charm but [pull_request_review_comment](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_review_comment) didn't get triggered. In any case `pull_request_review` is the one linked to a particular sha commit, hence `github.event.pull_request.head.sha` is the correct data to be used when checking out the source code.

.github/workflows/tav.yml
v1v:
```suggestion
      - id: transform
        name: Load matrix from tav.json
```

reakaleek (approved):

reakaleek (approved):

david-luna:
Thanks @v1v ! this will help a lot to improve testing of the agent.

I wonder if you considered to have the scripts in separate file, se it in t[he docs](https://github.com/actions/github-script#run-a-separate-file)

david-luna (approved):

david-luna (approved):

v1v:
> I wonder if you considered to have the scripts in separate file, se it in t[he docs](https://github.com/actions/github-script#run-a-separate-file)

No strong opinion. If that's better I'm happy to change it.

david-luna:
> > I wonder if you considered to have the scripts in separate file, se it in t[he docs](https://github.com/actions/github-script#run-a-separate-file)
> 
> No strong opinion. If that's better I'm happy to change it.

No need to. We can move it later if the script grows enough.

Looks more than good to me :)

.github/workflows/tav.yml
trentm: I don't understand what you mean here. Our test matrix is approximately 250 jobs now. If we run only 15 in parallel it may take twice as long to complete them all. Currently (see some examples at https://github.com/elastic/apm-agent-nodejs/actions/workflows/tav.yml) a TAV run is taking 1.5h to complete. It would be unfortunate if it takes 3h to complete.

Is there a reason to want lower max-parallel? Is there added cost?

trentm (approved):
@v1v Thanks! Just the one question. I'm fine with merging before that question is answered, if that is easier. We can always adjust later.

v1v:
> Is there a reason to want lower max-parallel? Is there added cost?

It's not about cost, since we use the existing GitHub runners and no hosted runners but the GitHub quota for the whole Elastic org. Back in the days was about 200 though it has been increased to 500. See https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

Using a lower max-parallel can help with an even distributed load between all the GitHub repositories using GitHub actions at elastic.

In this particular case, since the whole `tav` does not run per PR but per merge, I thought that limiting the max-parallel can help to distribute the workload since it's not part of the critical path.

trentm:
> In this particular case, since the whole `tav` does not run per PR but per merge, I thought that limiting the max-parallel can help to distribute the workload since it's not part of the critical path.

Okay, fair. Thanks.

trentm:
Thanks!

PR #3266, opened 2023-04-13
chore: re-add require-in-the-middle dep

trentm:
It was temporarily inlined some releases ago while changes were
separately upstreamed. Since then there has been some dev on
the module that it would be good to keep up with.


trentm:
For the record, the CI was failing for Node.js 12.0 for the https://github.com/elastic/apm-agent-nodejs/pull/3266/commits because it included a git branch dependency, and Node.js 12.0 came with npm v6.9.0 and [npm v6.10.0](https://github.com/npm/cli/pull/205) fixed this error that the `npm install` was hitting:

```
[92f5e4c](https://github.com/npm/pacote/commit/92f5e4c) https://github.com/zkat/pacote/pull/170 fix(errors): Fix "TypeError: err.code.match is not a function" error ([@jviotti](https://github.com/jviotti))
```

such is the pain of supporting ancient releases.

PR #3276, opened 2023-04-18
test: fix s3.test.js failure in Docker on macos due to timeout

trentm:
It was found that running this test in Docker on mac -- e.g. by running `.ci/scripts/test.sh -b release 18` -- would fail. It would hit the set timeout for the `execFile` call. THis bumps the timeout and improves some logging output around that so it is easier to diagnose if it happens again.

### DIFF
--- a/test/instrumentation/modules/aws-sdk/s3.test.js
+++ b/test/instrumentation/modules/aws-sdk/s3.test.js
@@ -45,7 +45,7 @@ tape.test('simple S3 usage scenario', function (t) {
       TEST_REGION: 'us-east-2'
     }
     t.comment('executing test script with this env: ' + JSON.stringify(additionalEnv))
-    console.time('exec use-s3')
+    console.time && console.time('exec use-s3')
     execFile(
       process.execPath,
       ['fixtures/use-s3.js'],
@@ -56,7 +56,7 @@ tape.test('simple S3 usage scenario', function (t) {
         env: Object.assign({}, process.env, additionalEnv)
       },
       function done (err, stdout, stderr) {
-        console.timeLog('exec use-s3')
+        console.timeLog && console.timeLog('exec use-s3')
         t.error(err, 'use-s3.js did not error out')
         if (err) {
           t.comment('err: ' + util.inspect(err))
--- a/test/instrumentation/modules/aws-sdk/s3.test.js
+++ b/test/instrumentation/modules/aws-sdk/s3.test.js
@@ -21,6 +21,7 @@ if (process.env.GITHUB_ACTIONS === 'true' && process.platform === 'win32') {
 //   cannot be tested.
 
 const { execFile } = require('child_process')
+const util = require('util')
 
 const tape = require('tape')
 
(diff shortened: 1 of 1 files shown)


trentm:
The execFile timeout had been 10s. It was found that on my macos laptop running in Docker Desktop the execFile would take ~13-14s.

Using this patch to gather timing info:

```diff
diff --git a/test/instrumentation/modules/aws-sdk/fixtures/use-s3.js b/test/instrumentation/modules/aws-sdk/fixtures/use-s3.js
index 3fdf8e5a..1e1eafa5 100644
--- a/test/instrumentation/modules/aws-sdk/fixtures/use-s3.js
+++ b/test/instrumentation/modules/aws-sdk/fixtures/use-s3.js
@@ -40,6 +40,9 @@
 // Output from a sample run is here:
 // https://gist.github.com/trentm/c402bcab8c0571f26d879ec0bcf5759c

+console.time('use-s3')
+console.timeLog('use-s3', '[use-s3] top')
+
 const apm = require('../../../../..').start({
   serviceName: 'use-s3',
   captureExceptions: false,
@@ -49,11 +52,15 @@ const apm = require('../../../../..').start({
   stackTraceLimit: 4, // get it smaller for reviewing output
   logLevel: 'info'
 })
+console.timeLog('use-s3', '[use-s3] after apm start')

 const crypto = require('crypto')
 const vasync = require('vasync')
+console.timeLog('use-s3', '[use-s3] before aws-sdk require')
 const AWS = require('aws-sdk')
+console.timeLog('use-s3', '[use-s3] after aws-sdk require')
 const assert = require('assert')
+console.timeLog('use-s3', '[use-s3] after all requires')

 const TEST_BUCKET_NAME_PREFIX = 'elasticapmtest-bucket-'

```

Here is timing from Docker Desktop on Mac:

```
$ .ci/scripts/test.sh -b "rc" -f 20
...
node_tests_1  | # use-s3: 0.055ms [use-s3] top
node_tests_1  | # use-s3: 1.737s [use-s3] after apm start
node_tests_1  | # use-s3: 1.806s [use-s3] before aws-sdk require
node_tests_1  | # use-s3: 13.260s [use-s3] after aws-sdk require
node_tests_1  | # use-s3: 13.260s [use-s3] after all requires
```

compared to running directly on my Mac:

```
% node test/instrumentation/modules/aws-sdk/s3.test.js
...
# use-s3: 0.065ms [use-s3] top
# use-s3: 226.955ms [use-s3] after apm start
# use-s3: 233.407ms [use-s3] before aws-sdk require
# use-s3: 897.404ms [use-s3] after aws-sdk require
# use-s3: 897.509ms [use-s3] after all requires
```


If I disable RITM (require-in-the-middle) handling, then the timing was closer to:

```
use-s3: 1.249s [use-s3] after apm
use-s3: 1.290s [use-s3] before require aws-sdk
use-s3: 2.408s [use-s3] after require aws-sdk
```

So what work is RITM doing? At a guess it is this:

    filename = Module._resolveFilename(id, this)

and that is doing file stats and those are DOG SLOW on Docker for Mac.
Not sure, though.


With the added output from this PR, we'll be able to see if it is a concern in CI tests. If not, then this isn't a high enough priority to dig into.

trentm:
For time comparison, in GH Actions CI for the node 18 test run the timing was:

```
exec use-s3: 976.25ms
```

FAR from the 13s in Docker for Mac.  So we should be all good here -- this isn't important to dig into.

PR #3278, opened 2023-04-19
chore: update versions tested now that Node.js v20 is out

trentm:
Node v20.0.0 is out (https://nodejs.org/en/blog/release/v20.0.0).   This updates to support that.

1. Change "engines" to just be `>=8.6.0` instead of explicitly listing all the even and current releases. This means that we'll avoid users getting this "npm WARN EBADENGINE" when a new release comes out:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'elastic-apm-node@3.44.1',
npm WARN EBADENGINE   required: { node: '^8.6.0 || 10 || 12 || 14 || 16 || 17 || 18 || 19' },
npm WARN EBADENGINE   current: { node: 'v20.0.0', npm: '9.6.4' }
npm WARN EBADENGINE }
```

at the cost of users of older elastic-apm-node versions perhaps assuming it supports all future Node.js releases.

In practice, I don't see any other npm packages using "engines" the way we have been doing. In typical practice it is, at best, used as a hint for the minimum node version supported. 

2. Update test workflows as appropriate for a v20 release.

On fall out here is that our full TAV test matrix is now greater than the GH Actions limit of 256 jobs for a workflow (8 node versions times 36 modules = 288). So for now I'm sacrificing testing against node v19. Node.js v19 will be EOL at 2023-06-01.

### DIFF
--- a/.ci/tav.json
+++ b/.ci/tav.json
@@ -1,5 +1,6 @@
 {
-  "versions": [ "19", "18", "16", "14", "12", "10", "8" ],
+  "// todo": "We want versions=['20','19','18','16','14','12','10','8'], but versions*modules needs to be <256 for the GH Actions jobs limit",
+  "versions": [ "20", "18", "16", "14", "12", "10", "8" ],
   "modules": [
     "@elastic/elasticsearch",
     "@elastic/elasticsearch-canary",
--- a/.github/workflows/edge.yml
+++ b/.github/workflows/edge.yml
@@ -37,7 +37,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [ "20" ]
+        node: [ "21" ]
         contextManager: [ 'patch', '' ]
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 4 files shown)


trentm:
Note: I expect that `test-vers (20) ...` will fail if tehre isn't yet a "node:20" image on Docker Hub. Often that takes a few days after the Node core release.

trentm:
> Note: I expect that `test-vers (20) ...` will fail if tehre isn't yet a "node:20" image on Docker Hub.

Yet they did *not* fail.

... the reason for this is that with the migration of CI from Jenkins to GH Actions the "test.yml" workflow tests no longer run via the ".ci/scripts/test.sh" driver that uses ".ci/docker/node-container/Dockerfile" which includes `FROM node:${NODE_VERSION}`.  Currently only the "tav.yml" tests are using this Dockerfile. So it is only TAV tests with v20 that will fail until the Docker container is available.

trentm:
There is a `node:20` docker image up now, so we can move ahead.

Issue #3279, opened 2023-04-20
"azure-functions.test.js" test failure: `Syntax error: "(" unexpected`


trentm:
Recently, there has been a new failure in the azure-functions test:

```
node_tests_1     | running test: cd . && node --unhandled-rejections=strict test/instrumentation/azure-functions/azure-functions.test.js > test_output/test-instrumentation-azure-functions-azure-functions.test.js.tap 2&>1
node_tests_1     |
node_tests_1     | TAP version 13
node_tests_1     | # SKIP setup: npm ci (in /app/test/instrumentation/azure-functions/fixtures/AJsAzureFnApp)
node_tests_1     | # azure functions
node_tests_1     | # setup
node_tests_1     | # mock APM apmServerUrl: http://localhost:43399
node_tests_1     | # setup: "func start" for AJsAzureFnApp fixture
node_tests_1     | # ["func start" stderr] /app/test/instrumentation/azure-functions/fixtures/AJsAzureFnApp/node_modules/azure-functions-core-tools/bin/func: 1: Syntax error: "(" unexpected
node_tests_1     | not ok 1 "func start" failed early: code=2
node_tests_1     |   ---
node_tests_1     |     operator: fail
node_tests_1     |     at: ChildProcess.onEarlyClose (/app/test/instrumentation/azure-functions/azure-functions.test.js:484:9)
node_tests_1     |     stack: |-
node_tests_1     |       Error: "func start" failed early: code=2
node_tests_1     |           at Test.assert [as _assert] (/app/node_modules/tape/lib/test.js:312:48)
node_tests_1     |           at Test.bound [as _assert] (/app/node_modules/tape/lib/test.js:95:17)
node_tests_1     |           at Test.fail (/app/node_modules/tape/lib/test.js:406:7)
node_tests_1     |           at Test.bound [as fail] (/app/node_modules/tape/lib/test.js:95:17)
node_tests_1     |           at ChildProcess.onEarlyClose (/app/test/instrumentation/azure-functions/azure-functions.test.js:484:9)
node_tests_1     |           at ChildProcess.emit (node:events:513:28)
node_tests_1     |           at maybeClose (node:internal/child_process:1091:16)
node_tests_1     |           at ChildProcess._handle.onexit (node:internal/child_process:302:5)
node_tests_1     |   ...
node_tests_1     | # make requests
node_tests_1     | ok 2 there is no fnAppProc # SKIP
node_tests_1     | # check all APM events
[...]

PR #3280, opened 2023-04-20
ci: allow the 'test.yml' CI to be run manually

trentm:
To be able to trigger a workflow manually requires `on: workflow_dispatch`.
Let's do that for the main "test" workflow. Then we can:

```
gh workflow run test.yml
```

Eventually it might be nice to have input params to limit the node versions run, etc. but that can be done separately.


trentm:
The `test-vers (18)` failures are due to https://github.com/elastic/apm-agent-nodejs/issues/3279 for which we have a workaround now.

Issue #3136, opened 2023-02-01
[META 753] Support Partial Transaction for AWS Lambda


elastic-apm-tech:
See meta issue for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/753

trentm:
See this coming lambda extension change: https://github.com/elastic/apm-aws-lambda/pull/384
We should work with and without an extension supporting that.

trentm:
FWIW, here is the Python PR adding this support: https://github.com/elastic/apm-agent-python/issues/1734

trentm:
![Screenshot 2023-04-21 at 12 59 50 PM](https://user-images.githubusercontent.com/46866/233724915-c5572212-0262-4697-af55-a46c9eae2335.png)

Here is an example of the coming PR for this showing how a timed-out lambda invocation now gets a reported (failed) transaction.

Issue #2379, opened 2021-10-18
lambda wrapper doesn't handle uncaughtException or unhandleRejection


trentm:
A lambda handler like this:

```js
exports.handler = apm.lambda(function handlerThrows3 (event, context) {
  console.warn('XXX start 100ms timeout')
  setTimeout(function () {
    console.warn('XXX go boom3')
    throw new Error('boom3')
  }, 100)
})
```

results in an uncaught exception. The current experimental Lambda wrapping doesn't trap this.

* * *

The Lambda Runtime does this:

```typescript
  private _setErrorCallbacks(invokeId: string): void {
    this.errorCallbacks.uncaughtException = (error: Error): void => {
      this.client.postInvocationError(error, invokeId, () => {
        process.exit(129);
      });
    };
    this.errorCallbacks.unhandledRejection = (error: Error): void => {
      this.client.postInvocationError(error, invokeId, () => {
        process.exit(128);
      });
    };
  }
```

I.e. it notes the error, then `process.exit(SPECIAL_VALUE)`s. I'm not sure if the runtime has overridden or is somehow handling those process.exit values. My limited observation is that the runtime node process is *not* restarting, because a subsequent function invocation uses the same VM and doesn't run through the APM agent startup again.

Some things to think about when handling this:

- [ ] Does the APM agent's optional `uncaughtException` handler (in agent.js) collide here? Ditto for a coming `unhandleRejection` handler.
- [ ] Do we have a timing race with `this.client.postInvocationError(...)` here if we want to note the error and send data to the extension?

trentm:
Some more details on my recent testing of this.

Summary:

| error case          | description                                | lambda response unchanged? | transaction reported? | error reported? |
| ------------------- | ------------------------------------------ | -------------------------- | --------------------- | --------------- |
| exception           | throw at top-level in a non-async function | yes (1) | yes | yes |
| uncaughtException   | throw in async part of non-async function  | yes (1) | no | no |
| unhandledRejection1 | rejecting a promise not handled with `.catch()` | yes | no | no |
| unhandledRejection2 | throwing inside an async function without a catch block | yes (1) | yes | yes |

Note (1): While the Lambda handler response not changed meaningfully (a good thing), there
is a change in the stacktrace due to the instrumentation. This is acceptable.


## exception (good)

Handler:

```js
exports.handler = function errorThrowSync () {
  throw new Error('errorThrowSync')
}
```

Response without agent (pretty-printed):

```
{
[...]

PR #3285, opened 2023-04-25
feat: pre-register partial transactions in AWS Lambda

trentm:
When used with the Lambda extension >=v1.4.0, this results in
transactions being reported for Lambda timeout, uncaughtException,
and unhandledRejection.

Closes: #3136
Closes: #2379

# checklist

- [x] depends on https://github.com/elastic/apm-nodejs-http-client/pull/199 and updating the package.json dep to a release of the http-client

# follow-up work

- Drop AgentMock in tests and re-add support for `contextManager: patch` usage (see https://github.com/elastic/apm-agent-nodejs/pull/3285#issuecomment-1523908892)
- Consider a separate issue to look at replacing lambda-local usage for local testing with https://github.com/aws/aws-lambda-runtime-interface-emulator that Jonas has been looking at for the Java agent. It runs the AWS-provided emulator in a Docker container, allows using the real Lambda extension. It doesn't support the Logs API and not sure about the newer Telemetry API, so it isn't a perfect representation, but is a lot closer. Java agent PR: https://github.com/elastic/apm-agent-java/pull/3120
- Add a follow-up issue (to #2379) to possibly do better with reporting an error APM object for `unhandleException`. With this change we'll be reporting a transaction with `transaction.outcome = "failure"`, which is good, but there won't be an "error" object with a stacktrace that points to the location of the throw error.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -40,6 +40,15 @@ Notes:
 [float]
 ===== Features
 
+* Improve error handling with AWS Lambda. When used together with the
+  https://github.com/elastic/apm-aws-lambda[Elastic AWS Lambda extension]
+  v1.4.0 or greater, the APM agent will pre-register a partial transaction
+  before the user's handler function is run. If the handler function fails
+  with a Lambda timeout, `uncaughtException`, or `unhandledRejection` then
+  the Lambda extension will report the failed transaction so it can be
+  seen in the Kibana APM app. ({issues}3136[#3136])
+
+
 [float]
 ===== Bug fixes
 
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,8 +44,8 @@ Notes:
   https://github.com/elastic/apm-aws-lambda[Elastic AWS Lambda extension]
   v1.4.0 or greater, the APM agent will pre-register a partial transaction
   before the user's handler function is run. If the handler function fails
-  with a Lambda timeout, `uncaughtException`, or `unhandledRejection` then
-  the Lambda extension will report the failed transaction so it can be
+  with a Lambda timeout, `uncaughtException`, `unhandledRejection`, or crash
+  then the Lambda extension will report the failed transaction so it can be
   seen in the Kibana APM app. ({issues}3136[#3136])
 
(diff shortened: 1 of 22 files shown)


trentm:
### A note on support for `contextManager: patch` usage

To support the "**synchronously** pre-register a transaction with the Lambda extension", were are necessarily using Promises in the `wrappedLambdaHandler` instrumentation now. This breaks async context tracking for the `contextManager: patch` config option usage. The "patch" context manager is deprecated, so this isn't considered a big deal.  **However**, it can be supported again via this diff:

```diff
diff --git a/lib/lambda.js b/lib/lambda.js
index 870c0b4f..df27d71a 100644
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -438,6 +438,7 @@ function setS3SingleData (trans, event, context, faasId, isColdStart) {

function elasticApmAwsLambda (agent) {
  const log = agent.logger
+  const ins = agent._instrumentation

  /**
    * Register this transaction with the Lambda extension, if possible.  This
@@ -644,13 +645,14 @@ function elasticApmAwsLambda (agent) {
      // Note: Wrapping context needs to happen *before any `await` calls* in
      // this function, otherwise the Lambda Node.js Runtime will call the
      // *unwrapped* `context.{succeed,fail,done}()` methods.
+      const runContext = ins.currRunContext()
      wrapContext(trans, event, context, triggerType)
      const wrappedCallback = wrapLambdaCallback(trans, event, context, triggerType, callback)

      await registerTransaction(trans, context.awsRequestId)

      try {
-        var retval = fn.call(this, event, context, wrappedCallback)
+        const retval = ins.withRunContext(runContext, fn, this, event, context, wrappedCallback)
        if (retval instanceof Promise) {
          return retval
        } else {
```

or similar.


The issue with applying that diff right now is that it adds usage of `agent._instrumentation` in this instrumentation, which is good and fine, but breaks the `AgentMock` being used in a large subset of `test/lambda/*.test.js`. It'll involve either another fake bandaid to `AgentMock` to have it start simulating a subset of `agent._instrumentation.*` behaviour **or** hundreds of lines of changes to the lambda tests to remove usage of AgentMock.   I want to do the latter.

However, I want to do the latter in a separate change to not add noise to this change.

trentm:
> However, I want to do the latter in a separate change to not add noise to this change.

https://github.com/elastic/apm-agent-nodejs/pull/3305 for this

Issue #2955, opened 2022-09-28
AWS instrumentation for SDK v3, S3 


estolfo:
Provide instrumentation for AWS SDK version 3 S3 client

See https://github.com/elastic/apm-agent-nodejs/pull/2124 for draft work.

david-luna:
Reviewed `@aws-sdk/node-http-handler` package which is used by the clients (S3, SNS, SQS) and instrumented it. The approach is to see if there is a way to identify the client from the request and only instrument `node-http-handler` module for all AWS clients

After instrumenting it I saw the requests have the following shapes

For `ListBucketsCommand` request
```
HttpRequest {
  method: 'GET',
  hostname: 's3.us-east-2.amazonaws.com',
  port: undefined,
  query: {},
  headers: {
    'content-type': 'application/xml',
    host: 's3.us-east-2.amazonaws.com',
    'x-amz-user-agent': 'aws-sdk-js/3.319.0',
    'user-agent': 'aws-sdk-js/3.319.0 os/darwin/22.4.0 lang/js md/nodejs/18.16.0 api/s3/3.319.0',
    'amz-sdk-invocation-id': '001e0d75-xxx',
    'amz-sdk-request': 'attempt=1; max=3',
    'x-amz-date': '20230425T171036Z',
    'x-amz-content-sha256': 'e3b...',
    authorization: 'AWS4-HMAC-SHA256 Credential=...'
  },
  body: '',
  protocol: 'https:',
  path: '/'
}
```

For `CreateBucketCommand` request
```
HttpRequest {
  method: 'PUT',
  hostname: 's3.us-east-2.amazonaws.com',
  port: undefined,
  query: {},
  headers: {
    'content-type': 'application/xml',
    'content-length': '497',
    Expect: '100-continue',
    host: 's3.us-east-2.amazonaws.com',
    'x-amz-user-agent': 'aws-sdk-js/3.319.0',
    'user-agent': 'aws-sdk-js/3.319.0 os/darwin/22.4.0 lang/js md/nodejs/18.16.0 api/s3/3.319.0',
    'amz-sdk-invocation-id': '50a4183f-xxx',
    'amz-sdk-request': 'attempt=1; max=3',
    'x-amz-date': '20230425T171037Z',
    'x-amz-content-sha256': '544f...',
    authorization: 'AWS4-HMAC-SHA256 Credential=...'
  },
[...]

PR #3287, opened 2023-04-26
feat: add instrumentation for aws-sdk S3 client

david-luna:
### Description

This change-set add instrumentation for `@aws-sdk/client-s3` which is available in v3. The approach is not to instrument the `S3Client` directly but the base `Client` class so we can add specific middleware functions based on each client without having to instrument them. 

Fixes: #2955

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
```diff
@@ -0,0 +1,337 @@
...
+      // https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/listbucketscommand.html
+      // Limitation: this doesn't handle paging.
+      function listAllBuckets (arg, next) {
+        const command = new ListBucketsCommand({})
+
+        toCallback(s3Client.send(command), function (err, data) {
+          log.info({ err, data }, 'listBuckets')
+          // TODO: need to check why this assertion fails
+          // assert(apm.currentSpan === null,
+          //   'S3 span should NOT be a currentSpan in its callback')
```
trentm, line 108: My first guess is that this will fixed if you change from using `agent.startSpan(...)` to `agent._instrumentation.createSpan(...)` -- as is now used in most (all?) instrumentations.

*Why* that makes a difference can be a long discussion, which we should have at some point. A starting point on using the Instrumentation API is the "Overview" section of https://github.com/elastic/apm-agent-nodejs/pull/2181

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -0,0 +1,207 @@
...
+      options: { step: 'initialize', priority: 'high', name: 'elasticAPMSpan' }
+    },
+    {
+      middleware: (next, context) => async (args) => {
+        context[elasticAPMStash] = {
+          hostname: args.request.hostname,
+          port: args.request.port,
+          protocol: args.request.protocol
+        }
+        return await next(args)
```
trentm, line 198: I'm no longer confident that the right way to write a aws-sdk middleware is to `await` the `next(args)` call. Do you know of a reference for that?  Looking at some examples for "middleware-sdk-s3":

- this one does *not* await: https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-sdk-s3/src/check-content-length-header.ts#L40
- this one *does* await: https://github.com/aws/aws-sdk-js-v3/blob/main/packages/middleware-sdk-s3/src/throw-200-exceptions.ts#L18

It would be good to have some comfort around that. For example, does this instrumentation work fine if this `await` is dropped?   I assume the await above in the "initialize" middleware is necessary.

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -0,0 +1,207 @@
...
+    if (resource.startsWith('arn:')) {
+      resource = bucket.split(':').slice(5).join(':')
+    }
+  }
+  return resource
+}
+
+/**
+ * We do alias them to a local type
+ * @typedef {import('@aws-sdk/types').InitializeMiddleware} InitalizeMiddleware
```
trentm, line 72: typo: s/InitalizeMiddleware/InitializeMiddleware/

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -0,0 +1,207 @@
...
+ * @typedef {import('@aws-sdk/types').FinalizeRequestHandlerOptions } FinalizeRequestHandlerOptions
+ *
+ * Then create our types
+ * @typedef {InitalizeMiddleware | FinalizeRequestMiddleware} S3Middleware
+ * @typedef {InitializeHandlerOptions | FinalizeRequestHandlerOptions} S3MiddlewareOptions
+
+ * @typedef {object} S3MiddlewareEntry
+ * @property {S3Middleware} middleware
+ * @property {import('@aws-sdk/types').HandlerOptions} options
+ */
```
trentm, line 84: This is a lot of content. ;)  Do you find it provides some useful completion in your editor?

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
trentm, line 1: Interesting, I got this error when running this with my current AWS config:

```
$ node ./test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js | ecslog
...
[2023-04-26T21:35:04.728Z]  INFO (elastic-apm-node): createBucket
    event.module: "app"
    bucketName: "elasticapmtest-bucket-20230426213504"
    data: null
    error: {
        "type": "S3ServiceException",
        "message":
            The async () => {
                        if (typeof region === "string") {
                            return (0, getRealRegion_1.getRealRegion)(region);
                        }
                        const providedRegion = await region();
                        return (0, getRealRegion_1.getRealRegion)(providedRegion);
                    } location constraint is incompatible for the region specific endpoint this request was sent to.,
        "stack_trace":
            IllegalLocationConstraintException: The async () => {
                        if (typeof region === "string") {
                            return (0, getRealRegion_1.getRealRegion)(region);
                        }
                        const providedRegion = await region();
                        return (0, getRealRegion_1.getRealRegion)(providedRegion);
                    } location constraint is incompatible for the region specific endpoint this request was sent to.
                at throwDefaultError (/Users/trentm/el/apm-agent-nodejs11/node_modules/@aws-sdk/smithy-client/dist-cjs/default-error-handler.js:8:22)
                at /Users/trentm/el/apm-agent-nodejs11/node_modules/@aws-sdk/smithy-client/dist-cjs/default-error-handler.js:18:39
                at de_CreateBucketCommandError (/Users/trentm/el/apm-agent-nodejs11/node_modules/@aws-sdk/client-s3/dist-cjs/protocols/Aws_restXml.js:3162:20)
                at processTicksAndRejections (node:internal/process/task_queues:96:5)
    }
```

I get this even when the APM agent isn't involved:

```
ELASTIC_APM_ACTIVE=false node ./test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js | ecslog
```

Do you know what might be causing this?

trentm (commented):
This is a great start.

Eventually, another thing to add to the v3 instrumentation is `span.setHttpContext(...)` as is done in the v2 instrumentation for S3, which adds this to the span:

```
    {
        "span": {
            // ...
            "context": {
                "http": {
                    "status_code": 200,
                    "response": {
                        "encoded_body_size": 11500
                    }
                }
            },
        }
    }
```

lib/instrumentation/index.js
```diff
@@ -31,6 +31,8 @@ const nodeSupportsAsyncLocalStorage = semver.satisfies(process.versions.node, '>
 const nodeHasInstrumentableFetch = typeof (global.fetch) === 'function'
 
 var MODULES = [
+  // Instrument the base package used by all AWS-SDK v3 clients
+  ['aws-sdk-v3', '@aws-sdk/smithy-client'],
```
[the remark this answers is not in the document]
david-luna: done!

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
```diff
@@ -0,0 +1,337 @@
...
+
+  vasync.pipeline({
+    arg: {},
+    funcs: [
+      // https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/listbucketscommand.html
+      // Limitation: this doesn't handle paging.
+      function listAllBuckets (arg, next) {
+        const command = new ListBucketsCommand({})
+
+        toCallback(s3Client.send(command), function (err, data) {
```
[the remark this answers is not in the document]
david-luna: oh, I probably did not scrolled enough 😅 . Changed!

lib/instrumentation/modules/aws-sdk-v3.js
```diff
@@ -0,0 +1,43 @@
...
+
+const middlewareFactories = {
+  S3Client: createMiddlewaresForS3Client
+}
+
+module.exports = function (mod, agent, { version, enabled }) {
+  if (!enabled) return mod
+
+  // We have to do this check since npm view reveals RC versions below 3
+  if (!semver.satisfies(version, '>=3')) {
```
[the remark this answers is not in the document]
david-luna: It makes a lot of sense. Done!

lib/instrumentation/modules/aws-sdk-v3.js
```diff
@@ -0,0 +1,43 @@
...
+    agent.logger.debug('@aws-sdk/smithy-client version %s not supported - aborting...', version)
```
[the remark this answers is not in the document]
david-luna: "aborting..." may be misunderstood and I also think the suggested message is more accurate.

Do not worry adding to many "nit" for now. I'm sure in the future there will be less of it ;)

lib/instrumentation/modules/aws-sdk-v3.js
```diff
@@ -0,0 +1,43 @@
...
+  shimmer.wrap(mod.Client.prototype, 'send', function (orig) {
+    return function _wrappedSmithyClientSend () {
+      if (!this[elasticAPMMiddlewares]) {
+        const clientName = this.constructor && this.constructor.name
+        const factory = middlewareFactories[clientName]
+        const middlewares = typeof factory === 'function' ? factory(this, agent) : []
+
+        this[elasticAPMMiddlewares] = middlewares
+        for (const item of this[elasticAPMMiddlewares]) {
+          this.middlewareStack.add(item.middleware, item.options)
```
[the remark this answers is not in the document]
david-luna: Sure!

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
david-luna: Indeed it was. I'm going to have a look at that PR to understand better. Thanks!

lib/instrumentation/modules/@aws-sdk/client-s3.js
david-luna: I think this is a leftover. It doesn't fail but I see no point on `await`ing a value to just return it. Next middleware will `then` on the returned promise.

Indeed the `await` on the prev middleware is necessary because we need the resolved result to extract info for the span.

lib/instrumentation/modules/@aws-sdk/client-s3.js
david-luna: The intent was to have this conversation. From this small experience I can tell
- it seems you cannot import several types at once (but haven't check all docs yet)
- autocompletion is good and worked out of the box
- makes more sense in types that will be exported, this was used internally

In that specific scenario I think it makes more sense to:
- move types to smithy-client (since they are generic to all clients)
- import and use in client-s3 
- future AWS instrumentations like sns & dynamodb will profit from them

I'm fine dropping these lines but I think in object we have more intensive use (Agent, Span, Transaction, ...) we might notice the benefits

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
david-luna: Not yet. I'm going to run my own tests too and post results here

lib/instrumentation/modules/@aws-sdk/client-s3.js
david-luna: I did move it to the smithy client so we can experiment on next clients (sqs, sns, dynamodb) and see if it' suseful

lib/instrumentation/modules/@aws-sdk/client-s3.js
david-luna: fixed!

david-luna:
> This is a great start.
> 
> Eventually, another thing to add to the v3 instrumentation is `span.setHttpContext(...)` as is done in the v2 instrumentation for S3, which adds this to the span:
> 
> ```
>     {
>         "span": {
>             // ...
>             "context": {
>                 "http": {
>                     "status_code": 200,
>                     "response": {
>                         "encoded_body_size": 11500
>                     }
>                 }
>             },
>         }
>     }
> ```

For the `encoded_body_size` property I've took it from the response header named `content-type`. Would that be ok?

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
```diff
@@ -0,0 +1,238 @@
...
+    function () {
+      tx.end()
+      process.exitCode = 0
+    },
+    function () {
+      tx.setOutcome('failure')
+      tx.end()
+      process.exitCode = 1
+    }
+  )
```
trentm, line 235: This script was hanging at the end ... which was a timeout on the HTTP client that `s3Client` is using (at least I strongly suspect so). I fixed that with:

```diff
@@ -225,11 +225,13 @@ function main () {
   useS3v3(s3Client, bucketName).then(
     function () {
       tx.end()
+      s3Client.destroy()
       process.exitCode = 0
     },
     function () {
       tx.setOutcome('failure')
       tx.end()
+      s3Client.destroy()
       process.exitCode = 1
     }
   )
```

test/instrumentation/modules/@aws-sdk/s3-v3.test.js
trentm, line 1: nit: maybe call this file "client-s3.test.js"?

test/instrumentation/modules/@aws-sdk/s3-v3.test.js
```diff
@@ -0,0 +1,349 @@
...
+// Execute 'node fixtures/use-s3js' and assert APM server gets the expected
+// spans.
+tape.test('simple S3 V3 usage scenario', function (t) {
+  const server = new MockAPMServer()
+  server.start(function (serverUrl) {
+    const additionalEnv = {
+      ELASTIC_APM_SERVER_URL: serverUrl,
+      AWS_ACCESS_KEY_ID: 'fake',
+      AWS_SECRET_ACCESS_KEY: 'fake',
+      TEST_BUCKET_NAME: 'elasticapmtest-bucket-1',
```
trentm, line 43:
```suggestion
      TEST_BUCKET_NAME: 'elasticapmtest-bucket-3',
```

Let's avoid re-using the same bucket name from s3.test.js in case we ever get into running test files in parallel -- which I hope we do at some point.

test/instrumentation/modules/@aws-sdk/s3-v3.test.js
```diff
@@ -0,0 +1,349 @@
...
+        env: Object.assign({}, process.env, additionalEnv)
+      },
+      function done (err, stdout, stderr) {
+        console.timeLog && console.timeLog('exec use-s3-v3')
+        t.error(err, 'use-s3-v3.js did not error out')
+        if (err) {
+          t.comment('err: ' + util.inspect(err))
+        }
+        t.comment(`use-s3-v3.js stdout:\n${stdout}\n`)
+        t.comment(`use-s3-v3<.js stderr:\n${stderr}\n`)
```
trentm, line 65:
```suggestion
        t.comment(`use-s3-v3.js stderr:\n${stderr}\n`)
```

test/instrumentation/modules/@aws-sdk/s3-v3.test.js
```diff
@@ -0,0 +1,349 @@
...
+          },
+          otel: {
+            attributes: {
+              'aws.s3.bucket': 'elasticapmtest-bucket-1',
+              'aws.s3.key': 'aDir/aFile.txt-does-not-exist'
+            }
+          },
+          outcome: 'failure'
+        }, 'getObjNonExistantObject produced expected span')
+        console.log(errors)
```
trentm, line 291:
```suggestion
```

test/instrumentation/modules/@aws-sdk/s3-v3.test.js
```diff
@@ -0,0 +1,349 @@
...
+ * compliance with the BSD 2-Clause License.
+ */
+
+'use strict'
+
+if (process.env.GITHUB_ACTIONS === 'true' && process.platform === 'win32') {
+  console.log('# SKIP: GH Actions do not support docker services on Windows')
+  process.exit(0)
+}
+
```
trentm, line 13:
```suggestion
if (process.env.ELASTIC_APM_CONTEXT_MANAGER === 'patch') {
  console.log('# SKIP @aws-sdk/* instrumentation does not work with contextManager="patch"')
  process.exit()
}
```

Because `async/await` is involved in the instrumentation there is no way it can work with the (now deprecated) `contextManager: "patch"` config option (see https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#context-manager for some details).  We can just skip the tests for this case.   This is why the CI test jobs with ", patch" in the name are failing.

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
trentm, line 1: nit: Could perhaps change this name to "use-client-s3.js". Then all the relevant files have "client-s3" in the name.

trentm (changes requested):
This is great.

Another couple things:

- You should add a `@aws-sdk/client-s3` line to the supported modules table here: https://www.elastic.co/guide/en/apm/agent/nodejs/current/supported-technologies.html#compatibility-tracing-and-instrumentation  (that is in "docs/supported-technologies.asciidoc").
- You should also add a note to the CHANGELOG.asciidoc.

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -0,0 +1,199 @@
...
+
+          // Set the httpContext
+          if (statusCode) {
+            const httpContext = {
+              status_code: statusCode
+            }
+
+            if (response && response.headers) {
+              const encodedBodySize = Number(response.headers['content-length'])
+              if (encodedBodySize) {
```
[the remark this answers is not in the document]
trentm: FWIW, here is how another instrumentation is doing the same: https://github.com/elastic/apm-agent-nodejs/blob/0a0e50a2f07f8938375a1a98a0f844f78ed6922a/lib/instrumentation/modules/%40elastic/elasticsearch.js#L193-L198

Useful differences from that handling is:  it is explicit about NaN handling, and if there is an actual `content-length: 0` then we'll include a possibly useful `encoded_body_size: 0` value.

trentm:
This will avoid the failures in CI tests with node less than v14 (the min supported node version for `@aws-sdk/*`):

```diff
diff --git a/test/instrumentation/modules/@aws-sdk/s3-v3.test.js b/test/instrumentation/modules/@aws-sdk/s3-v3.test.js
index ecfe0dc3..453502f2 100644
--- a/test/instrumentation/modules/@aws-sdk/s3-v3.test.js
+++ b/test/instrumentation/modules/@aws-sdk/s3-v3.test.js
@@ -6,10 +6,15 @@

 'use strict'

+const semver = require('semver')
 if (process.env.GITHUB_ACTIONS === 'true' && process.platform === 'win32') {
   console.log('# SKIP: GH Actions do not support docker services on Windows')
   process.exit(0)
 }
+if (semver.lt(process.version, '14.0.0')) {
+  console.log(`# SKIP @aws-sdk min supported node is v14 (node ${process.version})`)
+  process.exit()
+}

 // Test S3 instrumentation of the '@aws-sdk/client-s3' module.
 //
```

lib/instrumentation/modules/@aws-sdk/client-s3.js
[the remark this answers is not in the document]
david-luna: I have to admit that I'll probably assumed too much. Indeed there is an operation doing a `HEAD` request (if I recall properly the waiter command) which returns no `content-length` header.

So accessing that property we get `undefined` and the `encondedBodySize` becomes `NaN` which is a falsy value. I see no harm in being more explicit but I think we can omit the `isNaN` check.

test/instrumentation/modules/@aws-sdk/s3-v3.test.js
david-luna: Thanks for this pearl of wisdom. I need to do a deep dive on this topic :)

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -0,0 +1,199 @@
...
+              }
+            }
+            span.setHttpContext(httpContext)
+          }
+
+          // Configuring `new S3Client({useArnRegion:true})` allows one to
+          // use an Access Point bucket ARN for a region *other* than the
+          // one for which the client is configured. Therefore, we attempt
+          // to get the bucket region from the ARN first.
+          const useArnRegion = await client.config.useArnRegion()
```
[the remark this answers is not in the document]
david-luna: the config properties `useArnRegion` and `region` are defined in `ClientInputEndpointParamenters` interface. [See here](https://github.com/aws/aws-sdk-js-v3/blob/b1433c46aa476cd5574fb6d2942fb554f0451dbf/clients/client-s3/src/endpoint/EndpointParameters.ts#LL4C10-L4C10)

We can see the properties are defined as
- `useArnRegion` can be a boolean or a `Provider<boolean>`
- `region` can be a string or a `Provider<string>`

The generic type `Provider` defines a function that returns a promise of the given type so it can be awaited. 

The client configuration in the test suite always have these properties set to providers (default ones) but they can be defined at construction time passing a different [configuration object](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/s3clientconfig.html)

We need to check if those props are values or functions and act accordingly. Thanks for the heads up :)

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -135,10 +135,20 @@ function s3MiddlewareFactory (client, agent) {
...
-          const region = useArnRegion && args.input.Bucket.startsWith('arn:')
-            ? regionFromS3Arn(args.input.Bucket)
-            : await client.config.region()
+          const config = client.config
+          let useArnRegion
+          if (typeof config.useArnRegion === 'boolean') {
+            useArnRegion = config.useArnRegion
+          } else {
+            useArnRegion = await config.useArnRegion()
+          }
```
trentm, line 144: Have you found this is necessary? I really hope not.
Did that change in some older `@aws-sdk/client-s3` version perhaps?

package-lock.json
trentm, line 1: This PR shouldn't have most (any?) of these package-lock diffs. I think some of these will be resolved by merging from recent "main" changes.

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
david-luna: I've added it. Thanks

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
```diff
@@ -0,0 +1,238 @@
...
+//    # and deleting it afterwards.
+//    node use-s3-v3.js | ecslog
+//
+//    # Testing against localstack.
+//    docker run --rm -it -e SERVICES=s3 -p 4566:4566 localstack/localstack
+//    TEST_ENDPOINT=http://localhost:4566 node use-s3-v3.js | ecslog
+//
+//    # Use TEST_BUCKET_NAME to re-use an existing bucket (and not delete it).
+//    # For safety the bucket name must start with "elasticapmtest-bucket-".
+//    TEST_BUCKET_NAME=elasticapmtest-bucket-1 node use-s3.js | ecslog
```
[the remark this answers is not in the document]
david-luna: there were other places too. Changed!

lib/instrumentation/modules/@aws-sdk/smithy-client.js
```diff
@@ -0,0 +1,121 @@
...
+  if (!semver.satisfies(version, '>=3 <4')) {
+    agent.logger.debug('cannot instrument @aws-sdk/client-*: @aws-sdk/smithy-client version %s not supported', version)
+    return mod
+  }
+
+  shimmer.wrap(mod.Client.prototype, 'send', function (orig) {
+    return function _wrappedSmithyClientSend () {
+      const clientName = this.constructor && this.constructor.name
+      const clientConfig = clientsConfig[clientName]
+
```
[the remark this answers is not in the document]
david-luna: yeah, it looks cleaner. Changed!

test/instrumentation/modules/@aws-sdk/fixtures/use-s3-v3.js
```diff
@@ -0,0 +1,238 @@
...
+  waitUntilObjectExists
+} = require('@aws-sdk/client-s3')
+
+const TEST_BUCKET_NAME_PREFIX = 'elasticapmtest-bucket-'
+
+// ---- support functions
+
+// https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/index.html
+async function useS3v3 (s3Client, bucketName) {
+  const region = s3Client.config.region
```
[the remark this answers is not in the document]
david-luna: I've confirmed that although the `region` is passed as string in the constructor call the config property is transformed to an async function. This does not happen with the `useArnRegion` configuration property.

I've applied your change but since [the documentation](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/s3clientconfig.html) says we may get a plain string for `region` I think is best to keep the type checking in the instrumentation.

trentm (approved):

package-lock.json
david-luna: The PR is adding `@aws-sdk/client-s3` as a dev dependency and this is its entry in the lock file

```
    "node_modules/@aws-sdk/client-s3": {
      "version": "3.325.0",
      "resolved": "https://registry.npmjs.org/@aws-sdk/client-s3/-/client-s3-3.325.0.tgz",
      "integrity": "sha512-S88q84bJ1Wba3jnpIrWxMD4AbDsG7l45Ey4ZDdyriEceEgIWvBCA90ZOgSdaObhFNa5pNFZG9C6UbiaCz3F3bA==",
      "dev": true,
      "dependencies": {
        "@aws-crypto/sha1-browser": "3.0.0",
        "@aws-crypto/sha256-browser": "3.0.0",
        "@aws-crypto/sha256-js": "3.0.0",
        "@aws-sdk/client-sts": "3.325.0",
        "@aws-sdk/config-resolver": "3.310.0",
        "@aws-sdk/credential-provider-node": "3.325.0",
        "@aws-sdk/eventstream-serde-browser": "3.310.0",
        "@aws-sdk/eventstream-serde-config-resolver": "3.310.0",
        "@aws-sdk/eventstream-serde-node": "3.310.0",
        "@aws-sdk/fetch-http-handler": "3.310.0",
        "@aws-sdk/hash-blob-browser": "3.310.0",
        "@aws-sdk/hash-node": "3.310.0",
        "@aws-sdk/hash-stream-node": "3.310.0",
        "@aws-sdk/invalid-dependency": "3.310.0",
        "@aws-sdk/md5-js": "3.310.0",
        "@aws-sdk/middleware-bucket-endpoint": "3.310.0",
        "@aws-sdk/middleware-content-length": "3.325.0",
        "@aws-sdk/middleware-endpoint": "3.325.0",
        "@aws-sdk/middleware-expect-continue": "3.325.0",
        "@aws-sdk/middleware-flexible-checksums": "3.325.0",
        "@aws-sdk/middleware-host-header": "3.325.0",
        "@aws-sdk/middleware-location-constraint": "3.325.0",
        "@aws-sdk/middleware-logger": "3.325.0",
        "@aws-sdk/middleware-recursion-detection": "3.325.0",
        "@aws-sdk/middleware-retry": "3.325.0",
        "@aws-sdk/middleware-sdk-s3": "3.325.0",
        "@aws-sdk/middleware-serde": "3.325.0",
        "@aws-sdk/middleware-signing": "3.325.0",
        "@aws-sdk/middleware-ssec": "3.325.0",
        "@aws-sdk/middleware-stack": "3.325.0",
        "@aws-sdk/middleware-user-agent": "3.325.0",
        "@aws-sdk/node-config-provider": "3.310.0",
        "@aws-sdk/node-http-handler": "3.321.1",
        "@aws-sdk/protocol-http": "3.310.0",
        "@aws-sdk/signature-v4-multi-region": "3.310.0",
        "@aws-sdk/smithy-client": "3.325.0",
        "@aws-sdk/types": "3.310.0",
        "@aws-sdk/url-parser": "3.310.0",
        "@aws-sdk/util-base64": "3.310.0",
        "@aws-sdk/util-body-length-browser": "3.310.0",
        "@aws-sdk/util-body-length-node": "3.310.0",
        "@aws-sdk/util-defaults-mode-browser": "3.325.0",
        "@aws-sdk/util-defaults-mode-node": "3.325.0",
        "@aws-sdk/util-endpoints": "3.319.0",
        "@aws-sdk/util-retry": "3.310.0",
        "@aws-sdk/util-stream-browser": "3.310.0",
        "@aws-sdk/util-stream-node": "3.321.1",
        "@aws-sdk/util-user-agent-browser": "3.310.0",
        "@aws-sdk/util-user-agent-node": "3.310.0",
        "@aws-sdk/util-utf8": "3.310.0",
        "@aws-sdk/util-waiter": "3.310.0",
        "@aws-sdk/xml-builder": "3.310.0",
        "fast-xml-parser": "4.1.2",
        "tslib": "^2.5.0"
      },
      "engines": {
        "node": ">=14.0.0"
      }
    },
```

There's a lot of dependencies and these packages also have their dependencies as well. I guess for SNS, SQS & DynamoDB it will be less impact on the lock file.

PR #3288, opened 2023-04-26
fix: do not log.warn when passing null args to span.setServiceTarget(...)

trentm:
This method allows null arguments to indicate "no value":

    setServiceTarget(type?: string | null, name?: string | null): void;

Before this change it would warn if null or undefined was passed, e.g.:

    WARN (elastic-apm-node): "name" argument to Span#setServiceTarget must be of type "string", got type "object": ignoring

### DIFF
--- a/lib/instrumentation/span.js
+++ b/lib/instrumentation/span.js
@@ -290,7 +290,7 @@ Span.prototype.setServiceTarget = function (type, name) {
     } else {
       this._serviceTarget.type = type
     }
-  } else {
+  } else if (type != null) {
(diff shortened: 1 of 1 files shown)


trentm:
I noticed this when running `use-s3.js`. It would emit this log.warn for the ListBuckets call:

```
{"log.level":"warn","@timestamp":"2023-04-26T15:39:20.916Z","log":{"logger":"elastic-apm-node"},"ecs":{"version":"1.6.0"},"message":"\"name\" argument to Span#setServiceTarget must be of type \"string\", got type \"object\": ignoring"}
```

david-luna (approved):

Issue #3286, opened 2023-04-26
TypeError - `req` is required to have `headers` property


Alastair-Spencer:
**Describe the bug**

Since **3.42.0** and **3.44.0** a change has been made which now throws lambda runtimes with TypeErrors due to headers being missing utilised from a low level package **basic-auth** - https://github.com/jshttp/basic-auth/blob/v1.1.0/index.js#L80-L82. Areas around this package haven't been changed for a while so the cause of the issue is not quite clear yet although i'm trawling through changes in https://github.com/elastic/apm-agent-nodejs/compare/v3.42.0...v3.44.1 to try and help diagnose the cause.

Stack trace:
```json
{
    "errorType": "TypeError",
    "errorMessage": "argument req is required to have headers property",
    "stack": [
        "TypeError: argument req is required to have headers property",
        "    at getAuthorization (/var/task/node_modules/basic-auth/index.js:88:11)",
        "    at auth (/var/task/node_modules/basic-auth/index.js:66:16)",
        "    at Object.getUserContextFromRequest (/var/task/node_modules/elastic-apm-node/lib/parsers.js:113:26)",
        "    at Immediate.<anonymous> (/var/task/node_modules/elastic-apm-node/lib/agent.js:525:24)",
        "    at processImmediate (node:internal/timers:466:21)",
        "    at process.callbackTrampoline (node:internal/async_hooks:130:17)"
    ]
}
```

**To Reproduce**

1. Use this config:

Standard configuration applies from here on as the lambda is wrapped within the lambda handler and utilises environment configuration.

```
  ELASTIC_APM_SECRET_TOKEN: ${env:ELASTIC_APM_SECRET_TOKEN, ""}
  ELASTIC_APM_LOG_LEVEL: ${env:ELASTIC_APM_LOG_LEVEL, "info"}
  ELASTIC_APM_SERVER_URL: ${env:ELASTIC_APM_SERVER_URL, "http://localhost:8200"}
  ELASTIC_APM_LAMBDA_APM_SERVER: ${env:ELASTIC_APM_LAMBDA_APM_SERVER, ""}
  ELASTIC_APM_SERVICE_NAME: ${env:ELASTIC_APM_SERVICE_NAME, "${self:service}"}
  ELASTIC_APM_VERIFY_SERVER_CERT: true
  ELASTIC_APM_DATA_RECEIVER_TIMEOUT_SECONDS: 15
```

2. Deploy and experience type errors from ALB traffic

[...]

PR #3289, opened 2023-04-27
fix: crash in Lambda handler with ELB or API Gateway event with no 'headers'

trentm:
If the lambda handler object for an ELB- or API Gateway-triggered Lambda
invocation did not have a 'headers' field, then the Lambda
instrumentation would crash. I'm not sure how to get an event with no
'headers' field, but we should still be defensive here.

Fixes: https://github.com/elastic/apm-agent-nodejs/issues/3286


david-luna (approved):

trentm:
Thanks for the review!

PR #3293, opened 2023-04-28
chore, test: test driver improvements

trentm:
This fixes a couple issues:
- The previous test.js code to find .test.js files to test could accidentally include .test.js files from lazily-created `node_modules` trees in some test packages. This could lead to a failure on a *second* run of the tests locally.
- avoid a 30s hang at the end of central-config-disabled.test.js

Also:
- Remove unneeded cruft from the driver and a few test files.
- Get the test/start/... test files to be able to run independently (without needing custom setup from test.js)
- Modernize the TESTING.md doc, and some updates to the DEVELOPMENT.md dev notes doc.
- Drop 'npm run test:docs', it is broken and the fix is to just use 'npm run build:docs'


TESTING.md
trentm, line 1: Reviewer note: This is basically a re-write of most of TESTING.md.

test/central-config-disabled.test.js
```diff
@@ -24,18 +25,15 @@ test('central config disabled', function (t) {
...
-      t.end()
       agent.destroy()
       server.close()
+      t.end()
     }, 1000)
   })
-
-  t.on('end', function () {
-    if (existingValue) process.env.ELASTIC_APM_CENTRAL_CONFIG = existingValue
-  })
```
trentm, line 40: Reviewer note: Since the time this test file was written, test files are now always run in a separate process. That means there is no point in restoring envvars for the next test file.

test/script/run_tests.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-set -eo pipefail # abort if any of the commands exit badly
+if [ "$TRACE" != "" ]; then
+    export PS4='${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+    set -o xtrace
+fi
+set -o errexit
+set -o pipefail
```
trentm, line 8: Reviewer note: This is bash boilerplate that I typically use in bash files (including some already in this repo). It adds support for `TRACE=1 .../run_tests.sh` to get xtrace log output.

david-luna (approved):

PR #3294, opened 2023-04-28
docs: correct version in which partial Lambda transactions will be released

trentm:
Refs: #3285


trentm:
In #3285 I misstated what APM agent version would have the partial Lambda transactions support.

PR #3308, opened 2023-05-02
chore: fix bitrot.js dev util for recent changes

trentm:
- There are a couple more .tav.yml files for bitrot.js to include in its processing now.
- Also fix a regex to avoid this error:

```
% ./dev-utils/bitrot.js
bitrot: error: could not parse this table cell text from docs/supported-technologies.asciidoc: " https://www.npmjs.com/package/@opentelemetry/sdk-metrics[@opentelemetry/sdk-metrics]"
```

### DIFF
--- a/dev-utils/bitrot.js
+++ b/dev-utils/bitrot.js
@@ -205,7 +205,7 @@ function loadSupportedDoc () {
         results.push({ name: n, versions: row[1] })
       })
     } else {
-      match = /^https:\/\/.*\[(.*)\]$/.exec(row[0])
+      match = /^https:\/\/.*\[(.*)\]$/.exec(row[0].trim())
       if (!match) {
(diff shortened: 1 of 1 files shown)


dev-utils/bitrot.js
```diff
@@ -205,7 +205,7 @@ function loadSupportedDoc () {
         results.push({ name: n, versions: row[1] })
       })
     } else {
-      match = /^https:\/\/.*\[(.*)\]$/.exec(row[0])
+      match = /^https:\/\/.*\[(.*)\]$/.exec(row[0].trim())
```
david-luna, line 208: Looks good. Out of curiosity: is there any reason we want to keep starting/trailing spaces in cell values?

david-luna (approved):

dev-utils/bitrot.js
trentm: Nope, there is no reason to keep surrounding spaces in cell values.

PR #3309, opened 2023-05-02
chore: update nvm

trentm:
We use nvm to install specific versions of node for "Edge tests"
and benchmarks. v0.34.0 was from 2019.


trentm:
One way to exercise this is:

```
.ci/scripts/test.sh -b nightly 21
```

The NVM parts of this work. I've hit some other unrelated failure that I'll look into separately.

PR #3315, opened 2023-05-03
test: drop node 14 from RC tests now that it is EOL

trentm:
Also add v21 because at some point there may be an RC
for a v21.0.0 release.


trentm:
Current https://github.com/nodejs/release#release-schedule.  Node v14 is now EOL.

Screenshot 2023-05-03 at 2 50 20 PM

david-luna (approved):

Issue #3273, opened 2023-04-17
Set-Cookie header is redacted, but Cookie header is not


dgieselaar:
From the [spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-http.md#http-transactions):

> Captured request and response headers, cookies, and form bodies MUST be sanitised (i.e. secrets removed) according to data sanitization rules.

However, this only happens for the `Set-Cookie` header, not the `Cookie` header, which seems to be a violation of the spec.

david-luna:
Hi @dgieselaar 

Thanks for filling the issue. I've checked how the filter works adding this small change on the test file for filters
```
diff --git a/examples/trace-http.js b/examples/trace-http.js
index f9fbb524..ee7e9dab 100755
--- a/examples/trace-http.js
+++ b/examples/trace-http.js
@@ -43,7 +43,8 @@ const server = http.createServer(function onRequest (req, res) {
         res.writeHead(200, {
           server: 'example-trace-http',
           'content-type': 'text/plain',
-          'content-length': Buffer.byteLength(resBody)
+          'content-length': Buffer.byteLength(resBody),
+          'set-cookie': [ 'password=test', 'foo=bar']
         })
         res.end(resBody)
       })
@@ -56,7 +57,8 @@ server.listen(3000, function () {
   //
   // Note that this there is no current "transaction" here, so this HTTP
   // request is not captured by APM. See "trace-http-request.js" for more.
-  const clientReq = http.request('http://localhost:3000/', function (clientRes) {
+  const options = { headers: { accept: '*/*', cookie: 'password=test;foo=bar' } }
+  const clientReq = http.request('http://localhost:3000/', options, function (clientRes) {
     console.log('client response: %s %s', clientRes.statusCode, clientRes.headers)
```

Running `./examples/trace-http.js` example we get a transaction with this shape

```
{
    "transaction": {
        "name": "GET /",
        "type": "request",
        "result": "HTTP 2xx",
        "id": "d5233c52c314d102",
        "trace_id": "d93fb3d7a8b54b9e862851473bc309c8",
        "duration": 17922.716,
        "timestamp": 1683125345650013,
        "sampled": true,
        "context": {
            "user": {},
            "tags": {},
            "custom": {},
            "service": {},
            "cloud": {},
            "message": {},
            "request": {
                "http_version": "1.1",
                "method": "GET",
                "url": {
                    "raw": "/",
                    "protocol": "http:",
[...]

PR #3322, opened 2023-05-05
Send cookies explicitly into `context.request`'s transaction

david-luna:
# Description
This PR changes the transaction serialisation to be more explicit about sanitization of sensitive data in request's `cookie` header.

**Current status**

cookies are sanitized in the header value based on [sanitizeFieldNames](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#sanitize-field-names) configuration. As a result you see clear values and redacted ones. This may lead to think the agent is not properly doing sanitization.

**After this PR**

cookies are going to be placed in `context.request.cookies` map, which is accepted by the [intake API](https://github.com/elastic/apm-server/blob/d18c559e4136b2213a4f6e206258c5d303ec1b0f/docs/spec/v2/transaction.json#L129-L135), and sanitized based on the configuration. The cookie header will be completely redacted to avoid duplication of data.

Fixes: #3273 
### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


package-lock.json
```diff
@@ -21594,6 +21612,29 @@
       "integrity": "sha512-OPs5WnnT1xkCBiuQrZA4+YAV4HEJejmHneyraIaxsbev5yCEr6KMwINNFP9wQeFIw8FWcoTqF3vQsa5CDaI+8Q==",
       "dev": true
     },
+    "@types/node-fetch": {
```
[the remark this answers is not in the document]
trentm: This should be from https://github.com/elastic/apm-agent-nodejs/pull/3318/files

CHANGELOG.asciidoc
```diff
@@ -68,6 +68,8 @@ Notes:
 [float]
 ===== Chores
 
+* Add cookie map in transactions' request context and redact cookie header ({pull}3304[#3322])
```
trentm, line 71:
```suggestion
* Add cookie map in transactions' request context and redact cookie header ({pull}3322[#3322])
```

lib/parsers.js
```diff
@@ -58,6 +60,15 @@ function getContextFromRequest (req, conf, type) {
       Object.assign({}, req.headers),
       conf.sanitizeFieldNamesRegExp
     )
+
+    // TODO: deprecate filterHttpHeaders in next major release
```
trentm, line 64: We might be disagreeing on the definition of "deprecate". For me, "deprecating" means that we would document (and mark in types) that the config var is "deprecated" -- hence users should avoid it and the docs should show users how to migrate away from it, but the field is still supported.

The "TODO" for the next major release would be to *remove* the `filterHttpHeaders` config var.
Thoughts?

If we agree, then would you like to do the deprecation tweaks in this PR, or prefer that we do it in a separate change?

trentm (approved):
One type on the changelog, and a question about the deprecation below. Otherwise this looks good.

lib/parsers.js
david-luna: I think your approach is much better. I shall do the changes in this PR.

Thanks!

PR #3333, opened 2023-05-08
chore: mark filterHttpHeaders config option as deprecated

### LINKED ISSUE #3320: Deprecate filterHttpHeaders configuration
Once #3273 is resolved [filterHttpHeaders config](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#filter-http-headers) will become redundant since the agent will be capable of filtering/redacting headers (and cookies as well) by the proper set up of [captureHeaders](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#capture-headers) and [sanitizeFieldNames](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#sanitize-field-names).

We can proceed to deprecate `filterHttpHeaders`. Since this is considered a breaking change we should plan it for the next major release.

david-luna:
# Description

Mark the `filterHttpHeaders config option as deprecated:

- Update documentation accordingly
- Add a warning log if the option is used when starting the agent

Closes #3320 

### Checklist

- [ ] Implement code
- [ ] Add tests
- [x] Update documentation


docs/configuration.asciidoc
```diff
@@ -972,15 +972,21 @@ The `sanitizeFieldNames` configuration value allows you to configure a list of w
 case-insensitive by default. You may make wildcard searches case-sensitive by
 using the `(?-i)` prefix. These patterns apply to the request and response HTTP headers, as well as any form field captured during an `application/x-www-form-urlencoded` data request.
 
-The `sanitizeFieldNames` will redact any matched _field names_.  If you wish to filter or _redact_ other data the <<filter-http-headers,`filterHttpHeaders`>> configuration field or the <<apm-add-filter,API filtering functions>> may be a better choice.
+The `sanitizeFieldNames` will redact any matched _field names_.  If you wish to filter or _redact_ other data the <<apm-add-filter,API filtering functions>> may be a better choice.
 
 [[filter-http-headers]]
 ==== `filterHttpHeaders`
 
+[small]#Deprecated in: v3.x.x
```
david-luna, line 980: not sure what's going to be the exact version. Will update when sure

docs/configuration.asciidoc
```diff
@@ -972,31 +972,19 @@ The `sanitizeFieldNames` configuration value allows you to configure a list of w
...
-Currently, the following HTTP headers are anonymized by default:
-
-* `Authorization` - The full value of this header is redacted.  In versions of the agent greater than v3.9, the authorization header is redacted by the default configuration of <<sanitize-field-names>>.
-* `Cookie` - The cookies inside the `Cookie` header are analyzed and their values redacted if they appear sensitive (like a session cookie).
-  See the https://github.com/watson/is-secret[is-secret] module for details about which patterns are considered sensitive.
-
-If you wish to filter or sanitize other data,
-use one of the <<apm-add-filter,filtering>> functions.
+This configuration is *deprecated* and will be removed in a future major version. If you whish to filter or or _redact_ other 
+data the <<apm-add-filter,API filtering functions>> or <<sanitize-field-names, Sanitize fields configuration>> may be a better choice.
```
trentm, line 987: It would be helpful for this description to explain why this is deprecated and what a user using this config var should replace it with.  In this case `filterHttpHeaders: true` can just be dropped because it is the default behaviour, and `filterHttpHeaders: false` can be mostly replicated by changing the values in `sanitizeFieldNames` (with the caveat that the parsed cookie values are now on `transaction.request.cookies` rather than `transaction.request.headers.cookie`).  Do I have that right?

docs/configuration.asciidoc
```diff
@@ -972,31 +972,19 @@ The `sanitizeFieldNames` configuration value allows you to configure a list of w
...
-
-Currently, the following HTTP headers are anonymized by default:
-
-* `Authorization` - The full value of this header is redacted.  In versions of the agent greater than v3.9, the authorization header is redacted by the default configuration of <<sanitize-field-names>>.
-* `Cookie` - The cookies inside the `Cookie` header are analyzed and their values redacted if they appear sensitive (like a session cookie).
-  See the https://github.com/watson/is-secret[is-secret] module for details about which patterns are considered sensitive.
-
-If you wish to filter or sanitize other data,
-use one of the <<apm-add-filter,filtering>> functions.
+This configuration is *deprecated* and will be removed in a future major version. If you whish to filter or or _redact_ other
```
trentm, line 986:
```suggestion
This configuration is *deprecated* and will be removed in a future major version. If you wish to filter or or _redact_ other 
```

CHANGELOG.asciidoc
```diff
@@ -68,6 +68,9 @@ Notes:
 [float]
 ===== Chores
 
+* The config option <<filter-http-headers, filterHttpHeaders>> is now *deprecated*. It will be
+  removed in a future major version (after an 18 month deprecation period). ({pull}3333[#3333])
```
trentm, line 72:
```suggestion
  removed in a future major version. ({pull}3333[#3333])
```

I'm not sure we need to mention an 18 month deprecation period. I expect we'll have a 4.x release before 18 months as well. Did you get the 18 months from some policy doc that I am forgetting?

lib/config.js
```diff
@@ -1125,6 +1139,13 @@ function normalizeKeyValuePairs (opts) {
   }
 }
 
+/**
+ * Normalize and validate the given values for boolean configuration properties
+ *
+ * - the `filterHttpHeaders` config option is now deprecated. The default values of
+ *   `sanitizeFieldNames` already provide the same filtering of senstive information
+ *   within the `cookie` info
```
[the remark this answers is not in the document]
david-luna: You're right. This is actually a leftover from the 1st approach which was inside normalizeBool function. The check in now done in the constructor just before doing the normalisation.

CHANGELOG.asciidoc
david-luna: Just looked a previous entries of this file and found deprecation ones came with that comment so I assumed it was the norm.

lib/config.js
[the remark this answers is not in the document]
david-luna: Changed!

docs/configuration.asciidoc
```diff
@@ -972,31 +972,19 @@ The `sanitizeFieldNames` configuration value allows you to configure a list of w
 case-insensitive by default. You may make wildcard searches case-sensitive by
 using the `(?-i)` prefix. These patterns apply to the request and response HTTP headers, as well as any form field captured during an `application/x-www-form-urlencoded` data request.
```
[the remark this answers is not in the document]
david-luna: I could and I think I should. Thanks for the head up

docs/configuration.asciidoc
david-luna: sanitization of the cookie string was actually using `sanitizeFieldNames` to decide which cookie needs to be redacted. So your assumption is correct. We can provide a value for `sanitizeFieldNames` which will be the equivalent of `filterHttpHeaders: false`

docs/configuration.asciidoc
```diff
@@ -970,33 +970,27 @@ Remove sensitive data sent to Elastic APM.
...
+This configuration is *deprecated* and will be removed in a future major version. If you wish to filter or or _redact_ other 
+data the <<apm-add-filter,API filtering functions>> or <<sanitize-field-names, Sanitize fields configuration>> may be a better choice.
 
-If you wish to filter or sanitize other data,
-use one of the <<apm-add-filter,filtering>> functions.
+In versions before REPLACEME this option was `true` by default. Agent redacted cookies whose name matched any of `sanitizaFieldNames` expressions.
+As of version REPLACEME this option has no longer use since request cookies are parsed and stored into
+`transaction.context.request.cookies` name/value pair. These new fields will be redacted if their name
+match with any of the `sanitizaFieldNames` expressions threfore having the same behaviour. To replicate `filterHttpHeaders = false` you just
+have to modify `sanitizaFieldNames` option to make sure no HTTP request cookie is redacted.
```
trentm, line 993: Alternative suggested language:

```
`filterHttpHeaders` indicates whether the agent should redact sensitive fields in the `Cookie` HTTP request
header. If true, the `Cookie` header will be parsed in to the `transaction.request.cookies` name/value
mapping and any cookie fields that match the patterns in <<sanitize-field-names>> will be replaced with the
string `[REDACTED]`. If false, the `Cookie` header will be left as is.

This config option is *deprecated* and will be removed in a future major version. Use <<sanitize-field-names>>
instead. To replicate `filterHttpHeaders: false`, modify `sanitizeFieldNames` so that no patterns match
cookie fields that you do not want redacted.

See also the <<apm-add-filter,the filtering API>> for programmatic control of filtering APM data.
```

Perhaps this is a bit clearer?

trentm (approved):
I have some wording suggestions.

Issue #3227, opened 2023-03-29
follow-up work for TAV tests now being in GH Actions


trentm:
https://github.com/elastic/apm-agent-nodejs/pull/3127 moved our TAV tests CI runs from Jenkins to GH Actions. There are a few things to follow-up on with this:

- [x] I would very much like a way to be able to *manually* run a subset of the TAV tests on a PR. I *can* manually run one or more `.ci/scripts/test.sh -b "release" -t "${{ module }}" "${{node }}"` on my dev machine. However, it is useful to have check results reflected on a PR. Also for community PRs it is nice to run relevant TAV tests and have the results available via the PR. Before, when using Jenkins, we had the `run module tests for <modules>` GH comment. It doesn't need to be a GitHub comment. See https://github.com/elastic/apm-agent-nodejs/pull/3127#issuecomment-1487192295 for earlier discussion on this.
- [x] Are the individual TAV job checks shown for commits on main?  For example [commit 6660cc3](https://github.com/elastic/apm-agent-nodejs/commit/6660cc3d05eb351947518571c1e7e5078bab0a52) is the commit that added the new workflow, its workflow *did* run last night (https://github.com/elastic/apm-agent-nodejs/actions/runs/4552294471), yet the summary of checks for that commit do *not* include the TAV job runs.
- [x] Some sort of lint checks on tav.yml would be worthwhile:
    - [x] Is the [set of modules](https://github.com/elastic/apm-agent-nodejs/blob/main/.github/workflows/tav.yml#L38-L73) up to date? This should match the full set of `.tav.yml` module/group names for all ".tav.yml" files in the repo (currently there are two).
    - [x] Does number-of-node-versions * number-of-tav-modules exceed the GitHub limit of 256? Currently we are at 245. If we pass the limit, then we'll need to cope somehow. See https://github.com/elastic/apm-agent-nodejs/pull/3127#issuecomment-1487213047
- [x] The "dev-utils/jenkins-build-slow-steps.sh" dev tool should be replaced with one that looks at the GH Action runs. See https://github.com/elastic/apm-agent-nodejs/pull/3127#issuecomment-1487497306 for a start.

/cc @v1v (Victor, I'm not suggesting *you* need to do these things. I'll do some of them. Mostly I would appreciate suggestions/advice on the first item.)

trentm:
> To add more clarity, there are some caveats, for instance, GitHub comments should trigger builds if they are coming from an Elastician user ...

[...]

PR #3340, opened 2023-05-10
chore: add 'npm run lint:tav'

trentm:
This checks that .ci/tav.json is correct: both that it covers all the
modules in all the .tav.yml files in the repo, and that it never
results in a TAV test matrix greater than the GH Actions 256 limit.
This is part of `npm run lint` which is run by the lint CI workflow.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/3227 (item 3)

---

This also drops the body-parser TAV tests: we don't instrument it, just use it in some tests.


dev-utils/lint-tav-json.js
```diff
@@ -0,0 +1,88 @@
...
+  tavJson.modules.forEach(m => {
+    m.split(',').forEach(moduleName => {
+      moduleNamesFromJson.add(moduleName)
+    })
+  })
+  // console.log('moduleNamesFromJson: ', moduleNamesFromJson)
+
+  // Matrix 256 limit.
+  const matrixSize = tavJson.versions.length * tavJson.modules.length
+  if (matrixSize > 256) {
```
v1v, line 61: That's pretty neat! If it helps the fix could be refactoring the existing TAV to run in chunks as done with https://github.com/elastic/apm-agent-python/blob/1e38ec53115edc70c36c6485259733a8cde02ed9/.github/workflows/test.yml#L59-L81

v1v (approved):

dev-utils/lint-tav-json.js
```diff
@@ -0,0 +1,88 @@
...
+  )
+  // console.log('tavYmlPaths:', tavYmlPaths)
+  tavYmlPaths.forEach(p => {
+    const tavCfg = yaml.load(fs.readFileSync(p, 'utf8'))
+    Object.keys(tavCfg).forEach(k => {
+      const v = tavCfg[k]
+      moduleNamesFromYaml.add(v.name || k)
+    })
+  })
+  // console.log('moduleNamesFromYaml: ', moduleNamesFromYaml)
```
david-luna, line 47: nit: do you want to keep these logs?

david-luna (approved):

dev-utils/lint-tav-json.js
trentm: Kind of, yes. :)  If I ever have to come back and debug this script, then those would be the first interesting things I'd want to re-enable. However, granted, this is a simple script and I don't expect we'll need to come back and debug. I don't care that strongly either way.

To get waaay off track here: I sometime wonder at code (perhaps a language could have specific features/syntax for this) having ways of indicating points of interest for the debugging process.

trentm: I guess the next (better?) step would be to add a logging lib, a command-line option and/or envvar to set the debug level, and make these log.debug statements.

PR #3343, opened 2023-05-12
chore: extract normalizers to its own module

david-luna:
# Description

This is the 1st of a short series of PRs with the intent of have better organisation of the code and set the grounds for:
- tackle the issue #3060 
- a possible refactoring in the short/mid term

Normalization logic is quite straightforward and does not depend much on any context. This PR:
- extract the normalization functions into a specific module
- adds unit tests on these extracted methods
- require & use the functions inside `config.js` file

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update ~TypeScript~ typings (internal)
- [ ] Update documentation

### DIFF
--- a/lib/config/normalizers.js
+++ b/lib/config/normalizers.js
@@ -1,0 +1,488 @@
+/*
+ * Copyright Elasticsearch B.V. and other contributors where applicable.
+ * Licensed under the BSD 2-Clause License; you may not use this file except in
+ * compliance with the BSD 2-Clause License.
+ */
+
+'use strict'
+
+const { WildcardMatcher } = require('../wildcard-matcher')
+
+// TODO: move this typedef to logger.js
+/**
+ * @typedef {Object} Logger
+ * @property {function(any, any, any, any): undefined} log
+ * @property {function(any, any, any, any): undefined} info
+ * @property {function(any, any, any, any): undefined} warn
+ * @property {function(any, any, any, any): undefined} error
+ */
+
+/**
+ * Normalizes the key/value pairs properties of the config options object
+ * KeyValuePairs config vars are either an object or a comma-separated string
+ * of key=value pairs (whitespace around the "key=value" strings is trimmed):
+ *    {'foo': 'bar', 'eggs': 'spam'} => [['foo', 'bar'], ['eggs', 'spam']]
+ *    foo=bar, eggs=spam             => [['foo', 'bar'], ['eggs', 'spam']]
+ *
+ * @param {Record<string, unknown>} opts the configuration options to normalize
+ * @param {String[]} fields the list of fields to normalize as key/value pair
+ * @param {Record<string, unknown>} defaults the configuration defaults
+ * @param {Logger} logger
+ */
+function normalizeKeyValuePairs (opts, fields, defaults, logger) {
+  for (const key of fields) {
+    if (key in opts) {
+      if (typeof opts[key] === 'object' && !Array.isArray(opts[key])) {
+        opts[key] = Object.entries(opts[key])
+        return
+      }
+
+      if (!Array.isArray(opts[key]) && typeof opts[key] === 'string') {
+        opts[key] = opts[key].split(',').map(v => v.trim())
+      }
+
+      if (Array.isArray(opts[key])) {
+        // Note: Currently this assumes no '=' in the value. Also this does not
+        // trim whitespace.
+        opts[key] = opts[key].map(
+          value => typeof value === 'string' ? value.split('=') : value)
+      }
+    }
+  }
+}
+
+/**
+ * Normalizes the number properties of the config options object
+ *
+ * @param {Record<string, unknown>} opts the configuration options to normalize
+ * @param {String[]} fields the list of fields to normalize as number
+ * @param {Record<string, unknown>} defaults the configuration defaults
+ * @param {Logger} logger
+ */
+function normalizeNumbers (opts, fields, defaults, logger) {
+  for (const key of fields) {
+    if (key in opts) opts[key] = Number(opts[key])
+  }
+}
+
+/**
+ * Normalizes the number properties of the config options object
+ *
+ * @param {Record<string, unknown>} opts the configuration options to normalize
+ * @param {String[]} fields the list of fields to normalize as number
+ * @param {Record<string, unknown>} defaults the configuration defaults
+ * @param {Logger} logger
+ */
+function normalizeInfinity (opts, fields, defaults, logger) {
+  for (const key of fields) {
+    if (opts[key] === -1) opts[key] = Infinity
+  }
+}
+
+/**
+ * Translates a string byte size, e.g. '10kb', into an integer number of bytes.
+ *
+ * @param {string} input
+ * @returns {number|undefined}
+ */
+function bytes (input) {
+  const matches = input.match(/^(\d+)(b|kb|mb|gb)$/i)
+  if (!matches) return Number(input)
+
+  const suffix = matches[2].toLowerCase()
+  let value = Number(matches[1])
+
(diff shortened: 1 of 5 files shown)


lib/config/normalizers.js
david-luna, line 1: functions in this file are moved from `config.js` file. The only changes made on them is the removal on globals inside the functions & the parameter list for 2 reasons:
- move away context inside of the functions (no references to `BOOL_PROPS` or other globals, they come as param)
- "normalize" the  signature so it will be easier to merge/refactor some of them

test/config/normalizers.test.js
```diff
@@ -0,0 +1,388 @@
...
+
+  normalizeBytes(opts, fields, defaults, logger)
+
+  t.deepEqual(opts, {
+    bytes: 1000,
+    kiloBytes: 102400,
+    megaBytes: 10485760,
+    gigaBytes: 1073741824,
+    numberBytes: 12345678,
+    badWithDefault: NaN,
```
david-luna, line 95: As for now the function does not check for a default value. There are other normalizers that fallback to default. should we do the same here?

test/config/normalizers.test.js
```diff
@@ -0,0 +1,388 @@
...
+    badWithDefault: 'not-duration',
+    badWithoutDefault: 'not-duration',
+    anotherProperty: 25
+  }
+
+  normalizeDurationOptions(opts, fields, defaults, logger)
+
+  t.deepEqual(opts, {
+    // TODO: this normalized deletes keys whose values are not valid (because format or contraint)
+    // other normalizers keep the property as is. Should we remove as well?
```
david-luna, line 130: review ☝️

test/config/normalizers.test.js
```diff
@@ -0,0 +1,388 @@
...
+    negativeNumber: -100,
+    minusOne: Infinity,
+    anotherProperty: 'value'
+  })
+
+  t.ok(logger.calls.length === 0, 'we got no warnings for bad ignore options')
+  t.end()
+})
+
+test.skip('#normalizeKeyValuePairs()', function (t) {
```
david-luna, line 201: this test is skipped because either there is a bug or I did not get the spec properly. See comment in the test corpus

test/config/_mock_logger.js
```diff
@@ -0,0 +1,48 @@
...
+ * @property {Array<any>} message
+ */
+
+/**
+ * SomeClass is an example class for my question.
+ * @class
+ * @constructor
+ * @public
+ */
+class MockLogger {
```
trentm, line 22: There is already a `class CaptureLogger` that is very similar. Is the intent to share those two?

test/config/_mock_logger.js
```diff
@@ -0,0 +1,48 @@
...
+      interpolation
+    })
+  }
+
+  fatal () { this._log('fatal', arguments[0], [].slice.call(arguments, 1)) }
+  error () { this._log('error', arguments[0], [].slice.call(arguments, 1)) }
+  warn () { this._log('warn', arguments[0], [].slice.call(arguments, 1)) }
+  info () { this._log('info', arguments[0], [].slice.call(arguments, 1)) }
+  debug () { this._log('debug', arguments[0], [].slice.call(arguments, 1)) }
+  trace () { this._log('trace', arguments[0], [].slice.call(arguments, 1)) }
```
trentm, line 45: A warning that `function (message, ...)` is *not* the general signature for the logging methods. For example sometimes the first argument is an object with fields:

```
log.info({someField: 42, err: anErrorObj}, "the message %s", "a value to interpolate")
```

https://getpino.io/#/docs/api?id=trace describes the general interface.  What you have may suffice for testing now, however.

test/config/normalizers.test.js
trentm: I think for this refactoring change you should maintain the same behaviour and that we should have a follow-up PR or issue that fixes this behaviour so that we are following: https://github.com/elastic/apm/blob/main/specs/agents/configuration.md#invalid-configuration-values

test/config/normalizers.test.js
trentm: Per https://github.com/elastic/apm/blob/main/specs/agents/configuration.md#invalid-configuration-values we should be "ignoring" the value. Because of how we currently merge the different config sources with `Object.assign(...)` that means we should be deleting invalid values -- and logging a warning.

I think this PR should try *not to change behaviour* and fix these issues in a separate change.

test/config/normalizers.test.js
trentm: Took me a while to see it. The `return` in `normalizeKeyValuePairs()` is a bug. It means that processing stops after the first key that is an object -- in this case only `objectProperty` is handled.  This was broken back in https://github.com/elastic/apm-agent-nodejs/pull/1027    Nice job finding issues with this refactoring. :)

It means that any user that was attempting to use both `globalLabels` and `addPatch` at the same time might hit the bug. I would be shocked if any users are using `addPatch` -- it is really low level.

I think you might want to *not* fix this, but wait for a separate change to fix it. Then this change can remain a "no functional change" PR.

The `return` should be `continue`, I think. Or better yet, this implementation should be rewritten to use a switch and not be so subtle in which branches are intended to be taken. It also blows up if a `null` value is given (because `typeof null === 'object'`):

With this change:

```diff
diff --git a/test/config/normalizers.test.js b/test/config/normalizers.test.js
index 275c0d6a..90155e2a 100644
--- a/test/config/normalizers.test.js
+++ b/test/config/normalizers.test.js
@@ -198,10 +198,11 @@ test('#normalizeInfinity()', function (t) {
   t.end()
 })

-test.skip('#normalizeKeyValuePairs()', function (t) {
+test.only('#normalizeKeyValuePairs()', function (t) {
   const logger = new MockLogger()
   const defaults = {}
   const opts = {
+    badNull: null,
     objectProperty: { foo: 'bar', eggs: 'spam' },
     stringProperty: 'foo=bar, eggs=spam',
     arrayProperty: ['foo=bar', 'eggs=spam'],
```

I get this crash:

```
% node test/config/normalizers.test.js
TAP version 13
# #normalizeKeyValuePairs()
/Users/trentm/el/apm-agent-nodejs12/lib/config/normalizers.js:39
        opts[key] = Object.entries(opts[key])
                           ^

TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at normalizeKeyValuePairs (/Users/trentm/el/apm-agent-nodejs12/lib/config/normalizers.js:39:28)
    at Test.<anonymous> (/Users/trentm/el/apm-agent-nodejs12/test/config/normalizers.test.js:213:3)
    at Test.bound [as _cb] (/Users/trentm/el/apm-agent-nodejs12/node_modules/tape/lib/test.js:95:17)
    at Test.run (/Users/trentm/el/apm-agent-nodejs12/node_modules/tape/lib/test.js:115:28)
    at Test.bound [as run] (/Users/trentm/el/apm-agent-nodejs12/node_modules/tape/lib/test.js:95:17)
    at Immediate.next (/Users/trentm/el/apm-agent-nodejs12/node_modules/tape/lib/results.js:157:7)
    at processImmediate (node:internal/timers:466:21)
```

trentm (approved):
Some comments and Qs, but looks good.

trentm:

lib/config/normalizers.js
```diff
@@ -0,0 +1,488 @@
...
+ */
+
+'use strict'
+
+const { WildcardMatcher } = require('../wildcard-matcher')
+
+// TODO: move this typedef to logger.js
+/**
+ * @typedef {Object} Logger
+ * @property {function(any, any, any, any): undefined} log
```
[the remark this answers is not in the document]
david-luna: Added!

PR #3355, opened 2023-05-15
test: enable tests for knex versions >1 <2

david-luna:
# Description

Enabling knex intsrumentation and tests for version range `>1 <2`. 

### Checklist

- [x] Implement code
- [x] ~Add~ Enable tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/knex.js
+++ b/lib/instrumentation/modules/knex.js
@@ -24,8 +24,12 @@ module.exports = function (Knex, agent, { version, enabled }) {
       agent._conf.spanStackTraceMinDuration)
     return Knex
   }
-  if (semver.gte(version, '1.0.0')) {
+  if (semver.gte(version, '2.0.0')) {
     agent.logger.debug('knex version %s not supported - aborting...', version)
+    return Knex
+  }
+  if (semver.gte(version, '1.0.0') && semver.lt(process.version, '12.0.0')) {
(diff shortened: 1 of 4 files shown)


lib/instrumentation/modules/knex.js
```diff
@@ -77,5 +72,19 @@ module.exports = function (Knex, agent, { version, enabled }) {
...
+      agent.logger.debug('shimming knexInstance.client.runner')
+      shimmer.wrap(knexInstance.client, 'runner', wrapRunner)
+      agent.logger.debug('shimming Knex.Client.prototype functions: %j', QUERY_FNS)
+      shimmer.massWrap(knexInstance.client, QUERY_FNS, wrapQueryStartPoint)
+    } else {
+      agent.logger.debug('could not shim Knex')
+    }
+
+    return knexInstance
+  }
```
trentm, line 89: As we discussed in a call, we'll sit on this for a little while until I have a chance to try it out with the ESM work. It is possible this new way of instrumenting will break with ESM instrumention. It is possible it will work.

trentm (approved):

lib/instrumentation/modules/knex.js
trentm: Yup, it works fine with the ESM wrapping. I think you can go ahead and look at the "nit" comments here if you like, then we can merge this.

david-luna: thanks for having a look at it :)

PR #3358, opened 2023-05-16
chore: move config.js to its own module

david-luna:
# Description

This is a continuation from a previous one (#3343) which aim for better organisation of the config logic without changing the behavior. What's don in this PR:
- move constant and defaults to `schema.js`. This file will contain in the future a schema object for easier config definitions (object shape to be discussed with \@trentm)
- moved last normalizers left in config file and added tests for them
- move `config.js` file in its own module alongside with `schema.js` and `normalizers.js` files
-  update `require` calls of the dependant modules

Relates to #3060 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,9 +28,9 @@ Notes:
         ==== x.x.x - YYYY/MM/DD
 ////
 
-
 [[release-notes-3.x]]
 === Node.js Agent version 3.x
+
 
 [[release-notes-3.46.0]]
 ==== 3.46.0 - 2023/05/15
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,24 +32,6 @@ Notes:
 [[release-notes-3.x]]
 === Node.js Agent version 3.x
 
-==== Unreleased
-
-[float]
-===== Breaking changes
-
-[float]
-===== Features
-
-[float]
-===== Bug fixes
-
-[float]
-===== Chores
-
-* Move configuration to its own module and add normalizer tests.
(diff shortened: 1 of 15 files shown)


lib/config/schema.js
david-luna, line 1: This is the place where we will consolidate a schema for our config that will help to better maintain config options and ease some tasks like:
- add/deprecate/remove properties
- control dependencies between config options
- compose normalization functions
- map options to environment variables

A draft of this schema (already discussed) may be

```
{
  oneConfigOption: {
    desc: 'config option description',  // optional
    defaultValue:  42,
    env: 'ENV_VAR_NAME', // optional: also it could be an array of names
    normalize: normalizeFn, // optional: no normalization if undefined, several normalizations if array of functions
    deps: ['anotherConfigOption'] // optional: must resolve 1st these options before resolving this one
  }
}
```

CHANGELOG.asciidoc
```diff
@@ -28,9 +28,28 @@ Notes:
...
+===== Features
+
+[float]
+===== Bug fixes
+
+[float]
+===== Chores
+
+* Move configuration to its own module and add normalizer tests.
+  ({pull}3358[#3358])
```
trentm, line 50: FWIW, I think if a change is completely internal and shouldn't impact users at all, then it would be fine to not mention it in the changelog.

trentm (approved):

CHANGELOG.asciidoc
david-luna: I agree, let's remove it.

PR #3367, opened 2023-05-17
test: fix Elasticsearch service setup, 8.7.1 requires a 'network.host' value

trentm:
Since the bump in Elasticsearch version in #3341, Elasticsearch TAV
tests in CI have been failing, because the ES service never comes up.
It is failing with:
    Exception in thread "main" org.elasticsearch.ElasticsearchParseException: null-valued setting found for key [network.host] found at line number [2], column number [14]

Special values are defined here:
https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#network-interface-values

Refs: #3341

### DIFF
--- a/.ci/docker/docker-compose.yml
+++ b/.ci/docker/docker-compose.yml
@@ -74,7 +74,7 @@ services:
     image: docker.elastic.co/elasticsearch/elasticsearch:8.7.1
     environment:
       - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
-      - "network.host="
+      - "network.host=_site_"
       - "transport.host=127.0.0.1"
       - "http.host=0.0.0.0"
       - "xpack.security.enabled=false"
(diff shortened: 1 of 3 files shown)


trentm (commented):
/test tav "@elastic/elasticsearch,elasticsearch"

trentm (commented):
/test tav @elastic/elasticsearch,elasticsearch

david-luna (approved):

trentm:
Hrm, the `tav-command` reporting is not great. After doing `/test tav @elastic/elasticsearch,elasticsearch` earlier the "Checks" area *did* show successfully running the relevant tav-command jobs. However, now those checks don't show ... because those only show the checks relevant for the latest review comment, which is now David's "approved these changes".  The older passing run can be seen here: https://github.com/elastic/apm-agent-nodejs/actions/runs/5008165366  and all the `tav-command` workflow runs can be seen here: https://github.com/elastic/apm-agent-nodejs/pull/3367/checks but you need to labouriously click through all those links to find possible relevant test runs.  

I'm not sure what a great answer is here. The current state is that our using `pull_request_review` as the trigger for `tav-command` isn't great.

PR #3372, opened 2023-05-23
chore: move transport to its own module

david-luna:
# Description

Transport resolution logic is embed inside configuration module but it doesn't need to be there. This is another PR on the series for #3060 in order to put configuration module into diet. The changes are:
- create a transport module which exports `createTransport(config, agent)` API
- move `NoopTransport` next to transport module
- create `elastic-apm-transport` module which contains all logic to instantiate an `ElasticAPMHttpClient`
- use `createTransport` API in the agent to set `_transport` property

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


trentm (approved):
Approving if you want to move forward with this PR and other PRs that depend on this. However, a note/question/suggestion:

One of things that I think causes some unneeded complexity is the multiple names that are used to refer to this thing: "transport", "ElasticAPMHttpClient", "Client", "http client", "apm-nodejs-http-client.git".  I think it would be nice to tend towards one favoured name. We can't go change that name everywhere because (a) `transport` *is* a supported config var, for better or worse and (b) we aren't going to change the "apm-nodejs-http-client.git" repo name.   However, longer term I could see us possibly deprecating support for the `transport` config var and, more likely, merging apm-nodejs-http-client.git into apm-agent-nodejs.git. Having it as a separate repo has just added coordination burden in the past.

My suggestion for a name to tend towards is "apm-client" ("ApmClient", "apmClient").  "Transport" is too generic -- though the APM specs [*do* call it "transport"](https://github.com/elastic/apm/blob/main/specs/agents/transport.md). "HTTP" in the name is too specific to the type of transport -- APM server has toyed with supporting gRPC for intake and *does* support OTLP.  "APM Server" in the name (e.g. "APM Server client") is slightly weird and the client *does* actually support making calls that are specific to the APM Lambda extension (e.g. `POST /register/transaction`).

The reason I bring this up here, is because I wonder if this would be a good time to think about this naming given that we are creating a new subdir and renaming files.  Apologies that this might mean more work for you on this PR.

Thoughts?

## a possible option

A first pass could consider keeping `agent._transport` instead of changing to `agent._apmClient` or `agent._apmclient`.

```
lib/agent.js
    this._apmClient = createApmClient(this._conf, this)

lib/apm-client/index.js
    The module that exports `createApmClient()`. I wonder if this could hold
    most of what you have in "lib/transport/elastic-apm-transport.js"
    and "lib/transport/transport.js" -- or it would still be fine to separate
    this into two files.

lib/apm-client/noop-apm-client.js
    Rename NoopTransport to NoopApmClient.

lib/apm-client/apm-client.js
    This could be the eventual home of apm-nodejs-http-client.git#index.js
```

david-luna:
I agree on being more explicit at naming and I find you're proposal a very good way to move forward. Maybe with a slight change

```
lib/agent.js
    this._apmClient = createApmClient(this._conf, this)

lib/apm-client/apm-client.js    <-- Do not use "index.js" here for better code search (discussed on previous PR for config)
    The module that exports `createApmClient()`. I wonder if this could hold
    most of what you have in "lib/transport/elastic-apm-transport.js"
    and "lib/transport/transport.js" -- or it would still be fine to separate
    this into two files.

lib/apm-client/noop-apm-client.js
    Rename NoopTransport to NoopApmClient.

lib/apm-client/http-apm-client.js  <-- Add the type of client in the name
    This could be the eventual home of apm-nodejs-http-client.git#index.js

Then we could add new type of clients like

lib/apm-client/grpc-apm-client.js 
    This eventually will be an APM client implementation with gRPC
```

I've also checked the change from `agent._transport` to `agent._apmClient` and mostly affects tests accessing directly to the property.

IMHO it's better to have it done in this PR so I'll do both and wait for your review

david-luna:
I withdraw what I said about renaming `agent._transport`. The change affects 50+ files and adds too much noise. When this is merged I'll do a explicit PR for that change

lib/apm-client/apm-client.js
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Returns an APM client suited for the configuration provided
+ *
+ * @param {Object} config The agent's configuration
+ * @param {Object} agent The agents instance
+ */
+function createApmClient (config, agent) {
+  let client
+  if (config.disableSend || config.contextPropagationOnly) {
+    client = new NoopApmClient()
```
trentm, line 27: This is a slight behaviour change here that we are adding the event listeners for this code path.
I would bias towards not changing the behaviour here because this is meant to be a refactoring.
There is also no point in having the event listeners added for a NoopApmClient that we don't expect to ever emit those events.

If you agree, then the added `on ()` method to NoopApmClient could be removed. If we want NoopApmClient to mirror that part of the client API, it might be clearer to have it extend EventEmitter.

lib/apm-client/apm-client.js
```diff
@@ -0,0 +1,113 @@
...
+        msg += `\n${err.response}`
+      }
+
+      agent.logger.error(msg)
+    })
+
+    return client
+  }
+
+  return config.transport(config, agent)
```
trentm, line 108: nit: I think it would be clearer if this case could be handled at the top of `createApmClient`. Perhaps with:

```js
  if (config.disableSend || config.contextPropagationOnly) {
    return new NoopApmClient()
  } else if (typeof config.transport === 'function') {
    return config.transport(config, agent)
  }
  
  const client = new ElasticAPMHttpClient(getHttpClientConfig(config, agent))
  // ...
  ```

trentm (approved):
Approved with a suggestion.

lib/apm-client/apm-client.js
david-luna: Indeed this is a change of behaviour for the `NoopApmClient` and you're right this is beyond the scope of this PR. I'll revert this change for now.

The intent on making the clients to implement `on()` API is to enable testability of the logic that updates configuration form central configuration. Also I think future client implementations (gPRC?) must implement `on()` in order to send config back to the agent.

PR #3373, opened 2023-05-23
chore: refactor config tests

david-luna:
# Description

Move config tests to a dedicated folder and also tried to simplify them. The tests rely on an array of config definitions with the following shape:
```javascript
{
  name: ' nameOfTheConfig',
  defaultValue: 'value',    // defined if has a default
  envVar: 'ENV_VAR_NAME',    // defined if option can be set via ENV
  types: [ 'boolean' ],    // defined if there is a normalizer method 
  centralConfig: ' central_config_name',    // defined if it has a central config
}
```

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


test/config/config.test.js
```diff
@@ -0,0 +1,103 @@
...
+    value = value2
+  }
+
+  process.env[option.envVar] = value.toString()
+  agent.start(agentOptsNoTransportNoLogger)
+  t.strictEqual(agent._conf[option.name], value)
+
+  process.env[option.envVar] = existingEnvValue
+  agent.destroy()
+  t.end()
```
trentm, line 59: nit: My preference would be that utility test functions do *not* call the `t.end()` themselves. While I know this means that the tests below need to write `t.end()` more times, it *does* mean that it is a bit easier to read a given test block:

```js
test(..., t => {
    ...
})
```

and see if it explicitly calls `t.end()`.

test/config/config.test.js
```diff
@@ -0,0 +1,103 @@
...
+      return new NoopTransport()
+    }
+  }
+)
+
+function hasType (def, type) {
+  return def.types && def.types.indexOf(type) !== -1
+}
+
+function testEnvVar (t, option, value1, value2) {
```
trentm, line 42: nit: Perhaps calling this `function assertEnvVar` or similar, which is more of a current naming pattern:

```
% rg 'function test' test | wc -l
       5
% rg 'function assert' test | wc -l
      50
```

test/config/config.test.js
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// Test BOOLS that can be set via ENV
+const BOOL_OPTS_WITH_ENV_EXCLUDED = [
+  // This ones depend on other options so we should have specific tests
+  'captureSpanStackTraces',
+  'asyncHooks',
+  // this one seems to have a bug
+  'breakdownMetrics'
+]
```
trentm, line 69: It would be nice if this was data-driven eventually. I.e. if the set of config vars tested here was from an export of `lib/config/schema.js`. Is that the eventual intention?

I see that is done for `NUM` tests below.

test/config/config.test.js
```diff
@@ -0,0 +1,103 @@
...
+  test(`${option.name} should be configurable by environment variable ${option.envVar}`, function (t) {
+    testEnvVar(t, option, !option.defaultValue)
+  })
+
+  test(`should overwrite option property ${option.name} by environment variable ${option.envVar}`, function (t) {
+    testEnvVar(t, option, !option.defaultValue, !!option.defaultValue)
+  })
+})
+
+// Test BOOLS that can be set via ENV
```
trentm, line 85:
```suggestion
// Test NUMS that can be set via ENV
```

test/config/fixtures/config-schema.js
```diff
@@ -0,0 +1,86 @@
...
+  BOOL_OPTS,
+  NUM_OPTS,
+  BYTES_OPTS,
+  MINUS_ONE_EQUAL_INFINITY,
+  ARRAY_OPTS,
+  KEY_VALUE_OPTS
+} = require('../../../lib/config/schema')
+
+// Create schema for tests
+const CONFIG_SCHEMA = Object.keys(DEFAULTS).reduce((acc, key) => {
```
trentm, line 24: Do you intend to define the `CONFIG_SCHEMA` in "./lib/config/..." files eventually?  I'm a little confused why we would have an intermediate schema object just defined in the tests.

david-luna: Indeed that's the intent although it may have some slight changes at the end. The place for `CONFIG_SCHEMA` shall be `./lib/config/schema.js`

I think this tests will give us more confidence to move to the new schema before doing the refactor (in which we will remove this file and use the actual schema)

test/config/config.test.js
david-luna: I think it might be even better for discussions and improvements to have `CONFIG_SCHEMA` written on a direct export instead of having in a procedural way. I'll keep  the generator code just in case we need to generate it again

test/config/fixtures/config-schema.js
```diff
@@ -0,0 +1,445 @@
...
+    envVar: 'ELASTIC_APM_ERROR_ON_ABORTED_REQUESTS',
+    types: ['boolean']
+  },
+  {
+    name: 'exitSpanMinDuration',
+    defaultValue: '0ms',
+    envVar: 'ELASTIC_APM_EXIT_SPAN_MIN_DURATION',
+    centralConfig: 'exit_span_min_duration',
+    types: [
+      'duration({"defaultUnit":"ms","allowedUnits":["us","ms","s","m"],"allowNegative":false})'
```
trentm, line 164: XXX

PR #3382, opened 2023-05-26
fix: sanitize field names shouldn't mutate the original object

lytc:
### Description
This PR changes the sanitize field name (redactKeysFromObject / redactKeysFromPostedFormVariables) to not mutate the original object

#### Current status

redactKeysFromObject / redactKeysFromPostedFormVariables mutate the original object which sanitized the original request / response  headers

For example, our stack running in the AWS Lambda functions it sanitized www-authenticate response header
Screenshot 2023-05-26 at 10 47 21

#### After this PR

redactKeysFromObject / redactKeysFromPostedFormVariables will return the new copy of the original object

### Checklist

- [x] Implement code
- [x] Add tests

### DIFF
--- a/lib/filters/sanitize-field-names.js
+++ b/lib/filters/sanitize-field-names.js
@@ -30,7 +30,7 @@ function redactKeysFromPostedFormVariables (body, requestHeaders, regexes) {
   // if body is a string, use querystring to create object,
   // pass to redactKeysFromObject, and reserialize as string
   if (typeof body === 'string') {
-    const objBody = redactKeysFromObject(querystring.parse(body), regexes);
+    const objBody = redactKeysFromObject(querystring.parse(body), regexes)
     return querystring.stringify(objBody)
   }
 
--- a/lib/filters/sanitize-field-names.js
+++ b/lib/filters/sanitize-field-names.js
@@ -30,31 +30,25 @@ function redactKeysFromPostedFormVariables (body, requestHeaders, regexes) {
   // if body is a string, use querystring to create object,
   // pass to redactKeysFromObject, and reserialize as string
   if (typeof body === 'string') {
-    const objBody = querystring.parse(body)
-    redactKeysFromObject(objBody, regexes)
+    const objBody = redactKeysFromObject(querystring.parse(body), regexes);
     return querystring.stringify(objBody)
   }
 
   return body
 }
 
-function redactKeyFromObject (obj, regex) {
-  for (const [key] of Object.entries(obj)) {
-    if (regex.test(key)) {
-      obj[key] = REDACTED
-    }
-  }
-  return obj
-}
-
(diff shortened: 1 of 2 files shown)


david-luna:
Hi @lytc 

Thanks for contributing to elastic nodejs agent. I've checked and `redactKeysFromObject` API is used in this ways
- as function returning a value [see](https://github.com/elastic/apm-agent-nodejs/blob/3fefa56e4ea713561640fae857783acf1f94ee94/lib/filters/sanitize-field-names.js#L27)
- as a function which mutates the referenced object [see](https://github.com/elastic/apm-agent-nodejs/blob/3fefa56e4ea713561640fae857783acf1f94ee94/lib/filters/sanitize-field-names.js#L34)

So returning a new object in `redactKeysFromObject` may break other parts. Maybe another way of solving this might me to clone the response headers before adding the into the context. This is [the code](https://github.com/elastic/apm-agent-nodejs/blob/3fefa56e4ea713561640fae857783acf1f94ee94/lib/parsers.js#L111) that does it.

I would change it to

```javascript
if (conf.captureHeaders) {
  context.headers = Object.assign({}, res.headers || parseHttpHeadersFromReqOrRes(res, true))
  context.headers = redactKeysFromObject(context.headers, conf.sanitizeFieldNamesRegExp)
}
```

ly-tc:
Hi @david-luna

Thank you for your review and the suggestion!

From the PR I already updated the redactKeysFromPostedFormVariables to use the returned value from redactKeysFromObject instead of the referenced object [see](https://github.com/elastic/apm-agent-nodejs/pull/3382/files#diff-bba5e44f2ac00f3a32a71a0a1fdd2532c73cf833b855b31b9899764709eee837R33). 

And I already search all code to make sure there is no place that uses the referenced object so it will not break other parts!

david-luna:
@lytc 

Moving forward on your API change for `redactKeysFromObject` I've done a couple of suggestions to document the function and test its new contract. Also with this new contract there is a place where the agent does not need to clone the `request.headers` object

Here is the patch to avoid the unnecessary `Objec.assing(...)`

```bash
git diff
diff --git a/lib/parsers.js b/lib/parsers.js
index 17103d54..99263467 100644
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -56,10 +56,7 @@ function getContextFromRequest (req, conf, type) {
   }

   if (conf.captureHeaders) {
-    context.headers = redactKeysFromObject(
-      Object.assign({}, req.headers),
-      conf.sanitizeFieldNamesRegExp
-    )
+    context.headers = redactKeysFromObject(req.headers, conf.sanitizeFieldNamesRegExp)

     // TODO: remove filterHttpHeaders option in next major release
     // https://github.com/elastic/apm-agent-nodejs/issues/3332
```

PR #3383, opened 2023-05-26
chore: add logging preamble

david-luna:
# Description

This PR add initial logging according to the spec
https://github.com/elastic/apm/blob/main/specs/agents/logging.md#logging-preamble

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


lib/config/config.js
```diff
@@ -439,6 +454,76 @@ function loadConfigFile (configFile) {
...
+  const offsetPad = Math.abs(offsetHours) < 10 ? '0' : ''
+  const timeZone = `UTC${offsetSign}${offsetPad}${Math.abs(offsetHours * 100)}`
+  const frameWorkName = nonDefaultOpts.frameWorkName || 'unknown'
+  const frameWorkVersion = nonDefaultOpts.frameWorkVersion || ''
+  const environmentMessage = [
+    '\nEnvironment:',
+    `- Process ID:  ${process.pid}`,
+    `- Process name: ${process.title}`,
+    // TODO: check sec concerns
+    // ` - command line: ${process.cwd()}`,
```
david-luna, line 482: This is also mentioned in the spec. Pointing to [this issue](https://github.com/elastic/apm-agent-nodejs/issues/1916)
Maybe it's better to not log this value

lib/config/config.js
```diff
@@ -50,6 +52,24 @@ const {
   normalizeSpanStackTraceMinDuration
 } = require('./normalizers')
 
+const EXCLUDE_FIELDS = {
```
david-luna, line 55: moved out from `toJSON()` to reuse them in the logging preamble

lib/config/config.js
```diff
@@ -439,10 +454,82 @@ function loadConfigFile (configFile) {
   return null
 }
 
+function printLoggingPreamble (config, sources, configFilePath) {
```
trentm, line 457: Perhaps link to https://github.com/elastic/apm/blob/main/specs/agents/logging.md#logging-preamble in a comment here.

lib/config/config.js
```diff
@@ -145,6 +169,14 @@ class Config {
       this.logger = logging.createLogger(this.logLevel, customLogger)
     }
 
+    // Print logging preamble now that the logger is completelly resolved
+    const sources = {
+      environment: envOptions,
+      start: opts,
+      file: confFile || {}
+    }
+    printLoggingPreamble(this, sources, configFilePath)
```
trentm, line 178: I suspect it will make this more difficult for processing, but I think the logging preamble should come *after* the fallback serviceName/serviceVersion block below and *after* the `normalize(...)` call. It should show the effective values for some fields. E.g.:

- If all of `envOptions`, `opts`, and `confFile` set a `serviceName`, but the winner is invalid, then the logging preamble should show `null` for the `serviceName`.
- If the serviceName isn't specified, but the default value is determined below, the logging preamble should show it.
- If an invalid value for some config var is shown, and `normalize(...)` ignores it, the logging preamble should **not** include that config var unless it is one of the "Log Always" vars.

I wonder if this could be printed in `Agent.prototype.start` ... before the `if (!this._conf.active) {` block.

(Aside: it would be nice to include `this._agentActivationMethod` and perhaps drop the existing `agent configured correctly` log message. However, that brings up other questions, like whether we drop the `startTrace`, `main` and `dependencies`. :)

lib/config/config.js
```diff
@@ -439,10 +454,82 @@ function loadConfigFile (configFile) {
...
+
+    // const origin = nonDefaultKeys.reduce((orig, key) => (optKey in sources[key]) ? key : orig, '')
+    const origin = nonDefaultKeys.find(key => (optKey in sources[key]))
+    let value
+
+    if (REDACT_FIELDS[optKey]) {
+      value = REDACTED
+    } else if (NICE_REGEXPS_FIELDS[optKey] && Array.isArray(nonDefaultOpts[optKey])) {
+      // JSON.stringify() on a RegExp is "{}", which isn't very helpful.
+      value = nonDefaultOpts[optKey].map(r => r instanceof RegExp ? r.toString() : r)
```
trentm, line 517: The regexp fields are all internal processed versions of the config vars, I think, so we shouldn't need to show them in the logging preamble.

trentm (commented):
Sorry, I have lots of feedback. :)

lib/config/config.js
```diff
@@ -439,10 +454,82 @@ function loadConfigFile (configFile) {
   return null
 }
 
+function printLoggingPreamble (config, sources, configFilePath) {
+  const { logger } = config
+  const nonDefaultKeys = ['environment', 'start', 'file']
+  const nonDefaultOpts = Object.assign.apply(null, nonDefaultKeys.map(k => sources[k]).reverse())
+  const isPreview = AGENT_VERSION.indexOf('-') !== -1
+
+  // Version and pre-release warning
```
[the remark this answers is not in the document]
david-luna: I have no preference so I went with the sample in the spec. Fine for me to change to a JSON format :) 

We can log 1st the `agent-env-config` and then the warn if necessary.

lib/config/config.js
david-luna: I think here there are two views of the configuration:
- the configuration given to the agent
- the configuration that finally is resolved

The spec is not clear about what we need to show here, on each case the value is different
- showing the given configuration users can inspect if correctly passed with the readable format for some options like durations (`2s`, ...)
- showing resolved configuration users can se in which final configuration the agent is working with.

Notice that with the current implementation `normalize()` already uses the logger and, although the spec does not say the preamble should be the very 1st message, it would look strange IMHO.

On a previous comment you shared a JSON structure for config
```
agentConfig: {
  "serviceName": {
    "value": "my-service",
    "source": "env",
    "normalizedName": "service_name"
  },
  ...
}
```

Is that `normalizedName` property the result of the normalization? Do you think we can show both provided and normalized value? If so, is it necessary to log warnings about configs?

lib/config/config.js
```diff
@@ -439,10 +454,82 @@ function loadConfigFile (configFile) {
...
+  logger.info(`Elastic APM Node.js Agent, version: ${AGENT_VERSION}`)
+
+  if (isPreview) {
+    logger.warn('This is a pre-release version and not intended for use in production environments!')
```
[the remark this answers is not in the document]
david-luna: > Perhaps include the package name elastic-apm-node in the message.

I see our logger already adds the package name

```bash
node ./examples/trace-fetch.js | ecslog
[2023-05-30T16:46:30.109Z]  INFO (elastic-apm-node): Elastic APM Node.js Agent, version: 3.46.0
    agentVersion: "3.46.0"
    env: {
        "pid": 29988,
        "pname": "node",
        "os": "22.5.0",
        "arch": "arm64",
        "host": "XXXX.home",
        "timezone": "UTC+0200",
        "runtime": "Node.js v18.16.0"
    }
    config: {
       ...
    }
```

Let's add it only if the customer is using a custom logger.

lib/config/config.js
david-luna: Thank for the extensive feedback. Following it and since the agent is also resolving some configurations I made some changes.
- config resolves a new property with information for the logging preamble
- agent gets that info at `start()` and enriches it
- agent becomes responsible to print the preamble through the logger



> if a config var is shadowed by a higher config level (e.g. ELASTIC_APM_SERVICE_NAME overrides a value in agent.start(opts)), then it might be nice to have a log.debug for this

I agree


> > Is that normalizedName property the result of the normalization?

> That's coming from the spec line that says the normalized snake_case name should be logged.
> However, I'm not sure if we need to do this, or need to for all config vars. It might be useful for some of them.

It was easier to normalize all so it's done for all configuration options.

lib/config/config.js
```diff
@@ -439,6 +455,101 @@ function loadConfigFile (configFile) {
...
+/**
+ * Transforms keys in camelCase format to snake_case format
+ *
+ * @param {String} key A string contining an object key
+ * @returns {String} the key trasformed to snake_case
+ */
+function formatOptionKey (key) {
+  return key.replace(/[A-Z]/g, function (char) {
+    return `_${char.toLowerCase()}`
+  })
```
trentm, line 538: Will this cause a surprise for the rare config var like `instrumentIncomingHTTPRequests`?
What about having an explicit list of these normalized (or perhaps call them "cross-agent config var") names?
Something like this in schema.js:

```js
const CROSS_AGENT_CONFIG_VAR_NAME = {
  logLevel: 'log_level',
  serviceName: 'service_name'
  // ...
}
```

or eventually this could be an attribute of the richer `CONFIG_VARS` object.

One reason I like having these explicitly written as strings in a richer `CONFIG_VARS` is that a `grep` will find these. It can be useful to have a grep hit that shows `service_name` next to `serviceName`.

We could perhaps restrict defining these `crossAgentName` values to just those that are in central config and/or listed in the shared specs.

trentm (commented):
This is looking great. I haven't had a chance to test it out yet. I'll do that tomorrow.

lib/config/config.js
```diff
@@ -439,6 +455,101 @@ function loadConfigFile (configFile) {
...
+ * - normalizedValue: the value normalized by the configuration which is used in the agent
+ * - file: if source === 'file' this proprty is defined with the path of the config file
+ *
+ * @param {Object} config The configuration object with normalized options
+ * @param {Array<{ name: string, ops: Object}>} sources The list of sources of options ordered by priority
+ * @param {String | undefined} configFilePath Path of the config file if loaded
+ * @returns {Object} Object with agent, environment, and configuration data.
+ */
+function getLoggingPreambleData (config, sources, configFilePath) {
+  const optsFromSources = sources.reduce((prev, s) => Object.assign({}, s.opts, prev), {})
```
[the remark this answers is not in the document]
david-luna: If I'm not mistaken the in the expression `Object.assign(a, b)`, the var `b` overrides `a` and in this case implicitly will give more priority to the last source from the list. And we want just the opposite.

But your observation made me think about a better API for that. I think we can use `Array.prototype.reduceRight`

lib/config/config.js
```diff
@@ -439,6 +455,101 @@ function loadConfigFile (configFile) {
...
+ * @param {any} value The value to be formatted if it's a string
+ * @param {Boolean} shouldRedact true if the value should be redacted
+ * @returns {any}
+ */
+function formatOptionValue (value, shouldRedact) {
+  if (shouldRedact) {
+    return REDACTED
+  }
+
+  return typeof value === 'string' ? `"${value}"` : value
```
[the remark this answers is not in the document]
david-luna: The spec is enforcing it but you're right. I guess is because the spec is assuming a string format (but not enforcing it) instead of a JSON one.
https://github.com/elastic/apm/blob/9fadb845e0f7d7c4da544719c734f3ebedf8fd8a/specs/agents/logging.md?plain=1#L138

lib/config/config.js
david-luna: > Will this cause a surprise for the rare config var like instrumentIncomingHTTPRequests?

It will display even a rarer name :) 

I'll apply your suggestion, thanks.

lib/config/config.js
```diff
@@ -119,11 +139,15 @@ class Config {
...
     const envOptions = readEnv()
+    let configFilePath
 
     // If we didn't find a config file on process boot, but a path to one is
     // provided as a config option, let's instead try to load that
     if (confFile === null && opts && opts.configFile) {
       confFile = loadConfigFile(opts.configFile)
+      configFilePath = confFile && path.resolve(opts.configFile)
+    } else if (confFile) {
+      configFilePath = path.resolve(process.env.ELASTIC_APM_CONFIG_FILE || 'elastic-apm-node.js')
```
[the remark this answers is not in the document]
david-luna: I did not want to change the behaviour of the existing APIs but I think is a good idea :)

lib/agent.js
```diff
@@ -250,52 +242,62 @@ Agent.prototype.start = function (opts) {
...
+  this.logger.info(preambleData, 'Elastic APM Node.js Agent, version: %s', version)
+
+  if (isPreviewVersion) {
+    this.logger.warn('Version %s is a pre-release and not intended for use in production environments', version)
+  }
+
+  // Check cases where we do *not* start.
+  if (!this._conf.active) {
+    this.logger.debug('Elastic APM agent disabled (`active` is false)')
+  } else if (!this._conf.serviceName) {
```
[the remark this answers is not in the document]
david-luna: Definitely!

lib/config/config.js
```diff
@@ -439,6 +455,101 @@ function loadConfigFile (configFile) {
...
+ * @param {String | undefined} configFilePath Path of the config file if loaded
+ * @returns {Object} Object with agent, environment, and configuration data.
+ */
+function getLoggingPreambleData (config, sources, configFilePath) {
+  const optsFromSources = sources.reduce((prev, s) => Object.assign({}, s.opts, prev), {})
+  const result = {
+    agentVersion: AGENT_VERSION,
+    env: {
+      pid: process.pid,
+      pname: process.title,
```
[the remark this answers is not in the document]
david-luna: Then I think we should have the `proc` prefix for `pid` also. I'm going to change both.

lib/config/config.js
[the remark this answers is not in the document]
trentm: > I want 1st element to preserve its values since is the one with most priority.

That's why I prefixed with "If you are willing to reverse the priority of the given `sources`".

trentm:
This is looking good. Some comments on this example run on my machine. Feel free to push back, I'm obsessing some.

```
% edeployment11  node examples/trace-http.js  | ecslog
[2023-05-31T18:24:32.697Z]  WARN (elastic-apm-node): Invalid "cloudProvider" config value null, falling back to default auto
[2023-05-31T18:24:32.700Z]  INFO (elastic-apm-node): Elastic APM Node.js Agent, version: 3.46.0
```

What about "Elastic APM Node.js Agent v3.46.0"? A little shorter is nice.

```
    agentVersion: "3.46.0"
    env: {
        "procid": 22779,
```

I really think this should stay as "pid"
        
```
        "proctitle": "node",
        "os": "22.5.0",
```

Probably want to include `os.platform()` as well. Then perhaps the code could have a comment linking to https://en.wikipedia.org/wiki/Darwin_%28operating_system%29#Release_history for translating from a darwin to a macOS version (per https://stackoverflow.com/a/14989382/14444044).

```
        "arch": "x64",
        "host": "pink.local",
        "timezone": "UTC-0700",
        "runtime": "Node.js v16.20.0"
    }
    config: {
        "serviceName": {
            "source": "start",
            "sourceValue": "example-trace-http",
            "normalizedValue": "example-trace-http",
```

What about (a) "value" rather than "normalizedValue" and (b) only include "sourceValue" if it is different?
My main goal here is brevity.

```
            "normalizedName": "service_name"
```

Thoughts on calling that "commonName" rather than "normalizedName"? I think the word is more accurate.

```
        },
        "serviceVersion": {
            "source": "default",
            "normalizedValue": "3.46.0",
            "normalizedName": "service_version"
        },
        "serverUrl": {
            "source": "environment",
            "sourceValue": "[REDACTED]",
            "normalizedValue": "[REDACTED]",
            "normalizedName": "server_url"
        },
```

I think we should include the serverUrl. Earlier discussion on sanitizing it is here: https://github.com/elastic/apm/pull/713#discussion_r1022196424   Note that we should do *some* sanitization of it to remove possible basic auth in it.

```
        "logLevel": {
            "source": "default",
            "normalizedValue": "info",
            "normalizedName": "log_level"
        },
        "cloudProvider": {
            "source": "file",
            "sourceValue": null,
            "normalizedValue": "auto",
            "file": "/Users/trentm/el/apm-agent-nodejs14/elastic-apm-node.js"
        },
        "centralConfig": {
            "source": "file",
            "sourceValue": false,
            "normalizedValue": false,
            "file": "/Users/trentm/el/apm-agent-nodejs14/elastic-apm-node.js"
        },
        "metricsInterval": {
            "source": "file",
            "sourceValue": "0s",
            "normalizedValue": 0,
            "file": "/Users/trentm/el/apm-agent-nodejs14/elastic-apm-node.js"
        },
        "logUncaughtExceptions": {
            "source": "file",
            "sourceValue": true,
            "normalizedValue": true,
            "file": "/Users/trentm/el/apm-agent-nodejs14/elastic-apm-node.js"
        },
        "secretToken": {
            "source": "environment",
            "sourceValue": "[REDACTED]",
            "normalizedValue": "[REDACTED]",
            "normalizedName": "secret_token"
        },
        "useElasticTraceparentHeader": {
            "source": "start",
            "sourceValue": false,
            "normalizedValue": false
        },
        "usePathAsTransactionName": {
            "source": "start",
            "sourceValue": true,
            "normalizedValue": true
        }
    }
    activationMethod: "require"
```


## minimum size

Here is the minimum size log.info line that we will be adding to all users:

```
[2023-05-31T18:42:40.010Z]  INFO (elastic-apm-node): Elastic APM Node.js Agent, version: 3.46.0
    agentVersion: "3.46.0"
    env: {
        "procid": 23166,
        "proctitle": "node",
        "os": "22.5.0",
        "arch": "x64",
        "host": "pink.local",
        "timezone": "UTC-0700",
        "runtime": "Node.js v16.20.0"
    }
    config: {
        "serviceName": {
            "source": "default",
            "normalizedValue": "elastic-apm-node",
            "normalizedName": "service_name"
        },
        "serviceVersion": {
            "source": "default",
            "normalizedValue": "3.46.0",
            "normalizedName": "service_version"
        },
        "serverUrl": {
            "source": "default",
            "sourceValue": "[REDACTED]",
            "normalizedValue": "[REDACTED]",
            "normalizedName": "server_url"
        },
        "logLevel": {
            "source": "default",
            "normalizedValue": "info",
            "normalizedName": "log_level"
        }
    }
    activationMethod: "require"
```

raw:

```
{"log.level":"info","@timestamp":"2023-05-31T18:42:28.239Z","log":{"logger":"elastic-apm-node"},"agentVersion":"3.46.0","env":{"procid":23158,"proctitle":"node","os":"22.5.0","arch":"x64","host":"pink.local","timezone":"UTC-0700","runtime":"Node.js v16.20.0"},"config":{"serviceName":{"source":"default","normalizedValue":"elastic-apm-node","normalizedName":"service_name"},"serviceVersion":{"source":"default","normalizedValue":"3.46.0","normalizedName":"service_version"},"serverUrl":{"source":"default","sourceValue":"[REDACTED]","normalizedValue":"[REDACTED]","normalizedName":"server_url"},"logLevel":{"source":"default","normalizedValue":"info","normalizedName":"log_level"}},"activationMethod":"require","ecs":{"version":"1.6.0"},"message":"Elastic APM Node.js Agent, version: 3.46.0"}
```

That's 792 chars. I don't think there is much we can do about that length, though shorter would be nicer.

lib/agent.js
```diff
@@ -250,33 +247,43 @@ Agent.prototype.start = function (opts) {
...
+    this._agentActivationMethod = agentActivationMethodFromStartStack(startStack, this.logger)
+    preambleData.activationMethod = this._agentActivationMethod
+
+    if (this._conf.logLevel === 'trace') {
+      // Attempt to load package.json from process.argv.
+      let pkg = null
+      try {
+        var basedir = path.dirname(process.argv[1] || '.')
+        pkg = require(path.join(basedir, 'package.json'))
+      } catch (e) {}
```
trentm, line 268: Side note: This only looks in the same dir as the script. I think it might be fairly common for scripts to be in a "scripts" or "bin" or other subdir. Looking up at least one dir might be interesting. For example `ELASTIC_APM_LOG_LEVEL=trace node examples/trace-http.js` results in `pkg` not being found.   This isn't something for *this* PR though.

lib/agent.js
```diff
@@ -250,33 +247,43 @@ Agent.prototype.start = function (opts) {
...
+    }
+  }
+
+  this.logger.info(preambleData, 'Elastic APM Node.js Agent, version: %s', version)
+
+  if (isPreviewVersion) {
+    this.logger.warn('Version %s is a pre-release and not intended for use in production environments', version)
+  }
+
+  // TODO: this deactivation should be handled as well when config resolution/normalizaiton process
```
trentm, line 282: Hrm. I'm second guessing myself on whether it is the job of "lib/config/..." to do this. <shrug/> It could be in either place.

test/config/config.test.js
```diff
@@ -0,0 +1,156 @@
...
+const test = require('tape')
+
+const AGENT_VERSION = require('../../package.json').version
+
+const Agent = require('../../lib/agent')
+const { createMockLogger } = require('../_mock_logger')
+const { NoopApmClient } = require('../../lib/apm-client/noop-apm-client')
+const { REDACTED } = require('../../lib/constants')
+
+test('#loggingPreamble', function (t) {
```
trentm, line 24: Perhaps:

```suggestion
test('logging preamble', function (t) {
```

The intent of the "#foo" naming with the hash sign was to indicate it was testing the "foo" property of a class/object. E.g. `test('#getServiceName()'` in "agent.test.js" was to test the `getServiceName` method of the `Agent` class.

test/config/config.test.js
trentm, line 1: Woudl this file be more appropriately named "test/logging-preamble.test.js"? The logging of the preamble, and adding extra data to it (like `activationMethod`) is handled by agent.js.  Also there is already "test/config.test.js" with the same basename.

trentm (approved):
I have a number of nits, but looks great.

test/config/config.test.js
david-luna: Added that file since all logic started only inside config and I the long term I think test folder should have same structure than lib (at least for module testing)

With the latests changes I think your proposal is a good one. That file may come back when we improve the config processing using a schema.

lib/config/config.js
```diff
@@ -523,6 +532,25 @@ function getLoggingPreambleData (config, sources, configFilePath) {
...
+  const url = parseUrl(urlStr)
+  let result = url.href
+
+  if (url.username || url.password) {
+    result = result.replace(url.username, REDACTED)
+    result = result.replace(url.password, REDACTED)
+  }
+
+  return result
+}
```
trentm, line 552: I think this may not be correct.

Doing:

```
% ELASTIC_APM_DISABLE_SEND=true ELASTIC_APM_SERVER_URL='http://user@example.com/file.html' node examples/trace-http.js | ecslog
...
    config: {
        "serviceName": {
            "source": "default",
            "value": "elastic-apm-node",
            "commonName": "service_name"
        },
        "serviceVersion": {
            "source": "default",
            "value": "3.46.0",
            "commonName": "service_version"
        },
        "serverUrl": {
            "source": "environment",
            "value": "[REDACTED]http://[REDACTED]@example.com/file.html",
            "commonName": "server_url"
        },
```

lib/config/config.js
```diff
@@ -506,14 +508,21 @@ function getLoggingPreambleData (config, sources, configFilePath) {
...
+    const optionDetails = { source: sourceName }
 
-    result.config[key] = {
-      source: sourceName,
-      sourceValue: shouldRedact ? REDACTED : optsFromSources[key],
-      normalizedValue: shouldRedact ? REDACTED : config[key]
+    result.config[key] = optionDetails
+
+    if (shouldRedact) {
+      optionDetails.value = key === 'serverUrl' ? sanitizeUrl(config[key]) : REDACTED
```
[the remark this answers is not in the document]
trentm: nm, I am wrong.

PR #3409, opened 2023-06-07
Chore (upgrade eslint config and packages)

### LINKED ISSUE #2583: upgrade to eslint@8
If after we switch to eslint (https://github.com/elastic/apm-agent-nodejs/pull/2579) is in, we should upgrade from the current eslint@6 to eslint@7.   https://eslint.org/docs/user-guide/migrating-to-7.0.0

- drops node v8 support
- Deal with "no-var" rule. A quick run with the update to eslint-config-standard@16 results in ~2000 of these. I'm not sure if `--fix` will automatically fix these.
```
  167:7   warning  Unexpected var, use let or const instead  no-var
```

Note: Why not eslint@8 yet? The current latest is eslint@8, but eslint-config-standard and standard aren't ready for eslint@8 until eslint-config-standard@17 (https://github.com/standard/standard/blob/master/CHANGELOG.md#1700-0---2022-01-31), which is current in pre-release.

Update (2022-07-14): eslint-config-standard@17 is released now, so I believe moving to eslint@8 would be possible now. See https://github.com/elastic/apm-agent-nodejs/pull/2820 and https://github.com/elastic/apm-agent-nodejs/pull/2817

jmetev1:
fixes #2583 
I updated linting packages. And of course that comes with some changes. I broke up the autofixes vs manual fixes by commit. Autofix did some of the vars, but not all of them. What do you want to do with 144 remaining? 

Also i see you're hiring. Can i get a referral?

### Checklist

- [x] Implement code


package.json
```diff
@@ -150,13 +150,12 @@
...
-    "eslint": "^6.8.0",
-    "eslint-config-standard": "^14.1.1",
-    "eslint-plugin-import": "^2.25.4",
+    "eslint": "^8.42.0",
+    "eslint-config-standard": "^16",
+    "eslint-plugin-import": "^2.27.5",
     "eslint-plugin-license-header": "^0.6.0",
     "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-promise": "^4.3.1",
-    "eslint-plugin-standard": "^4.1.0",
```
jmetev1, line 159: no longer needed with this eslint-config-standard

jmetev1:
i have now signed cla

trentm:
Hi @jmetev1. Thanks very much for the PR.  There are a lot of changes here, including lint/style changes, so I have a fair amount of feedback. 
Thank you for splitting manual and auto changes between commits.

# no-var

(https://github.com/standard/standard/issues/633)

We discussed internally for a bit, and we'd like to keep this off. I.e. we'd like to *not* take the large set of changes to switch to `const/let`. The exception is with the "examples/" dir, where we'll apply the rule and change current `var` usages over.  I have a patch below that explains our reasoning.


# array-callback-return

(https://github.com/standard/standard/issues/859)

I'd like to keep this set to "error", and then revisit the few places we hit this:

```
./lib/instrumentation/modules/graphql.js
  175:33  error  Expected to return a value at the end of function  array-callback-return
  176:83  error  Function expected a return value                   array-callback-return
  177:42  error  Function expected a return value                   array-callback-return

./lib/redact-secrets/index.js
  41:30  error  Expected to return a value in function  array-callback-return

./test/_utils.js
  38:12  error  Expected to return a value at the end of function  array-callback-return
```


# Coordinating

1. Here is a suggested starter patch for setting "no-var" and a couple other small changes: https://gist.github.com/trentm/d610966e3cdd02ec8ff810f1b30da511
   If you would prefer, I can make commits to your feature branch. Please let me know.

2. Then your commit 1727021 with auto lint fixes should be reverted.

3. Then there will be some autofixes for "no-var" just in the "examples/" subdir to apply.

4. Then we can work through the array-callback-return issues. I am happy to work through those.  The one in redact-secrets is the only one I have to poke around to relearn.

jmetev1:
I'm afraid with this patch you've gone out of my git depth. I tried to google but am unsure if this is something I'm supposed to be able to apply automatically. But feel free to update my branch.

trentm:
> But feel free to update my branch.

Will do. Thanks.

trentm:
> but i am curious, what was i supposed to do with your gist? i was looking around here, but had no luck https://gist.github.com/nepsilon/22bc62a23f785716705c

That gist has it right. You could have need to download my gist to a file (using the "Raw" link): https://gist.githubusercontent.com/trentm/d610966e3cdd02ec8ff810f1b30da511/raw/10433a76f380c9724b932fe0ef9205dc97940375/the.patch
and then applied that to your git clone with `git apply path/to/downloaded/the.patch`.

jmetev1:
ohhhh i forget about raw option. i tried that but without raw part.

trentm:

trentm:

trentm (approved):

trentm:
Thanks very much @jmetev1

jmetev1:
no problem

PR #3410, opened 2023-06-09
fix: do not start the Agent in "start.js" when in a Worker thread

trentm:
This avoids the, arguably, surprise case of an additional APM agent
being started whenever a new Worker is created (see Node.js
`worker_thread` docs) if `node -r elastic-apm-node/start.js ...`
or `import 'elastic-apm-node/start.js'` is being used.

The former is especially surprising when using a loader with node v20:
    node --experimental-loader=elastic-apm-node/loader.mjs -r elastic-apm-node/start.js ...
because Node v20 has moved loader processing to a Worker... so the
"start.js" is executed again in that worker. The visible side-effect
is that the logging preamble is printed twice.

This change isn't ideal because an explicit use of
    import 'elastic-apm-node/start.js'
in TypeScript or ESM code for a worker does *not* start the APM agent.
That's a subtlety that we rely on documentation to explain away.

---

This is the first of a couple smaller PRs extracting changes from my ESM PR (https://github.com/elastic/apm-agent-nodejs/pull/3381).

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -56,6 +56,14 @@ const {
   normalizeSpanStackTraceMinDuration
 } = require('./normalizers')
 
+let threadId
+try {
+  const workerThreads = require('worker_threads')
+  threadId = workerThreads.threadId
+} catch (_err) {
+  // pass
+}
+
 const EXCLUDE_FIELDS = {
   logger: true,
   transport: true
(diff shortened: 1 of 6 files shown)


lib/config/config.js
```diff
@@ -56,6 +56,14 @@ const {
   normalizeSpanStackTraceMinDuration
 } = require('./normalizers')
 
+let threadId
+try {
```
david-luna, line 60: nit: Both `isMainThread` & `threadId` properties were added in v10.5.0 so what's your opinion on checking on the version

```javascript
const threadsSupported = semver.gte(process.version, '10.5.0')
let threadId
if (threadsSupported) {
  threadId = require('worker_threads').threadId
}
```

if you prefer the try/catch approach I would drop a comment telling why it may throw

lib/config/config.js
```diff
@@ -483,12 +491,12 @@ function loadConfigFile (filePath) {
  * @returns {Object} Object with agent, environment, and configuration data.
  */
 function getLoggingPreambleData (config, sources, configFilePath) {
-  // const optsFromSources = sources.reduce((prev, s) => Object.assign({}, s.opts, prev), {})
   const optsFromSources = sources.reduceRight((prev, s) => Object.assign(prev, s.opts), {})
   const result = {
     agentVersion: AGENT_VERSION,
     env: {
       pid: process.pid,
+      tid: threadId, // worker_threads docs do *not* specify if the main thread is always tid 0
```
david-luna, line 499: nit: in logging preamble we discussed the size of it. So maybe avoid adding the property if it's undefined would be nice

david-luna (approved):

lib/config/config.js
trentm: Good point about the size of the preamble. Given that I expect very little worker thread usage, and even less a Worker with the APM agent running in it, this `tid` field will rarely be useful. I'm inclined to not add it for now.

Issue #3428, opened 2023-06-13
Start supporting `response.writableEnded` in transaction's context


david-luna:
According to [nodejs docs](https://nodejs.org/api/http.html#responsefinished) `response.finished` property is deprecated since: v13.4.0, v12.16.0. There is another property recommended by the docs named `writableEnded`. Agent is relying on this property to fill the `transaction.contex.response.finished` field.

Although latest version v20.3.0 still supports it agent should anticipate the removal of that property and use the new one when needed. So some expression like the one below may be enough

```javascript
context.finshed = res.finished || res.writableEnded
```

thesixpathguy:
Hi @david-luna, I am interested to work on the issue. Can you please assign it to me?

PR #3430, opened 2023-06-14
fix: Started Supporting response.writableEnded in transaction's context.

thesixpathguy:
Since response.finished property is deprecated as of the newer node versions and another property i.e. response.writableEnded is currently used in place of it. So started supporting the response.writableEnded in transaction's context. Also, the change is backward compatible as the response.finished is not removed yet. 
Fixes: #3428

### Checklist

- [x] Implement code
- [ ] Update documentation

### DIFF
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -115,7 +115,7 @@ function getContextFromResponse (res, conf, isError) {
 
   if (isError) {
     context.headers_sent = res.headersSent
-    context.finished = res.finished
+    context.finished = res.finished || res.writableEnded
   }
 
   return context
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -115,7 +115,11 @@ function getContextFromResponse (res, conf, isError) {
 
   if (isError) {
     context.headers_sent = res.headersSent
-    context.finished = Boolean(res.finished || res.writableEnded)
+    if (typeof res.finished === 'boolean') {
+      context.finished = res.finished
+    } else {
+      context.finished = res.writableEnded
+    }
   }
 
   return context
--- a/lib/parsers.js
+++ b/lib/parsers.js
@@ -115,7 +115,7 @@ function getContextFromResponse (res, conf, isError) {
 
   if (isError) {
     context.headers_sent = res.headersSent
-    context.finished = res.finished || res.writableEnded
+    context.finished = Boolean(res.finished || res.writableEnded)
   }
 
   return context


lib/parsers.js
```diff
@@ -115,7 +115,7 @@ function getContextFromResponse (res, conf, isError) {
 
   if (isError) {
     context.headers_sent = res.headersSent
-    context.finished = res.finished
+    context.finished = res.finished || res.writableEnded
```
david-luna, line 118: @thesixpathguy 

There are some tests failing with the following message for node v12 and below

```
not ok 567 should be deeply equivalent
  ---
    operator: deepEqual
    expected: |-
      { status_code: 204, headers: { 'content-length': '4', string: 'foo', number: '42', array: [ 'foo', '42' ], password: '[REDACTED]' }, headers_sent: false, finished: false }
    actual: |-
      { status_code: 204, headers: { 'content-length': '4', string: 'foo', number: '42', array: [ 'foo', '42' ], password: '[REDACTED]' }, headers_sent: false }
    at: <anonymous> (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/test/agent.test.js:1793:11)
```

This is due to `writableEnded` not being defined on those versions therefore when response is not finished the expression becomes
```javascript
context.finished = false || undefined  // the result of the expression is undefined
```
then at serialisation of the span the process omits undefined properties to save bandwidth.

We can wrap the expression with a `Boolean` constructor to make sure we have the right type.
```suggestion
    context.finished = Boolean(res.finished || res.writableEnded)
```

david-luna (changes requested):

lib/parsers.js
```diff
@@ -115,7 +115,7 @@ function getContextFromResponse (res, conf, isError) {
 
   if (isError) {
     context.headers_sent = res.headersSent
-    context.finished = res.finished
+    context.finished = Boolean(res.finished || res.writableEnded)
```
david-luna, line 118: We hit another test failure, this time in a different scenario. The failing tests shows this log

```
not ok 6 pseudo-res testCase 1
  ---
    operator: deepEqual
    expected: |-
      { status_code: 500, headers: {}, headers_sent: undefined, finished: undefined }
    actual: |-
      { status_code: 500, headers: {}, headers_sent: undefined, finished: false }
    at: <anonymous> (/home/runner/work/apm-agent-nodejs/apm-agent-nodejs/test/parsers.test.js:116:9)
```

There is an scenario we did not take in consideration. AOM agent can run not only on HTTP servers but also in other contexts like AWS Lambda. In such scenarios agent does not get an HttpResponse but an object with a similar shape. That object does not have `finished` or `writableEnded` therefore the `context.finished` shouldn't have a value so we need to drop this one liner for something with the following behaviour
- if both properties are `undefined` => context.finished = undefined
- if any of the properties is `boolean` => context.finished = the_boolean_value

I'll drop a suggestion but do not feel obligated to apply it directly, we can have a discussion :) 

```suggestion
    if (typeof res.finished === 'boolean') {
      context.finished = res.finished
    } else {
      context.finished = res.writableEnded
    }
```

david-luna (changes requested):

lib/parsers.js
thesixpathguy: Hey @david-luna, please tell me if I understand it correctly. 
1. If `res.finished` is of type boolean then either `res.writableEnded` will be the same boolean value or `undefined`. So anyways it will evaluate to `res.finished`. 
2. I am assuming that if `res.finished` is `undefined` then `res.writableEnded` will also be `undefined` by default because it will not be an HTTP server then. Is this assumption correct?
3. If we come to "else" statement, then we have `res.finished` as `undefined` and this implies from 2nd statement that `res.writableEnded` is also `undefined`. So can't we directly evaluate the expression to `undefined` over here. 


And hence the compiling above three points, I am thinking something like this:
```suggestion
    if (typeof res.finished === 'boolean') {
      context.finished = Boolean(res.finished || res.writableEnded)
    } else {
      context.finished = undefined
    }
    
```

david-luna: Hi @thesixpathguy 

sorry for the delay on the feedback. I was off for a week

> If `res.finished` is of type boolean then either `res.writableEnded` will be the same boolean value or `undefined`. So anyways it will evaluate to `res.finished`.

So if `finished === false` and `writableEnded === undefined` we got the expression `false || undefined` resulting in `undefined` and this is not what we need. Using the `Boolean` constructor solves this but if we already know that `res.finished` contains a valid value we could use it directly.

> I am assuming that if `res.finished` is `undefined` then `res.writableEnded` will also be `undefined` by default because it will not be an HTTP server then. Is this assumption correct?

At this moment your assumption is good but a future node release will remove `finished` and `writableEnded` will be the only property available in the HTTP response object. The issue is about covering this possible scenario

The code I've suggested before was hiding some context by merging the new scenario (only writableEnded defined) with the scenario of non HTTP Responses (lambda instrumentation). We can use some extra keystrokes to be even more explicit

```javascript
if (typeof res.finished === 'boolean') {
    // Scenario: versions of node with finished defined, we can still use it
    context.finished = res.finished
} else if (typeof res.writableEnded === 'boolean') {
    // Scenario: new versions of node with finished removed, we use writableEnded
    context.finished = res.writableEnded
}
// If no branch was executed there is no need to set finished property for context
// because it was a response like object from lambda instrumentation
```

thesixpathguy: @david-luna 
Oh, I see. 
Yup, the above code seems correct and concise. Let me implement that one.

david-luna (approved):

david-luna:

Issue #1830, opened 2020-10-15
HTTP2 Instrumentation and Distributed Tracing


astorm:
**Is your feature request related to a problem? Please describe.**

Our HTTP 2 instrumentation does not support distributed tracing -- transactions [are started without passing in a TraceParent or TraceState header](https://github.com/elastic/apm-agent-nodejs/blob/4fe301d66a1ef748a77a0cde7584931e176c618b/lib/instrumentation/modules/http2.js#L51) 

This is a placeholder issue to gauge interest in such a feature. 

**Describe the solution you'd like**

For incoming HTTP2 servers to scan for traceparent and tracestate headers, and for outgoing http2 requests to include these headers. 

**Describe alternatives you've considered**

Manually starting a transaction [is an inelegant](https://www.elastic.co/guide/en/apm/agent/nodejs/master/distributed-tracing.html#tracing-custom-protocol) solution that could work.

**Additional context**

PR #3451, opened 2023-06-29
fix: set the transaction parent for an incoming HTTP/2 request

paulish:
With distributed tracing incoming http2 requests have no parent ids. This behavior can be observed using grpc/js module which is using http2 transport.

The change extracts parent ids and state from the request headers the similar way as it is done for http/1.

This fixes https://github.com/elastic/apm-agent-nodejs/issues/1830

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


trentm:

trentm:
@paulish  I don't have permission to push a test case to your repo. (I'm guessing that's because it is your "main" branch instead of a separate feature branch.   I'll merge this as is, then follow up with an added test case.  Thanks.

Issue #3464, opened 2023-07-04
`Cannot read properties of undefined (reading 'hostname')` in @aws-sdk/client-s3 instrumentation


trentm:
In https://github.com/elastic/apm-agent-nodejs/issues/3454#issuecomment-1619569176 it was reported that instrumentation of `@aws-sdk/client-s3` could throw:

> I'm not sure if directly related to your test failures but the upgrade of elastic-apm-node from 3.45.0 to 3.47.0 broke S3 interactions like generating pre-signed URLs, throwing:
>
> ```
> Cannot read properties of undefined (reading 'hostname')
> at (node_modules/elastic-apm-node/lib/instrumentation/modules/@aws-sdk/client-s3.js:155)
> ```
>
> Using client-s3 version 3.332.0.

---

@AndKiel Thanks for reporting this!   Instrumentation of `@aws-sdk/client-s3` was first added in v3.46.0 of this APM agent. I'm almost certain this is a separate issue from #3454, so I have moved it here.

**If you had a small reproduction script**, or even just a small snippet of JavaScript code showing the S3 Client API calls that hit this, that would be very helpful for me attempting to reproduce.

---

Our instrumentation is adding [aws-sdk v3 middleware](https://github.com/aws/aws-sdk-js-v3#middleware-stack) -- to the "initialize" and "finalizeRequest" steps. The error above could happen if the "finalizeRequest" step is *not run* for a particular client call.

trentm:
@AndKiel  For what it is worth, I cannot repro using:

```js
const {
  S3Client,
  GetObjectCommand,
} = require('@aws-sdk/client-s3')
const { getSignedUrl } = require('@aws-sdk/s3-request-presigner')
...
  command = new GetObjectCommand({ Bucket: bucketName, Key: key })
  const signedUrl = await getSignedUrl(s3Client, command, { expiresIn: 3600 })
  console.log('signedUrl: ', signedUrl)
...
```

I'm using the latest elastic-apm-node (the HEAD of the "main" git branch), and `@aws-sdk/client-s3@3.363.0` and `@aws-sdk/s3-request-presigner@3.363.0`.

AndKiel:
The full stack trace we got would be:

```
Cannot read properties of undefined (reading 'hostname')
  at (node_modules/elastic-apm-node/lib/instrumentation/modules/@aws-sdk/client-s3.js:155)
  at processTicksAndRejections (node:internal/process/task_queues:95)
[...]

PR #3467, opened 2023-07-05
fix: `@aws-sdk/client-s3` instrumentation could cause `getSignedUrl` to crash

trentm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/3464

### DIFF
--- a/lib/instrumentation/modules/@aws-sdk/client-s3.js
+++ b/lib/instrumentation/modules/@aws-sdk/client-s3.js
@@ -61,14 +61,15 @@ function s3MiddlewareFactory (client, agent) {
   return [
     {
       middleware: (next, context) => async (args) => {
+        // Ensure there is a span from the wrapped `client.send()`.
+        const span = agent._instrumentation.currSpan()
+        if (!span || !(span.type === TYPE && span.subtype === SUBTYPE)) {
+          return await next(args)
+        }
+
         const input = args.input
         const bucket = input && input.Bucket
         const resource = resourceFromBucket(bucket)
-        const span = agent._instrumentation.currSpan()
-
-        if (!span) {
-          return await next(args)
-        }
         // The given span comes with the operation name and we need to
         // add the resource if applies
         if (resource) {
@@ -152,17 +153,18 @@ function s3MiddlewareFactory (client, agent) {
 
           // Destination context.
           const destContext = {
-            address: context[elasticAPMStash].hostname,
-            port: context[elasticAPMStash].port,
             service: {
               name: SUBTYPE,
               type: TYPE
             }
           }
+          if (context[elasticAPMStash]) {
+            destContext.address = context[elasticAPMStash].hostname
+            destContext.port = context[elasticAPMStash].port
+          }
           if (resource) {
             destContext.service.resource = resource
           }
-
           if (region) {
             destContext.cloud = { region }
           }
@@ -190,7 +192,6 @@ function s3MiddlewareFactory (client, agent) {
         }
 
         context[elasticAPMStash] = {
-          protocol: req.protocol,
           hostname: req.hostname,
           port
         }
--- a/test/instrumentation/modules/@aws-sdk/fixtures/use-client-s3.js
+++ b/test/instrumentation/modules/@aws-sdk/fixtures/use-client-s3.js
@@ -63,10 +63,33 @@ const {
   waitUntilBucketExists,
   waitUntilObjectExists
 } = require('@aws-sdk/client-s3')
+const { getSignedUrl } = require('@aws-sdk/s3-request-presigner')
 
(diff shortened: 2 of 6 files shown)


lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -61,14 +61,15 @@ function s3MiddlewareFactory (client, agent) {
...
     {
       middleware: (next, context) => async (args) => {
-        const input = args.input
-        const bucket = input && input.Bucket
-        const resource = resourceFromBucket(bucket)
+        // Ensure there is a span from the wrapped `client.send()`.
         const span = agent._instrumentation.currSpan()
-
-        if (!span) {
+        if (!span || !(span.type === TYPE && span.subtype === SUBTYPE)) {
```
trentm, line 66: Reviewer note: This is the fix for the first surprise. We do a slightly weak check that the current span is one likely created by our `client.send()` wrapper (in smithy-client.js).

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -152,17 +153,18 @@ function s3MiddlewareFactory (client, agent) {
...
-            port: context[elasticAPMStash].port,
             service: {
               name: SUBTYPE,
               type: TYPE
             }
           }
+          if (context[elasticAPMStash]) {
+            destContext.address = context[elasticAPMStash].hostname
+            destContext.port = context[elasticAPMStash].port
+          }
```
trentm, line 164: Reviewer note: This is the fix for the second surprise. It is possible for middleware (if it doesn't call `next()`) to result in our "finalizeRequest" middleware not being called, so `context[elasticAPMStack]` wouldn't be set.

lib/instrumentation/modules/@aws-sdk/client-s3.js
```diff
@@ -190,7 +192,6 @@ function s3MiddlewareFactory (client, agent) {
         }
 
         context[elasticAPMStash] = {
-          protocol: req.protocol,
```
trentm, line 193: Reviewer note: This field wasn't being used, so I dropped it.

test/instrumentation/modules/@aws-sdk/fixtures/use-client-s3.js
```diff
@@ -129,14 +152,30 @@ async function useClientS3 (s3Client, bucketName) {
   // https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/classes/getobjectcommand.html
   command = new GetObjectCommand({ Bucket: bucketName, Key: key })
   data = await s3Client.send(command)
-  log.info({ data }, 'getObject')
+  // `data.Body` is a *stream*, so we cannot just log it
+  // https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-s3/interfaces/getobjectcommandoutput.html#body
+  const body = await slurpStream(data.Body)
+  assert(body === content)
+  delete data.Body
+  log.info({ data, body }, 'getObject')
```
trentm, line 160: Reviewer note: This change is unrelated to the bug. With aws-sdk **v3** that `data.Body` is a Node.js *stream* object. Logging it results in an unhelpful and huge log record. This changes to fully read that stream so we can log the S3 object content instead.

test/instrumentation/modules/@aws-sdk/fixtures/use-client-s3.js
```diff
@@ -129,14 +152,30 @@ async function useClientS3 (s3Client, bucketName) {
...
+  // This is interesting to test, because `getSignedUrl` uses the command
+  // `middlewareStack` -- including our added middleware -- **without** calling
+  // `s3Client.send()`. The test here is to ensure this doesn't break.
+  const customSpan = apm.startSpan('get-signed-url')
+  const signedUrl = await getSignedUrl(
+    s3Client,
+    new GetObjectCommand({ Bucket: bucketName, Key: key }),
+    { expiresIn: 3600 })
+  log.info({ signedUrl }, 'getSignedUrl')
+  customSpan.end()
```
trentm, line 172: Reviewer note: Adding a test case. Before the fixes above, this would throw.

test/instrumentation/modules/@aws-sdk/fixtures/use-client-s3.js
```diff
@@ -228,7 +267,8 @@ function main () {
       s3Client.destroy()
       process.exitCode = 0
     },
-    function () {
+    function (err) {
+      apm.logger.error(err, 'useClientS3 rejected')
```
trentm, line 271: Reviewer note: Before this change the error, if any, was silently not logged -- making debugging more difficult.

david-luna (approved):

PR #3470, opened 2023-07-07
fix: identifying the "sp_prepare" stored procedure changed for tedious@16.2.0

trentm:
tedious@16.2.0 changed the "name" it uses internally for RpcRequestPayload objects, from a string to the SQL Server-defined numeric identifiers. For example, to identify a "prepare":

```diff
- new RpcRequestPayload('sp_prepare', ...);
+ new RpcRequestPayload(Procedures.Sp_Prepare, ...);
```

where that `Procedures.Sp_Prepare` is a hardcoded 11, corresponding to https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-prepare-transact-sql

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,24 @@ Notes:
 === Node.js Agent version 3.x
 
 
+==== Unreleased
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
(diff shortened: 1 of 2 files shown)


trentm (commented):
/test tav tedious

david-luna (approved):

Issue #2990, opened 2022-10-17
drop `opts.timer` option passed to Span or Transaction creation


trentm:
Delete any `opts.timer` option passed to Span or Transaction creation. While I don't expect anyone to have ever used this option, it is theoretically a breaking change, so should be left to a major ver bump.

https://github.com/elastic/apm-agent-nodejs/pull/2939 is adding a deprecation warning for passing in `opts.timer` to Span or Transaction creation. It *is* an internal option used to pass to the internal `GenericSpan` based class, but users should never want or need to use this option.

trentm:
FWIW the current warning looks like this:

```
> var s2 = apm.startSpan('s2', {timer: {}})
undefined
> (node:72351) [ELASTIC_APM_SPAN_TIMER_OPTION] DeprecationWarning: specifying the `timer` option to `new Span()` was never a public API and will be removed
```

trentm:
Closed by https://github.com/elastic/apm-agent-nodejs/pull/3537 on the "dev/4.x" branch, which will become the 4.x active branch soonish.

PR #3537, opened 2023-08-01
chore!: drop "timer" option to startTransaction/startSpan

trentm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/2990


david-luna (approved):
Should we warn about the agent ignoring/removing the option so app devs can drop it in their apps config?

trentm:
> can drop it in their apps config?

I might be misunderstanding you, but this isn't an app config thing. This could only impact a user if they were doing:

```
apm.startSpan('my-span-name', {timer: <some custom Timer object>})
```

where that `timer` option was never documented.

david-luna:
Thanks for the clarification :)

Issue #2682, opened 2022-05-09
s/npm install/npm ci/ in tests and builds


trentm:
We should use [`npm ci`](https://docs.npmjs.com/cli/v8/commands/npm-ci) where ever we are using `npm install` in our CI and tests:

```
% rg 'npm install' .ci test dev-utils
dev-utils/make-lambda-layer-zip.sh
36:npm install --global-style elastic-apm-node-*.tgz

test/script/docker/run_tests.sh
39:      npm install

.ci/Jenkinsfile
541:                bat 'npm install'

.ci/scripts/docker-test.sh
34:  until npm install; do
39:      printf "Retry of 'npm install' %s/%s exited %s, retrying in %s seconds...\n" "$count" "$retries" "$exit" "$wait" >&2

.ci/scripts/test_basic.sh
5:npm install

.ci/scripts/test_types_babel_esm.sh
5:npm install

.ci/scripts/prepare-benchmarks-env.sh
4:# to use this) and `npm install`.
29:npm install

.ci/Makefile
12:		npm install --global-style https://github.com/elastic/apm-agent-nodejs#$(BRANCH_NAME); \
```

Note that we can only do this with npm v6 and later, which means we have to wait until we drop node v8 support.

- [ ] Does `npm ci` support `--global-style` used for the lambda layer and docker container?

trentm:
> * Does `npm ci` support `--global-style` used for the lambda layer and docker container?

It appears, yes: https://docs.npmjs.com/cli/v8/commands/npm-ci#global-style

PR #3538, opened 2023-08-01
chore: use 'npm ci' in tests

trentm:
Closes: https://github.com/elastic/apm-agent-nodejs/issues/2682

### DIFF
--- a/test/opentelemetry-metrics/fixtures.test.js
+++ b/test/opentelemetry-metrics/fixtures.test.js
@@ -153,26 +153,26 @@ async function checkHasPrometheusMetrics(t) {
 
 // ---- tests
 
-// We need to `npm install` for a first test run.
+// We need to `npm ci` for a first test run.
 const haveNodeModules = fs.existsSync(path.join(fixturesDir, 'node_modules'));
 tape.test(
-  `setup: npm install (in ${fixturesDir})`,
+  `setup: npm ci (in ${fixturesDir})`,
   { skip: haveNodeModules },
   (t) => {
     const startTime = Date.now();
     exec(
-      'npm install',
+      'npm ci',
       {
         cwd: fixturesDir,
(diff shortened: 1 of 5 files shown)


test/opentelemetry-metrics/fixtures.test.js
```diff
@@ -153,26 +153,26 @@ async function checkHasPrometheusMetrics(t) {
 
 // ---- tests
 
-// We need to `npm install` for a first test run.
+// We need to `npm ci` for a first test run.
 const haveNodeModules = fs.existsSync(path.join(fixturesDir, 'node_modules'));
 tape.test(
-  `setup: npm install (in ${fixturesDir})`,
+  `setup: npm ci (in ${fixturesDir})`,
```
david-luna, line 159: There is a similar test in `activation-method.test.js`. Should we change it too?

https://github.com/elastic/apm-agent-nodejs/blob/bcd0fa684f72edc899cfe474da07e35dc6c92156/test/activation-method/activation-method.test.js#L28

david-luna (commented):
LGTM. We can merge as soon we answer that question about `activation-method` tests :)

test/opentelemetry-metrics/fixtures.test.js
trentm: That fixtures dir doesn't have a package-lock.json, so cannot use `npm ci`.  And the only dep there is the agent itself:

```
[./test/activation-method/fixtures (git:main)]
% cat .npmrc
package-lock=false

[./test/activation-method/fixtures (git:main)]
% cat package.json
{
  "name": "activation-method-fixtures",
  "version": "1.0.0",
  "private": true,
  "dependencies": {
    "elastic-apm-node": "file:../../.."
  }
}
```

So no need for `npm ci`.

PR #3547, opened 2023-08-03
chore: make agent destroy async

### LINKED ISSUE #3222: make `Agent.destroy()` async
Currently in index.d.ts:

```typescript
  export interface Agent {
    // ...
    destroy (): void;
  }
```

In the [coming OTel Metrics work](https://github.com/elastic/apm-agent-nodejs/pull/3152), `Agent.destroy()` will possibly call an async method to shutdown an OTel MeterProvider. This means that agent shutdown code can possibly be running after the sync `Agent.destroy()` method returns. That's unfortunate.

The fix is to make `Agent.destroy()` async -- i.e. returns a Promise -- then the user can `await` it. However, is changing that return type in the TypeScript types a breaking change? I'm not sure.

david-luna:
By making it async now the agent's users can a `await` it to make sure any OTel MeterProviders has finalised its shutdown process.

Closes #3222 

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


lib/agent.js
```diff
@@ -135,14 +135,11 @@ Agent.prototype.destroy = function () {
...
-      .shutdown({ timeoutMillis: 1000 })
-      .catch((reason) => {
-        this.logger.warn('failed to shutdown OTel MeterProvider:', reason);
-      });
+    try {
+      await this._otelMeterProvider.shutdown({ timeoutMillis: 1000 });
+    } catch (reason) {
+      this.logger.warn('failed to shutdown OTel MeterProvider:', reason);
+    }
     this._otelMetricsProvider = null;
```
trentm, line 143: I just noticed this var is misspelled.
```suggestion
    this._otelMeterProvider = null;
```

david-luna (commented):
/test tav

lib/agent.js
```diff
@@ -135,14 +135,11 @@ Agent.prototype.destroy = function () {
...
-    // Dev Note: It is unfortunate that `.destroy()` can return while this
-    // shutdown work might still be ongoing. See
-    // https://github.com/elastic/apm-agent-nodejs/issues/3222 to fix that.
-    this._otelMeterProvider
-      .shutdown({ timeoutMillis: 1000 })
-      .catch((reason) => {
-        this.logger.warn('failed to shutdown OTel MeterProvider:', reason);
-      });
+    try {
+      await this._otelMeterProvider.shutdown({ timeoutMillis: 1000 });
```
trentm, line 139: This isn't necessarily the job of *this* change, but here is a concern. This `.shutdown` call, effectively results in await'ing `ElasticApmMetricExporter.flush()` (https://github.com/elastic/apm-agent-nodejs/blob/d86e26a5bb452455550105c226ff1298dabb56db/lib/opentelemetry-metrics/ElasticApmMetricExporter.js#L159-L161), which calls `_agent.flush()`.

At the top of Agent._flush is:

```js
  if (!this._apmClient) {
    // Log an *err* to provide a stack for the user.
    const err = new Error('cannot flush agent before it is started');
    this.logger.warn({ err }, err.message);
    if (cb) {
      process.nextTick(cb);
    }
    return;
  }
```

and earlier in `Agent.destroy` we've done `this._apmClient = null`.   So now I am curious if we are getting this `'cannot flush agent before it is started'` log warning when attempting to `apm.destroy()`. If not, then why not?

So, I think TODOs here are:

1. Add a test for awaiting this `apm.destroy()`.
2. Understand why we are or are not getting that warning.  Note that we will only run this `this._otelMeterProvider.shutdown` if we have instrumented `@opentelemetry/sdk-metrics`, so I'm guessing we just haven't tested this.
3. Likely we want to move the `await this._otelMeterProvider.shutdown(...)` to the *top* of `Agent.prototype.destroy()`.

trentm:
> but here is a concern. 

So indeed this is an existing issue.  With this code:

```
// foo.js
const apm = require('elastic-apm-node');
const otel = require('@opentelemetry/api');

const meter = otel.metrics.getMeter('my-meter');
const numReqs = meter.createCounter('num_requests', {
  description: 'number of HTTP requests',
});

let interval = setInterval(() => {
  numReqs.add(1);
}, 1000);

setTimeout(async () => {
  console.log('XXX destroy and exit');
  clearInterval(interval);
  await apm.destroy();
  console.log('XXX destroyed');
}, 5000);
```

Running that results in the warning about "cannot flush" that I mentioned above:

```
$ node -r elastic-apm-node/start.js foo.js | ecslog
...
XXX destroy and exit
XXX destroy: start
[2023-08-04T19:10:12.850Z]  WARN (elastic-apm-node): cannot flush agent before it is started
    error: {
        "type": "Error",
        "message": "cannot flush agent before it is started",
        "stack_trace":
            Error: cannot flush agent before it is started
                at Agent._flush (/Users/trentm/el/apm-agent-nodejs6/lib/agent.js:850:17)
                at /Users/trentm/el/apm-agent-nodejs6/lib/agent.js:822:12
                at new Promise (<anonymous>)
                at Agent.flush (/Users/trentm/el/apm-agent-nodejs6/lib/agent.js:821:12)
                at ElasticApmMetricExporter.shutdown (/Users/trentm/el/apm-agent-nodejs6/lib/opentelemetry-metrics/ElasticApmMetricExporter.js:160:24)
                at PeriodicExportingMetricReader.onShutdown (/Users/trentm/el/apm-agent-nodejs6/node_modules/@opentelemetry/sdk-metrics/build/src/export/PeriodicExportingMetricReader.js:102:30)
                at PeriodicExportingMetricReader.shutdown (/Users/trentm/el/apm-agent-nodejs6/node_modules/@opentelemetry/sdk-metrics/build/src/export/MetricReader.js:102:53)
                at MetricCollector.shutdown (/Users/trentm/el/apm-agent-nodejs6/node_modules/@opentelemetry/sdk-metrics/build/src/state/MetricCollector.js:53:34)
                at /Users/trentm/el/apm-agent-nodejs6/node_modules/@opentelemetry/sdk-metrics/build/src/MeterProvider.js:77:30
                at Array.map (<anonymous>)
                at MeterProvider.shutdown (/Users/trentm/el/apm-agent-nodejs6/node_modules/@opentelemetry/sdk-metrics/build/src/MeterProvider.js:76:62)
                at Agent.destroy (/Users/trentm/el/apm-agent-nodejs6/lib/agent.js:140:37)
                at Timeout._onTimeout (/Users/trentm/el/apm-agent-nodejs6/examples/opentelemetry-metrics/foo.js:22:13)
    }
XXX destroyed
```

trentm (approved):

PR #3553, opened 2023-08-04
chore: warn for missing units in config and for `errorMessagesMaxLength` usage

### LINKED ISSUE #2121: warn on missing unit for duration/size config vars, require unit on new config vars
There was a discussion recently and suggestion from @felixbarny that the APM agents make the units on time/duration config vars (e.g. `metricsInterval`) *required* instead of optional.  The history there is that some Java agent config vars that were later shared with other agents (e.g. `span_frames_min_duration`) used to have the units in the name (e.g. `span_frames_min_duration_ms`). The switch to drop the units in the name added support for units in the value, but for backward compatibility supported default units.

**The problem:** There are some cases where the user can get surprised:
- different default units between agents (e.g. the recently added `spanFramesMinDuration` in the node.js agent defaults to *seconds* with no units, but the default is *milliseconds* for other agents)
- the default units are not obvious to the user doing the configuration

Making the units required is obviously a breaking change, so would need to wait for a major version bump.

(See also the related #2120 for handling invalid config values.)

### LINKED ISSUE #2314: warn on `errorMessageMaxLength` deprecated config option usage
The `errorMessageMaxLength` option was deprecated by `longFieldMaxLength` in #2193. The option should be removed for the next major version.

Update (2023-07): Let's just warn with `process.emitWarning` (... or log.warn?)

david-luna:
Spit out a warning whenever a size/bytes config option is given without its units. The log shows the value provided and the name of the option.

This PR also adds a warning log if the deprecated `errorMessagesMaxLength` is used.

Closes #2121 
Closes #2314 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation

### DIFF
--- a/docs/upgrade-to-v4.asciidoc
+++ b/docs/upgrade-to-v4.asciidoc
@@ -85,7 +85,7 @@ unit, to avoid ambiguity the APM agent will now warn if the units are not
 provided.
 
 [[v4-warning-size-units]]
-=====
+===== "units missing in size value" 
 
 Byte size config options like `apiRequestSize` expect to have the size
 units specified in the value (e.g. `"10kb"`, `"1gb"`). If the unit is not
--- a/docs/upgrade-to-v4.asciidoc
+++ b/docs/upgrade-to-v4.asciidoc
@@ -70,7 +70,7 @@ For log correlation with _structured_ logs, see <<log-correlation-ids>>.
 This section documents some new log output warnings from the APM agent, and how to avoid them.
 
 [[v4-warning-duration-units]]
-=====
+===== "units missing in duration value"
 
 
 [source,json]
--- a/docs/upgrade-to-v4.asciidoc
+++ b/docs/upgrade-to-v4.asciidoc
@@ -81,6 +81,6 @@ the default value.
 [[v4-warning-size-units]]
 =====
 
-`apiRequestSize` and `errorMessageMaxLength` config options expect to
-have the size specified in the value (e.g. `10kb`, `1gb`). If the unit
-is not in the value tha agent will warn about it and fallback to `bytes`.
+Byte size config options like `apiRequestSize` expect to have the size
+units specified in the value (e.g. `10kb`, `1gb`). If the unit is not
+in the value tha agent will warn about it and fallback to `bytes`.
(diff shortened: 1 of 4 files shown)


docs/upgrade-to-v4.asciidoc
```diff
@@ -59,7 +59,34 @@ via:
...
+
+Byte size config options like `apiRequestSize` expect to have the size
+units specified in the value (e.g. `10kb`, `1gb`). If the unit is not
+in the value tha agent will warn about it and fallback to `bytes`.
+
+[[v4-warning-error-message-max-length]]
+=====
+
+Agent will warn if the deprecated config option `errorMessageMaxLength`
+is set
```
trentm, line 92:
```suggestion
```

I know this looks inconsistent. :)  I'm leaning towards not documenting *all* added new warnings. Only interesting ones.  A warning for a deprecated config option should be self-explanatory.  Sorry for getting you to write more docs.

trentm (approved):
Just some suggestions for the docs.

docs/upgrade-to-v4.asciidoc
```diff
@@ -59,7 +59,34 @@ via:
...
+----
+
+Configuration options that define a duration, like `metricsInterval` or 
+`exitSpanMinDuration`, expect to have their units specified in the value
+(e.g. `"10s"`, `"100ms"`). While current duration options have a default
+unit, to avoid ambiguity the APM agent will now warn if the units are not
+provided.
+
+[[v4-warning-size-units]]
+=====
```
trentm, line 88:
```suggestion
===== "units missing in size value" 
```

lib/config/normalizers.js
```diff
@@ -86,11 +86,20 @@ function normalizeInfinity(opts, fields, defaults, logger) {
...
+ * @param {import('../logging.js').Logger} logger
  * @returns {Number|undefined}
  */
-function bytes(input) {
+function bytes(input, key, logger) {
   const matches = input.match(/^(\d+)(b|kb|mb|gb)$/i);
-  if (!matches) return Number(input);
+  if (!matches) {
+    logger.warn(
+      'units missing in value "%s" for "%s" config option. Use one of b|kb|mb|gb',
```
trentm, line 97:
```suggestion
      'units missing in size value "%s" for "%s" config option. Use one of b|kb|mb|gb',
```

docs/upgrade-to-v4.asciidoc
trentm: Done in commit b65305c2f80ff980e81cb57fd9b4e9866e96ff80.

PR #3554, opened 2023-08-04
chore: return noop transaction if agent not started

### LINKED ISSUE #2429: `apm.startTransaction()` should return a NoopTransaction if the Agent is not yet started
See discussion on #2425.  It was decided there that, while the TypeScript types in `index.d.ts` allow `apm.startTransaction()` returning null is some cases, it could be a breaking change to JS users of `apm.startTransaction(...)` that did not handle this case -- especially because the Agent API docs for `startTransaction` don't mention the possibility of getting `null`.

Update: [This comment](https://github.com/elastic/apm-agent-nodejs/issues/2425#issuecomment-963594136) has the best description for why we are dropping `| null` from the return type of `apm.startTransaction()`.

david-luna:
This PR adds a check in the `agent.startTransaction()` API which looks if the agent has been started. If so it lets the method run otherwise returns a noop transaction (same a approach as OTel does).

Closes #2429 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/docs/upgrade-to-v4.asciidoc
+++ b/docs/upgrade-to-v4.asciidoc
@@ -67,6 +67,9 @@ For log correlation with _structured_ logs, see <<log-correlation-ids>>.
 [[v4-api-start-transaction]]
 ===== `agent.startTransaction()`
 
-The `agent.startTransaction()` method has been changed to return an invalid
-or noop transaction if the agent is not yet started.
+The `agent.startTransaction()` method has been changed to return a do-nothing
+no-op Transaction, if the agent is not yet started. The return type has changed to
+no longer include `| null`. The intent of these changes is to allow the user to use
+`.startTransaction()` without having to worry if the agent is yet started, nor to
+have to handle a possible `null` return value.
 
--- a/docs/upgrade-to-v4.asciidoc
+++ b/docs/upgrade-to-v4.asciidoc
@@ -63,3 +63,10 @@ console.log( stringify(span.ids, ' ', '=')) );
 ----
 
 For log correlation with _structured_ logs, see <<log-correlation-ids>>.
+
+[[v4-api-start-transaction]]
+===== `agent.startTransaction()`
+
+The `agent.startTransaction()` method has been changed to return an invalid
+or noop transaction if the agent is not yet started.
+
--- a/lib/instrumentation/noop-transaction.js
+++ b/lib/instrumentation/noop-transaction.js
@@ -19,7 +19,7 @@ const NOOP_TRACEPARENT = '00-00000000000000000000000000000-0000000000000000-00';
 function NoopTransaction() {
   // Public properties:
   // https://www.elastic.co/guide/en/apm/agent/nodejs/current/transaction-api.html
-  this.name = 'Noop Transaction';
+  this.name = 'unnamed';
   this.type = 'noop';
   this.subtype = undefined;
   this.action = undefined;
--- a/lib/instrumentation/noop-transaction.js
+++ b/lib/instrumentation/noop-transaction.js
@@ -8,70 +8,71 @@
 
 const constants = require('../constants');
 
-module.exports = NoopTransaction;
-
 const NOOP_TRANSACTION_ID = '0000000000000000';
 const NOOP_TRACEID = '00000000000000000000000000000000';
 const NOOP_TRACEPARENT = '00-00000000000000000000000000000-0000000000000000-00';
 
-// Usage:
-//    new NoopTransaction()
-function NoopTransaction() {
-  // Public properties:
-  // https://www.elastic.co/guide/en/apm/agent/nodejs/current/transaction-api.html
-  this.name = 'unnamed';
-  this.type = 'noop';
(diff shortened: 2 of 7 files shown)


david-luna:
Possible spurious test failing on Windows for `http-client` module

```
not ok 1066 should have received 2 requests completely
  ---
    operator: equal
    expected: 2
    actual:   1
    at: Timeout._onTimeout (D:\a\apm-agent-nodejs\apm-agent-nodejs\test\apm-client\http-apm-client\basic.test.js:344:11)
    stack: |-
      Error: should have received 2 requests completely
          at Test.assert [as _assert] (D:\a\apm-agent-nodejs\apm-agent-nodejs\node_modules\tape\lib\test.js:309:48)
          at Test.strictEqual (D:\a\apm-agent-nodejs\apm-agent-nodejs\node_modules\tape\lib\test.js:473:7)
          at Timeout._onTimeout (D:\a\apm-agent-nodejs\apm-agent-nodejs\test\apm-client\http-apm-client\basic.test.js:344:11)
```

docs/upgrade-to-v4.asciidoc
```diff
@@ -63,3 +63,10 @@ console.log( stringify(span.ids, ' ', '=')) );
 ----
 
 For log correlation with _structured_ logs, see <<log-correlation-ids>>.
+
+[[v4-api-start-transaction]]
+===== `agent.startTransaction()`
+
+The `agent.startTransaction()` method has been changed to return an invalid
+or noop transaction if the agent is not yet started.
```
trentm, line 71:
```suggestion
The `agent.startTransaction()` method has been changed to return an do-nothing
no-op Transaction, if the agent is not yet started. Before this change it returned a
Transaction object that could result in API errors. The return type has changed to
no longer include `| null`.
```

^^ Does that description work?

lib/instrumentation/noop-transaction.js
```diff
@@ -0,0 +1,77 @@
...
+// Public methods:
+// https://www.elastic.co/guide/en/apm/agent/nodejs/current/transaction-api.html
+NoopTransaction.prototype.setType = returnVoid;
+NoopTransaction.prototype.setLabel = returnTrue;
+NoopTransaction.prototype.addLabels = returnTrue;
+NoopTransaction.prototype.setOutcome = returnVoid;
+NoopTransaction.prototype.startSpan = returnNull;
+NoopTransaction.prototype.end = returnVoid;
+NoopTransaction.prototype.ensureParentId = function () {
+  return '00000000';
```
trentm, line 51:
```
> var t = apm.startTransaction('t1')
undefined
> t.ensureParentId()
'dccd55f81e5a251c'
```

I think the retval of `ensureParentId` needs to be 8-bytes (16 hex chars).
It could re-use `NOOP_TRANSACTION_ID`.

```suggestion
  return NOOP_TRANSACTION_ID;
```

lib/instrumentation/noop-transaction.js
```diff
@@ -0,0 +1,77 @@
...
+
+module.exports = NoopTransaction;
+
+const NOOP_TRANSACTION_ID = '0000000000000000';
+const NOOP_TRACEID = '00000000000000000000000000000000';
+const NOOP_TRACEPARENT = '00-00000000000000000000000000000-0000000000000000-00';
+
+// Usage:
+//    new NoopTransaction()
+function NoopTransaction() {
```
trentm, line 19: nit: We *could* use `class NoopTransaction`. Either is totally fine.

lib/instrumentation/noop-transaction.js
```diff
@@ -0,0 +1,77 @@
...
+ * Copyright Elasticsearch B.V. and other contributors where applicable.
+ * Licensed under the BSD 2-Clause License; you may not use this file except in
+ * compliance with the BSD 2-Clause License.
+ */
+
+'use strict';
+
+const constants = require('../constants');
+
+module.exports = NoopTransaction;
```
trentm, line 11: nit: Would you object to this being a named export, rather than a default export? I know that existing files of ours use default exports, but I've been trying to move away from them. I don't have a strong argument against default exports other than when mixing CommonJS and ESM, default exports can be problematic. Obviously that argument doesn't apply in this case.

```suggestion
module.exports = {
  NoopTransaction
};
```

nit(style): My preference is for exports to be listed at the bottom. I guess someday I'll need to adapt to ESM-style `export ...` statements throughout the module. :)

trentm (changes requested):

trentm:
> I wonder if we should update `agent-api.asciidoc` to mention about this specific return value.

Yes, that is a good idea.

lib/instrumentation/noop-transaction.js
trentm: Done in commit cf60004d.

lib/instrumentation/noop-transaction.js
trentm: Done in commit cf60004d

trentm (approved):

PR #3555, opened 2023-08-04
chore: set `useElasticTraceparentHeader`'s default value to `false`

### LINKED ISSUE #3510: change `useElasticTraceparentHeader` to default to false
In 3.x versions the `useElasticTraceparentHeader` (https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#use-elastic-traceparent-header) defaults to *true*, which means that by default the APM agent will add the `elastic-apm-traceparent` header to outgoing requests -- along with the usual W3C `traceparent` and `tracestate` headers. This is a legacy header now that the W3C trace-context headers are well-established.

The new behaviour, intended for 4.x, is that this header will no longer be set on outgoing requests by default. The get the old default behaviour, one must explicitly set the config var.

Note that this config var never impacted the *reading* of an `elastic-apm-traceparent` header on incoming requests. That header will still be used as a fallback of `traceparent` is not set.

- [ ] Also remove the few examples where we explicitly set `useElasticTraceparentHeader: false`.

david-luna:
As part of the config changes we're doing for next major version `4.x` we're disabling the elastic traceparent header by default. This PR does that change and also updates its tests & docs accordingly.

Closes #3510

### Checklist

- [x] Implement code
- [x] Update tests
- [x] Update documentation

### DIFF
--- a/CHANGELOG4.asciidoc
+++ b/CHANGELOG4.asciidoc
@@ -27,8 +27,6 @@
 * Remove instrumentation support for the old 'hapi' package -- the current
   '@hapi/hapi' package is still instrumented. ({issues}2691[#2691])
 
-* Change default value of `useElasticTraceparentHeader` config option to `false`.
-
 [float]
 ===== Features
 
@@ -40,6 +38,8 @@
 
 * Add a warning message when a duration config option is provided
   without units. ({issues}2121[#2121])
+
+* Change default value of `useElasticTraceparentHeader` config option to `false`.
 
 [[release-notes-3.x]]
 === Node.js Agent version 3.x
--- a/CHANGELOG4.asciidoc
+++ b/CHANGELOG4.asciidoc
@@ -27,6 +27,8 @@
 * Remove instrumentation support for the old 'hapi' package -- the current
   '@hapi/hapi' package is still instrumented. ({issues}2691[#2691])
 
+* Change default value of `useElasticTraceparentHeader` config option to `false`.
+
 [float]
 ===== Features
 
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -1423,7 +1423,7 @@ The captured request body (if any) is stored on the `span.db.statement` field. C
 ==== `useElasticTraceparentHeader`
 
 * *Type:* Boolean
-* *Default:* `true`
+* *Default:* `false`
 * *Env:* `ELASTIC_APM_USE_ELASTIC_TRACEPARENT_HEADER`
 
 To enable {apm-guide-ref}/apm-distributed-tracing.html[distributed tracing], the agent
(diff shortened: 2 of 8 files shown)


CHANGELOG4.asciidoc
```diff
@@ -39,6 +39,8 @@
 * Add a warning message when a duration config option is provided
   without units. ({issues}2121[#2121])
 
+* Change default value of `useElasticTraceparentHeader` config option to `false`.
```
trentm, line 42:
```suggestion
* Change default value of `useElasticTraceparentHeader` config option to `false`.
  This means that for outgoing HTTP requests, the APM agent will no longer add the
  `elastic-apm-traceparent` header. This vendor-specific header was used in the past
  while the https://w3c.github.io/trace-context/[W3C trace-context] spec was still
  in development. Now that it is in wide use, the `elastic-apm-traceparent` header is
  only useful for interaction with very old Elastic APM agents.
```

docs/configuration.asciidoc
```diff
@@ -1423,7 +1423,7 @@ The captured request body (if any) is stored on the `span.db.statement` field. C
 ==== `useElasticTraceparentHeader`
```
trentm, line 1424:
```suggestion

[small]#Change default in v4.0.0, in v3.x the default was `true`#

```

docs/upgrade-to-v4.asciidoc
```diff
@@ -34,6 +34,12 @@ Support for `filterHttpHeaders` config option has been removed. Redaction of
 HTTP headers and also request cookies is controlled by the existing config
 option <<sanitize-field-names, `sanitizeFieldNames`>>.
 
+===== `useElasticTraceparentHeader`
+
+The default value of `useElasticTraceparentHeader` config option has changed
+to `false` since the standard `traceparent` header is added in outgoing
+HTTP requests. If you want the agent to continue sending `elastic-apm-traceparent`
+HTTP header you can set to `true` via env var or start options.
```
trentm, line 42:
```suggestion
The default value of the <<use-elastic-traceparent-header>> config option has changed
to `false`. This means that the vendor-specific `elastic-apm-traceparent` header will no
longer by added to outgoing HTTP requests. The `traceparent` header (from the 
https://w3c.github.io/trace-context/[W3C trace-context standard]) is added by the
APM agent. If you want the agent to continue sending `elastic-apm-traceparent`
HTTP header you can set it to `true` via env var or start options.
```

test/instrumentation/modules/http2.test.js
```diff
@@ -602,8 +602,6 @@ function assertPath(t, trans, secure, port, path, httpVersion) {
   if (trans.context.request.headers.traceparent) {
     expectedReqHeaders.traceparent = trans.context.request.headers.traceparent;
     expectedReqHeaders.tracestate = trans.context.request.headers.tracestate;
-    expectedReqHeaders['elastic-apm-traceparent'] =
```
david-luna, line 605: Note for reviewer: http2 tests were assuming `useElasticTraceparent` default to `true`. I wonder if we have to refactor it so we tests both cases.

trentm: I am fine not bothering with a test of `useElasticTraceparentHeader: true` for http2. We have one for http tests and that's good enough.

trentm (approved):

Issue #3161, opened 2023-02-16
Mongoose spans are getting attached to wrong parent transaction


sefiros1:
**Describe the bug**

I'm using Koa with MongoDB via mongoose. With multiple simultaneous requests, mongoose spans begin to attach to the wrong parents. There is no such problem with one simultaneous request.

**To Reproduce**

Steps to reproduce the behavior: 

1. Start MongoDB server and run Koa server with mongoose (`node index.js`):
```js
// index.js
/* eslint-disable import/order */
const elasticApmNode = require('elastic-apm-node');
elasticApmNode.start({
	serverUrl: 'http://localhost:8200',
});

const mongoose = require('mongoose');
const Koa = require('koa');
const app = new Koa();

const ip = 'localhost';
const port = 27017;
const username = 'root';
const password = 'password';
const database = 'database';

async function bootstrap() {
	await mongoose.connect(`mongodb://${ip}:${port}/admin`, {
		auth: {
			username,
			password,
		},
		dbName: database,
	}).catch(e => {
		console.error('Mongo connection error', e);
		// eslint-disable-next-line no-process-exit
		process.exit(1);
	});

	const Cat = mongoose.model('Cat', { name: String });

	// response
	app.use(async ctx => {
		const { pathname } = ctx.request.URL;
		if (pathname === '/create') {
			const kitty = new Cat({ name: 'Zildjian' });
			await kitty.save();
			ctx.body = 'Meow';
		} else if (pathname === '/getAll') {
			ctx.body = await Cat.find();
		}
	});

	app.listen(3000);
	console.log('listening on 3000');
}

bootstrap();
```
2. Fill database with some data:
```bash
for i in `seq 1 500`; do curl http://localhost:3000/create; done
```
3. Run first simple test:
```bash
ab -n 10 -c 1 http://localhost:3000/getAll
```
4. Run second test with simultaneous requests:
```bash
ab -n 50 -c 50 http://localhost:3000/getAll
```

In kibana you should see normal traces for first test (step 3), but after second test (step 4) you will see
broken traces:
- some with no mongo spans at all
![Screenshot from 2023-02-16 12-59-14](https://user-images.githubusercontent.com/29738400/219335823-72b93a1a-1ed9-4378-9d13-8e010157a51b.png)

- some with too many spans
[...]

PR #3586, opened 2023-08-23
fix: remove double instrumentation of mongob commands

david-luna:
While doing investigations for #3161 some added logs revealed we were creating two spans for each command it was started. The root cause of this was because the instrumentation was registering the event listeners twice and as a result
- 2 spans where created
- the 2nd span overwrites the 1st on the map of active spans
- the 2nd span is ended but not the 1st

Since we are patching the `MongoClient#constructor` and also the static `connect` method whenever a user was calling `connect` to get a client this happens:
- the `patched connect` calls its original function
- the `original connect` method creates a new instance using `new MongoClient(...)`
- the `patched constructor` of MongoClient adds listeners
- the `patched connect` adds listeners to the client returned by its original function (duplication)

In conclusion there is no need to patch `connect` if already adding listeners in the constructor.

NOTE: this behavior started to happen from v14.11.0 when the creation of the instance changed from `new MongoClient(...)` to `new this(...)` making the patched constructor to run

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


david-luna (commented):
/test tav mongodb

david-luna (commented):
/test tav mongodb

david-luna (commented):
/test tav mongodb

david-luna (commented):
/test tav mongodb

CHANGELOG.asciidoc
```diff
@@ -38,6 +38,11 @@ Notes:
 [float]
 ===== Bug fixes
 
+* Fix instrumentation of `mongodb` to avoid multiple command handler
+  registrations when client is created via `MongoClient.connect` static
+  method. ({pull}3586[#3586])
+
+
```
trentm, line 45: This should be added to a new `        ==== Unreleased` block.

test/_spy_method.js
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * Returns an object with data about the usage of the spied method
+ *
+ * @param {any} test the test where the spy is created
+ * @param {Object} target the object which holds the method to spy
+ * @param {String} method the method to spy on
+ * @returns {SpyObject} object with usage data
+ */
+function spyOn(test, target, method) {
```
trentm, line 38: nit: I would rename this `test` to `t`, which is the typical var name used in tape and tap tests.

test/_spy_method.js
```diff
@@ -0,0 +1,73 @@
...
+    throw Error('spyOn: target or method not defined');
+  }
+  if (!target[method]) {
+    throw Error(`spyOn: method ${method} is not defined on target`);
+  }
+  if (typeof target[method] !== 'function') {
+    throw Error(`spyOn: method ${method} is not a function`);
+  }
+  if (typeof target[method][kOrigFunction] === 'function') {
+    throw Error(`spyOn: method ${method} is already used`);
```
trentm, line 49: nit: perhaps this?

```suggestion
    throw Error(`spyOn: method ${method} is already being spied on`);
```

test/instrumentation/modules/mongodb.test.js
```diff
@@ -206,6 +216,11 @@ test('ensure run context', async function (t) {
         `span ${s.type}.${s.subtype} "${s.name}" is a child of the transaction`,
       );
     });
+    // Internal calls to `createSpan` are not duplicated
+    t.ok(
+      createSpanSpy.hasBeenCalled(3),
+      `spans created: expected 3, actual ${createSpanSpy.calls.length}`,
```
trentm, line 222: nit: I was a bit confused as to why this is 3 and not 4.  It is because `agent.startSpan('manual').end();` doesn't happen to call `agent._instrumentation.createSpan()`, because of an implementation detail.

This is totally optional. A potential alternative would be to use the existing `transaction.span_count.started` value reported on serialized transactions. Like this: https://gist.github.com/trentm/032e3b560dc8ce4c3d1092253a956890

Then `spyOn` wouldn't be needed in this case. I think `spyOn` is still a useful test utility that could be added here, or saved for a separate commit.

trentm (changes requested):
Nice find!  The changelog message position should move up. The rest of my comments are nits.

CHANGELOG.asciidoc
trentm: If https://github.com/elastic/apm-agent-nodejs/pull/3590 goes in first and this merges in those changes, then this might be solved.

test/instrumentation/modules/mongodb.test.js
david-luna: Oh, I did not know about `transaction.span_count.started`. If its just a counter and there is no internal logic that tweaks its value then I think its better to use that. I'll apply your suggestion, thanks

test/_spy_method.js
david-luna: finally I decided to remove `spyOn` method since it's not necessary in this tests

test/instrumentation/modules/mongodb.test.js
trentm: > Oh, I did not know about `transaction.span_count.started`.

Yah, totally fair. It exists for collecting data on dropped spans when hitting `transactionMaxSpans` limits, so it is a little indirect and lucky to be usable for this.

trentm (approved):

PR #3590, opened 2023-08-25
chore: add default ports to `context.service.target.name` for exit HTTP spans 

david-luna:
This PR is conform  with the agreements of the discussion in https://github.com/elastic/apm/pull/700 which in summary says that default ports should be explicit in `serviceTarget.name` property of the span.

Updatecli workflows are failing because of that and JSON specs cannot be updated #3589

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/test/service-resource-inference.test.js
+++ b/test/service-resource-inference.test.js
@@ -51,8 +51,11 @@ testData.forEach((testDatum) => {
         // service_resource_inference.json oddly provides context.http.url as
         // an object with "host" and "port" fields, rather than as a URL string
         // as accepted by the intake API. So we need to construct it.
+        // UPDATE: there is a PR fixing it https://github.com/elastic/apm-agent-nodejs/pull/3589
+        // making this test fail so for a short period we need to accept both cases
+        // before removing this tweak.
         const httpContext = Object.assign({}, testDatum.span.context.http);
-        if (httpContext.url) {
+        if (httpContext.url && typeof httpContext.url !== 'string') {
           const u = new URL('', 'http://example.com');
           if (httpContext.url.host) {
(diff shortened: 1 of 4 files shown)


david-luna (commented):
/test tav

test/service-resource-inference.test.js
```diff
@@ -51,8 +51,11 @@ testData.forEach((testDatum) => {
         // service_resource_inference.json oddly provides context.http.url as
         // an object with "host" and "port" fields, rather than as a URL string
         // as accepted by the intake API. So we need to construct it.
+        // UPDATE: there is a PR fixing it https://github.com/elastic/apm-agent-nodejs/pull/3589
+        // making this test fail so for a short period we need to accept both cases
+        // before removing this tweak.
```
trentm, line 56: I'm a little confused. Should we take https://github.com/elastic/apm-agent-nodejs/pull/3589 *first* and then this fix?  Or perhaps add the `lib/instrumentation/span.js` changes as part of #3589?  Then the shared test file change, and the APM agent changes to match it are in the same commit.

If you prefer to do them separately (this one first), then that's cool too.

trentm (approved):

test/service-resource-inference.test.js
david-luna: What I had in mind is 1st to add the default port and then let `updatecli` to update the JSON spec file in #3589 
With this change in the conditional the test accepts both formats in the spec file (URL as object or as string) so the steps I had planned are:
- merge this PR
- merge #3589 when tests pass
- remove the `if` block in a new PR (no longer needed)

I'll check if I can do the last stem on #3589 so we avoid an extra PR

Issue #2760, opened 2022-06-06
tracking delayed dep updates because of old Node support


trentm:
I'm using this issue to track dependencies that we are *not* updating while we still support old major versions of node. Sometimes it is because the newer dependency version actually breaks with the older node, and sometimes it is just because that dep dropped support for the older node (but might not yet actually break).

# node v8

- dep lru-cache@6 dropped support for node v8
- devDep ajv@8 dropped support for node v8 (https://github.com/elastic/apm-agent-nodejs/pull/2755)
- depDep eslint@7 dropped support for node v8 (https://github.com/elastic/apm-agent-nodejs/issues/2583)
- dep pino@8 dropped support for node <v14, pino@7 only tests with node v12 and later (it mentions dropping node v10 support in [its 7.0.0 release notes](https://github.com/pinojs/pino/releases/tag/v7.0.0)). We are using pino@6 which seems to work with node v8.6 and up, but is only tested with node v10 and up.
- devDep mkdirp@1 dropped support for node v8 (https://github.com/elastic/apm-agent-nodejs/pull/2792#issuecomment-1171488188), mkdirp@2 still has `node >=10` in "engines", but only tests with v16+ now (https://github.com/elastic/apm-agent-nodejs/pull/3113)
- devDep ndjson@2 dropped support for node <v10 (https://github.com/elastic/apm-agent-nodejs/pull/2800)
- dep semver@7 dropped support for node<v10 (https://github.com/elastic/apm-agent-nodejs/pull/2808)
- dep load-source-map@3 now requires node v12 (https://github.com/rexxars/load-source-map/pull/6), we've inlined it to lib/load-source-map.js to not have to cope with node ver incompat for now.

# node v10

- dep lru-cache@7 dropped support for node v10 (https://github.com/elastic/apm-agent-nodejs/pull/2756)
- dep pino@8 dropped support for node <v14, pino@7 only tests with node v12 and later (it mentions dropping node v10 support in [its 7.0.0 release notes](https://github.com/pinojs/pino/releases/tag/v7.0.0))
- dep load-source-map@3 now requires node v12 (https://github.com/rexxars/load-source-map/pull/6), we've inlined it to lib/load-source-map.js to not have to cope with node ver incompat for now.

# node v12

- dep pino@8 dropped support for node <v14

# ESM-only 

- devDep get-port@6 is ESM-only and dropped support for node <v12.20

trentm:
[...]

PR #3611, opened 2023-09-05
chore: bump min Node.js to v14.17; update deps

trentm:
This bumps our min Node.js from v14.5 to v14.17; and bumps a number
of dependencies.

- v14.5 was for AsyncLocalStorage
- v14.6 brings a V8 with spport for private methods syntax, FinalizationRegistry (used by pino@8)
- v14.8 brings AsyncResource.bind
- v14.17 brings AbortController, AbortSignal.reason (used by readable-stream@4, glob@10, etc.)

Closes: https://github.com/elastic/apm-agent-nodejs/issues/2760


trentm:
Cannot use pino@8 either because it dep `on-exit-leak-free` uses `FinalizationRegistry` which isn't in node v14.5. It *is* in a later Node v14 -- v14.6.0 per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/FinalizationRegistry

trentm:
> This makes me want to _consider_ bumping our min Node.js to v14.17.

After some discussion: let's do it.

david-luna (approved):

PR #3618, opened 2023-09-08
fix: changes to metadata.cloud.project.* fields collected for GCP

trentm:
- cloud.project.id per the shared APM spec change
- fix the bignum parsing of instance.id (we were losing precision before
  this change)
- add instance.name (missed in the original impl?)

Closes: https://github.com/elastic/apm-agent-nodejs/issues/3614


lib/cloud-metadata/gcp.js
```diff
@@ -5,9 +5,13 @@
  */
 
 'use strict';
+
 const URL = require('url').URL;
+const JSONBigInt = require('json-bigint');
```
trentm, line 10: Reviewer note: This is the same lib that OTel is indirectly using. They use the Google-provided "gcp-metadata" package. I didn't want to use that because it adds yet another HTTP lib that we don't particularly need (gaxios, a wrapper around node-fetch that provides an interface like `axios`).  

The "bignumber.js" dep (used by `json-bigint`) adds about 1MB to the install size, I think, which is unfortunate. We *could* get into shenanigans like using a regex to parse out the "id". :)

lib/cloud-metadata/gcp.js
```diff
@@ -44,11 +48,25 @@ function getMetadataGcp(
...
+            'error fetching gcp metadata: unexpected statusCode: ' +
+              res.statusCode,
+          ),
+        );
+        return;
+      }
+      // Note: We could also guard on the response having the
+      // 'Metadata-Flavor: Google' header as done by:
+      // https://github.com/googleapis/gcp-metadata/blob/v6.0.0/src/index.ts#L109-L112
+
```
trentm, line 64: Reviewer note: Add a sanity guard on the response from the Google metadata API. If it isn't a 200, we shouldn't get into parsing it.  This is my justification for dropping some of the (unnecessary, IMO) defensive coding and testing cases.

lib/cloud-metadata/gcp.js
```diff
@@ -78,76 +96,37 @@ function getMetadataGcp(
...
+  const az = data.instance.zone.split('/').pop();
 
   const metadata = {
-    availability_zone: null,
-    region: null,
+    provider: 'gcp',
     instance: {
-      id: null,
+      id: data.instance.id.toString(), // We expect this to be a BigInt.
+      name: data.instance.name,
```
trentm, line 118: Reviewer note: This function has been simplified (IMO). The `instance.name` had been missed before (or perhaps the APM spec changed after original impl?).  The `data.instance.id` is the interesting BigInt case.

test/cloud-metadata/_fixtures.js
```diff
@@ -64,128 +64,11 @@ module.exports = {
...
-              index: 0,
-              interface: 'SCSI',
-              mode: 'READ_WRITE',
-              type: 'PERSISTENT',
-            },
-          ],
-          guestAttributes: {},
-          hostname:
-            'username-temp-delete-me-cloud-metadata.c.elastic-apm.internal',
-          id: 7684572792595385000,
```
trentm, line 99: Reviewer note: I've changed this test fixture to use an **unparsed** actual response on a test GCP VM. The problem with having an already parsed object, is that it accidentally included the BigInt loss of precision. That `7684572792595385000` is wrong. The actual instance.is was some `7684572792595385???` value with something other than zeros. It was just not noticed originally.

test/cloud-metadata/index.test.js
```diff
@@ -254,28 +254,6 @@ tape('cloud metadata: aws empty data', function (t) {
   });
 });
 
-tape('cloud metadata: gcp empty data', function (t) {
```
trentm, line 257: Reviewer note: I've dropped a couple unnecessarily defensive test cases, IMHO. See the note above about the statusCode==200 guard.

lib/cloud-metadata/gcp.js
david-luna: > This is my justification for dropping some of the (unnecessary, IMO) defensive coding and testing cases.

I agree that we can assume safely that a 200 status comes with the proper metadata described in their docs. Also it simplifies the code so IMO this is a good change :)

lib/cloud-metadata/gcp.js
david-luna: Checking the [metadata docs](https://cloud.google.com/compute/docs/metadata/default-metadata-values#instance-attributes-metadata) and also the JSON fixture you added I think we can be sure that there are only two numeric values in the response: `numericProjectId` and `id` (within instance)

```json
{
  "instance": {
    "attributes": {},
    "id": 5737554347302044216,
    ...
  },
  "project": {
    "attributes": {
      ...
    },
    "numericProjectId": 523926462582,
    "projectId": "acme-eng"
    ...
  }
}
```

Since it is so specific I think we can opt-in to a custom solution to capture that info. You mentioned `RegExp` and maybe is the way to go using capture groups with `String.match`. We can capture the numeric values with something similar to the snippet below and use `JSON.parse` for the rest

```js
const respBody = '...' // The response text with all the metadata

const instanceIdRegExp = /"id":(\d+),/;
const projectIdRegExp = /"numericProjectId":(\d+),/;

const instanceIdMatch = respBody.match(instanceIdRegExp)
const projectIdMatch = respBody.match(projectIdRegExp)

// The match is already an String so no need to convert
const instanceId = instanceIdMatch && instanceMatch[1];
const projectId = projectIdMatch && projectIdMatch[1];

```

NOTE: regexes should be reviewed since they do not cover the case of being the last property of the object

david-luna (approved):

lib/cloud-metadata/gcp.js
trentm: David and I discussed. The added "bignumber.js" dep *is* a bit of size, but still relatively low compared to a number of other deps, so we think it is better to have the clearer implementation and not get into possibly brittle regex usage just to save some install size.

PR #3623, opened 2023-09-13
feat: add observability for lambdas that call other lambdas

fdaciuk:
This Pull Request enables the visualization of Lambda-to-Lambda invocations in the Service Map of Elastic Search and provides detailed insights into the "spans" and "transactions" used by the internal Lambda function. The way we use to identify Lambda-to-Lambda invocations is by checking for the presence of the `traceparent` and `tracestate` in the invocation payload.

I welcome any insights or recommendations to make it even more robust and user-friendly.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/http-shared.js
+++ b/lib/instrumentation/http-shared.js
@@ -207,7 +207,10 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
   return function wrapHttpRequest(orig) {
     return function wrappedHttpRequest(input, options, cb) {
       const parentRunContext = ins.currRunContext();
-      var span = ins.createSpan(null, 'external', 'http', { exitSpan: true });
+      var span =
+        input?.headers?.['x-amz-invocation-type'] === 'RequestResponse'
+          ? null
+          : ins.createSpan(null, 'external', 'http', { exitSpan: true });
       var id = span && span.transaction.id;
       agent.logger.debug('intercepted call to %s.%s %o', moduleName, method, {
         id,
--- a/lib/lambda.js
+++ b/lib/lambda.js
@@ -706,6 +706,14 @@ function elasticApmAwsLambda(agent) {
       let traceparent;
       let tracestate;
       if (
+        triggerType === TRIGGER_GENERIC &&
+        event.traceparent &&
+        event.tracestate
+      ) {
+        traceparent = event.traceparent;
+        tracestate = event.tracestate;
+      }
+      if (
         (triggerType === TRIGGER_API_GATEWAY || triggerType === TRIGGER_ELB) &&
         event.headers
       ) {


lib/lambda.js
```diff
@@ -705,6 +705,14 @@ function elasticApmAwsLambda(agent) {
...
       let traceparent;
       let tracestate;
+      if (
+        triggerType === TRIGGER_GENERIC &&
+        event.traceparent &&
+        event.tracestate
+      ) {
+        traceparent = event.traceparent;
+        tracestate = event.tracestate;
+      }
```
trentm, line 715: @fdaciuk Thanks for the PR. 

Can you show some simple Lambda function code that uses this? That would help me understand a bit better.
Does your code that is doing the Lambda Invoke API call (https://docs.aws.amazon.com/lambda/latest/dg/API_Invoke.html) manually pass in `traceparent` and `tracestate` in the payload?

lib/instrumentation/http-shared.js
```diff
@@ -207,7 +207,10 @@ exports.traceOutgoingRequest = function (agent, moduleName, method) {
   return function wrapHttpRequest(orig) {
     return function wrappedHttpRequest(input, options, cb) {
       const parentRunContext = ins.currRunContext();
-      var span = ins.createSpan(null, 'external', 'http', { exitSpan: true });
+      var span =
+        input?.headers?.['x-amz-invocation-type'] === 'RequestResponse'
+          ? null
+          : ins.createSpan(null, 'external', 'http', { exitSpan: true });
```
trentm, line 213: Was this change to avoid the extra span necessary for the Service Map connection to work? Or was this just to avoid having an addition span in the trace waterfall?

trentm:
cla/check

lib/lambda.js
fdaciuk: Hi @trentm. Exactly! I don't know if there is better way to do this, but that update worked for my use case.

I'm invoking the lambda like this:

```js
  const command = new InvokeCommand({
    FunctionName: "my-lambda",
    InvocationType: "RequestResponse",
    Payload: new TextEncoder().encode(JSON.stringify({ traceparent, tracestate })),
  })
```

lib/instrumentation/http-shared.js
fdaciuk: Both I guess. If this span is created when a lambda is invoked, I can see two entries on Service Map: one empty (created by this span) and other with the full tracing (created from the code send in `lib/lambda.js`.

This code just removes the empty one.

lib/lambda.js
trentm: There isn't a current way to do this ... except possibly Amazon's X-Ray support might enable making the connection. We aren't currently doing anything with X-Ray in our Lambda support.

Providing some way for this manual distributed tracing in Lambda is interesting. I like the idea.

I wonder about instead using `ClientContext` in the [InvokeCommand](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-lambda/classes/invokecommand.html) call.   Theoretically, if the APM agent automatically instrumented `@aws-sdk/client-lambda` -- which I'm guessing you are using here -- then we could (perhaps optionally) support automatically adding the trace-context info (`traceparent` et al) to the ClientContext or Payload.

I'm not sure if `ClientContext` and `Payload`, IIUC, are unstructured. Or does the provided `ClientContext` need to be a JSON object that can be added to the `context` object passed to the Lambda function?  If these are unstructured, in general, then it would be a hard sell to have a default feature added to the APM agent that tried to impose some structure on it.

fdaciuk: > I'm not sure if ClientContext and Payload, IIUC, are unstructured. Or does the provided ClientContext need to be a JSON object that can be added to the context object passed to the Lambda function?

That's a good question. I'm not sure, but love the idea this to be automatic. We're creating a library on our company to abstract the use of `InvokeCommand` and inject `traceparent` and `tracestate` into Payload, so if we have an option to do this from `elastic-apm-node`, even manually, would be great =)

trentm: Just a heads up that I won't be able to get to this soon. :/

fdaciuk: No problem. For now, we're using a forked version of this repo with these updates. I hope it's not a problem until you decide whether or not it's a valuable feature in this lib =)

PR #3643, opened 2023-09-26
feat: add @aws-sdk/client-sqs instrumentation

### LINKED ISSUE #2957: AWS instrumentation for SDK v3, SQS
Provide instrumentation for AWS SDK version 3 SQS client

david-luna:
This PR add a new instrumentation for the package `@aws-sdk/client-sqs`, the v3 of the client for the queuing system. The implementation pursues to have the same behaviour than the previous instrumentation. This means operations or command names are mapped to specific actions. 

Close #2957 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


test/instrumentation/modules/@aws-sdk/client-sns.test.js
```diff
@@ -248,7 +248,7 @@ const testFixtures = [
       TEST_REGION: 'us-east-2',
     },
     versionRanges: {
-      node: NODE_VER_RANGE_IITM,
+      node: NODE_VER_RANGE_IITM_GE14,
```
david-luna, line 251: Note for reiewer: leftover from SNS PR

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
```diff
@@ -0,0 +1,299 @@
...
+        },
+        'deleteMessageBatch',
+      );
+
+      t.deepEqual(
+        spanLinks,
+        [
+          { trace_id: tx.trace_id, span_id: sendMessageSpanId },
+          // XXX: not sure why this one is not appearing
+          { trace_id: tx.trace_id, span_id: sendMessagesBatchSpanId },
```
david-luna, line 247: Note for the reviewer: reason why it's still a draft. I have to check why we do not have the 2nd link. Probably not adding the message attributes properly when sending the batch

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
```diff
@@ -0,0 +1,299 @@
...
+              },
+            },
+            message: { queue: { name: 'elasticapmtest-queue-1.fifo' } },
+          },
+          outcome: 'success',
+        },
+        'sendMessageBatch',
+      );
+
+      // There will be one or more `SQS ReceiveMessage ...` spans for the ReceiveMessage
```
trentm, line 160:
```suggestion
      // There will be one or more `SQS POLL ...` spans for the ReceiveMessage
```
?

lib/instrumentation/modules/@aws-sdk/client-sqs.js
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  return 'unknown';
+}
+
+/**
+ * @typedef {import('@aws-sdk/client-sqs').Message} Message
+ */
+/**
+ * // Extract span links from up to 1000 messages in this batch.
```
trentm, line 220:
```suggestion
 * Extract span links from up to 1000 messages in this batch.
```

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
trentm: Comparing to sqs.test.js (for v2), I think you should expect to get *three* span links. An issue with this test (it might not be the reason for this failure) is that it is assuming that the order spans will be:
- all of the ReceiveMessage spans first,
- then a single DeleteMessageBatch span

That's wrong. See the handling in sqs.test.js. There can be one to three DeleteMessageBatch calls interspersed with the ReceiveMessage calls.

trentm (commented):
Mostly looks good. A couple nits and one comment about the failing test.

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
david-luna: Got it. I was not handling properly the `Entries` of `SendMessageBatch` command and not propagating the attributes properly. The fix has been pushed :)

trentm: Great. 

FWIW, I think there is still the potential logic error. The spans might be:
- ReceiveMessage (no messages received)
- ReceiveMessage (1 of the 3 messages received)
- DeleteMessageBatch (to delete that 1 received message)
- RecieveMessage (the remaining 2 messages received)
- DeleteMessageBatch (to delete those last 2 messages)

IIUC the current test code will fail on this case.

test/instrumentation/modules/@aws-sdk/client-s3.test.js
```diff
@@ -456,7 +456,7 @@ const testFixtures = [
     versionRanges: {
       node: NODE_VER_RANGE_IITM,
     },
-    verbose: true,
+    verbose: false,
```
david-luna, line 459: Note for reviewer: leftover. We do not want to be verbose on the CI

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
david-luna: You're right. I've been lucky the order was the same in my tests. I've changed to check the span nam within the loop like `sqs.test.js` file does.

david-luna:
@trentm last changes include also `use-client-sqs.mjs` to check ESM compat and its tests. Have a look at it please :)

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
david-luna: done!

lib/instrumentation/modules/@aws-sdk/client-sqs.js
```diff
@@ -0,0 +1,332 @@
...
+                msgAttrs[name] = { DataType: 'String', StringValue: value };
+              },
+            );
+          });
+        }
+
+        // Ensure there is a span from the wrapped `client.send()`.
+        if (!span || !(span.type === TYPE && span.subtype === SUBTYPE)) {
+          return await next(args);
+        }
```
trentm, line 95: Should this block be up at the top, as it is in client-s3.js and client-dynamodb.js?

I have the same question for client-sns.js.

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
```diff
@@ -0,0 +1,306 @@
...
+      );
+
+      // There will be one or more `SQS POLL ...` spans for the ReceiveMessage
+      // API calls until all messages are retrieved -- with interspersed
+      // `SQS DELETE_BATCH ...` spans to delete those messages as they are
+      // received.
+      const spanLinks = [];
+
+      while (spans.length > 0) {
+        const currSpan = spans.shift();
```
trentm, line 167: nit: If the `break` case is hit below -- i.e. `spans` includes some span other than `SQS POLL ...` or `SQS DELETE_BATCH ...`, e.g. if `use-client-sqs.js` is testing some other SQS commands -- then this `spans.shift()` will have removed that span from `spans` for subsequent testing.  This is why the while-loop in "sqs.test.js" isn't `.shift()`ing until `topSpanName` has been identified as one we want to process.

Just a nit though because, it works here because `spans` is empty after the `POLL/DELETE_BATCH` spans.

trentm (approved):
Approved, though I have a Q and a nit below.

lib/instrumentation/modules/@aws-sdk/client-sqs.js
david-luna: Checking the v2 implementations of both the trace context propagation is done always 1st and then there is the check of the `span` to decide if we have to extract info and add it to its context. I guess we want the same behaviour for v3

test/instrumentation/modules/@aws-sdk/client-sqs.test.js
david-luna: You're right. I've interpreted the condition of the `while` as "let's consume all the remaining spans".

lib/instrumentation/modules/@aws-sdk/client-sqs.js
trentm: You are correct. Thanks.

trentm (approved):

PR #3651, opened 2023-09-29
Add GitHub action to add issue to agents project board when milestoned

estolfo:
Copied from the [java agent repo](https://github.com/elastic/apm-agent-java/blob/main/.github/workflows/addToProject.yml), with the team name replaced.
I replaced what was in this file already, assuming that it's no longer relevant. Please let me know if you still need what was in the file-- it adds an issue to the Old Agents project board.

### DIFF
--- a/.github/workflows/addToProject.yml
+++ b/.github/workflows/addToProject.yml
@@ -1,20 +1,86 @@
-name: Auto Assign to Project(s)
-
+name: Add to APM Agents Project
 on:
   issues:
-    types: [opened, edited, milestoned]
-env:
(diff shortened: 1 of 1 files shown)


david-luna (approved):

david-luna:
LGTM but I think we should let @trentm to have a 👀

trentm (approved):
@estolfo Looks good. Thanks!

Separate thing: It would be nice to also update our labeler.yml (and related config files) at some point, as apm-agent-java has done.  (The the labeler workflow runs are generating deprecation warnings: https://github.com/elastic/apm-agent-nodejs/actions/runs/6349252109)

Issue #2992, opened 2022-10-18
Allow Logical Grouping of Instrumentation by Name


astorm:
**Is your feature request related to a problem? Please describe.**

Instrumentation increasingly requires us to hook multiple modules, which means it's no longer easy/intuitive to enable/disable a single module's instrumentation. 

**Describe the solution you'd like**

Modify module wrapping and disabling to allow for logic grouping of instrumentations. 

**Describe alternatives you've considered**

Passing complexity on to users and being sad :(

**Additional context**

Traditionally, there's been a 1:1 relationship between the module being instrumented and the individual instrumentation-module we write.  Put another way, for each entry [in the `MODULES`](https://github.com/elastic/apm-agent-nodejs/blob/c55b1d8c32b2574362fc24d81b8e173ce2f75257/lib/instrumentation/index.js#L32-L69) array, there's been a single instrumentation file [in the `instrumentation/modules` folder](https://github.com/elastic/apm-agent-nodejs/tree/main/lib/instrumentation/modules). Traditionally, access to the object retured by requiring a module has been enough for us to replace the methods we need.

However, modern javascript practices are moving towards modules having more private members.  This is both a cultural change, and a technical change brought about by TypeScript.  Additionally -- many projects are moving towards having multiple packages.  

One example of this is Node Redis.  Where older versions packaged everything under a single `redis` package

    require('redis')

[Modern versions have many packages](https://github.com/redis/node-redis/tree/d0bfa771a85b61fbdfda0401bc422d98ea1103a9#packages)

    require('@redis/client')
    require('@redis/bloom')
    require('@redis/graph')
    require('@redis/json')
    require('@redis/search')
    require('@redis/time-series')
    
Additionally -- the redis instrumentation requires that we instrument _multiple_ individual exports to get at the methods and objects we need.

    var MODULES = [
      [
      '@node-redis/client/dist/lib/client',
      '@node-redis/client/dist/lib/client/commands-queue',
      '@redis/client/dist/lib/client',
      '@redis/client/dist/lib/client/commands-queue',
      
We should modify our module loading code to allow for a _logical_ grouping of names.  Something like 

    var MODULES = {
        'redis': [
          'redis',
          '@node-redis/client/dist/lib/client',
          '@node-redis/client/dist/lib/client/commands-queue',
          '@redis/client/dist/lib/client',
          '@redis/client/dist/lib/client/commands-queue',
        ]
        // ...
    }
    
[...]

PR #3657, opened 2023-10-03
fix: mongodb async context fix

### LINKED ISSUE #3161: Mongoose spans are getting attached to wrong parent transaction

david-luna:
This PR add extra instrumentation to Mongodb's `ConectionPool` class to ensure the context is not lost when the pool class is queueing the commands. The agent makes use of the `AsyncLocalStiorage.bind` static function to inject the context into the queued callbacks.

The PR also migrates to the new test implementations making use of `runTestFixtures` util.

Fixes #3161 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation


lib/instrumentation/index.js
```diff
@@ -115,6 +116,8 @@ const IITM_MODULES = {
   https: { instrumentImportMod: true },
   ioredis: { instrumentImportMod: false },
   knex: { instrumentImportMod: false },
+  // XXX: check this
+  // mongodb: { instrumentImportMod: false },
```
david-luna, line 120: Note for reviewer: this is a work in progress though I do not mind to have it in a separate PR.

lib/instrumentation/index.js
```diff
@@ -358,35 +361,43 @@ Instrumentation.prototype._startHook = function () {
...
-          'could not shim %s module: %s',
-          name,
-          e.message,
-        );
-        return exports;
+  this._ritmHook = new RitmHook(
+    this._patches.keys,
+    // XXX: found this was not enabled so our redis instrumentations
+    // may not be working as for now
+    { internals: true },
```
david-luna, line 368: Note for reviewer: check the comment and give feedback please. Was okay to pass the new option? what about other instrumentations.

lib/instrumentation/modules/mongodb.js
```diff
@@ -4,185 +4,6 @@
...
-  }
-
-  function collectionFor(event) {
-    const collection = event.command[event.commandName];
-    return typeof collection === 'string' ? collection : '$cmd';
-  }
-};
+// Since we are also instrumenting submodules we moved this instrumentation
+// to a dedicated folder
+module.exports = require('./mongodb/lib/index');
```
david-luna, line 9: Note to reviewer: moved to `modules/mongodb/lib/index.js` to keep all instrumentation logic together. The path is mimicking the one in the lib

lib/instrumentation/index.js
trentm: Adding ESM in a separate PR would be helpful. Keeping this PR to a small change for using AsyncResource.bind might be helpful to compare with an equivalent OTel change.

lib/instrumentation/index.js
trentm: `internals: true` isn't needed if the sub-module path, e.g. `@redis/client/dist/lib/client`, is an entry in the first argument to `RitmHook` -- i.e. if it is on the explicit list of modules to hook into. If I add these debug-prints to RITM:

```diff
--- /Users/trentm/el/require-in-the-middle/index.js	2023-07-13 15:32:26
+++ ./node_modules/require-in-the-middle/index.js	2023-10-03 09:26:14
@@ -126,6 +126,7 @@
   const self = this
   const patching = new Set()
   const internals = options ? options.internals === true : false
+  console.log('XXX ritm: internals=%s', internals);
   const hasWhitelist = Array.isArray(modules)

   debug('registering require hook')
@@ -204,6 +205,7 @@
       }
       moduleName = filename
     } else if (hasWhitelist === true && modules.includes(filename)) {
+      console.log('XXX ritm abspath in allowlist: hasWhitelist=%s modules.includes(filename)=%s filename=%s', hasWhitelist, modules.includes(filename), filename);
       // whitelist includes the absolute path to the file including extension
       const parsedPath = path.parse(filename)
       moduleName = parsedPath.name
@@ -227,6 +229,7 @@
       if (hasWhitelist === true && modules.includes(moduleName) === false) {
         if (modules.includes(fullModuleName) === false) return exports // abort if module name isn't on whitelist

+        console.log('XXX ritm allowlisted sub-module: moduleName=%s fullModuleName=%s filename=%s', moduleName, fullModuleName, filename)
         // if we get to this point, it means that we're requiring a whitelisted sub-module
         moduleName = fullModuleName
       } else {
diff --git a/lib/instrumentation/index.js b/lib/instrumentation/index.js
index b90152f7..29e4bbb7 100644
--- a/lib/instrumentation/index.js
+++ b/lib/instrumentation/index.js
@@ -363,6 +363,7 @@ Instrumentation.prototype._startHook = function () {
     name,
     basedir,
   ) {
+    console.log('XXX RitmHook called:', name, basedir);
     const enabled = self._isModuleEnabled(name);
     var pkg, version;

```

and run examples/trace-redis.js:

```
% node examples/trace-redis.js
{"log.level":"info","@timestamp":"2023-10-03T16:30:11.754Z","log":{"logger":"elastic-apm-node"},"agentVersion":"4.0.0","env":{"pid":19696,"proctitle":"node","os":"darwin 22.6.0","arch":"x64","host":"pink.local","timezone":"UTC-0700","runtime":"Node.js v16.20.2"},"config":{"serviceName":{"source":"start","value":"example-trace-redis4","commonName":"service_name"},"serviceVersion":{"source":"default","value":"4.0.0","commonName":"service_version"},"serverUrl":{"source":"default","value":"http://127.0.0.1:8200/","commonName":"server_url"},"logLevel":{"source":"default","value":"info","commonName":"log_level"},"spanCompressionEnabled":{"source":"start","value":false}},"activationMethod":"require","ecs":{"version":"1.6.0"},"message":"Elastic APM Node.js Agent v4.0.0"}
XXX ritm: internals=false
XXX ritm allowlisted sub-module: moduleName=@redis/client fullModuleName=@redis/client/dist/lib/client/commands-queue filename=/Users/trentm/el/apm-agent-nodejs/node_modules/@redis/client/dist/lib/client/commands-queue.js
XXX RitmHook called: @redis/client/dist/lib/client/commands-queue /Users/trentm/el/apm-agent-nodejs/node_modules/@redis/client
XXX RitmHook called: generic-pool /Users/trentm/el/apm-agent-nodejs/node_modules/generic-pool
XXX ritm allowlisted sub-module: moduleName=@redis/client fullModuleName=@redis/client/dist/lib/client filename=/Users/trentm/el/apm-agent-nodejs/node_modules/@redis/client/dist/lib/client/index.js
XXX RitmHook called: @redis/client/dist/lib/client /Users/trentm/el/apm-agent-nodejs/node_modules/@redis/client
XXX RitmHook called: redis /Users/trentm/el/apm-agent-nodejs/node_modules/redis
PING res:  PONG
SET res:  OK
GET res:  bar
MULTI res:  [ 'OK', 'eggs' ]
```

The RITM `opts.internals` value is only used later for internal module paths that are *not* on the explicitly set of modules to hook.

lib/instrumentation/index.js
```diff
@@ -358,35 +361,43 @@ Instrumentation.prototype._startHook = function () {
...
+    function (exports, name, basedir) {
+      // RITM returns the path with extension for sub-modules
+      // so we need to make sure is not there when patching
+      if (name.endsWith('.js')) {
+        name = name.substring(0, name.length - 3);
       }
```
trentm, line 374: This was only needed when `internals: true` was being used?

lib/instrumentation/modules/mongodb.js
trentm: nit: I understand wanting to group all "mongodb-related instrumentation code" into one area. However, this currently makes the mongodb instrumentation special in that the patching of `require('mongodb')` is *not* at "lib/instrumentation/modules/mongodb.js".  Mostly, though, this (and the ESM changes, and the test file refactoring) is contributing to this being a rather large patch where the main point is the `AsyncResource.bind` usage.

trentm (commented):
The async context fix is a little lost in the 2000 lines of change due to the ESM work, the mongodb.js move and the test file refactoring. It would be helpful to have these in 3 separate PRs. (However, I hate that that is a pain because of our review policies.)

lib/instrumentation/index.js
david-luna: That's correct. I'll check again how the hook works with using your diff and remove the RITM options

david-luna:
> nit: I understand wanting to group all "mongodb-related instrumentation code" into one area. However, ...

I'm okay with leaving the implementation where it was. I'll put it back in a new commit

lib/instrumentation/index.js
david-luna: I did not get those logs for `mongodb` and I think I know why. The last log inside `./node_modules/require-in-the-middle/index.js` is inside a conditional

```js
if (hasWhitelist === true && modules.includes(moduleName) === false) {
    if (modules.includes(fullModuleName) === false) return exports // abort if module name isn't on whitelist

    console.log('XXX ritm allowlisted sub-module: moduleName=%s fullModuleName=%s filename=%s', moduleName, fullModuleName, filename)
    // more code here ...     
}
```

In my case
- moduleName === 'mongodb'
- fullModuleName === 'mongodb/lib/cmap/connection_pool'

The expression allows to pass sub-modules only if the root module is not in the whitelist. I this case the agent is instrumenting the main module and one sub-module. I think it works on `@redis` because its instrumention only sub-modules.

I see two options here. Let me know what you think
- change the instrumentation and go for sub-modules (need to check if feasible)
- make use of the `internals` option.

trentm: > * make use of the `internals` option.

I'd prefer to avoid this. It will mean that we will be calling our `onRequire` hook function for every require'd file under each of the modules we instrument (every lib file in ioredis, in express, in fastify, etc.). There *may* be a perf impact of this.

---

Option 3: Remove that limitation in RITM.

I think this is a bug in RITM that you cannot do this.  Having 'mongodb' on the list of modules to patch shouldn't mean that you cannot patch `mongodb/lib/cmap/connection_pool`.  I'll have to stare at this longer... the if/else's in RITM make my eyes swim a bit.

lib/instrumentation/index.js
```diff
@@ -142,7 +150,9 @@ function Instrumentation(agent) {
       // Lazy require so that we don't have to use `require.resolve` which
       // would fail in combination with Webpack. For more info see:
       // https://github.com/elastic/apm-agent-nodejs/pull/957
-      return require(`./modules/${pathName}.js`)(...args);
+      const modExt = pathName.endsWith('.js') ? '' : '.js';
```
david-luna, line 153: Note for reviewer: now we can get a plain module name or a sub-module with the `.js` extension. The patch function needs to be aware of the extesion

lib/instrumentation/index.js
david-luna: 1. That's okay. There is no need to rush in a fix there and we still have to check what's the impact since no one hasn't noticed.
2. I've done the changes to have sub-modules with the `.js` extension. I've noticed our Hook is reading the `package.json` file each time is called so enabling internals will do many reads. To avoid this I made use of a `Map` to cache versions.

lib/instrumentation/index.js
```diff
@@ -358,35 +368,32 @@ Instrumentation.prototype._startHook = function () {
...
+    { internals: true },
+    function (exports, name, basedir) {
+      const enabled = self._isModuleEnabled(name);
+      var version;
+
+      const isHandlingLambda =
+        self._lambdaHandlerInfo && self._lambdaHandlerInfo.module === name;
+
+      if (!isHandlingLambda && basedir) {
+        version = self._getPackageVersion(name, basedir);
```
david-luna, line 385: Note for reviewer: `_getPackageVersion` returns `undefined` if there is an exception reading the package file.

lib/instrumentation/index.js
```diff
@@ -412,22 +419,26 @@ Instrumentation.prototype._startHook = function () {
 };
 
 Instrumentation.prototype._getPackageVersion = function (modName, modBaseDir) {
+  const modPath = path.join(modBaseDir, 'package.json');
+  if (this._versionsCache.has(modPath)) {
```
david-luna, line 423: Note for reviewer: both RITM and IITM hooks use this function and will get cached result if they request many times the version for the same `pacakge.json` file

lib/instrumentation/modules/mongodb.js
david-luna: put back in place :)

lib/instrumentation/index.js
david-luna: With the latest commit this is not necessary. The hook delegates the extension handling to the patch function.

david-luna:
IMO this PR should wait until #2992 is done, so I'll leave it as a draft

lib/instrumentation/index.js
david-luna: This [PR](https://github.com/elastic/apm-agent-nodejs/pull/3658) solves our use case for instrumenting module and sub-module at the same time. which is what we need for `mongodb`.

We still have to look at RITM but since the issue is only reported by us I think we have no rush.

lib/instrumentation/index.js
david-luna: This is already handled in https://github.com/elastic/apm-agent-nodejs/pull/3658

david-luna:
This PR is closed in favor of #3665

PR #3658, opened 2023-10-06
chore: refactor module patch handling in Instrumentation class

trentm:
This refactors how module patch handlers are loaded and the Instrumentation class file. There is now a PatcherRegistry that is a little smarter than the `_patches` object before it.

The main change is that we now use RITM's internals:true option so that our `onRequire` hook is called with module sub-paths. This will allow #3657 to hook both "mongodb" and "mongodb/lib/cmap/connection_pool.js", and avoid a RITM bug where "mongodb" &
"mongodb/lib/cmap/connection_pool" (no extension, indicating a "sub-module") don't work together.

Benefits and other changes:
- With the smarter "PatcherRegistry" the module name can be used in `disableInstrumentations` to disable any patchers for that module. E.g. `disableInstrumentations=next` will disable all four `next`-related patchers.
- A custom "lambda" name is now usable to disable the Lambda instr, FWIW.
- (Note: we *could* also add custom disableInstrumentations keys for other modules, e.g. perhaps "aws-sdk" for all those.)
- Some edge cases with the (rare, and undocumented) `config.addPatch` config var have been cleaned up.
- Loading of the lambda-handler has been refactored a bit to no longer need to know internal details of the instrumentation module.

Refs: #3657
Closes: #2992


test/config.test.js
```diff
@@ -1194,140 +1194,6 @@ usePathAsTransactionNameTests.forEach(function (usePathAsTransactionNameTest) {
   );
 });
 
-test('disableInstrumentations', function (t) {
```
trentm, line 1197: Reviewer note: I have dropped this test. It has always been a thorn. All this knowledge in here for dropping certain packages (because of node version compat) from the test is silly. The test provides basically nil value. It also requires deep internal knowledge of "lib/instrumentation/index.js" impl details.

david-luna: I agree is cumbersome to keep this test up to date and thanks to this PR I think it should be easier since instrumentation enablement relies on the proper grouping of modules and it sub-modules by their `diKey`.

Perhaps a test on `PatchRegistry` ensuring keys are properly resolved would be a good substitute for this test.

david-luna (dismissed):

david-luna:
Looks good. I'm leaving already the approval so you can merge when you finish the the caching of module versions

test/config.test.js
trentm: I added some testing in commit [eeed31c](https://github.com/elastic/apm-agent-nodejs/pull/3658/commits/eeed31c95337c30e25243e81eff2e2a88a2a5acd)

david-luna (approved):

PR #3680, opened 2023-10-17
chore: tav versions update script

david-luna:
Include a new optional config for modules which can define in which way the `versions` property should be updated regularly. The initial set of modes would be:

- `latest-minors`: from each minor version the highest patch
- `latest-majors`: from each major version the highest minor.patch
- `max-n`: first version, n versions in between (the number is configurable) and last version

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


david-luna (commented):
/test tav mongodb

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,130 @@
...
+
+/*
+ * Copyright Elasticsearch B.V. and other contributors where applicable.
+ * Licensed under the BSD 2-Clause License; you may not use this file except in
+ * compliance with the BSD 2-Clause License.
+ */
+
+'use strict';
+
+// Update version property for each module defined in .tav.yml file
```
trentm, line 11: nit: I think it would be helpful if the top comment here identified the "versions-selection" field as the important one that it is using for this.

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,130 @@
...
+    // TODO: maybe it depends on the mode
+    cfg.versions = newVers.join(' || ');
+
+    // Fill the yamlText wihtout loosing comments and format
+    let verIdx, modIdx;
+
+    for (let i = 0; i < tavLines.length; i++) {
+      const l = tavLines[i];
+      if (!modIdx) {
+        modIdx = l.startsWith(`${name}:`) ? i : undefined;
```
trentm, line 69: Sometimes the name is quoted, e.g. if it includes a `@`:

```yaml
'@koa/router-v8-v11':
  name: '@koa/router'
  versions: '>=8 <11'
  node: '>=8.0.0'
  peerDependencies: koa@2
  commands: node test/instrumentation/modules/koa-router/new-name.test.js
'@koa/router':
  name: '@koa/router'
  versions: '>=11 <13'
  node: '>=12.0.0'
  peerDependencies: koa@2
  commands: node test/instrumentation/modules/koa-router/new-name.test.js

elasticsearch:
  versions: '>=8.0.0'
  commands: node test/instrumentation/modules/elasticsearch.test.js
```

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,130 @@
...
+
+async function main() {
+  const tavContent = readFileSync(TAV_PATH, { encoding: 'utf-8' });
+  const tavConfig = yaml.load(tavContent);
+  const tavEntries = Object.entries(tavConfig);
+  const modNames = new Set();
+  const versionsMap = new Map();
+
+  const versionModes = {
+    'latest-minors': getLatestMinors,
```
trentm, line 33: Did you intend to do "latest-majors" and "latest" (as mentioned in the issue description) in this PR? They could come in separate PRs later when/if we need them.

.tav.yml
```diff
@@ -138,8 +138,11 @@ pg-new-node:
 # Latest mongodb-core release (v3.2.7) was released in 2019. We test a subset
 # of the full supported range.
 mongodb-core:
-  versions: '1.2.19 || 1.2.32 || 1.3.21 || 2.0.14 || 2.1.20 || 3.0.11 || 3.1.11 || 3.2.7' # latest minors subset of '>=1.2.19 <4'
+  versions: '1.2.19 || 1.2.32 || 1.3.21 || 2.0.14 || 2.1.20 || 3.0.11 || 3.1.11 || 3.2.7'
```
trentm, line 141: We talked about adding the caret to the last version. Do you intend to do that for this PR, or in a later one?

trentm (dismissed):

dev-utils/update-tav-versions.js
david-luna: thanks, I've changed the code to also cover this case :)

dev-utils/update-tav-versions.js
david-luna: modes are quite straightforward so I've added them all. They are separated in functions for better review

.tav.yml
david-luna: and in some cases we talked about ending with something like `>3.2.7 <4`. Could we have one way of adding last versions? I'm planning to add it.

.tav.yml
```diff
@@ -94,18 +94,27 @@ redis:
 # subset (the first, plus then the latest in each major.minor).
 ioredis-v2-v4:
   name: ioredis
-  versions: '2.0.0 || 2.0.1 || 2.1.0 || 2.2.0 || 2.3.1 || 2.4.3 || 2.5.0 || 3.0.0 || 3.1.4 || 3.2.2 || >3.2.2 <4'
+  versions: '2.0.0 || 2.0.1 || 2.1.0 || 2.2.0 || 2.3.1 || 2.4.3 || 2.5.0 || 3.0.0 || 3.1.4 || ^3.2.2'
```
david-luna, line 97: Note for reviewer: this caret version is added by the script for all modes. We should decide if
- we use caret or range for all modes
- caret/range depends on the selected mode

.tav.yml
```diff
@@ -476,25 +488,32 @@ fastify-v2:
...
     - node test/sanitize-field-names/fastify.test.js
+  update-versions:
+    mode: max-5
+    range: '>=3 <4'
 fastify:
   name: fastify
   # Update versions periodically with "./dev-utils/tav-versions-fastify.sh".
-  versions: '4.0.0 || 4.2.1 || 4.5.2 || 4.8.1 || 4.10.1 || 4.14.0 || 4.17.0 || >4.17.0' # subset of '>=4 <4.0.1 || >4.0.1 <4.16.0 || >4.16.2'
+  versions: '4.0.0 || 4.5.0 || 4.9.1 || 4.14.1 || 4.19.1 || 4.23.2 || ^4.24.3'
+  versions-stash: '4.0.0 || 4.2.1 || 4.5.2 || 4.8.1 || 4.10.1 || 4.14.0 || 4.17.0 || >4.17.0' # subset of '>=4 <4.0.1 || >4.0.1 <4.16.0 || >4.16.2'
```
david-luna, line 506: Note for reviewer: this range here is a bit tricky and current implementation does not support it. I think it can be done by enriching the `update-versions` config. Instead of having a `range` configuration we should split in two configs:
- `include` (required): works exactly as `range`
- `exclude` (optional): provides a range of versions we want to filter out

In this case the range `'>=4 <4.0.1 || >4.0.1 <4.16.0 || >4.16.2'` can be translated to

```
- include: '>=4 <5'
- exclude: '4.0.1 || >=4.16.0 <=4.6.2'
```

david-luna:
@trentm sorry for the big change I did in the last commits. I did not crate a new PR since these changes address the questions/comments you wrote. We can discuss it later today

.tav.yml
trentm: Either form (caret or a range with `>... <...') sounds good to me.

.tav.yml
trentm: Perhaps prefer the `>3.2.2 <4` style? At least in the case where you want to catch a new major release, it can handle it. For example:   `^1.2.3` will not catch a possible new `2.x` release. However only can write `>=1.2.3` or `>=1.2.3 < 2` or `>=1.2.3 <3` to specify a larger range as needed.

trentm: Caret if fine though unless we run into a case where we *need* the "next major" range.

.tav.yml
david-luna: I'll go for caret :)

dev-utils/update-tav-versions.js
david-luna: comment updated

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,224 @@
...
+ * @property {String} node
+ * @property {String[]} commands
+ * @property {Object} update-versions
+ * @property {String} update-versions.mode
+ * @property {String} update-versions.range
+ */
+
+'use strict';
+
+// This script scans the `tav.yml` file placed at the root folder and for each
```
trentm, line 23: nit:

```suggestion
// This script scans the `.tav.yml` file placed at the root folder and for each
```

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,224 @@
...
+// entry checks if the `update-versions` property is defined. In that case
+// the script will update the versions property according to the `update-versions`
+// configuration:
+// - range: only versions satisfiying this range will be in the update
```
trentm, line 27: This should be updated to discuss `include`, `exclude` instead now.

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,224 @@
...
+/** @typedef {import('semver').SemVer} SemVer */
+/**
+ * @typedef {Object} TavConfig
+ * @property {String} [name]
+ * @property {String} versions
+ * @property {String} node
+ * @property {String[]} commands
+ * @property {Object} update-versions
+ * @property {String} update-versions.mode
+ * @property {String} update-versions.range
```
trentm, line 18: Is `.range` type wrong now?

dev-utils/update-tav-versions.js
```diff
@@ -0,0 +1,224 @@
...
+        versions = getMax(filteredVers, Number(num));
+      }
+    } else {
+      console.error(
+        `Error: Version selection mode for TAV config ${name} unknown (${mode})`,
+      );
+      continue;
+    }
+
+    // Assuming `include` is always in the form ">={Lower_limit} <{Higher_Limit}"
```
trentm, line 107: nit: Might be nice to check that and throw if it isn't the case.

trentm (approved):
Nice! Just a couple nits.

dev-utils/update-tav-versions.js
david-luna: I've added the format validations at the top of the loop so the script throws before doing an expensing operation like fetching the list of versions from npm.

dev-utils/update-tav-versions.js
david-luna: Yes! Thanks for pointing it out :)

dev-utils/update-tav-versions.js
david-luna: done!

trentm (approved):

PR #3690, opened 2023-10-23
chore: test with Node v21

trentm:
Node.js v21 was first released on 2023-10-17. v22 nightlies should
start coming as well.


trentm (commented):
/test tav all 21

david-luna (approved):

Issue #3691, opened 2023-10-24
AsyncResource.bind usage in mongodb instrumentation is broken for node v14 and earlier


trentm:
Recently the TAV=mongodb tests for node v14 have started failing, e.g. https://github.com/elastic/apm-agent-nodejs/actions/runs/6567481980/job/17840287524

A repro of the failure:

```
% npm install --no-save mongodb@5
% nvm use 14
Now using node v14.21.2 (npm v6.14.17)

% node test/instrumentation/modules/mongodb/fixtures/use-mongodb-connect.js
...
[...]

PR #3692, opened 2023-10-24
fix: usage of AsyncResource.bind in mongodb instrumentation for Node.js v14 and earlier

trentm:
Fixes: #3691


trentm (commented):
/test tav mongodb

david-luna (approved):

Issue #3182, opened 2023-03-02
mongodb instrumentation should set `span.outcome` and `captureError()` on `commandFailed`


trentm:
From https://github.com/elastic/apm-agent-nodejs/pull/3168#issuecomment-1451859280

> About span outcome - the same `onEnd` event handler is in use for `commandSucceeded` and `commandFailed` events:
> 
> ```js
>           client.on('commandSucceeded', onEnd)
>           client.on('commandFailed', onEnd)
> ```
> 
> And it's possible to set outcome by checking `event.failure` like:
> 
> ```js
>   function onEnd (event) {
>     if (!activeSpans.has(event.requestId)) return
>     const span = activeSpans.get(event.requestId)
>     activeSpans.delete(event.requestId)
>     span.setOutcome(event.failure ? 'failure' : 'success')
>     span.end((span._timer.start / 1000) + event.duration)
>   }
> ```
> 
> Doc: [v4.0](https://www.mongodb.com/docs/drivers/node/v4.0/fundamentals/monitoring/command-monitoring/#commandfailed) [v5.0](https://www.mongodb.com/docs/drivers/node/current/fundamentals/monitoring/command-monitoring/#commandfailed)

This diff to "examples/trace-mongodb.js" shows a way to get a `commandFailed` event -- a bogus "hint" to a `.deleteOne()` call (https://www.mongodb.com/docs/manual/reference/method/db.collection.deleteOne/#std-label-ex-deleteOne-hint),

```diff
diff --git a/examples/trace-mongodb.js b/examples/trace-mongodb.js
index 2f4f8992..733d8241 100755
--- a/examples/trace-mongodb.js
+++ b/examples/trace-mongodb.js
@@ -42,6 +42,12 @@ async function main () {
     ])
     console.log('insertMany:', res)

+    try {
+      await coll.deleteOne({ item: 'ham' }, { hint: 'foo' })
+    } catch (err) {
+      console.log('error in .deleteOne() with bogus "hint":', err)
+    }
+
     res = await coll.findOne({ item: 'eggs' })
     console.log('findOne eggs:', res)

```

Running that:

```
% node examples/trace-mongodb.js
insertMany: {
  acknowledged: true,
  insertedCount: 3,
  insertedIds: {
    '0': new ObjectId("6400e8780cc1a4a785975123"),
    '1': new ObjectId("6400e8780cc1a4a785975124"),
    '2': new ObjectId("6400e8780cc1a4a785975125")
  }
[...]

PR #3695, opened 2023-10-27
fix: set outcome for mongodb spans

david-luna:
This PR sets span outcome for `mongodb` commands. The tests realised (see related issue) reveals that a failed command triggers:
- a **CommandFailed** event with a property named `failure`
- a **CommandSucceeded** event with the a property named `reply`. The presence of a sub-property named `writeErrors` indicates the command failed.

Closes: #3182 

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -61,7 +61,7 @@ See the <<upgrade-to-v4>> guide.
 * Fix `mongodb` instrumentation to avoid loosing context when multiple cursors
   are running concurrently. ({issues}3161[#3161])
 
-* Set `mongodb` span's outcome acordingly to the result of the command being traced.
+* Set `mongodb` span's outcome according to the result of the command being traced.
   ({pull}3695[#3695])
 
 [float]
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -61,6 +61,8 @@ See the <<upgrade-to-v4>> guide.
 * Fix `mongodb` instrumentation to avoid loosing context when multiple cursors
   are running concurrently. ({issues}3161[#3161])
 
+* Set `mongodb` span's outcome acordingly to the result of the command being traced.
+  ({pull}3695[#3695])
 
 [float]
 ===== Chores
--- a/lib/instrumentation/modules/mongodb.js
+++ b/lib/instrumentation/modules/mongodb.js
@@ -33,8 +33,8 @@ module.exports = (mongodb, agent, { version, enabled }) => {
   if (mongodb.instrument) {
     const listener = mongodb.instrument();
     listener.on('started', onStart);
-    listener.on('succeeded', onEnd);
-    listener.on('failed', onEnd);
+    listener.on('succeeded', onSuccess);
+    listener.on('failed', onFailure);
   } else if (mongodb.MongoClient) {
     // mongodb 4.0+ removed the instrument() method in favor of
     // listeners on the instantiated client objects. There are two mechanisms
@@ -52,8 +52,8 @@ module.exports = (mongodb, agent, { version, enabled }) => {
         }
         super(...args);
         this.on('commandStarted', onStart);
-        this.on('commandSucceeded', onEnd);
-        this.on('commandFailed', onEnd);
+        this.on('commandSucceeded', onSuccess);
+        this.on('commandFailed', onFailure);
         this[kListenersAdded] = true;
       }
     }
@@ -103,8 +103,8 @@ module.exports = (mongodb, agent, { version, enabled }) => {
             } else {
               if (!client[kListenersAdded]) {
                 client.on('commandStarted', onStart);
-                client.on('commandSucceeded', onEnd);
-                client.on('commandFailed', onEnd);
+                client.on('commandSucceeded', onSuccess);
+                client.on('commandFailed', onFailure);
                 client[kListenersAdded] = true;
               }
               callback(err, client);
@@ -116,8 +116,8 @@ module.exports = (mongodb, agent, { version, enabled }) => {
         p.then((client) => {
           if (!client[kListenersAdded]) {
             client.on('commandStarted', onStart);
-            client.on('commandSucceeded', onEnd);
-            client.on('commandFailed', onEnd);
+            client.on('commandSucceeded', onSuccess);
+            client.on('commandFailed', onFailure);
             client[kListenersAdded] = true;
           }
         });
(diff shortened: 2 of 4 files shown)


david-luna (commented):
/test tav mongodb

david-luna (commented):
/test tav mongodb

david-luna:
TAV tests failing for `mongo@3..5.11` on the bogus deleteOne test because it does not report error:
- event fired is a `success`
- there is no `writeError` in the reply

I think it was not considered an error before so:
- should we adapt the test?
- or should we reduce our versions to test?

CHANGELOG.asciidoc
```diff
@@ -61,6 +61,8 @@ See the <<upgrade-to-v4>> guide.
 * Fix `mongodb` instrumentation to avoid loosing context when multiple cursors
   are running concurrently. ({issues}3161[#3161])
 
+* Set `mongodb` span's outcome acordingly to the result of the command being traced.
```
trentm, line 64:
```suggestion
* Set `mongodb` span's outcome according to the result of the command being traced.
```

trentm:
> I think it was not considered an error before so:
> 
>     * should we adapt the test?
> 
>     * or should we reduce our versions to test?

If we could find a reliable way to have a mongo client failure for testing, that would be great. I would try for a little bit to find one, but if you cannot after 5-10 minutes, then it is fine to reduce the error test to just the versions on which we know how to trigger one.

lib/instrumentation/modules/mongodb.js
```diff
@@ -176,8 +177,11 @@ module.exports = (mongodb, agent, { version, enabled }) => {
 
   function onEnd(event) {
     if (!activeSpans.has(event.requestId)) return;
+    const { failure, reply } = event;
+    const hasFailed = failure || (reply && reply.writeErrors);
     const span = activeSpans.get(event.requestId);
     activeSpans.delete(event.requestId);
+    span.setOutcome(hasFailed ? OUTCOME_FAILURE : OUTCOME_SUCCESS);
```
trentm, line 184: See my comment at https://github.com/elastic/apm-agent-nodejs/issues/3182#issuecomment-1783217513

david-luna: I like the idea of having a specific function fir each type of event. It will prepare the code for the changes you're proposing i the long term.

david-luna:
> If we could find a reliable way to have a mongo client failure for testing, that would be great.

I've tried both promise and callback ways
```js
// with Promises
try {
  await collection.deleteOne({ item: 'eggs' }, { hint: 'foo' });
  console.info('success in .deleteOne() with bogus "hint"'); // <- it call this log
} catch (err) {
  console.info('error in .deleteOne() with bogus "hint"');
}

// with callback
collection.deleteOne({ item: 'eggs' }, { hint: 'foo' }, function (err, res) {
  if (err) {
    console.log('error on callback', err);
  } else {
    console.log('no error on callback', res); // <- it call this log
  }
});
```

And inspecting the `CommandSuccessEvent` I don't see any hint

```js
// both Promises & callbacks return the same shape
{
  "connectionId": "localhost:27017",
  "requestId": 1,
  "commandName": "delete",
  "duration": 1,
  "reply": {
    "n": 0,
    "ok": 1
  }
}
```

I think we can tweak the test to take the version number in consideration while asserting the spans so we could keep testing in `<=3.5.11`.

david-luna (commented):
/test tav mongodb

lib/instrumentation/modules/mongodb.js
```diff
@@ -175,13 +173,49 @@ module.exports = (mongodb, agent, { version, enabled }) => {
...
     if (!activeSpans.has(event.requestId)) return;
-    const { failure, reply } = event;
-    const hasFailed = failure || (reply && reply.writeErrors);
+
+    const span = activeSpans.get(event.requestId);
+    activeSpans.delete(event.requestId);
+
+    // From mongodb@3.6.0 and up some commands like `deleteOne` may contain
+    // error data inside the `reply` property. It makes sense to capture it.
+    const writeErrors = event && event.reply && event.reply.writeErrors;
```
trentm, line 193: FWIW, I think we could use optional chaining with our new min-node version now: https://node.green/#ES2020-features-optional-chaining-operator-----
No need to. Just saying it is an option for our new code now.

lib/instrumentation/modules/mongodb.js
```diff
@@ -175,13 +173,49 @@ module.exports = (mongodb, agent, { version, enabled }) => {
...
+
+    if (writeErrors && writeErrors.length) {
+      // TODO: this gets linked to the parent transaction. How to link it to the span?
```
trentm, line 196: `captureError` has a `parent` option for just this case:

```
//   - `opts.parent` - A Transaction or Span instance to make the parent of
//     this error. If not given (undefined), then the current span or
//     transaction will be used. If `null` is given, then no span or transaction
//     will be used.
```

I guess the emitted `CommandSucceededEvent`s (and the other events) are not in the context of the created span, which is fine.

test/instrumentation/modules/mongodb/mongodb.test.js
```diff
@@ -43,6 +43,9 @@ const TEST_DB = 'elasticapm';
 const TEST_COLLECTION = 'test';
 const TEST_USE_CALLBACKS = semver.satisfies(MONGODB_VERSION, '<5');
 
+// From mongodb@3.6.0 and up CommandSuccessEvents may contain errors
+const MONGODB_SUCESS_WITH_ERRORS = semver.satisfies(MONGODB_VERSION, '>=3.6.0');
```
trentm, line 47: nit: typo

```suggestion
const MONGODB_SUCCESS_WITH_ERRORS = semver.satisfies(MONGODB_VERSION, '>=3.6.0');
```

lib/instrumentation/modules/mongodb.js
david-luna: IMO this a proper use case for optional chaining. I'll add it in the next commit. Thanks for pointing that out :)

test/instrumentation/modules/mongodb/mongodb.test.js
david-luna: At this pace I'll earn a `typo master` badge from github 😅

lib/instrumentation/modules/mongodb.js
david-luna: done!

david-luna (commented):
/test tav mongodb

trentm (approved):

PR #3705, opened 2023-11-02
chore: enable dependabot version updates of github-actions yaml files

v1v:
## What does this PR do?

* Update dependabot to run on [Sundays](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups) and use [groups](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups).
* Add dependabot for github actions.

Unfortunately it's not possible to pass different directories yet, see https://github.com/dependabot/dependabot-core/issues/2178

## Why is it important?

Reduce the noise during the week and create one PR with all the changes related to the GitHub actions/composites


.github/dependabot.yaml
trentm, line 1: This repo already has a "dependabot.yml" file. This block will need to be added to that file.

Aside: was there a reason to use ".yaml" instead of the ".yml" suffix that GH docs (e.g. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file) refer to?

.github/dependabot.yaml
```diff
@@ -0,0 +1,16 @@
+version: 2
+updates:
+
+  # GitHub actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    reviewers:
+      - "elastic/observablt-ci"
```
trentm, line 8: FWIW, I'd be happy for this to be or to include "elastic/apm-agent-node-js" for reviewing these updates.

.github/dependabot.yaml
```diff
@@ -0,0 +1,16 @@
...
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "22:00"
```
trentm, line 12: Is the reason to specify Sunday and a time so that all `package-ecosystem: "github-actions"` dependabot PRs across Obs repos tend to be at the same time?

trentm (changes requested):

trentm:
> this means all package updates will come at once in a single PR right?

No. The title was a bit misleading. IIUC, this is about telling dependabot to *also* do version update PRs for github-actions used in our .github/workflows/*.yml files.  And yes, *those* updates will be grouped into a single PR. The dependabot "group:"-config for grouping will only group updates for a single ["package-ecosystem"](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem) -- i.e. version update PRs for "github-actions" will be separate from those for "npm".

.github/dependabot.yaml
v1v: I missed that file... 👼

.github/dependabot.yaml
v1v: to help with knowing mondays are the days for reviewing those dependencies

.github/dependabot.yaml
v1v: > Aside: was there a reason to use ".yaml" instead of the ".yml" suffix that GH docs

Not really, in some places I saw `yaml` to help with the full name extension. No strong opinion, it supports both anyway

PR #3715, opened 2023-11-09
chore: unify configuration options definition

david-luna:
PR to unify the way configuration options are defined and processed.

NOTE: this is a draft and so its description, do not review yet

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update TypeScript typings
- [ ] Update documentation
- [ ] Add CHANGELOG.asciidoc entry
- [x] Commit message follows [commit guidelines](https://github.com/elastic/apm-agent-nodejs/blob/main/CONTRIBUTING.md#commit-message-guidelines)

PR #3733, opened 2023-11-15
chore: move preamble data function to config schema

david-luna:
Follow up PR to refactor the configuration of the agent. this time we use the new schema to extract the data that is shown in the log preamble.

### Checklist

- [x] Refactor code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -77,9 +77,6 @@ const NICE_REGEXPS_FIELDS = {
 
 const confDefaultOptions = getDefaultOptions();
 const confEnvOpts = getEnvironmentOptions();
-let confFilePath = path.resolve(
-  confEnvOpts.configFile || confDefaultOptions.configFile,
-);
 let confFileOpts = getFileOptions();
 
 // Configure a logger for the agent.
--- a/lib/config/config.js
+++ b/lib/config/config.js
@@ -33,6 +33,7 @@ const {
   getFileOptions,
   setStartOptions,
   getPreambleData,
+  readEnvOptions,
 } = require('./schema');
 
 const {
@@ -122,22 +123,18 @@ function createConfig(opts, logger) {
 
 class Config {
   constructor(opts, logger) {
+    // TODO: remove this call once config and logging-preamble tests are refactored
+    readEnvOptions();
+
     const isLambda = isLambdaExecutionEnvironment();
     const envOptions = getEnvironmentOptions();
-
-    // If we didn't find a config file on process boot, but a path to one is
-    // provided as a config option, let's instead try to load that
-    if (confFileOpts === null && opts && opts.configFile) {
-      confFilePath = path.resolve(opts.configFile);
-      confFileOpts = getFileOptions(opts.configFile);
-    }
 
     setStartOptions(opts);
 
     Object.assign(
       this,
       confDefaultOptions, // default options
-      confFileOpts, // options read from config file
+      getFileOptions(), // options read from config file
       opts, // options passed in to agent.start()
       envOptions, // options read from environment variables
     );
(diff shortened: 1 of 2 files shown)


lib/config/config.js
```diff
@@ -112,16 +109,6 @@ function configLogger(opts = {}) {
 function initialConfig(logger) {
   const cfg = Object.assign({}, confDefaultOptions);
 
-  // Reproduce the generated properties for `Config`.
```
david-luna, line 115: Note to reviewer: leftover from the [previous PR](https://github.com/elastic/apm-agent-nodejs/pull/3715). The schema already has values for this internal properties

lib/config/config.js
```diff
@@ -465,118 +435,6 @@ function truncateOptions(opts) {
...
- * - sourceValue: the raw value provided im the source
- * - normalizedValue: the value normalized by the configuration which is used in the agent
- * - file: if source === 'file' this proprty is defined with the path of the config file
- *
- * @param {Object} config The configuration object with normalized options
- * @param {Array<{ name: string, opts: Object}>} sources The list of sources of options ordered by priority, highest priority first.
- * @param {String | undefined} configFilePath Path of the config file if loaded
- * @returns {Object} Object with agent, environment, and configuration data.
- */
-function getLoggingPreambleData(config, sources, configFilePath) {
```
david-luna, line 482: Note for reviewer: this is the function moved and refactored. There are tests that check them implementation in `./tests/logging-preamble.test.js`

trentm (approved):
I like the `redactFn` addition to CONFIG_SCHEMA.

I'm a little uncomfortable with using CONFIG_SCHEMA to hold configuration values, and that functions named `get*`, like `getFileOptions`, now have the side-effect of modifying the CONFIG_SCHEMA global.

However, given the eventual goal is merging these two files, and these are all internal implementation details in `lib/config/...`, it is fine.

david-luna:
> I'm a little uncomfortable with using CONFIG_SCHEMA to hold configuration values, and that functions named `get*`, like `getFileOptions`, now have the side-effect of modifying the CONFIG_SCHEMA global.

I do agree on the side effects so I'm going to remove them in this PR. Gathering env and file options could be done at the same time the module is checking for dependency integrity

And for the schema I think we could agree to:
- leave `CONFIG_SCHEMA` pristine only with data to define the options (name, enVar, redatcFn...)
- create a new object which holds the values from different sources and its normalized value (data needed for preamble)
- a new getter which returns the final config object that is assigned to `agent._config`

this will be done in a next PR

lib/config/config.js
```diff
@@ -135,30 +123,18 @@ function createConfig(opts, logger) {
...
-    this.ignoreUrlStr = [];
-    this.ignoreUrlRegExp = [];
-    this.ignoreUserAgentStr = [];
-    this.ignoreUserAgentRegExp = [];
-    this.elasticsearchCaptureBodyUrlsRegExp = [];
-    this.transactionIgnoreUrlRegExp = [];
-    this.sanitizeFieldNamesRegExp = [];
-    this.ignoreMessageQueuesRegExp = [];
+    // TODO: remove this call once config and logging-preamble tests are refactored
+    readEnvOptions();
```
david-luna, line 127: Note for reviewer: env should be read once but some tests are setting env dinamically to tests different scenarios so its mandatory to read again. In a follow up PR I will move the tests to fixtures (using `runTestFixtures`) so we have better control of the env.

lib/config/config.js
```diff
@@ -135,30 +123,18 @@ function createConfig(opts, logger) {
...
     const isLambda = isLambdaExecutionEnvironment();
     const envOptions = getEnvironmentOptions();
 
-    // If we didn't find a config file on process boot, but a path to one is
-    // provided as a config option, let's instead try to load that
-    if (confFileOpts === null && opts && opts.configFile) {
-      confFilePath = path.resolve(opts.configFile);
-      confFileOpts = getFileOptions(opts.configFile);
-    }
+    setStartOptions(opts);
```
david-luna, line 132: Note for reviewer: now `setStartOptions` handles if a new config file is set during agent start so no need for the previous code.

lib/config/config.js
trentm: FWIW, I'm okay with env options being read from `process.env` when `Config`s constructor is called. I think it should be okay for a *user* to tweak `process.env.*` before importing and starting the APM agent.

trentm (approved):

david-luna:

PR #3751, opened 2023-11-24
chore: refactor test of config options from `start` and `env`

david-luna:
This PR tries to make easier the testing of config by
- running the tests in a separate process
- querying the schema instead of hardcoded arrays to check for options

The tests moved are the ones checking the defaults, env vars & priority of env over start & file options.

NOTE: the point of using `runTestFixtures` is to be as close as possible to a real scenario and also to make sure test are running in isolation so no leftover may affect other tests

### Checklist

- [ ] Implement code
- [x] Refactor tests
- [ ] Update documentation


test/config/config.test.js
```diff
@@ -0,0 +1,295 @@
...
+    timeout: 20000, // sanity guard on the test hanging
+    maxBuffer: 10 * 1024 * 1024, // This is big, but I don't ever want this to be a failure reason.
+    verbose: true,
+    checkScriptResult: (t, err, stdout) => {
+      t.error(err, `use-agent.js script succeeded: err=${err}`);
+
+      const useAgentLogs = stdout
+        .split('\n')
+        .filter((l) => l.startsWith('use-agent log:'))
+        .map((l) => l.replace('use-agent log:', ''));
```
trentm, line 64: nit: I guess this could use `getUseAgentLogs()`?

test/config/config.test.js
```diff
@@ -0,0 +1,295 @@
...
+// ---- tests
+
+/** @type {import('../_utils').TestFixture[]} */
+const testFixtures = [
+  {
+    name: 'use agent - shoud have defaults',
+    script: 'fixtures/use-agent.js',
+    cwd: __dirname,
+    noConvenienceConfig: true, // we want full control of the env
+    timeout: 20000, // sanity guard on the test hanging
```
trentm, line 55: nit: Is the 20s timeout something you've found necessary? Or copypasta? There is a default of 10s on `runTestFixtures`. If 20s is commonly needed in our CI, we could bump the default.

test/config/config.test.js
```diff
@@ -0,0 +1,295 @@
...
+    maxBuffer: 10 * 1024 * 1024, // This is big, but I don't ever want this to be a failure reason.
```
trentm, line 56: nit: Similarly, I wonder if we should bump the default here. The default maxBuffer (in Node.js's spawn/exec) is 1MiB. I don't know if we get close to that typically. I *do* recall hitting it once in one test when I was turning on lots of debug/trace logging in the fixture, which is why I had that  `This is big, but I don't ever want this to be a failure reason.` case.   (This can be done in a separate PR.)

test/config/config.test.js
```diff
@@ -0,0 +1,295 @@
...
+          t.deepEqual(
+            resolvedValue,
+            expectedValue,
+            `${key} config is resolved to its environment value`,
+          );
+        });
+    },
+  },
+  {
+    name: 'use agent - shoud have prioroty of env, start, file',
```
trentm, line 251: nit:
```suggestion
    name: 'use agent - should have priority of env, start, file',
```

test/config/config.test.js
```diff
@@ -0,0 +1,295 @@
...
+/** @type {import('../_utils').TestFixture[]} */
+const testFixtures = [
+  {
+    name: 'use agent - shoud have defaults',
+    script: 'fixtures/use-agent.js',
+    cwd: __dirname,
+    noConvenienceConfig: true, // we want full control of the env
+    timeout: 20000, // sanity guard on the test hanging
+    maxBuffer: 10 * 1024 * 1024, // This is big, but I don't ever want this to be a failure reason.
+    verbose: true,
```
trentm, line 57: nit: was `verbose: true` for debugging here?

trentm (approved):
If you just want to move ahead quickly with the commit for other refactoring work, definitely feel free to skip over all the nits and just merge.

david-luna:
> feel free to skip over all the nits and just merge.

I'm going to merge this one and add all of them in the next PR. Thanks for your thorough review :)

Issue #3759, opened 2023-11-27
Add configuration option to allow injecting headers on all outgoing requests


lukeelmers:
We can use `apm?._apmClient?.config({ headers: { Foo: 'Baz' } });` to inject headers in requests from the agent, however it doesn't apply these headers to the first two requests that are sent when bootstrapping:
1. `GET $apmServer/`
2. `GET $apmServer/config/v1/agents?...` if using central config

It would be great to have a configuration option that could be set before any requests are sent, thus giving us a guarantee that all outgoing requests will have these headers.

Our use case is running an app in an environment where outbound network traffic is carefully controlled, and we need to add some special headers to ensure APM requests are allowed through.

PR #3760, opened 2023-11-27
feat: add 'apmClientHeaders' config option for custom headers in APM server comms

trentm:
Closes: #3759

---

I haven't added any tests here. FWIW there are existing tests in the *http-apm-client tests* for the `headers` config option that we are using here:
https://github.com/elastic/apm-agent-nodejs/blob/f590ef67c5b016520b844ce3a87164a4b7bbbfbd/test/apm-client/http-apm-client/config.test.js#L171-L194

I don't think there is much value in a test case that a `apmClientHeaders` value given to `apm.start()` makes it to the HttpApmClient() instance. For example, there isn't an equivalent test for these related config vars being passed through:

```js
    rejectUnauthorized: conf.verifyServerCert,
    serverTimeout: conf.serverTimeout * 1000,
```

### DIFF
--- a/examples/trace-http-request.js
+++ b/examples/trace-http-request.js
@@ -45,10 +45,6 @@ function makeARequest(url, opts, cb) {
 // an HTTP server, we manually start a transaction. More details at:
 // https://www.elastic.co/guide/en/apm/agent/nodejs/current/custom-transactions.html
 const t0 = apm.startTransaction('t0');
-makeARequest(
-  'http://httpstat.us/200',
-  { headers: { accept: '*/*' } },
-  function () {
-    t0.end();
-  },
-);
+makeARequest('http://google.com/', { headers: { accept: '*/*' } }, function () {
+  t0.end();
+});
--- a/lib/config/schema.js
+++ b/lib/config/schema.js
@@ -590,6 +590,12 @@ const CONFIG_SCHEMA = [
     defaultValue: [],
     deps: ['ignoreUserAgents'],
   },
+  {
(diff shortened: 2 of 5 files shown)


examples/trace-http-request.js
```diff
@@ -45,10 +45,6 @@ function makeARequest(url, opts, cb) {
...
 // https://www.elastic.co/guide/en/apm/agent/nodejs/current/custom-transactions.html
 const t0 = apm.startTransaction('t0');
-makeARequest(
-  'http://httpstat.us/200',
-  { headers: { accept: '*/*' } },
-  function () {
-    t0.end();
-  },
-);
+makeARequest('http://google.com/', { headers: { accept: '*/*' } }, function () {
```
trentm, line 48: Reviewer note: I changed the URL used for this example, because httpstat.us (at least for HTTP) has been hanging for me of late.

lib/config/schema.js
```diff
@@ -590,6 +590,12 @@ const CONFIG_SCHEMA = [
     defaultValue: [],
     deps: ['ignoreUserAgents'],
   },
+  {
+    name: 'apmClientHeaders',
+    configType: 'stringKeyValuePairs',
+    defaultValue: undefined,
+    envVar: 'ELASTIC_APM_APM_CLIENT_HEADERS',
+  },
```
trentm, line 598: BTW, kudos to your config refactoring so far, David. Having this change be localized to just this is wonderful.

david-luna (approved):

PR #3762, opened 2023-11-28
chore: another batch of config tests being refactored

david-luna:
This PR refactors tests for:
- truthy/falsy values
- byte values
- -1 === Infinity values
- key/value pairs values

### Checklist

- [ ] Implement code
- [x] Refactor tests
- [ ] Update documentation

### DIFF
--- a/test/config/config.test.js
+++ b/test/config/config.test.js
@@ -316,12 +316,12 @@ const testFixtures = [
       t.equal(
         resolvedConfig.breakdownMetrics,
         false,
-        'boolean value is parsed',
+        '"false" value for breakdownMetrics is parsed',
       );
       t.equal(
         resolvedConfig.apiRequestSize,
         1024 * 1024,
-        'bytes value is parsed for apiRequestSize',
+        '"1mb" value is parsed for apiRequestSize',
       );
       t.equal(
         warnLogs[0].message,
@@ -366,7 +366,7 @@ const testFixtures = [
       KUBERNETES_POD_NAME: 'kube-pod-name',
       KUBERNETES_POD_UID: 'kube-pod-id',
     },
-    verbose: true,
+    verbose: false,
     checkScriptResult: (t, err, stdout) => {
       t.error(err, `use-agent.js script succeeded: err=${err}`);
       const useAgentLogs = getUseAgentLogs(stdout);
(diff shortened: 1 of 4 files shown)


test/config/config.test.js
```diff
@@ -287,6 +286,141 @@ const testFixtures = [
...
+        warnLogs[1].message,
+        'unrecognized boolean value "nope" for "active"',
+        'got a warning about bogus boolean value',
+      );
+      t.equal(
+        debugLogs[0].message,
+        'Elastic APM agent disabled (`active` is false)',
+        'got a debug log about agent disabled',
+      );
+      // XXX: normalization turns this value to `undefined` because "bogus"
```
david-luna, line 345: Note to reviewer: since `undefined` is falsy we may want to actually have `false` so we ensure we have the right type

test/config/config.test.js
```diff
@@ -43,6 +43,19 @@ function getUseAgentLogs(stdout) {
     .map((l) => l.replace(prefix, ''));
 }
 
+/**
+ * Returns the warnings emitted by the logger
```
trentm, line 47: nit: The description isn't quite accurate about it being specifically about "warn" level.

test/config/config.test.js
trentm: I think `undefined` should be treated the same as "not specified". What happens if the same config option is specified at multiple levels, e.g.: `ELASTIC_APM_ACTIVE=nope` and `apm.start({ active: true })`. Normally the envvar is a high priority, so it should win. But in this case, I think we'd expect to see:

- a warning that the `nope` value is invalid, and hence ignored
- active is `true` because the value from `apm.start(...)` was used.

I'm not sure if that answers your question. :)

trentm: nit: Can we avoid commiting "XXX", or at least for very long, please? :) That's a string I use heavily as a marker for "this code is for local debugging/dev notes that is incomplete and shouldn't go in the release".

trentm (approved):

test/config/config.test.js
david-luna: sorry I left this on the code :(

I'll merge this one and create a new PR today moving more tests and removing this comment.

PR #3783, opened 2023-12-11
test: lock Node.js v18 testing to v18.18.2 as a workaround for IITM double-import issue

trentm:
In Nodejs v18.19.0 ESM module loading was moved to being off-thread
(a backport from Node.js v20, https://github.com/nodejs/node/pull/44710).
This breaks current import-in-the-middle usage, resulting in ESM modules
being double-loaded when the IITM hook is active (once on the loader
worker thread, and then again later on the main thread).

This change is a workaround to skip testing recent Node.js v18
versions until the IITM issue is resolved..


david-luna (approved):

trentm (commented):
/test tav knex

Issue #2905, opened 2022-08-31
Support for KafkaJS publishers and consumers


michaelhyatt:
It would be great to have NodeJS agent support [KafkaJS](https://kafka.js.org/) publishers and consumers to create transactions and spans, as well as trace context propagation.

astorm:
@michaelhyatt is there a specific Node.js kafka library you're interested in supporting? It looks like there's a few of them.

michaelhyatt:
Yep, https://kafka.js.org/

Gauravism2017:
Are there any progress/plans to support this library?

trentm:
@Gauravism2017 No update on this currently. Comments/notes/reactions on this issue (like yours) will help decide if we can prioritize this work.

PR #3786, opened 2023-12-12
feat: add kafkajs instrumentation

david-luna:
PR to add `kafkajs` instrumentation. Summary of additions
- instrument `send`& `sendBatch` methods of `kafkajs producers`
  -  creating a spans if apply (topic not ignored by config)
  - propagating traceparent info into message headers

- instruments `eachMessage` of `kafkajs consumers`
  - creating a transaction if apply (topic not ignored by config)
  - setting the `trace_id` if traceparent present in message headers
- instruments `eachBatch` of `kafkajs consumers`
  - creating a transaction if apply (topic not ignored by config)
  - setting the transaction `links` if traceparent present in message headers

Closes: #2905 

### Checklist

- [x] Implement code
- [x] Add tests
- [x] Update documentation


david-luna:
note: concurrency is an issue that make test fail

oleh-kravchuk:
Hi, wondering about any updates for this MR, when it will be merged approx?

oleh-kravchuk:
@trentm Thanks for the quick reply

We are curious in the new solution (native kafkajs support)
cuz currently we reused the next approach:

```
import { registerInstrumentations } from "@opentelemetry/instrumentation";
import { NodeTracerProvider } from "@opentelemetry/sdk-trace-node";
import { KafkaJsInstrumentation } from "opentelemetry-instrumentation-kafkajs";
import { Span } from "@opentelemetry/api";

const tracerProvider = new NodeTracerProvider();
tracerProvider.register();

const instrumentations = [];
instrumentations.push(
  new KafkaJsInstrumentation({
    producerHook: function (span: Span, _topic: string, message) {
      span.setAttribute("kafka_producer_hook_attribute", customParser(message));
    },
    consumerHook: function (span: Span, _topic: string, message) {
       span.setAttribute("kafka_consumer_hook_attribute", customParser(message));
    }
  })
);

registerInstrumentations({ instrumentations });
```

We expect a native support will speed up the initialisation phase and simplify the code base, than this one ^.

Could you compare both ways and share with us pros&cons? Thanks in advance

max-zirkonzahn:
Hi there, are there any news / updates on the state of this MR?

ImadSai:
I'm interested too about this feature.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,451 @@
...
+  function wrapEachBatch(origEachBatch) {
+    return async function ({ batch }) {
+      if (shouldIgnoreTopic(batch.topic, config)) {
+        return origEachBatch.apply(this, arguments);
+      }
+
+      const trans = ins.startTransaction(`${NAME} RECEIVE from batch`, TYPE);
+
+      // TODO: not sure if this should be here but we got batches for only one topic
+      // https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#context-fields
```
trentm, line 198: Sounds good. It looks like a batch can only receive messages from a single topic. I thikn you can remove this TODO.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,451 @@
...
+   * @param {ConsumerRunConfig['eachBatch']} origEachBatch
+   * @returns {ConsumerRunConfig['eachBatch']}
+   */
+  function wrapEachBatch(origEachBatch) {
+    return async function ({ batch }) {
+      if (shouldIgnoreTopic(batch.topic, config)) {
+        return origEachBatch.apply(this, arguments);
+      }
+
+      const trans = ins.startTransaction(`${NAME} RECEIVE from batch`, TYPE);
```
trentm, line 195:
```suggestion
      const trans = ins.startTransaction(`${NAME} RECEIVE from ${batch.topic}`, TYPE);
```

?   I'm not sure if there is a way to distinguish this transaction name that is receiving a batch from a transaction name for receiving a *single message*?  The spec doesn't really say: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#naming
Maybe a non-issue for now.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,451 @@
...
+      };
+      trans.setServiceContext(serviceContext);
+
+      // Extract span links from up to 1000 messages in this batch.
+      // https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#receiving-trace-context
+      // A span link is created from a `traceparent` header in a message.
+      const messages = batch && batch.messages;
+
+      if (messages) {
+        const links = [];
```
trentm, line 213: See my comment on the test case for this. Currently this is resulting in duplicate links (same trace_id and span_id) because it is receiving messages sent from the same outgoing span. Perhaps use a `Set` of seen traceparent values.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,451 @@
...
+      const parentSpan =
+        span || runContext.currSpan() || runContext.currTransaction();
+
+      if (parentSpan) {
+        record.messages.forEach((msg) => {
+          const newHeaders = Object.assign({}, msg.headers);
+          parentSpan.propagateTraceContextHeaders(
+            newHeaders,
+            function (carrier, name, value) {
+              // TODO: why doing it everywhere???
```
trentm, line 302: Because there is a share impl of `propagateTraceContextHeaders` in lib/tracecontext/index.js that is meant to handle bwcompat cases supporting `elastic-apm-traceparent`.  The idea here is that even if `useElasticTraceparentHeader: true` we still don't want to propagate that header for this *new* instrumentation -- only for HTTP instrumentation.  At least in SQS there is a low limit on the number of headers, so adding `elastic-apm-traceparent` that will almost certainly never get used could cause harm.

It certainly does feel silly to have this check everywhere however. Perhaps `.propagateTraceContextHeaders` could grow a `{legacyHeaders: true}` option (or a separate method for this) that would only be called in the HTTP instrumentation. Only that code path would consider adding `elastic-apm-traceparent`.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,451 @@
...
+   * - creates an exit span for the operation
+   * - propagates trace context via message headers
+   *
+   * @param {Producer['sendBatch']} origSendBatch
+   * @returns {Producer['sendBatch']}
+   */
+  function wrapProducerSendBatch(origSendBatch) {
+    return async function (batch) {
+      // TODO: discuss this. if all topics are ignored should the agent avoid
+      // creating a span?
```
trentm, line 360: Yes, I think so.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,451 @@
...
+        });
+      }
+
+      if (!span) {
+        return origSendBatch.apply(this, arguments);
+      }
+
+      const destCtx = {
+        // `service.name` and `service.resource` cannot be set since `producer.sendBatch`
+        // is meant to send to several topics at a time
```
trentm, line 407: This might be worth discussing with the team. What does the Java Kafka instr do in this case? If there is a single topic in the outgoing batch, then should this use that value?

trentm (commented):
This is great.  I am sorry for taking so long to review.

lib/instrumentation/modules/kafkajs.js
david-luna: All my tests suggested there is only one topic per batch so I?m going to apply this suggestion

lib/instrumentation/modules/kafkajs.js
david-luna: Thanks for the explanation. Makes a lot of sense :)

lib/instrumentation/modules/kafkajs.js
david-luna: I can check the Java agent Implementation 1st before raising this topic in the next meeting.

test/docker-compose.yml
```diff
@@ -132,6 +132,42 @@ services:
...
+      - nodekafkadata:/var/lib/kafka/data
+    environment:
+      - KAFKA_BROKER_ID=1
+      - KAFKA_CFG_ZOOKEEPER_CONNECT=zookeeper:2181
+      - ALLOW_PLAINTEXT_LISTENER=yes
+      - KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP=CLIENT:PLAINTEXT,EXTERNAL:PLAINTEXT
+      - KAFKA_CFG_LISTENERS=CLIENT://:9092,EXTERNAL://:9093
+      - KAFKA_CFG_ADVERTISED_LISTENERS=CLIENT://kafka:9092,EXTERNAL://localhost:9093
+      - KAFKA_CFG_INTER_BROKER_LISTENER_NAME=CLIENT
+      - KAFKA_CFG_DELETE_TOPIC_ENABLE=true
```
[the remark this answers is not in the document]
david-luna: I'm going to add the ref link. Thanks for pointing it out :)

test/instrumentation/modules/kafkajs/kafkajs.test.js
```diff
@@ -0,0 +1,637 @@
...
+        delete t.result;
+        delete t.context.user;
+        delete t.context.tags;
+        delete t.context.custom;
+        delete t.context.cloud;
+      });
+
+      // Check message handling transactions
+      t.deepEqual(transactions.shift(), {
+        name: 'Kafka RECEIVE from batch',
```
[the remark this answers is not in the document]
david-luna: As said in a comment above. Although is not in the documentation the batch handler is called for each topic. I'll run come extra tests for that to be sure.

test/instrumentation/modules/kafkajs/kafkajs.test.js
```diff
@@ -0,0 +1,637 @@
...
+            span_id: spanId,
+          },
+          {
+            trace_id: tx.trace_id,
+            span_id: spanId,
+          },
+          {
+            trace_id: tx.trace_id,
+            span_id: spanId,
+          },
```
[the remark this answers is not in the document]
david-luna: okay, now I get it. SQS was one of the sources I've looked when I've started this part of the implementation. I'll do the changes you suggested. Thanks! :)

lib/instrumentation/modules/kafkajs.js
david-luna: changed!

test/instrumentation/modules/kafkajs/kafkajs.test.js
```diff
@@ -0,0 +1,637 @@
...
+    name: 'simple Kafkajs usage scenario for headers and body capturing on message reception',
+    script: 'fixtures/use-kafkajs-each-message.js',
+    cwd: __dirname,
+    timeout: 20000,
+    env: {
+      TEST_CLIENT_ID: 'elastic-kafka-client',
+      TEST_GROUP_ID: `elastictest-kafka-group-${rand}`,
+      TEST_TOPIC: kafkaTopic,
+      TEST_KAFKA_HOST: kafkaHost,
+      TEST_CAPTURE_HEADERS: 'true',
```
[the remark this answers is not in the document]
david-luna: That nit surfaced that actually the instrumentation was not checking `captureHeaders` configuration. It is fixed now :)

test/instrumentation/modules/kafkajs/kafkajs.test.js
```diff
@@ -0,0 +1,637 @@
...
+    },
+    checkApmServer(t, apmServer) {
+      t.ok(apmServer.events[0].metadata, 'metadata');
+      const events = sortApmEvents(apmServer.events);
+      const tx = events.shift().transaction;
+
+      // First the transaction.
+      t.ok(tx, 'got the send transaction');
+
+      // Check topic is ignored
```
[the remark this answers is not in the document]
david-luna: Previous tests are already checking the context propagation when the message sending is instrumented so I wanted to have the other case as well (no span generated from the instrumentation, therefore the trace context comes from the transaction)

I'm going keep the ignore and stash the `transaction_id` to check `traceparent` headers in next test

david-luna (commented):
/test tav kafkajs

lib/instrumentation/modules/kafkajs.js
trentm: The `topic` is a property of the `batch` object, so I assume it must be just one topic -- unless the types or Kafka core docs suggest otherwise.  With a short attempt, I can't find any Kafka docs on this.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -356,13 +367,15 @@ module.exports = function (mod, agent, { version, enabled }) {
...
-      const shouldIgnoreBatch =
-        topics && topics.every((t) => shouldIgnoreTopic(t, config));
+      let shouldIgnoreBatch = true;
+      const messages = batch.topicMessages || [];
+      for (const msg of messages) {
+        if (!shouldIgnoreTopic(msg.topic, config)) {
+          shouldIgnoreBatch = false;
+          break;
+        }
+      }
```
trentm, line 378: nit: This nit might be premature optimization. If the number of messages is very large (say 1000), then we'll be running the possibly-relatively-slow `shouldIgnoreTopic` for every message, when those many messages are likely all from a small set of topics. Perhaps this could get the set of topics and only `shouldIgnoreTopic(topic)` once for each topic.

test/instrumentation/modules/kafkajs/fixtures/use-kafkajs-each-message.js
```diff
@@ -80,7 +78,7 @@ async function useKafkajsClient(kafkaClient, options) {
     messages: [
       { value: 'each message 1', headers: { foo: 'string' } },
       { value: 'each message 2', headers: { foo: Buffer.from('buffer') } },
-      { value: 'each message 3' },
+      { value: 'each message 3', headers: { auth: 'secret_token' } },
```
trentm, line 81:
```suggestion
      { value: 'each message 3', headers: { auth: 'this_is_a_secret' } },
```

super-nerd-crazy-nit: To avoid a grep hit with `secret_token` in lib/config/schema.js.

test/instrumentation/modules/kafkajs/kafkajs.test.js
```diff
@@ -344,7 +341,7 @@ const testFixtures = [
     },
   },
   {
-    name: 'simple Kafkajs usage scenario for headers and body capturing on message reception',
+    name: 'simple Kafkajs usage scenario for `captureHeaders=false` and `capturebody=all` on message reception',
```
trentm, line 344:
```suggestion
    name: 'simple Kafkajs usage scenario for `captureHeaders=false` and `captureBody=all` on message reception',
```

nit

trentm (dismissed):
A couple of nits.

lib/instrumentation/modules/kafkajs.js
david-luna: yes, the previous code was doing the same and also allocating a new array of strings and using a callback function as an argument for `.every` method. So this version tackles only the memory allocation.

What I've missed is a topic could be the same for many messages so we could use a `Set` and iterate over it. I'm going to add this change

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,472 @@
...
+      for (const t of topics) {
+        if (!shouldIgnoreTopic(t, config)) {
+          shouldIgnoreBatch = false;
+          break;
+        }
+      }
+
+      if (!shouldIgnoreBatch) {
+        span = ins.createSpan(
+          `${NAME} send messages batch`,
```
trentm, line 390: I guess this doesn't conform to https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-messaging.md#naming
However, that spec doesn't cover the batch case.  This is basically the same discussion as https://github.com/elastic/apm-agent-nodejs/pull/3786/files#r1488754514 below.

lib/instrumentation/modules/kafkajs.js
```diff
@@ -0,0 +1,472 @@
...
+      }
+
+      // W3C trace-context propagation.
+      const runContext = ins.currRunContext();
+      const parentSpan =
+        span || runContext.currSpan() || runContext.currTransaction();
+
+      if (parentSpan && batch.topicMessages) {
+        batch.topicMessages.forEach((topicMessage) => {
+          topicMessage.messages.forEach((msg) => {
```
trentm, line 405: Should we avoid propagating trace-context on messages in a topic that we should ignore?

lib/instrumentation/modules/kafkajs.js
david-luna: note to self: As agreed, use the following rule for naming batches `Kafaka SEND to A,B,C,D...` where `A,B,C,D...` are the topics sorted and truncated at a certain length

lib/instrumentation/modules/kafkajs.js
trentm: Nevermind, I am wrong. This should still propagate the trace-context.

lib/instrumentation/modules/kafkajs.js
trentm: ... removing ignored topic names from the span/transaction name.

oleh-kravchuk:
@david-luna @trentm 

I see that this MR is already at the final stage, isn't it?

Thanks for your efforts for shipping such a valuable changes.
We look forward to use `native kafkajs instrumentation` on our side ;-)

lib/instrumentation/modules/kafkajs.js
david-luna: when brought to the agents meeting there was a consensus to:
- add the topic name if there is only ONE topic not being ignored
- add no topic names in any form if there is more than one topic not being ignored

david-luna: the implementation is done and a specific test is added in the last commit. @trentm could you do a last review (hopefully)?

test/instrumentation/modules/kafkajs/kafkajs.test.js
```diff
@@ -280,16 +281,143 @@ const testFixtures = [
         delete s.sample_rate;
       });
 
+      // The 1st batch has only one topic which is not ignored
+      // so the span has message context and also
+      // - service.target.name
+      // - destination.service.resource
       t.deepEqual(spans.shift(), {
-        name: 'Kafka send messages batch',
+        name: `Kafka send to ${kafkaTopic}`,
```
trentm, line 289:
```suggestion
        name: `Kafka SEND to ${kafkaTopic}`,
```

david-luna:
@trentm I've also updated the test expectations for the latest changes in span names. Now tests are green :)

david-luna (commented):
/test tav kafkajs

trentm (approved):
Awesome. Nice work.

PR #3789, opened 2023-12-15
fix fastify deprecated routerMethod、routerPath

aimidy:
FastifyDeprecation: You are accessing the deprecated "request.routerMethod" property. Use "request.routeOptions.method" instead. Property "req.routerMethod" will be removed in `fastify@5`.
FastifyDeprecation: You are accessing the deprecated "request.routerPath" property. Use "request.routeOptions.url" instead. Property "req.routerPath" will be removed in `fastify@5`.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/fastify.js
+++ b/lib/instrumentation/modules/fastify.js
@@ -70,8 +70,8 @@ module.exports = function (
 
     agent.logger.debug('adding onRequest hook to fastify');
     _fastify.addHook('onRequest', (req, reply, next) => {
-      const method = req.routerMethod || req.raw.method; // Fallback for fastify >3 <3.3.0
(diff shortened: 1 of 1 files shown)


trentm:
@aimidy Thanks very much for digging into this and opening this PR!

Your patch would, I think, break for fastify 4.x versions before 4.10.0 when `req.routeOptions` was added. "I think" because I haven't tested that.

I fixed this same issue in the OpenTelemetry repo a few months back https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1763
I've opened https://github.com/elastic/apm-agent-nodejs/pull/3814 here that should fix the issue.

trentm:
Handled the equivalent via https://github.com/elastic/apm-agent-nodejs/pull/3814

Issue #3800, opened 2023-12-19
update-tav-versions.js needs update for tav@5 syntax


trentm:
See #3796

PR #3812, opened 2024-01-02
chore: adapt `update-tav-versions.js` to the new TAV@5 syntax

david-luna:
In a previous [dependabot update](https://github.com/elastic/apm-agent-nodejs/pull/3796) the `test-all-versions` package was updated two major versions. One of this majors had a breaking change in the syntax of `.tav.yml` file so the script used to keep that file updated (update-tav-versions.js) no longer works.

This PR adapts the existing script to the new `.tav.yml` syntax and also:
- updates the `.tav.yml` file
- removes `tav-versions-fastify.sh` which was used to update fastify and is no longer needed.

Closes: #3800 

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


dev-utils/update-tav-versions.js
```diff
@@ -150,7 +137,17 @@ async function main() {
...
+      }
+    } else if (tavToUpdate && line.startsWith('  - versions:')) {
+      console.log(`Updating versions of ${tavToUpdate}`);
+      const tavMatrix = tavVersMap.get(tavToUpdate);
+      const tavVers = tavMatrix.shift();
+
+      if (tavVers.length > 0) {
+        tavLines[idx] = `  - versions: '${tavVers.join(' || ')}'`;
+      }
     }
```
trentm, line 151: It might be helpful to have a guard that will error out if there are any unused `tavVersMap` entries. Could this happen if there is a block with multiple versions sections, but the first field is not `versions`, e.g. something like:

```yaml
mysql2:
  - foo: bar
    versions: '>=1.0.0 <1.6.0'
    commands: "..."
```

trentm (approved):

PR #3814, opened 2024-01-02
fix(fastify): avoid the FSTDEP017 and FSTDEP018 deprecation warnings

trentm:
Obsoletes: #3789

* * *

This works on the same issue as #3789. It re-uses some lessons from the similar https://github.com/open-telemetry/opentelemetry-js-contrib/pull/1763 that I did a few months back.


trentm (commented):
/test tav fastify

david-luna (approved):

PR #3835, opened 2024-01-18
fix(mongodb): fix #3834

### LINKED ISSUE #3834: fix <database>.$cmd.getMore span on mongodb
we have spans like this `<database>.$cmd.getMore`, but it should be `<database>.<collection>.getMore`

correct getCollection to handle `getMore` command events

there is not way to fix this

**Additional context**
improved span name

```
function collectionFor(event) {
    const collection = event.command[event.commandName];
    return typeof collection === 'string' ? collection : '$cmd';
  }
  ```
  
  https://github.com/elastic/apm-agent-nodejs/blob/main/lib/instrumentation/modules/mongodb.js#L224

### PR DESCRIPTION
**Pull request by sibelius**
command started event example:

```jsx
{
  event: CommandStartedEvent {
    name: 'commandStarted',
    address: '172.20.0.2:27017',
    connectionId: 12,
    serviceId: undefined,
    requestId: 686,
    databaseName: 'mydatabase',
    commandName: 'getMore',
    command: {
      getMore: new Long('1769182660590360229'),
      collection: 'Interaction',
      batchSize: 1000,
      lsid: [Object],
      '$clusterTime': [Object],
      '$db': 'mydatabase'
    }
  },
  commandName: 'getMore',
  collection: new Long('1769182660590360229')
}
```

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/lib/instrumentation/modules/mongodb.js
+++ b/lib/instrumentation/modules/mongodb.js
@@ -222,6 +222,10 @@ module.exports = (mongodb, agent, { version, enabled }) => {
   }
 
   function collectionFor(event) {
+    if (event.commandName === 'getMore') {
+      return event.command.collection;
+    }
+    
     const collection = event.command[event.commandName];
     return typeof collection === 'string' ? collection : '$cmd';
   }


david-luna:
@sibelius thanks for your contribution into `elastic-apm-node` :)

What you propose enforces the specs related to `mongodb` instrumentation ([link](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-db.md#mongodb)) for that specific command so indeed we should find a way to add the collection.

In order to move forward we need to complete the following tasks:
-  [ ] be defensive on how we extract the collection name. The [mongodb specs](https://mongodb.github.io/node-mongodb-native/6.3/classes/CommandStartedEvent.html#command) refers to command as a BSON's `Document` which does not specify the properties this PR inspects to get the collection name. Since the agent supports the range `>=3.3.0 <7` we can assume that these properties may change across versions so an `if (event.commandName === 'getMore' && event.command.collection)` would fallback to old implentation if the `collection` property does not exist or contains an empty name.
- [ ] add a test to assert the change does what is supposed to do. Would come handy if there are changes in newer versions of `mongodb` that could break our logic to extract the name from `getMore` internal command. With the CI that test we can catch if this situation happens.
- [ ] make sure your code follow the linting rules. You can use `npm run lint:fix` to fix any auto-fixable issue.

sibelius:
It looks correct for me, tks

PR #3837, opened 2024-01-18
chore: clarify koa-router instrumentation range

trentm:
No functional change. However, I was confused for a little while that
'12.0.1' satisfies '>=5.2.0 <=12'. To me it isn't self-explanatory
that '<=12' is not the same as '<=12.0.0'.

### DIFF
--- a/lib/instrumentation/modules/mongodb.js
+++ b/lib/instrumentation/modules/mongodb.js
@@ -19,7 +19,7 @@ const HOSTNAME_PORT_RE = /^(.+):(\d+)$/;
 
 module.exports = (mongodb, agent, { version, enabled }) => {
   if (!enabled) return mongodb;
-  if (!semver.satisfies(version, '>=3.3 <7.0')) {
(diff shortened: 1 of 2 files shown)


lib/instrumentation/modules/mongodb.js
```diff
@@ -19,7 +19,7 @@ const HOSTNAME_PORT_RE = /^(.+):(\d+)$/;
 
 module.exports = (mongodb, agent, { version, enabled }) => {
   if (!enabled) return mongodb;
-  if (!semver.satisfies(version, '>=3.3 <7.0')) {
+  if (!semver.satisfies(version, '>=3.3 <7')) {
```
trentm, line 22: Note for reviewer: THis change is a necessary part of this, but was a style change I made to match our typical `... <$majorVer` style that we use in module instrumentations when I was reviewing them all.

david-luna (approved):

PR #3899, opened 2024-03-04
Fix incorrect hostname detection on Windows

mpoc:
In some cases, the PowerShell command that is used to detect the hostname will contain extra output before the hostname, e.g. the Microsoft copyright banner or any output that is printed in the PowerShell profiles. This causes invalid hostnames to be sent to the APM server.

This change ensures that the PowerShell output is limited only to the hostname.

### DIFF
--- a/lib/apm-client/http-apm-client/detect-hostname.js
+++ b/lib/apm-client/http-apm-client/detect-hostname.js
@@ -30,7 +30,15 @@ function detectHostname() {
       // https://learn.microsoft.com/en-us/dotnet/api/system.net.dns.gethostentry
       out = spawnSync(
         'powershell.exe',
-        ['[System.Net.Dns]::GetHostEntry($env:computerName).HostName'],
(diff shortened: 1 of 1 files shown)


mpoc:
Not sure, I couldn't find a definitive answer. 

I just wanted to clarify that I wasn't suggesting the banner shows up inconsistently. What I meant was that log messages in profile scripts will break the hostname detection. Likewise, anything printed before the hostname, like the logo banner, will break it for the same reason.

trentm:
run docs-build

trentm:
> I just wanted to clarify [...]

Understood. Thanks.

trentm (approved):

trentm:
> I'll follow up with a PR on our specs for this:
> https://github.com/elastic/apm/blob/main/specs/agents/metadata.md#hostname

https://github.com/elastic/apm/pull/851 for that.

Issue #3884, opened 2024-02-27
issue labeler workflow failing


david-luna:
latest dependabot PRs are failing in the issue labeler CI workflow and blocking the merge. The POST request failed with a 403 so we should start by checking if the token has expired or was rotated

Ref: https://github.com/elastic/apm-agent-nodejs/actions/runs/8042141883/job/21962393814?pr=3883

excerpt of the error
```
RequestError [HttpError]: Resource not accessible by integration
    at /home/runner/work/_actions/AlexanderWert/issue-labeler/v2.3/node_modules/@octokit/request/dist-node/index.js:66:23
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  status: 403,
  headers: {
    'access-control-allow-origin': '*',
    'access-control-expose-headers': 'ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset',
    connection: 'close',
    'content-encoding': 'gzip',
    'content-security-policy': "default-src 'none'",
    'content-type': 'application/json; charset=utf-8',
    date: 'Mon, 26 Feb 2024 01:02:07 GMT',
    'referrer-policy': 'origin-when-cross-origin, strict-origin-when-cross-origin',
    server: 'GitHub.com',
    'strict-transport-security': 'max-age=3[15](https://github.com/elastic/apm-agent-nodejs/actions/runs/8042141883/job/21962393814?pr=3883#step:3:16)36000; includeSubdomains; preload',
    'transfer-encoding': 'chunked',
    vary: 'Accept-Encoding, Accept, X-Requested-With',
    'x-accepted-github-permissions': 'issues=write; pull_requests=write',
    'x-content-type-options': 'nosniff',
    'x-frame-options': 'deny',
    'x-github-api-version-selected': '2022-11-28',
    'x-github-media-type': 'github.v3; format=json',
    'x-github-request-id': 'BCF4:797F:65ED267:68FBC11:65DBE30E',
    'x-ratelimit-limit': '15000',
    'x-ratelimit-remaining': '14993',
    'x-ratelimit-reset': '1708912857',
[...]

PR #3921, opened 2024-03-19
ci: fix/refactor labeler workflow

trentm:
- use more reliable 'elastic' org membership test for triage, etc. labels
- bypass 'action-add-labels' action to avoid another dep
- drop agent-nodejs label that we no longer require for use in GH projects

Fixes: https://github.com/elastic/apm-agent-nodejs/issues/3884

* * *

The `actions/github-script@v7` usage is from apm-agent-java.
The dropping `agent-nodejs` label adding is pending on team discussion (https://elastic.slack.com/archives/C9A6ANR5Y/p1710809441508399).

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -15,18 +15,6 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
-    # XXX This might be dropped, pending team discussion.
-    - name: Add agent-nodejs label
-      uses: actions/github-script@v7
-      with:
-        script: |
-          github.rest.issues.addLabels({
-            issue_number: context.issue.number,
-            owner: context.repo.owner,
-            repo: context.repo.repo,
-            labels: ["agent-nodejs"]
-          })
-
     - id: is_elastic_member
       uses: elastic/apm-pipeline-library/.github/actions/is-member-elastic-org@current
       with:
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -5,36 +5,42 @@ on:
   pull_request_target:
     types: [opened]
 
-# 'issues: write' for https://docs.github.com/en/rest/issues/labels?apiVersion=2022-11-28#add-labels-to-an-issue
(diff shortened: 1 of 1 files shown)


.github/workflows/labeler.yml
```diff
@@ -5,36 +5,30 @@ on:
...
-          apmmachine
-          dependabot
-          dependabot[bot]
-        GITHUB_TOKEN: ${{ secrets.APM_TECH_USER_TOKEN }}
+        token: ${{ secrets.APM_TECH_USER_TOKEN }}
 
     - name: Add community and triage labels
-      if: steps.checkUserMember.outputs.isTeamMember != 'true' && steps.checkUserMember.outputs.isExcluded != 'true'
-      uses: actions-ecosystem/action-add-labels@v1
+      if: contains(steps.is_elastic_member.outputs.result, 'false') && github.actor != 'dependabot[bot]'
```
reakaleek, line 25: Did you omit @apmmachine for a reason?

trentm: I am copying what apm-agent-java is doing, e.g.: https://github.com/elastic/apm-agent-java/blob/main/.github/workflows/labeler.yml#L33

Do we use apmmachine anymore for creating issues or PRs? I don't see any: https://github.com/elastic/apm-agent-nodejs/issues?q=author%3A%40apmmachine+

reakaleek: We are still using it for updatecli workflows.

I think you need to remove the `@` in the search. 

https://github.com/elastic/apm-agent-nodejs/issues?q=author%3Aapmmachine+ shows the last PR in january.

trentm: Thanks.

Presumably apm-agent-java will have the same issue, then.

trentm: @reakaleek Oh, actually, `@apmmachine` *is* a member of "elastic", so we don't need to explicitly handle excluding `github.actor === 'apmmachine'` in this test.

reakaleek: Ah, right. Makes sense.

PR #3928, opened 2024-03-22
test: add healthcheck for zookeeper test service so 'npm test' works outside of CI

trentm:
Running 'npm test' outside of CI will start all test services in docker
*then wait for them to all be healthy*. Without a healthcheck for zk
the 'wait_for_healthy' in test/script/run_tests.sh would never pass.

### DIFF
--- a/.ci/docker/docker-compose.yml
+++ b/.ci/docker/docker-compose.yml
@@ -149,6 +149,13 @@ services:
       - nodezookeeperdata:/bitnami
     environment:
       - ALLOW_ANONYMOUS_LOGIN=yes
+    healthcheck:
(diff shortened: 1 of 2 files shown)


david-luna (approved):
A clean checkout was not enough to made it run for me. I needed to tweak the `docker_compose.yml` file
Ref: https://github.com/elastic/apm-nodejs-dev/issues/27#issuecomment-2017441589

trentm:
> A clean checkout was not enough to made it run for me. I needed to tweak the `docker_compose.yml` file

Let's do that in a separate change.

trentm:
> Let's do that in a separate change.

https://github.com/elastic/apm-agent-nodejs/pull/3939 for that.

PR #3938, opened 2024-03-25
release: add provenance with GitHub actions

v1v:

Run the release steps partially for `main` - to validate things work smoothly before the release. It skips the below steps:
  - npm publish
  - github release creation 
  
## Further details

Use the GitHub provenance in this project as explained in https://docs.npmjs.com/generating-provenance-statements

## Test

I created a test feature branch and https://github.com/elastic/apm-agent-nodejs/actions/runs/8819809747 is now running in `dry-run` mode:

- https://github.com/elastic/apm-agent-nodejs/attestations


.github/workflows/release.yml
```diff
@@ -4,14 +4,20 @@ on:
   push:
     tags:
       - v*.*.*
+    branches:
+      - main
+      - provenance
```
v1v, line 9: unrequired

v1v:
```suggestion
```

reakaleek (dismissed):

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+          tags: |
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.latest }}
+          build-args: |
+            AGENT_DIR=/build/dist/nodejs
+
+      - name: Attest image
+        uses: github-early-access/generate-build-provenance@main
+        with:
+          subject-name: "${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}"
```
trentm, line 76:
```suggestion
          subject-name: "${{ env.DOCKER_IMAGE_NAME }}"
```

https://github.com/github-early-access/generate-build-provenance/tree/main?tab=readme-ov-file#container-image says:

> ... Do NOT include a tag as part of the image name -- the specific image being attested is identified by the supplied digest. ...

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+          subject-digest: ${{ steps.push.outputs.digest }}
+          push-to-registry: false
```
trentm, line 78: Why not `true` here?

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+          context: .
+          push: true
+          tags: |
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.latest }}
+          build-args: |
+            AGENT_DIR=/build/dist/nodejs
+
+      - name: Attest image
+        uses: github-early-access/generate-build-provenance@main
```
trentm, line 74: Ew that `github-early-access/generate-build-provenance` doesn't have any releases. We are okay running "whatever happens to be latest main at the time" here?

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+        with:
+          context: .
+          push: true
+          tags: |
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.latest }}
+          build-args: |
+            AGENT_DIR=/build/dist/nodejs
+
+      - name: Attest image
```
trentm, line 73:
```suggestion
      - name: Attest Docker image
```

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+            echo "latest=test-latest" >> "${GITHUB_OUTPUT}"
+          else
+            # version without v prefix (e.g. 1.2.3)
+            echo "tag=${GITHUB_REF_NAME/v/}" >> "${GITHUB_OUTPUT}"
+            echo "latest=latest" >> "${GITHUB_OUTPUT}"
+          fi
+
+      # Use docker/build-push-action@v5.3.0 in conjunction with github-early-access/generate-build-provenance@main
+      - name: Build and Push Docker Image
+        id: push
```
trentm, line 62: nit: I'd prefer that the `id`s used for these are changed to `docker-push` (rather than `push`) and perhaps `docker-setup` or `docker-vars`.   

- The `push` was a little hard for me to notice the `steps.push.outputs.digest` connection below.
- Having `docker-` as a common leading prefix seems helpful.

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
 
       - run: make -C .ci dist
 
-      - run: make -C .ci push-docker
+      - name: generate build provenance
```
trentm, line 42:
```suggestion
      - name: Attest Lambda layer zip
```

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+        uses: github-early-access/generate-build-provenance@main
+        with:
+          subject-path: "${{ github.workspace }}/build/aws/*.zip"
```
trentm, line 45:
```suggestion
          subject-path: "${{ github.workspace }}/build/aws/elastic-apm-node-lambda-layer-*.zip"
```

Q: Is there much value in this attestation? We don't upload the .zip file as an artifact... other than to AWS regions as lambda layers. I suppose the attestation will be there if someone wants to download those layer zip files (https://stackoverflow.com/questions/56614037/how-to-download-aws-lambda-layer) and check them.

.ci/Makefile
```diff
@@ -37,10 +37,12 @@ env:
 
 # Create Docker image
 build-docker: validate-version
+	@echo "Obsoleted, use GitHub actions"
 	docker build -t $(DOCKER_REGISTRY)/$(DOCKER_IMAGE_NAME):$(AGENT_VERSION) \
   --build-arg AGENT_DIR=$(DIST_DIR) ..
 
 push-docker: build-docker
+	@echo "Obsoleted, use GitHub actions"
```
trentm, line 45: Moving away from the push-docker.sh script drops support for only *sometimes* applying the "latest" Docker tag to an image. This handling was used so that we could use the same push-docker.sh script for the current latest major releases (4.x) and for the 3.x maintenance branch.

I suppose this is fine, given that we won't add provenance support to the 3.x branch.  At some point if/when tehre is a 5.x major, then we'd need to make sure that a maintenance 4.x release does *NOT* push the "latest" Docker tag.

.ci/Makefile
```diff
@@ -37,10 +37,12 @@ env:
 
 # Create Docker image
 build-docker: validate-version
+	@echo "Obsoleted, use GitHub actions"
```
trentm, line 40: I would like this PR to remove the sections of this Makefile and scripts (like push-docker.sh), if they will become unused. In a separate PR is fine if helpful/necessary.

trentm (changes requested):

.github/workflows/release.yml
v1v: I decided to create an attestation for anything we publish, regardless of where those artifacts are stored. I think it's valuable we also do that for the lambda layers

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+
+      - id: setup-docker
```
v1v, line 47:
```suggestion
      - id: docker-vars
```

.github/workflows/release.yml
v1v, line 62:
```suggestion
        id: docker-push
```

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+      # Use docker/build-push-action@v5.3.0 in conjunction with github-early-access/generate-build-provenance@main
+      - name: Build and Push Docker Image
+        id: push
+        uses: docker/build-push-action@v5.3.0
+        with:
+          context: .
+          push: true
+          tags: |
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}
+            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.latest }}
```
v1v, line 69:
```suggestion
            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.docker-vars.outputs.tag }}
            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.docker-vars.outputs.latest }}
```

.github/workflows/release.yml
```diff
@@ -33,7 +39,43 @@ jobs:
...
+          build-args: |
+            AGENT_DIR=/build/dist/nodejs
+
+      - name: Attest image
+        uses: github-early-access/generate-build-provenance@main
+        with:
+          subject-name: "${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}"
+          subject-digest: ${{ steps.push.outputs.digest }}
```
v1v, line 77:
```suggestion
          subject-digest: ${{ steps.docker-push.outputs.digest }}
```

.ci/Makefile
v1v: Shall we move this topic to a follow-up?

.github/workflows/release.yml
v1v: I took the example from the official docs, this is a GitHub supported initiative, so I probably wrongly assumed `main` was good.

I'll use a fixed version to play safe, thanks!

v1v: OTOH, we might need to monitor if they fix anything that can be needed

v1v: As we are trialing this... I'd say we keep with `main` for now and then use a fixed version when they move from early access to something else, does it sound good?

.github/workflows/release.yml
v1v: I've been told that this should work, and it worked
- https://github.com/elastic/apm-agent-nodejs/actions/runs/8822486532/job/24220785121#step:10:99

although no clue where those attestations are now pushed in the registry, but the `push-to-registry: true` does something 

```suggestion
          push-to-registry: true
```

.ci/Makefile
trentm: Yes, this PR doesn't need to handle this.

.ci/Makefile
```diff
@@ -35,14 +35,6 @@ clean:
 env:
 	@env | sort
 
-# Create Docker image
```
trentm, line 38: nit: these lines above could now be removed:

```
DIST_DIR = /build/dist/nodejs
DOCKER_IMAGE_NAME = observability/apm-agent-nodejs
DOCKER_REGISTRY = docker.elastic.co
```

.github/workflows/release.yml
```diff
@@ -33,7 +38,43 @@ jobs:
...
+            # for testing purposes
+            echo "tag=test" >> "${GITHUB_OUTPUT}"
+            echo "latest=test-latest" >> "${GITHUB_OUTPUT}"
+          else
+            # version without v prefix (e.g. 1.2.3)
+            echo "tag=${GITHUB_REF_NAME/v/}" >> "${GITHUB_OUTPUT}"
+            echo "latest=latest" >> "${GITHUB_OUTPUT}"
+          fi
+
+      # Use docker/build-push-action@v5.3.0 in conjunction with github-early-access/generate-build-provenance@main
```
trentm, line 59: nit: This comment can be dropped.

.github/workflows/release.yml
```diff
@@ -33,7 +38,43 @@ jobs:
 
       - run: make -C .ci dist
 
-      - run: make -C .ci push-docker
+      - name: Attest Lambda layer zip
+        uses: github-early-access/generate-build-provenance@main
+        with:
+          subject-path: "${{ github.workspace }}/build/aws/elastic-apm-node-lambda-layer-*.zip"
```
trentm, line 44: Q: Do we really want to (a) publish attestations (this goes on a public immutable ledger of some sort, IIUC), and (b) docker build and push for **every commit to main**?  My concern is that this is very wasteful.

.github/workflows/release.yml
v1v: see https://github.com/elastic/apm-agent-nodejs/pull/3938/commits/a8430f4bdb72b9abdf8dbb28699689d353b6b3b7

.ci/Makefile
v1v: see https://github.com/elastic/apm-agent-nodejs/pull/3938/commits/a8430f4bdb72b9abdf8dbb28699689d353b6b3b7

.github/workflows/release.yml
v1v: I'm probably biased, but I prefer to run the release on the main branch with a `dry-run` or a subset of the jobs; this can detect any breaking changes in workflows.

Therefore, I enabled the provenance for those artefacts - I see it might be useful, and those artefacts will be available and with their attestations.

From the disc space, it should be ok, there is no limitation in our elastic docker registry

trentm (approved):

PR #3972, opened 2024-04-12
chore: add a script for, and update the release process for updating the 4.x docs branch

trentm:
Now that we need to use a PR to update the 4.x branch for a release, I've added a script to help with this.

This script was used for https://github.com/elastic/apm-agent-nodejs/pull/3971 for the v4.5.2 release.

### DIFF
--- a/dev-utils/update-4x-branch.sh
+++ b/dev-utils/update-4x-branch.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+#
+# Create a PR to update the 4.x branch to match the state of "main" for the
+# just-tagged release. The 4.x branch needs to be updated for the current docs
+# build.
+#
+
+if [ "$TRACE" != "" ]; then
+    export PS4='${BASH_SOURCE}:${LINENO}: ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+    set -o xtrace
+fi
+set -o errexit
+set -o pipefail
+
+function fatal {
+    echo "$(basename $0): error: $*"
+    exit 1
+}
+
+# json=./node_modules/.bin/json
+# if [ ! -f "$json" ]; then
+#     echo "Requirements:"
+#     echo "   - install npm dependencies"
+#     echo "   - cd into root folder of the project"
+#     echo "Usage:"
+#     echo "   ./dev-utils/ci-tav-slow-jobs.sh"
+#     exit 1;
+# fi
+
+
+# branch=main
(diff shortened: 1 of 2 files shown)


dev-utils/update-4x-branch.sh
```diff
@@ -0,0 +1,92 @@
...
+fi
+set -o errexit
+set -o pipefail
+
+function fatal {
+    echo "$(basename $0): error: $*"
+    exit 1
+}
+
+# json=./node_modules/.bin/json
```
david-luna, line 20: Do we need this commented block?

david-luna (dismissed):

dev-utils/update-4x-branch.sh
trentm: nope, forgotten copypasta

trentm:
Note to self: When I merged https://github.com/elastic/apm-agent-nodejs/pull/3971, by default the commit message included the commit messages from all the cherry-picked changes.  I don't know if we really want that. (I removed all the commit message body content before merging this particular PR.) My curr preference would be to have these squashed commits to 4.x not have all these details repeated in the commit messages.

An option would be to have the script squash all the commits into one before creating the PR. We *would* then lose all the separation of commits on the PR for browsing them, but I don't think that is that valuable.  It *is* somewhat valuable to keep the listing of included cherry-picked commits in the PR description.

dev-utils/update-4x-branch.sh
```diff
@@ -0,0 +1,71 @@
...
+    git log --pretty=format:%h -30 | tail -n +2 | while read sha; do
+        possible=$(git tag --points-at $sha)
+        if [[ "$possible" =~ ^v[0-9]+\.[0-9]+\.[0-9]$ ]]; then
+            echo $possible
+            break
+        fi
+    done
+)
+if [[ -z "$LASTTAG" ]]; then
+    fatal "could not find previous release tag in last 30 commits"
```
david-luna, line 46: does this means we have a limit of commits between releases? I'm okay with that but then there should be a reminder somewhere

david-luna (approved):

dev-utils/update-4x-branch.sh
trentm: I can bump it. 30 *does* seem a little low. This was just meant as a sanity guard.

trentm: commit 41216321

david-luna (approved):
thanks for making releases easier for us :)

Issue #3980, opened 2024-04-15
KafkaJS instrumentation throw exception on "tombstone message"


wilski-b:
**Describe the bug**

Kafkajs support added in v4.5.2 throwing error for "tombstone messages" (messages with no payload value).

**To Reproduce**

1.  With APM enabled, using a kafkajs consumer, consume a tombstone message (message with null value) see script below.

**Expected behavior**

1. To be able to consume a "tombstone message" using KafkaJS and APM.

**Environment (please complete the following information)**

- OS: Linux
- Node.js version: 18
- APM Server version: 8.9.1
- Agent version: 4.5.2

**How are you starting the agent? (please tick one of the boxes)**

- [x] Calling `agent.start()` directly (e.g. `require('elastic-apm-node').start(...)`)
- [ ] Requiring `elastic-apm-node/start` from within the source code
- [ ] Starting node with `-r elastic-apm-node/start`

**Additional context**

Here is a script that reproduces the issue (note process.env's that must be set):
```
const apm = require('elastic-apm-node')

apm.start({
  secretToken: process.env.APM_TOKEN,
  serverUrl: process.env.APM_HOST,
  environment: process.env.ENV,
  captureBody: 'all',
})

const { Kafka } = require('kafkajs')

const kafka = new Kafka({
  clientId: 'test-app',
  brokers: [process.env.KAFKA_CLUSTER_ENDPOINTS],
})
const consumer = kafka.consumer({ groupId: 'my-group' })
consumer.on('consumer.crash', event => {
  const error = event?.payload?.error
  console.error(error)
  process.exit(1)
})
const producer = kafka.producer()
const TEST_TOPIC = 'test-topic'

/**
 *
 */
async function run() {
  await consumer.connect()
  await consumer.subscribe({ topics: [TEST_TOPIC] })
  await consumer.run({
    eachMessage: async ({ message }) => {
      console.log(`Received message with value: ${JSON.stringify(message.value?.toString())}`)
    },
  })

  await producer.connect()
  await producer.send({
    topic: TEST_TOPIC,
    messages: [
      { value: 'This message and others with a value not NULL work fine.' },
      { value: null }, // this message will end up causing elastic-apm-node to throw an uncaught exception from elastic-apm-node\lib\instrumentation\modules\kafkajs.js:141:41
    ],
  })
}

run()

```
error being thrown:
```
undefined,
  [cause]: TypeError: Cannot read properties of null (reading 'toString')
[...]

PR #3985, opened 2024-04-22
fix(kafkajs): handle kafka tombstone messages

david-luna:
The `kafkajs` instrumentation was calling `toString` method for all of the messages received. This approach wasn't taking in consideration `tombstone messages` which is defines as

> A message with a key and a null payload will be treated as a delete from the log. Such a record is sometimes referred to as a _tombstone_.

Ref: https://kafka.apache.org/documentation/#design_compactionbasics

Fixes: #3980

### Checklist

- [x] Implement code
- [x] Add tests
- [ ] Update documentation

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,23 @@ Notes:
 
 See the <<upgrade-to-v4>> guide.
 
+==== Unreleased
+
+[float]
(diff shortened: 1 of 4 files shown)


trentm (commented):
/test tav kafkajs

trentm (approved):

PR #3998, opened 2024-04-29
ci: use docker/metadata-action to gather tags/labels

v1v:
### What

Use https://github.com/docker/metadata-action/tree/8e5442c4ef9f78752691e2d8f8d19755c6f78e81/?tab=readme-ov-file#semver to get the tag version and [open-container](https://github.com/opencontainers/image-spec/blob/main/annotations.md) labels.

Use sha commit for the GitHub action related to docker.

### Test

See https://github.com/elastic/apm-agent-nodejs/actions/runs/8878357355, didn't run with `push`

```
Run docker/build-push-action@2cdde995de11925a030ce8070c3d77a52ffcf1c0
  with:
    context: .
    push: false
    tags: docker.elastic.co/observability/apm-agent-nodejs:test-latest
    labels: org.opencontainers.image.created=[2](https://github.com/elastic/apm-agent-nodejs/actions/runs/8878357355/job/24373834641#step:9:2)024-04-29T12:16:2[3](https://github.com/elastic/apm-agent-nodejs/actions/runs/8878357355/job/24373834641#step:9:3).000Z
  org.opencontainers.image.description=Elastic APM Node.js Agent
  org.opencontainers.image.licenses=BSD-2-Clause
  org.opencontainers.image.revision=f9f28681ae69736728d1[5](https://github.com/elastic/apm-agent-nodejs/actions/runs/8878357355/job/24373834641#step:9:5)aaa93880fffba5f2e6d
  org.opencontainers.image.source=https://github.com/elastic/apm-agent-nodejs
  org.opencontainers.image.title=apm-agent-nodejs
  org.opencontainers.image.url=https://github.com/elastic/apm-agent-nodejs
  org.opencontainers.image.version=test-latest
    build-args: AGENT_DIR=/build/dist/nodejs
```

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -50,7 +50,7 @@ jobs:
         with:
           images: ${{ env.DOCKER_IMAGE_NAME }}
           tags: |
-            type=raw,value=latest,enable={{is_default_branch}}
+            type=raw,value=latest,prefix=test-,enable={{is_default_branch}}
             type=ref,event=tag,prefix=,suffix=
 
       - name: Build and Push Docker Image
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,9 +49,13 @@ jobs:
         uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81  # v5.5.1
         with:
           images: ${{ env.DOCKER_IMAGE_NAME }}
+          flavor: |
+            latest=auto
           tags: |
-            type=raw,value=latest,prefix=test-,enable={{is_default_branch}}
+            # "1.2.3" and "latest" Docker tags on push of git tag "v1.2.3"
             type=semver,pattern={{version}}
+            # "edge" Docker tag on git push to default branch
+            type=edge
 
       - name: Build and Push Docker Image
         id: docker-push
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,7 +51,7 @@ jobs:
           images: ${{ env.DOCKER_IMAGE_NAME }}
           tags: |
             type=raw,value=latest,prefix=test-,enable={{is_default_branch}}
-            type=ref,event=tag,prefix=,suffix=
+            type=semver,pattern={{version}}
 
       - name: Build and Push Docker Image
         id: docker-push
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -44,28 +44,23 @@ jobs:
         with:
           subject-path: "${{ github.workspace }}/build/aws/elastic-apm-node-lambda-layer-*.zip"
 
-      - id: docker-vars
-        name: Set up docker variables
-        run: |-
-          if [ "${{ startsWith(github.ref, 'refs/tags') }}" == "false" ] ; then
-            # for testing purposes
-            echo "tag=test" >> "${GITHUB_OUTPUT}"
-            echo "latest=test-latest" >> "${GITHUB_OUTPUT}"
-          else
-            # version without v prefix (e.g. 1.2.3)
-            echo "tag=${GITHUB_REF_NAME/v/}" >> "${GITHUB_OUTPUT}"
-            echo "latest=latest" >> "${GITHUB_OUTPUT}"
-          fi
+      - name: Extract metadata (tags, labels)
+        id: docker-meta
+        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81  # v5.5.1
+        with:
+          images: ${{ env.DOCKER_IMAGE_NAME }}
+          tags: |
+            type=raw,value=latest,enable={{is_default_branch}}
+            type=ref,event=tag,prefix=,suffix=
 
       - name: Build and Push Docker Image
         id: docker-push
-        uses: docker/build-push-action@v5.3.0
+        uses: docker/build-push-action@2cdde995de11925a030ce8070c3d77a52ffcf1c0  # v5.3.0
         with:
           context: .
           push: true
-          tags: |
-            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.docker-vars.outputs.tag }}
-            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.docker-vars.outputs.latest }}
+          tags: ${{ steps.docker-meta.outputs.tags }}
+          labels: ${{ steps.docker-meta.outputs.labels }}
           build-args: |
             AGENT_DIR=/build/dist/nodejs
 


.github/workflows/release.yml
```diff
@@ -44,28 +44,23 @@ jobs:
...
-            echo "latest=latest" >> "${GITHUB_OUTPUT}"
-          fi
+      - name: Extract metadata (tags, labels)
+        id: docker-meta
+        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81  # v5.5.1
+        with:
+          images: ${{ env.DOCKER_IMAGE_NAME }}
+          tags: |
+            type=raw,value=latest,prefix=test-,enable={{is_default_branch}}
+            type=ref,event=tag,prefix=,suffix=
```
v1v, line 54: See https://github.com/elastic/apm-agent-python/pull/2033#discussion_r1583086509

```suggestion
            type=semver,pattern={{version}}
```

.github/workflows/release.yml
```diff
@@ -44,28 +44,23 @@ jobs:
...
-            echo "tag=${GITHUB_REF_NAME/v/}" >> "${GITHUB_OUTPUT}"
-            echo "latest=latest" >> "${GITHUB_OUTPUT}"
-          fi
+      - name: Extract metadata (tags, labels)
+        id: docker-meta
+        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81  # v5.5.1
+        with:
+          images: ${{ env.DOCKER_IMAGE_NAME }}
+          tags: |
+            type=raw,value=latest,prefix=test-,enable={{is_default_branch}}
```
trentm, line 53: Doesn't this one *always* result in having a "test-" prefix, even for a real release?

Why does `enable={{is_default_branch}}` resolve to `enabled=true` in this example run: https://github.com/elastic/apm-agent-nodejs/actions/runs/8878357355/job/24373834641

```
Run docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81
  with:
    images: docker.elastic.co/observability/apm-agent-nodejs
    tags: type=raw,value=latest,prefix=test-,enable={{is_default_branch}}
  type=raw,value=latest,prefix=test-,enable=true
  type=ref,event=tag,prefix=,suffix=
  
    context: workflow
    github-token: ***
  env:
    DOCKER_IMAGE_NAME: docker.elastic.co/observability/apm-agent-nodejs
    DOCKER_USERNAME: 
    DOCKER_PASSWORD: 
```

.github/workflows/release.yml
```diff
@@ -44,28 +44,23 @@ jobs:
...
-        uses: docker/build-push-action@v5.3.0
+        uses: docker/build-push-action@2cdde995de11925a030ce8070c3d77a52ffcf1c0  # v5.3.0
         with:
           context: .
           push: true
-          tags: |
-            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.docker-vars.outputs.tag }}
-            ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.docker-vars.outputs.latest }}
+          tags: ${{ steps.docker-meta.outputs.tags }}
+          labels: ${{ steps.docker-meta.outputs.labels }}
```
trentm, line 63: For the record this results in adding a number of labels that we never had before:

Before:

```
/usr/bin/docker buildx build --build-arg AGENT_DIR=/build/dist/nodejs --iidfile /home/runner/work/_temp/docker-actions-toolkit-fyNsIK/iidfile --tag docker.elastic.co/observability/apm-agent-nodejs:test --tag docker.elastic.co/observability/apm-agent-nodejs:test-latest --metadata-file /home/runner/work/_temp/docker-actions-toolkit-fyNsIK/metadata-file --push .
```

after:

```
/usr/bin/docker buildx build --build-arg AGENT_DIR=/build/dist/nodejs --iidfile /home/runner/work/_temp/docker-actions-toolkit-zpfy5E/iidfile --label org.opencontainers.image.created=2024-04-29T12:16:23.000Z --label org.opencontainers.image.description=Elastic APM Node.js Agent --label org.opencontainers.image.licenses=BSD-2-Clause --label org.opencontainers.image.revision=f9f28681ae69736728d15aaa93880fffba5f2e6d --label org.opencontainers.image.source=https://github.com/elastic/apm-agent-nodejs --label org.opencontainers.image.title=apm-agent-nodejs --label org.opencontainers.image.url=https://github.com/elastic/apm-agent-nodejs --label org.opencontainers.image.version=test-latest --tag docker.elastic.co/observability/apm-agent-nodejs:test-latest --metadata-file /home/runner/work/_temp/docker-actions-toolkit-zpfy5E/metadata-file .
```

trentm (commented):
> Use https://github.com/docker/metadata-action/tree/8e5442c4ef9f78752691e2d8f8d19755c6f78e81/?tab=readme-ov-file#typeref to get the tags/annotatiions/labels

You link to `#typeref` there, and link to:

> See https://github.com/elastic/apm-agent-nodejs/actions/runs/8878357355

which *does* show using a config of the "docker-metadata" action with:

```
type=ref,event=tag,prefix=,suffix=
```

However, the current diff in this PR does *not* use `type=ref,...` at all, so I am confused.

.github/workflows/release.yml
v1v: That's the expected behaviour for `main` but no for git tags.

I'm trying to do the same as done in

```
          if [ "${{ startsWith(github.ref, 'refs/tags') }}" == "false" ] ; then
            # for testing purposes
            echo "tag=test" >> "${GITHUB_OUTPUT}"
            echo "latest=test-latest" >> "${GITHUB_OUTPUT}"
          else
            # version without v prefix (e.g. 1.2.3)
            echo "tag=${GITHUB_REF_NAME/v/}" >> "${GITHUB_OUTPUT}"
            echo "latest=latest" >> "${GITHUB_OUTPUT}"
          fi
```

.github/workflows/release.yml
v1v: Those are the opencontainer default labels:
- https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys

We can remove them, sorry if I was not clear about this change

v1v:
> However, the current diff in this PR does not use type=ref,... at all, so I am confused.

I initially used but discovered that the `type: ref` uses the name of the tag and the tags in this particular case use the format `v<version>`, therefore the image name will contain `v<version>` rather than `<version>`


Those details were mentioned in a similar PR that I raised in the `apm-agent-python`:

- https://github.com/elastic/apm-agent-python/pull/2033#discussion_r1583086509

.github/workflows/release.yml
trentm: I don't have a strong argument either way. If we add these, then they somewhat become a promise that we will always define them in the future. I don't think it likely that that will be an issue.

On balance, I would be inclined to *not* add these labels unless there is a known use case to have them. I am fine either way however.

.github/workflows/release.yml
trentm: > I'm trying to do the same as done in

1. Okay, so first we want a `{{version}}` tag on "push tag":

    ```
    type=semver,pattern={{version}}
    ```

2. Then we want a `latest` tag on "push tag". Per https://github.com/docker/metadata-action/#latest-tag this is already handled by default. 

    Ideally we only want this "latest" if this is the latest major version, like we'd had with "push-docker.sh", but we are skipping that requirement for now. I *think* we could make this explicit by using the "flavor" input. On the current latest major branch (i.e. on "main") we want the default:

    ```
    flavor: |
      latest=auto
    ```

    and on other branches (like a future "4.x" branch if/when we have a "5.x" version) we'd want:

    ```
    flavour: |
      latest=false
    ```

3. Then we want a `test` (or perhaps `test-latest` or some other name) Docker tag for pushes to a branch. (We could limit this to branch "main" or to any branch possibly used for testing when the "on:" section is edited temporarily)

    ```
    type=ref,event=branch,value=test
    ```

    However, I'm not sure that works. The `value=` argument is not documented for https://github.com/docker/metadata-action/#typeref

    An alternative might be to limit to only "wanting" to have an `edge` tag for pushes to the default branch (i.e. to "main"), then we could use https://github.com/docker/metadata-action/?tab=readme-ov-file#typeedge

    ```
    type=edge
    ```


Assuming we'd be okay with the "edge" thing, then we could have this:


```yaml
      - name: Determine Docker tags and labels
        id: docker-meta
        uses: docker/metadata-action@8e5442c4ef9f78752691e2d8f8d19755c6f78e81  # v5.5.1
        with:
          images: ${{ env.DOCKER_IMAGE_NAME }}
          flavor: |
            latest=auto
          tags: |
            # "1.2.3" and "latest" on push of tag "v1.2.3"
            type=semver,pattern={{version}}
            # "edge" tag on push to default branch
            type=edge
```

reakaleek (approved):

.github/workflows/release.yml
v1v: Let me add your suggestion then

v1v: See 479d97b8

trentm (approved):

PR #4056, opened 2024-06-04
github-action: run buildkite action with GH secrets

v1v:

Run the Buildkite pipeline for the microbenchmarks without waiting for them. 

There is no need to wait since it uses a different CI—logs can be accessed through the UI; the GitHub action will provide the link to them.

Slack notifications are delegated to the Buildkite pipeline itself.

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation

### DIFF
--- a/.github/workflows/microbenchmark.yml
+++ b/.github/workflows/microbenchmark.yml
@@ -18,33 +18,16 @@ permissions:
 jobs:
   microbenchmark:
     runs-on: ubuntu-latest
-    # wait up to 1 hour
-    timeout-minutes: 60
+    timeout-minutes: 5
     steps:
-      - id: buildkite
-        name: Run buildkite pipeline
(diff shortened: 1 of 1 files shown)


reakaleek (approved):

david-luna:
@v1v just a tiny question.

> Slack notifications are delegated to the Buildkite pipeline itself.

does that mean that the action would pass always because is not waiting? so we rely on slack to get any failure notifications?

v1v:
> does that mean that the action would pass always because is not waiting? 

That's correct

> so we rely on slack to get any failure notifications?

The idea is to delegate the error reporting through the Buildkite pipeline, which supports Slack notifications.

ATM, the microbenchmarks are broken since a recent change in the `git https cloning`. I plan to fix it sooner rather than later and revisit the notifications to be smart enough to ping the right team in Slack if something goes sideways.

david-luna (approved):

Issue #4077, opened 2024-06-10
add support for synchronous gauge to OTel Bridge metrics support


trentm:
In `@opentelemetry/api` v1.9.0 support was added to the metrics API for creating a synchronous gauge: https://github.com/open-telemetry/opentelemetry-js/blob/main/api/CHANGELOG.md#190
See: https://github.com/open-telemetry/opentelemetry-js/issues/4296

We should see what we need to do to add support for that to our OTel bridge metrics support. At the least adding to test/opentelemetry-metrics/...

PR #4078, opened 2024-06-10
feat: support OTel span.addLink, span.addLinks; add similar APIs to the APM agent API

trentm:
This also updates a few places for the new otel/api v1.9.0.

Refs: #4071 (the dependabot update doesn't get everything)
Refs: #4070
Refs: #4069
Refs: #4077 (a separate issue for this other new feature in otel/api@1.9.0)
Refs: https://github.com/elastic/apm/pull/870

---

`@opentelemetry/api@1.9.0` added support for `span.addLink()` and `span.addLinks()`.
This updates our OTel bridge tracing support to work with those APIs.

This also adds public `addLink()` and `addLinks()` equivalent methods on our Transaction and Span APIs.
Note that the span links support still has limitations: only trace_id and span_id are sent/stored (no span link attributes, no 'tracestate' in the span context).

I created a separate issue for the other feature added to otel/api@1.9.0.

### DIFF
--- a/lib/instrumentation/generic-span.js
+++ b/lib/instrumentation/generic-span.js
@@ -281,7 +281,7 @@ GenericSpan.prototype._serializeOTel = function (payload) {
 // span link as it will be serialized and sent to APM server. If the linkArg is
 // invalid, this will return null.
 //
-// @param {Array} linkArg - An object with a `context` property that is
+// @param {Object} linkArg - An object with a `context` property that is
 //    a Transaction, Span, or TraceParent instance; an OTel SpanContext object;
 //    or a W3C trace-context 'traceparent' string.
 function linkFromLinkArg(linkArg) {
--- a/lib/instrumentation/generic-span.js
+++ b/lib/instrumentation/generic-span.js
@@ -170,25 +170,28 @@ GenericSpan.prototype.addLabels = function (labels, stringify) {
   return true;
 };
 
-// This method is private because the APM agents spec says that (for OTel
-// compat), adding links after span creation should not be allowed.
-// https://github.com/elastic/apm/blob/main/specs/agents/span-links.md
-//
-// To support adding span links for SQS ReceiveMessage and equivalent, the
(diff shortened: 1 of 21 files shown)


lib/opentelemetry-bridge/OTelSpan.js
```diff
@@ -167,6 +167,14 @@ class OTelSpan {
     return this;
   }
 
+  addLink(link) {
```
david-luna, line 170: According to the [Span interface](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/trace/span.ts#L80-L98) both `addLink` and `addLinks` must return the span itself. You already did it in [lib/opentelemetry-bridge/OTelBridgeNonRecordingSpan.js](https://github.com/elastic/apm-agent-nodejs/pull/4078/files#diff-d05e9b0314bfa0a1928fe1612ae124355bcbc74005f626adee5880ea4cdf90f2) so why not here?

david-luna (dismissed):
Thanks for this PR. I went too quick bumping the OTEL API package :)

All looks good, just couple of comments about types

lib/opentelemetry-bridge/OTelSpan.js
trentm: Fixed in commit c0340b24, along with a test for it.

lib/instrumentation/generic-span.js
```diff
@@ -278,9 +281,9 @@ GenericSpan.prototype._serializeOTel = function (payload) {
 // span link as it will be serialized and sent to APM server. If the linkArg is
 // invalid, this will return null.
 //
-// @param {Object} linkArg - An object with a `context` property that is a
-//    Transaction, Span, or TraceParent instance, or a W3C trace-context
-//    'traceparent' string.
+// @param {Array} linkArg - An object with a `context` property that is
```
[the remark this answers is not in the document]
trentm: Commit c0340b24

david-luna (approved):

PR #4121, opened 2024-07-05
updatecli: use shared policy

v1v:
### What does this PR do?

Use a shared [updatecli policy](https://www.updatecli.io/docs/core/shareandreuse/). The exact configuration is used in different GitHub repositories.

I ran https://github.com/elastic/apm-agent-nodejs/actions/runs/9806067124

### Checklist

- [x] Implement code
- [ ] Add tests
- [ ] Update documentation


.ci/updatecli/values.d/scm.yml
```diff
@@ -0,0 +1,7 @@
+scm:
+  enabled: true
+  owner: elastic
+  repository: apm-agent-nodejs
```
reakaleek, line 4: I'm starting to wonder if we could utilize `$GITHUB_` env vars for this.

But in the shared policy.

reakaleek (dismissed):

.ci/updatecli/values.d/scm.yml
v1v: They are there, but I forced it to play safe.

```

# {{ $GitHubRepositoryList := env "GITHUB_REPOSITORY" | split "/"}}

...

{{ if or (.scm.enabled) (env "GITHUB_REPOSITORY") }}
scms:
  default:
    kind: "github"
    spec:
      # Priority set to the environment variable
      user: '{{ default $GitHubUser .scm.user}}'
      email: '{{ .scm.email }}'
      owner: '{{ default $GitHubRepositoryList._0 .scm.owner }}'
      repository: '{{ default $GitHubRepositoryList._1 .scm.repository}}'
...
```

reakaleek: Ah okay. 👍

update-compose.yaml
```diff
@@ -0,0 +1,12 @@
+policies:
```
trentm, line 1: Does this file need to be at the top-level?

... after some digging, I gather that it does (per https://www.updatecli.io/docs/core/compose/).
IMO, it is unfortunate that the tool is called "updatecli" but the config file just has "update" in it, which is ambiguous.

Could we please have a comment at the top of the file giving context. The file "update-compose.yaml" is definitely not a well-known file name. Something like:

```suggestion
# Config file for `updatecli compose ...`.
# https://www.updatecli.io/docs/core/compose/
policies:
```

update-compose.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - name: Handle apm-data server specs
```
trentm, line 2: Unfortunately these "name:"  `Handle ...` strings do not appear in the workflow run log at all, e.g.: https://github.com/elastic/apm-agent-nodejs/actions/runs/9806099782/job/27077125378

update-compose.yaml
```diff
@@ -0,0 +1,12 @@
...
+    policy: ghcr.io/elastic/oblt-updatecli-policies/apm/apm-data-spec:0.2.0@sha256:7069c0773d44a74c4c8103b4d9957b468f66081ee9d677238072fe11c4d2197c
```
trentm, line 3: These policy files are behind credentials that I don't have, so are uninspectable by me. Do these live in a repo somewhere that I could look at? Here I guess:
https://github.com/elastic/oblt-updatecli-policies/tree/main/updatecli/policies/apm/apm-data-spec

The "0.2.0" container image label doesn't have a matching git tag in that repository. I see there is a changelog, but it is a bit of a guess which commit a particular container image build corresponds to.

Perhaps at least a clickable URL to be able to go look at the related code? Something like:

```suggestion
    # https://github.com/elastic/oblt-updatecli-policies/tree/main/updatecli/policies/apm/apm-data-spec
    policy: ghcr.io/elastic/oblt-updatecli-policies/apm/apm-data-spec:0.2.0@sha256:7069c0773d44a74c4c8103b4d9957b468f66081ee9d677238072fe11c4d2197c
```

Though, that is of limited help, because it is very difficult to understand what this workflow is attempting to do and to follow what the policy is doing. I can make guesses, but for example, it isn't clear to me what build process in oblt-updatecli-policies is rendering the https://github.com/elastic/oblt-updatecli-policies/blob/main/updatecli/policies/apm/apm-data-spec/updatecli.d/default.tpl template to a useable final thing that updatecli eventually executes.

trentm (approved):
Approved with some reservations. The end result here feels like a *very* complex solution to the problem of periodically checking for updates to some files vendored from another repo.

update-compose.yaml
v1v: I'll raise an issue in the upstream project so it can be honoured or at least they know this case. Thanks

update-compose.yaml
v1v: > These policy files are behind credentials that I don't have

Not at all, you can login to GitHub itself:
- https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

>  Do these live in a repo somewhere that I could look at? 

I didn't want to explicitly write the GitHub repository, it's private at the moment, I'm waiting for an approval to set it public available.

> [elastic/oblt-updatecli-policies@main/updatecli/policies/apm/apm-data-spec](https://github.com/elastic/oblt-updatecli-policies/tree/main/updatecli/policies/apm/apm-data-spec)

v1v: > Though, that is of limited help, because it is very difficult to understand what this workflow is attempting to do and to follow what the policy is doing.

Do you know how I can help with this? I'm open to making this better.

The idea behind the updatecli shareable policies is to share updatecli pipelines between different GitHub projects and then use 'values' to configure the behaviour. That's why the '[default.tpl](https://github.com/elastic/oblt-updatecli-policies/blob/main/updatecli/policies/apm/apm-data-spec/updatecli.d/default.tpl)' exists and then infer the values.

In this particular case, these changes are all related to syncing up the apm and apm-data specs/gherkins in all the APM Agents, see https://github.com/elastic/observability-robots/issues/1751 (Only for Elastic employees)

v1v:
> The end result here feels like a very complex solution to the problem of periodically checking for updates to some files vendored from another repo.

I'm not sure what's the reason for the complex solution:
- Configure the values in specific YAML files (one per policy)
- Configure the shared policies in the root file `update-compose.yaml`
- GitHub action to configure the environment (updatecli, credentials) and run the updatecli policies.

It's idempotence and will create Pull Requests with changes if needed.

All the above is required in 8 or 9 different GitHub projects. I already went through changing behaviour in the updatecli pipeline, and had to deal with several PRs.

- https://github.com/elastic/apm-agent-dotnet/pull/2263
- https://github.com/elastic/apm-agent-go/pull/1570
- https://github.com/elastic/apm-agent-java/pull/3497
- https://github.com/elastic/apm-agent-nodejs/pull/3845
- https://github.com/elastic/apm-agent-php/pull/1123
- https://github.com/elastic/apm-agent-python/pull/1957
- https://github.com/elastic/apm-agent-ruby/pull/1439

Using a centralised shared place for well defined policies will help us by different means:
- easy to onboard new projects
- reduce duplication
- tests those polices in isolation.
- version them

The very end goal is to use automation, therefore if a new version is released for the updatecli policies, the file `update-compose.yaml` will be automatically updated itself, see https://github.com/elastic/apm-agent-go/pull/1634


Please let me know if I need to clarify anything, I'll raise a few GitHub issues in the updatecli project based on your valuable feedback, 🙏

update-compose.yaml
v1v: > Does this file need to be at the top-level?

I tried to use a different location but it didn't handle well certain configurations, see https://github.com/updatecli/updatecli/issues/2266

I guess, it will eventually support it.

trentm:
> Please let me know if I need to clarify anything

All good. Thanks for feedback.

Perhaps add that comment at the top of "update-compose.yaml" stating that the file is related to `updatecli`? If that is a pain because you need this to be common across the many apm-agent repos, then you can skip it. Thanks.


Note to self: I can browse the published ghcr.io packages here:
https://github.com/orgs/elastic/packages?tab=packages&q=oblt
https://github.com/elastic/oblt-updatecli-policies/pkgs/container/oblt-updatecli-policies%2Fapm%2Fapm-data-spec

v1v:
> Perhaps add that comment at the top of "update-compose.yaml" stating that the file is related to updatecli? 

Not at all, see [`5981a5e` (#4121)](https://github.com/elastic/apm-agent-nodejs/pull/4121/commits/5981a5e843c7cf59721a02801df8b4533766f8cf)

reakaleek (approved):

trentm (approved):

PR #4139, opened 2024-07-15
chore(deps): pin lru-cache ver, *downgrade* rimraf and glob to ensure compat with node 14.17; block updates to newer majors

trentm:
isaacs aggressively dropped support for Node.js versions in the latest majors
of these packages. Let's stop dependabot creating PRs for them again.


jackshirazi (dismissed):

jackshirazi (dismissed):

trentm:
More complexity to support Node.js 14.17.

## glob

The penultimate glob@10 release supported these Node.js versions:

```
  "engines": {
    "node": "14 >=14.21 || 16 >=16.20 || 18 || 20 || >=22"
  },
```

but then "engines" was dropped entirely in latest "10" in maintainer frustration.

We need Node.js 14.17 support currently and glob@10 latest doesn't work with that (uses `require('node:url')`).
Options:

1. pin to 10.3.4, or a slightly later but not latest glob@10
2. use glob@9 (9.3.5) which supports:

```
  "engines": {
    "node": ">=16 || 14 >=14.17"
  }
```

I'm choosing option 2: `glob@9`.

## rimraf

Likewise we may need rimraf@4, because latest rimraf@5 is min Node.js 14.20 and uses glob@10.


## lru-cache

lru-cache is slightly more painful

- We cannot use lru-cache@11 because it drops Node.js 14, 16, 18.
- lru-cache@10 is what we have been using (10.0.1), but its later minor versions toyed with dropping node 14 and 16, then back to `"node": "14 >= 14.21 || 16 >= 16.20 || 18 >=18.20 || 20 || >=22"`, and finally settling on not having an "engines" specified at all (presumably in some frustration to just move on).
- We *do potentially* want lru-cache@10, rather than lru-cache@9 for the https://github.com/isaacs/node-lru-cache/issues/309 bug fix, which was in 10.0.1.  We use lru-cache at runtime (it isn't just a devDep), and we *do* use the LRUCache `dispose` functionality, albeit without a `ttl` that is part of that bug report.
- glob@9 deps on path-scurry@^1.6.1, which deps on lru-cache@^10.2.0. Fun fun.  (Dropping or toying with dropping major Node.js vers in minors really complicates things.)

Practically I think lru-cache@10 will no longer change and I think it will work with 14.17 (our current min-node).

Solution: Update to lru-cache@10.2.0 and lock to that version. This satisfies the *current* glob@9 deps, has the lru-cache@10.0.1 fix we want and is before lru-cache@10.x started dropping versions.

trentm:
Annnnnd further back we go.

The issue now is that glob@9 depends on path-scurry@1. Later versions of path-scurry@1 bumped the min-supported Node.js version to 14.18:

```
  "engines": {
    "node": ">=16 || 14 >=14.18"
  },
```

for the legitimate reason that using `node:`-prefixed imports helps with adding support for the deno runtime (see: https://github.com/isaacs/path-scurry/pull/16). 

It would have been nice if this had been done in a new major version, but (a) that ship has sailed and (b) I can see why bumping from "14" to "14.18" as a min-supported node feels like something one could get away with in a minor.

# options

1. Move our devDeps back to `glob@8` and `rimraf@7`, which don't use path-scurry and work with Node.js 14.17 (pretty sure).
2. Bump the min supported Node.js for this package from 14.17 to 14.18 on the assumption/basis that the package ecosystem is assuming that is fine to allow `node:`-prefix usage to allow bun support.
3. Bump the min supported Node.js to later 14 versions, perhaps the latest 14.x release, or a vague "14" and don't specify.

I think options 2 and 3 are *reasonable* to do, in general. However, I'd rather not for *this PR* that was just about bumping some deps. Also, doing this just for *devDeps* feels a bit like the tail wagging the dog, so let's not bump our min-node for now.

That means option 1. Let's see if that works. It will mean this warning on npm install:

```
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```

jackshirazi (approved):

Issue #4171, opened 2024-07-31
Version v4.7.1 Lambda Layer Misconfigured: All Regions Pointing to us-east-1


changboyu:
**Describe the bug**

I found that the ARN table for Lambda Layer version v4.7.1 is incorrect. All regions are pointing to the us-east-1 region, as depicted in the attached image. 

[Release page](https://github.com/elastic/apm-agent-nodejs/releases/tag/v4.7.1)

<img width="1666" alt="image" src="https://github.com/user-attachments/assets/f4bc2702-7f49-42ed-a728-6a5fa834ef75">

This impact page [Monitoring AWS Lambda Node.js Functions](https://www.elastic.co/guide/en/apm/agent/nodejs/current/lambda.html) 

<img width="938" alt="image" src="https://github.com/user-attachments/assets/2d01d404-23de-44ce-90d0-4e1a84ee9aaa">

**To Reproduce**

1. Visit [Monitoring AWS Lambda Node.js Functions](https://www.elastic.co/guide/en/apm/agent/nodejs/current/lambda.html) stop to load the lambda layer arn for other regions.
2. Choose a region rather than us-east-1
3. The guide's APM agent layer is missing and shows <SELECTED_AGENT_LAYER_ARN>

**Expected behavior**
Each Region should have their own lambda layer
</details>

trentm:
Thanks for reporting this! I think I've found the issue. We'll try to get a new release out in the next few days.

trentm:
The v4.7.1 release published the lambda layers incorrectly. Instead of
publishing the layer zip to each region once, it published to us-east-1 every
time, resulting in:

```
arn:aws:lambda:us-east-1:267093732750:layer:elastic-apm-node-ver-4-7-1:1
arn:aws:lambda:us-east-1:267093732750:layer:elastic-apm-node-ver-4-7-1:2
arn:aws:lambda:us-east-1:267093732750:layer:elastic-apm-node-ver-4-7-1:3
...
arn:aws:lambda:us-east-1:267093732750:layer:elastic-apm-node-ver-4-7-1:22
```

## tl;dr 

This happened because:
1. PR #4109 changed the "release.yml" workflow to use `uses: elastic/oblt-actions/aws/auth@...` for AWS auth, from a different auth mechanism. This new mechanism ultimately uses `aws-actions/configure-aws-credentials` which sets the `AWS_DEFAULT_REGION` and `AWS_REGION` envvars (by default to `us-east-1`).
[...]

PR #4172, opened 2024-07-31
fix(release): ensure we publish Lambda Layers to the intended AWS region

trentm:
Before this change an errant AWS_REGION envvar would override the
AWS_DEFAULT_REGION envvar. Explcitly using --region=... will win.

Fixes: #4171
Refs: #4109


trentm:
cla/check

david-luna (approved):

Issue #4238, opened 2024-09-16
Add support for express@5


trentm:
express@5.0.0 was released recently (2024-09-10). We will need to do some work on our instrumentation to support it.

Just attempting to instrument v5 with the current instrumentation results in tests *crashing*:

```diff
diff --git a/lib/instrumentation/modules/express.js b/lib/instrumentation/modules/express.js
index d7784cfa..417e27de 100644
--- a/lib/instrumentation/modules/express.js
+++ b/lib/instrumentation/modules/express.js
@@ -17,9 +17,9 @@ module.exports = function (express, agent, { version, enabled }) {

   agent.setFramework({ name: 'express', version, overwrite: false });

-  if (!semver.satisfies(version, '^4.0.0')) {
+  if (!semver.satisfies(version, '>=4.0.0 <6.0.0')) {
     agent.logger.debug(
-      'express version %s not supported - aborting...',
+      'instrumenting express@%s is not supported, skipping',
       version,
     );
     return express;
```

Test crash:

```
% node test/instrumentation/modules/express/basic.test.js
...
# expose app.use handle properties
TypeError: Cannot read properties of undefined (reading 'stack')
    at handle (/Users/trentm/el/apm-agent-nodejs3/test/instrumentation/modules/express/basic.test.js:512:35)
    at handle (/Users/trentm/el/apm-agent-nodejs3/lib/instrumentation/modules/express.js:80:25)
    at Layer.handleRequest (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/lib/layer.js:145:17)
    at trimPrefix (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/index.js:337:13)
    at /Users/trentm/el/apm-agent-nodejs3/node_modules/router/index.js:295:9
    at processParams (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/index.js:575:12)
    at next (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/index.js:289:5)
    at Function.handle (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/index.js:183:3)
    at router (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/index.js:59:12)
    at handle (/Users/trentm/el/apm-agent-nodejs3/lib/instrumentation/modules/express.js:80:25)
    at Layer.handleRequest (/Users/trentm/el/apm-agent-nodejs3/node_modules/router/lib/layer.js:145:17)
[...]

PR #4239, opened 2024-09-16
test(express): drop express@5 from TAV tests

trentm:
We don't yet support instrumenting express@5. 5.0.0 was recently released.
Our TAV config was including express@5 in tests.

Refs: https://github.com/elastic/apm-agent-nodejs/issues/4238


trentm (commented):
/test tav express

david-luna (approved):

PR #4291, opened 2024-10-31
chore: better logging for config state and transaction sampling

trentm:
This adds a periodic (once per minute) dump of effective config changes
from the default, once per minute at 'trace'-level.

This also fixes an issue where a central-config update of `log_level`
did not update the *child* logger on the APM client. Only the agent's
own logger's level was updated.

This adds trans.sampled to some debug logging output.

---

@david-luna Note this is targeting 3.x. If this looks reasonable, I'll adapt the patch for "main" as well.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -24,6 +24,7 @@ const errors = require('./errors');
 const { InflightEventSet } = require('./InflightEventSet');
 const Instrumentation = require('./instrumentation');
 const { elasticApmAwsLambda } = require('./lambda');
+const logging = require('./logging');
 const Metrics = require('./metrics');
 const parsers = require('./parsers');
 const symbols = require('./symbols');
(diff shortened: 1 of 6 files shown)


trentm:
I need to work through the test errors.

trentm:
The v20 test failures are fixed by https://github.com/elastic/apm-agent-nodejs/pull/4295

lib/agent.js
```diff
@@ -324,6 +325,23 @@ Agent.prototype.start = function (opts) {
 
   this.logger.info(preambleData, 'Elastic APM Node.js Agent v%s', version);
 
+  if (!logging.isLoggerCustom(this.logger)) {
+    // Periodically dump the current config (delta from defaults) when logging
+    // at "trace"-level. This allows getting the effective config from a running
+    // agent by setting trace-level logging and getting 1 minute of logs.
+    // (Sometimes getting logs from application *start* is no possible.)
+    setInterval(() => {
+      if (this.logger.isLevelEnabled('trace')) {
```
david-luna, line 334: could the level be changed at runtime? If not I think we could `setInterval` within this block.

david-luna (approved):

david-luna:
> @david-luna Note this is targeting 3.x. If this looks reasonable, I'll adapt the patch for "main" as well.

makes sense to have record of the diif in config and also the sampled transactions. This is `main` material :)

lib/agent.js
trentm: Yes it can be changed at runtime via central config, and in the use case I needed this for that *was* the case.
I balked at integrating this change with central config handling. 

However, I may do that for the *main* branch version of this change, because having a `setInterval` running all the time kinda sucks. On the other hand, having a setInterval (change in behaviour, however slight) that only turns on in "trace" made could mean more subtle bugs. It would be awful if there was a crash that only resulted from code that enabled when in "trace" mode. I'm talking generally.

Issue #4293, opened 2024-11-01
Lambda instrumentation does not take if _HANDLER has a period other in the module path


trentm:
A Lambda function has a "Handler" string (set on the `_HANDLER` envvar in the Lambda invocation process) that indicates the location of the handler function (both its module file location, and the export lookup path inside that module).
E.g. `some/subdir/filename.handlerFunctionName` points to a JS module at `$LAMBDA_TASK_ROOT/some/subdir/filename.{js,cjs,mjs}` and the `handlerFunctionName` export from that module.

The APM agent's processing of that handler string breaks if there is a period (`.`) in the module path. For example, this works:

```bash
_HANDLER="src/functions/myfunc/handler.main"
```

but this does not:

```bash
_HANDLER="./src/functions/myfunc/handler.main"
```

I suspect also that this would fail, but I have not tested.

```bash
_HANDLER="src/functions/my.other.func/handler.main"
```

The symptom is that a Lambda function setup to be instrumented by the Node.js APM agent does not show this log.debug:

```
{"log.level":"debug",...,"log.logger":"elastic-apm-node","message":"instrumenting /var/task/...@$LATEST module (enabled=true, ...)"}
```

Then no Lambda transaction is created, and then (typically) any additional spans from other instrumentation will not get created for lack of a current transaction.

PR #4294, opened 2024-11-01
fix(lambda): fix lambda instr to work with a '.' in the handler module path

trentm:
Before this if the _HANDLER string had a '.' in the module path (the
part before the 'moduleName.functionExport'), then the parsing of
that handler string would silently produce bogus 'lambdaHandlerInfo'
that would result in a RITM path that would never actually get loaded,
hence no Lambda instrumentation.

Fixes: #4293

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -33,6 +33,24 @@ Notes:
 
 See the <<upgrade-to-v4>> guide.
 
+==== Unreleased
+
+[float]
+===== Breaking changes
+
+[float]
+===== Features
+
+[float]
+===== Bug fixes
+
+* Fix AWS Lambda instrumentation to work with a "handler" string that includes
+  a period (`.`) in the module path. E.g. the leading `.` in `Handler: ./src/functions/myfunc/handler.main`. ({issues}4293[#4293]).
+
+[float]
+===== Chores
+
+
 [[release-notes-4.8.0]]
 ==== 4.8.0 - 2024/10/08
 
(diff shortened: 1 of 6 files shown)


trentm:
I've verified manually that this works in a real Lambda with a leading './' on the handler path.
I see this log output line:

```
2024-11-01T22:57:51.242000+00:00 2024/11/01/[$LATEST]395d9c62d54d40c08e373171c8fad153 {"log.level":"debug","@timestamp":"2024-11-01T22:57:51.238Z","log.logger":"elastic-apm-node","ecs.version":"8.10.0","message":"instrumenting /var/task/src/functions/hi/handler.js@$LATEST module (enabled=true, isImportMod=false)"}
```

and instrumentation works.

## testing the added log.warn

I then broke the implementation again, but kept the log.warn's. I published a new layer and ran my test Lambda with that.
Instrumentation, again, isn't working (as expected) and I see this log.warn:

```
2024-11-01T23:01:13.136000+00:00 2024/11/01/[$LATEST]279be659b3e14d49aa948e43ecbfb74e {"log.level":"warn","@timestamp":"2024-11-01T23:01:13.128Z","log.logger":"elastic-apm-node","ecs.version":"8.10.0","fullHandlerString":"./src/functions/hi/handler.main","moduleRoot":"./src/functions/hi/","module":"","message":"could not find Lambda handler module file (ESM not yet supported)"}
```

This hopefully, will give us a clue if we hit an issue again. E.g. we don't support .mjs handler modules yet, so we expect to see a warning.

david-luna (approved):

Issue #1966, opened 2021-02-05
TypeError: Cannot set property 'created_at' of undefined


htjuliantara:
Versions:

node: 10.17.0
apm-agent-nodejs: 3.10.0
elasticsearch, apm-server: 7.9.2
the nodejs client throws (briefly after start when first sample is apparently processed) with

```
TypeError: Cannot set property 'created_at' of undefined
    at breadthFilter (/home/gateway/node_modules/breadth-filter/index.js:35:21)
    at Object.truncTransaction [as transaction] (/home/gateway/node_modules/elastic-apm-http-client/lib/truncate.js:107:18)
    at Client._encode (/home/gateway/node_modules/elastic-apm-http-client/index.js:333:34)
    at Client.encodeObject (/home/gateway/node_modules/elastic-apm-http-client/index.js:275:15)
    at Array.map (<anonymous>)
    at Client._writevCleaned (/home/gateway/node_modules/elastic-apm-http-client/index.js:279:22)
    at processBatch (/home/gateway/node_modules/elastic-apm-http-client/index.js:253:12)
    at Client._writev (/home/gateway/node_modules/elastic-apm-http-client/index.js:271:3)
    at doWrite (/home/gateway/node_modules/readable-stream/lib/_stream_writable.js:409:97)
    at clearBuffer (/home/gateway/node_modules/readable-stream/lib/_stream_writable.js:505:5)
    at Client.Writable.uncork (/home/gateway/node_modules/readable-stream/lib/_stream_writable.js:321:94)
    at Timeout._corkTimer.setTimeout [as _onTimeout] (/home/gateway/node_modules/elastic-apm-http-client/index.js:302:14)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
```

config:

```
ELASTIC_APM_SERVICE_NAME=xxx
ELASTIC_APM_SECRET_TOKEN=xxx
ELASTIC_APM_SERVER_URL=http://
ELASTIC_APM_VERIFY_SERVER_CERT=false
ELASTIC_APM_CAPTURE_BODY=all
```

gladly tell me what other information would be helpful :-)

astorm:
[...]

PR #4359, opened 2024-12-10
fix: guard against apm-clients encoding/truncation of trace objects throwing

trentm:
We have observed an exception from 'truncate.transaction()' twice,
to my knowledge. I have been unable to reproduce it so far.
A possible side-effect of client._encode throwing in the
apm-client is that it effectively gets wedged ('corked()').

Refs: #1966

### DIFF
--- a/lib/apm-client/http-apm-client/index.js
+++ b/lib/apm-client/http-apm-client/index.js
@@ -405,6 +405,15 @@ Client.prototype._resetEncodedMetadata = function () {
 
   // This is the only code path that should set `_encodedMetadata`.
   this._encodedMetadata = this._encode({ metadata }, Client.encoding.METADATA);
+  if (!this._encodedMetadata) {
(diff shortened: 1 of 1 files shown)


trentm:
This guard is, I believe, the right thing to do. However it is slightly frustrating because I've not been able to reproduce a case where `truncate.transaction(...)` hits the exception we've observed.

The second instance I've seen in the wild was *slightly* different in that it threw on node_modules/breadth-filter/index.js:40, instead of line 35 (as in the stacktrace in https://github.com/elastic/apm-agent-nodejs/issues/1966).

This is a wild guess, but in both instances the attribute name (`created_at`, `person_type`) *suggests* a possible link to usage of `sequelize`. I *do* see some `Object.defineProperty` handling in sequelize that could possibly be related, but I've not managed to reproduce anything yet.

david-luna (approved):

david-luna (approved):

PR #4441, opened 2025-01-27
test: test FIPS 140 compliance

trentm:
Periodically (weekly) run the test suite (most of it) inside the Chainguard Base FIPS container to test FIPS 140 compliance. If that run fails, slack is notified.

Re "most of it": The test run skips tests that depend on an external service (databases, kafka, et al.). 1. I don't know how to get GH Actions "services" to be available to a test job running inside a Docker container. 2. Skipping these tests is file for FIPS checking because none of the instrumentations do any crypto.


trentm:
During dev of this PR I was running the "test-fips" workflow for every change. The final intended state is to only run "test-fips" weekly on a schedule; therefore the "test-fips" check doesn't show here.  

However, you can see a successful run here: https://github.com/elastic/apm-agent-nodejs/actions/runs/13018228802/job/36312550102
This is from commit 7b10b3d.

package.json
```diff
@@ -172,7 +172,6 @@
     "got": "^11.8.5",
     "graphql": "^16.6.0",
     "handlebars": "^4.7.3",
-    "https-pem": "^3.0.0",
```
trentm, line 175: Reviewer note:
This devDep was used to provide a self-signed TLS cert to use from some HTTPS testing. It was problematic because it generated the cert in an npm postinstall script, which is icky, and it used 1024 bits for an RSA key, which is too small for FIPS 140.

The test suite had 3-4 separate places where static certs or cert generation was being used for tests. I've consolidated all those into one place: test/fixtures/certs/
It uses a 2048-bit RSA key, which satisfies the strict FIPS OpenSSL provider.

test/sanitize-field-names/_shared.js
```diff
@@ -96,6 +96,7 @@ function createAgentConfig(values = {}) {
     metricsInterval: 0,
     centralConfig: false,
     captureBody: 'all',
+    apmServerVersion: '8.17.0',
```
trentm, line 99: Reviewer note: This removes a spurious log.error from the APM agent used in the test/sanitize-field-names/*.test.js tests. The APM agent is started without an APM server having been created. The APM agent then complains when trying to get the APM server's version.  We use this `apmServerVersion` technique in many other tests.

test/sanitize-field-names/hapi.test.js
```diff
@@ -64,7 +64,7 @@ async function runTest(
   agent._config(agentConfig);
   const server = Hapi.server({
     port: 0,
-    host: 'localhost',
+    host: '127.0.0.1',
```
trentm, line 67: Reviewer note: The test env inside the chainguard container was hitting the failing case we've seen before where (a) the test server thinks "localhost" is IPv6 and (b) the test client uses IPv4 (127.0.0.1). This fixes that by dropping the ambiguity.

david-luna (approved):

PR #4578, opened 2025-04-22
feat: support sourcemap resolution for *ESM* stack frames in captured APM errors

trentm:
This also modernizes the examples/typescript/... a bit and (notably) changes it to use ESM.
This was the repro example I was using for this fix -- though I removed the `apm.captureError(...)` I was manually using in examples/typescript/index.ts for dev.

Yes, I don't have a test case here. We should discuss if we're going to require that.

### DIFF
--- a/docs/reference/starting-agent.md
+++ b/docs/reference/starting-agent.md
@@ -80,12 +80,12 @@ apm.start({
   serverUrl: 'https://...',
   secretToken: '...',
   // ...
-})
+});
 ```
 
 ```ts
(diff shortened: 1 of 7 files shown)


david-luna:
> Yes, I don't have a test case here. We should discuss if we're going to require that.

Thew change is quite straight forward and not changing existing behavior. I'm okay to not have a test if it requires to have another TypeScript project as a fixture.

david-luna (approved):

PR #4648, opened 2025-06-11
Create doc-preview-comment.yml

theletterf:
This adds a flow that automatically renders links to docs previews in PRs that edit docs.

Contributes to https://github.com/elastic/docs-projects/issues/492

### DIFF
--- a/.github/workflows/doc-preview-comment.yml
+++ b/.github/workflows/doc-preview-comment.yml
@@ -1,0 +1,72 @@
+name: "Docs preview comment"
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+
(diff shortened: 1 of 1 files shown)


trentm (approved):

trentm:
Approved, then noticed this is only in draft.  Do as you will. :)

v1v:
@theletterf , I see some potential to actually include this logic in the `elastic/oblt-actions` or a similar centralised repository, for two main reasons:
1) Reuse components
2) Simplify maintenance


We did something 'similar' for https://github.com/elastic/apm-agent-nodejs/blob/main/.github/workflows/github-commands-comment.yml

What do you think?

PR #4757, opened 2025-09-04
chore!: drop experimental Next.js instrumentation

trentm:
This removes Next.js instrumentation, which was only ever in tech preview
and only supported up to Next.js 13.3.0, which is now obsolete. The
built-in OpenTelemetry instrumentation in Next.js is a better option
users.


trentm:
Aside: Most of our dependabot "security" issues are from the out-dated examples/nextjs and Next.js instr *tests*. While this is breaking, it only ever supported Next.js versions up to 13.3.0, which stopped getting support in 2022. Also, our instrumentation support here was only ever a tech preview, so I believe we can drop it in a minor.

david-luna (approved):

david-luna:
nit: `dependabot.yml` has an ignore rule for next in https://github.com/elastic/apm-agent-nodejs/blob/f09ef2a75e0ca8efb77d58bc3ac015db7002c7b8/.github/dependabot.yml#L37

trentm:
> `dependabot.yml` has an ignore rule for next in

Thanks. I need to remove that.

david-luna (dismissed):

docs/reference/nextjs.md
```diff
@@ -1,149 +0,0 @@
...
-
-Open [http://localhost:3000](http://localhost:3000) in your browser to load your Next.js app. If you used the `create-next-app` tool above, it defines an [/api/hello](http://localhost:3000/api/hello) API endpoint. You can provide some artificial load by running the following in a separate terminal:
-
-```bash
-while true; do sleep 1; curl -i http://localhost:3000/api/hello; done
-```
-
-Visit your Kibana APM app and, after a few seconds, you should see a service entry for your Next.js app. The service name will be pulled from the "name" field in "package.json". It can be overriden with [`serviceName`](/reference/configuration.md#service-name). Here is an example:
-
-![Kibana APM app showing Next.js my-app](images/nextjs-my-app-screenshot.png)
```
david-luna, line 105: nit: the image referenced here is not deleted in this PR.

david-luna (approved):

david-luna (approved):

akselinurmio:
This change unfortunately breaks existing Next.js applications that rely on the `elastic-apm-node/start-next.js` module. While the instrumentation may have originally been marked as experimental, it was fully functional and documented in Elastic’s public documentation, effectively making it part of the public API surface.

Given this, removing it in a minor (or non-major) release introduces a breaking change for users still depending on it. I would strongly recommend reverting this removal and deferring it to the next major release, ensuring users have sufficient time to migrate to the built-in OpenTelemetry instrumentation in Next.js.

akselinurmio:
Wait a minute: is the new recommended way also experimental? Just read the [_OpenTelemetry bridge_ page](https://www.elastic.co/docs/reference/apm/agents/nodejs/opentelemetry-bridge) in your documentation now, and it says: "Integration with the OpenTelemetry Tracing API was added as _experimental_ in v3.34.0.". Is there any stable (non-experimental) way of instrumenting Next.js?

trentm:
@akselinurmio  Thanks for bring this up.

> This change unfortunately breaks existing Next.js applications that rely on the `elastic-apm-node/start-next.js` module. While the instrumentation may have originally been marked as experimental, it was fully functional and documented in Elastic’s public documentation, effectively making it part of the public API surface.

The Next.js support was documented as a "technical preview", which is intended to indicate that it is **separate** from finalized functionality that is subject to typical support mechanisms. However, I don't have a link to an official Elastic support document that spells out what is meant by "technical preview".

From the [old "Get started with Next.js" document](https://github.com/elastic/apm-agent-nodejs/pull/4757/files#diff-b2ec28042380cbca66dfff777849685ac12bf456371861546a974863d8c74badL18-L20):

>> This Next.js instrumentation is a *technical preview* while we solicit feedback from Next.js users. Currently `next` versions `>=12.0.0 <13.3.0` are supported. If you are a Next.js user, please help us provide a better Next.js observability experience with your feedback on our [Discuss forum](https://discuss.elastic.co/tags/c/apm/nodejs). 

---

I'd like to understand exactly what the breakage is for you to better suggest a fix or workaround.

- What version of Next.js are you using? The instrumentation was intentionally only active for `next` versions `>=12.0.0 <13.3.0` which, IIUC, are all obsolete versions. The last version supported, `next@13.2.4`, was released in 2023 (`'13.2.4': '2023-03-10T00:50:30.690Z',`).
- If you have pinned your `next` version used to something less-than 13.3.0, would you be able to pin your used `elastic-apm-node` version to `4.13.0` -- the last version with the Next.js instrumentation? I realize this isn't ideal, in case the `4.x` release line includes new functionality or fixes that you might later want. FWIW, our development focus is now primarily on a native-OTel-based agent ([EDOT Node.js](https://www.elastic.co/docs/reference/opentelemetry/edot-sdks/node).
- Could you use `start.js` rather than `start-next.js`? The difference was small. The only addition in `start-next.js` was a [small block of code to special-case calling `apm.flush()` at shutdown](https://github.com/elastic/apm-agent-nodejs/pull/4757/files#diff-a3ad0ecbc071a79a97226cba5f266ed6ccb20c0d0f4a5ee829dc4a787a4091e7L14-L27).   If that flushing code is necessary for your usage, you *could* copy that `start-next.js` to your deployment and call `node --require ./my-start-next.js app.js` rather than `node --require elastic-apm-node/start-next.js app.js`.

---



> Wait a minute: is the new recommended way also experimental? Just read the [_OpenTelemetry bridge_ page](https://www.elastic.co/docs/reference/apm/agents/nodejs/opentelemetry-bridge) in your documentation now, and it says: "Integration with the OpenTelemetry Tracing API was added as _experimental_ in v3.34.0.". Is there any stable (non-experimental) way of instrumenting Next.js?

No, the "OpenTelemetry bridge" in the Elastic APM Node.js Agent (npm package `elastic-apm-node`) is **not** what is being recommended. Although the terminology is overlapping so it is confusing.

Hopefully some history can help clarify:

- Before the OpenTelemetry (OTel) project existed, Elastic had its APM agents for various languages, including the Node.js APM agent housed in this repo.
- When OTel was getting mature and popular, Elastic added "bridges" to its non-OTel APM agents. These bridges allowed a user to (a) use the non-OTel Elastic APM agent and (b) use the **OpenTelemetry API** (the OTel equivalent to the [Elastic APM agent API](https://www.elastic.co/docs/reference/apm/agents/nodejs/api)) instead of the APM agent API. The idea was that the OpenTelemetry API is vendor-agnostic, and in some cases richer, so users might prefer to use that API for their manual instrumentation.  Underneath the hood the bridge is translating OpenTelemetry API calls to the non-OTel APM agent's internal API, e.g. translating [the creation of a Span described by the OTel spec](https://opentelemetry.io/docs/specs/otel/trace/api/#span) to the Node.js API agent's [`startSpan()` API](https://www.elastic.co/docs/reference/apm/agents/nodejs/agent-api#apm-start-span).
- In the Node.js APM agent, this bridge never went further than an experiment. It became clear to Elastic that the better future was full *native* support for OpenTelemetry. That means users would get telemetry from their Node.js apps using either (a) the OpenTelemetry JS SDK  (https://opentelemetry.io/docs/languages/js/) or (b) Elastic's [distribution](https://opentelemetry.io/docs/concepts/distributions/) of the OTel JS SDK. Our distribution is called [EDOT Node.js](https://www.elastic.co/docs/reference/opentelemetry/edot-sdks/node).

The recommendation in [the v4.14.0 breaking changes note](https://www.elastic.co/docs/release-notes/apm/agents/nodejs/breaking-changes#4-14-0) **for Next.js applications** is to use the OpenTelemetry JS SDK support built directly into recent versions of Next.js and maintained by Vercel: https://nextjs.org/docs/app/guides/open-telemetry

Vercel maintains instrumentation for Next.js and has, effectively, a distribution of the OTel JS SDK in their `@vercel/otel` package.

PR #4771, opened 2025-09-08
chore: update to  v5

david-luna:
### Checklist

- [x] Update tests
- [ ] Update documentation

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -48,7 +48,7 @@
         "unicode-byte-truncate": "^1.0.0"
       },
       "devDependencies": {
-        "@apollo/server": "^4.2.2",
+        "@apollo/server": "^5.0.0",
         "@aws-sdk/client-dynamodb": "^3.379.1",
         "@aws-sdk/client-s3": "^3.379.1",
         "@aws-sdk/client-sns": "^3.379.1",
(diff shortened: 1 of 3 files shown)


lib/instrumentation/modules/graphql.js
```diff
@@ -181,7 +181,7 @@ module.exports = function (graphql, agent, { version, enabled }) {
       if (trans._graphqlRoute) {
         var name =
           queries.length > 0 ? queries.join(', ') : 'Unknown GraphQL query';
-        if (trans.req) var path = getPathFromRequest(trans.req, true);
+        if (trans.req) var path = getPathFromRequest(trans.req, true, true);
```
david-luna, line 184: note to reviewer: The 3rd param is to use the path as transaction name and makes the test pass but I?m not 100% sure about the implications of changing this.

trentm: I don't either. The tests are only testing cases where the path is '/', so kinda hard to know if there are real implications.

`usePathAsTransactionName` is *in general* dangerous because it can lead to high cardinality transaction names. But I'm not sure if that is a real concern for GraphQL endpoints where I **think** the GraphQL calls are all hanging off of a single HTTP path.

trentm:
Closing. Will follow-up with a separate PR limiting to v4.

PR #4795, opened 2025-09-25
feat(lambda): Add 'nodejs22.x' to the set of 'compatible runtimes' for our Lambda layers

trentm:
Note that my understanding is that the 'compatible runtimes' for Lambda layers is
advisory only. In other words, our Lambda layers before this *will work* with
an AWS Lambda that uses Node.js 22, but the AWS Console user interface just
doesn't list 'nodejs22.x' as compatible.

---

Reviewer note: I'll add a changelog entry on the [current release PR](https://github.com/elastic/apm-agent-nodejs/pull/4794) after this is merged.


trentm:
image

^^ and example of what this looks like in the AWS Console

PR #4885, opened 2025-12-01
chore: revert rc update of updatecli

david-luna:
reverts https://github.com/elastic/apm-agent-nodejs/pull/4879 which was merged by mistake


david-luna:
After internal discussion we decided to close this.

PR #5129, opened 2026-06-30
chore: skip publishing of AWS Lambda layers to AWS regions known to be disrupted

trentm:
https://health.aws.amazon.com/health/status
This will require eventually restoring publishing to these regions when
service is restored.

See https://github.com/elastic/apm-agent-python/pull/2658 for an
alternative take used for the Python APM agent.


trentm:
https://github.com/elastic/apm-agent-nodejs/issues/5130 opened to eventually restore Lambda publishing to those regions.

david-luna (approved):

